PR #9, opened 2026-08-18
chore: enforce workspace architecture contracts

yuezengwu:
## Summary

- add a declarative contract for all pnpm workspaces and permitted internal dependency edges
- reject unregistered workspaces, invalid internal version ranges, forbidden dependencies, deep imports, and cross-workspace relative imports
- validate package identity and artifact metadata, with six focused fixture tests
- run the architecture verifier from the existing required `pnpm check` CI path

## Why

OpenTag previously documented its package boundaries but relied on review discipline to preserve them. This change turns those boundaries into a deterministic repository gate while keeping the current four-workspace graph unchanged.

## Context Tree

- Paired draft PR: https://github.com/first-tree-ai/first-tree-context/pull/953

The Context Tree PR follows the code-first workflow: merge this code PR first, then reconcile and mark the tree PR ready.

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`

All 12 script tests and 19 workspace tests pass.

## Non-goals

- no runtime behavior or dependency versions change
- no new workspace or package is introduced

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "scripts": {
     "build": "turbo run build",
     "dev": "turbo run dev",
-    "check": "biome check . && node scripts/third-party-notices.mjs --check",
+    "check": "biome check . && node scripts/verify-workspace-contracts.mjs && node scripts/third-party-notices.mjs --check",
+    "check:workspace": "node scripts/verify-workspace-contracts.mjs",
     "lint": "biome lint .",
     "format": "biome format --write .",
     "typecheck": "turbo run typecheck",
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   "devDependencies": {
     "@biomejs/biome": "^2.4.0",
     "@types/node": "^22.15.0",
+    "semver": "^7.8.5",
     "tsdown": "^0.21.4",
     "tsx": "^4.21.0",
     "turbo": "^2.5.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,6 +23,9 @@ importers:
       '@types/node':
         specifier: ^22.15.0
         version: 22.20.1
+      semver:
+        specifier: ^7.8.5
+        version: 7.8.5
       tsdown:
         specifier: 0.21.4
         version: 0.21.4(@emnapi/core@1.10.0)(@emnapi/runtime@1.10.0)(typescript@5.9.3)
--- a/scripts/__tests__/workspace-contracts.test.mjs
+++ b/scripts/__tests__/workspace-contracts.test.mjs
@@ -80,6 +80,13 @@ const workspaceManifests = {
   },
 };
 
+const workspaceReferences = {
+  "apps/cli": [{ path: "../../packages/client" }, { path: "../../packages/shared" }],
+  "packages/client": [{ path: "../shared" }],
+  "packages/server": [{ path: "../shared" }],
+  "packages/shared": [],
+};
+
 async function writeJson(filePath, value) {
   await mkdir(dirname(filePath), { recursive: true });
   await writeFile(filePath, `${JSON.stringify(value, null, 2)}\n`);
@@ -92,6 +99,9 @@ async function createFixture() {
 
   for (const [workspacePath, manifest] of Object.entries(workspaceManifests)) {
     await writeJson(join(rootDirectory, workspacePath, "package.json"), manifest);
+    await writeJson(join(rootDirectory, workspacePath, "tsconfig.json"), {
+      references: workspaceReferences[workspacePath],
+    });
     await mkdir(join(rootDirectory, workspacePath, "src"), { recursive: true });
     await writeFile(join(rootDirectory, workspacePath, "src/index.ts"), "export {};\n");
   }
@@ -146,6 +156,20 @@ test("rejects forbidden workspace dependencies and non-workspace ranges", async 
   });
 });
 
+test("rejects TypeScript project references that drift from the contract", async () => {
+  await withFixture(async (rootDirectory) => {
+    await writeJson(join(rootDirectory, "packages/client/tsconfig.json"), {
+      references: [{ path: "../server" }],
+    });
+
+    await assert.rejects(verifyWorkspaceContracts({ rootDirectory }), (error) => {
+      assert.match(error.message, /references forbidden workspace package "@opentag\/server"/);
+      assert.match(error.message, /missing reference to allowed workspace package "@opentag\/shared"/);
+      return true;
+    });
+  });
+});
+
 test("rejects package metadata that weakens the artifact boundary", async () => {
   await withFixture(async (rootDirectory) => {
     const manifestPath = join(rootDirectory, "packages/shared/package.json");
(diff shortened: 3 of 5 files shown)


baixiaohang (changes requested):
- Rationale: The proposed gate represents the current graph correctly, but several independent dependency and artifact-boundary paths can still drift without being rejected.

**Risk level**: B-low
- Path baseline: C — the diff is limited to root configuration and repository scripts/tests.
- Semantic lift: B-low — the configuration establishes dependency-boundary and required-CI enforcement.

**PR summary**
- Problem: Contributors currently rely on review discipline to keep OpenTag workspaces on the intended dependency graph and public package surfaces.
- Approach: Add a declarative workspace contract and run a TypeScript-AST-based verifier from the required `pnpm check` path, so boundary violations receive deterministic CI feedback.
- Impacted modules: root package scripts, workspace architecture contract, repository verification scripts and fixtures.

**Review findings**
❌ 1. The contract is not checked against the per-workspace TypeScript project references, even though those files independently encode the same build dependency graph (`apps/cli` references client/shared; client and server reference shared). A new `packages/client -> packages/server` reference, or a stale reference after removing an edge from the contract, passes this verifier and leaves two topology definitions able to drift. Please validate workspace `tsconfig.json` references against `allowedWorkspaceDependencies`, or derive one representation from the other. [R1 / `scripts/verify-workspace-contracts.mjs:313`]

❌ 2. `collectModuleSpecifiers` does not recognize TypeScript `ImportTypeNode`, and it only accepts call expressions with exactly one argument. For example, `type Result = import("@opentag/server/internal").Result` inside the client never reaches `validateImports`, so it evades both the forbidden-edge and deep-import checks; a dynamic import with an options argument is skipped for the same reason. Please collect all literal module-reference forms covered by the boundary and add focused fixtures for them. [R5 / `scripts/verify-workspace-contracts.mjs:160`]

❌ 3. The artifact checks only assert that an `exports["."]` key exists and that `files` is a non-empty array. A manifest with `exports: { ".": null }` and `files: ["README.md"]` therefore passes while exposing no root artifact, so the claimed artifact boundary is not actually enforced. Please either declare the expected export targets/files coverage in the contract or validate that the root export resolves to artifacts included by `files`. [R3 / `scripts/verify-workspace-contracts.mjs:290`]

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: The project-reference and import-AST gaps are closed, but the metadata gate still accepts package targets and versions that the runtimes it is modeling reject.

**Risk level**: B-low
- Path baseline: C — the diff remains limited to root configuration and repository scripts/tests.
- Semantic lift: B-low — the configuration establishes dependency-boundary, package-artifact, and required-CI enforcement.

**PR summary**
- Problem: Contributors currently rely on review discipline to keep OpenTag workspaces on the intended dependency graph and public package surfaces.
- Approach: Add a declarative workspace contract and run a TypeScript-AST-based verifier from the required `pnpm check` path, so boundary violations receive deterministic CI feedback.
- Impacted modules: root package scripts, workspace architecture contract, repository verification scripts and fixtures.

**Review findings**
❌ 1. `target.startsWith("./")` is not sufficient to establish a valid package-relative export target. Node rejects targets that escape the package or contain `.` / `..` / `node_modules` segments after the initial prefix, but this verifier accepts, for example, `"./dist/node_modules/private.mjs"` with `files: ["dist"]`; the coverage helper returns true even though Node rejects the export and npm always excludes `node_modules`. Please validate target URL/segments according to the [Node package-target rules](https://nodejs.org/api/packages.html#path-rules-and-validation-for-export-targets) before applying `files` coverage, with a focused fixture. [R3 / `scripts/verify-workspace-contracts.mjs:387`]

❌ 2. `SEMVER_PATTERN` is not SemVer-compliant: it accepts values such as `01.0.0`, `1.0.0-alpha..1`, and `1.0.0-01`, although npm requires `version` to be parseable by `node-semver`. Please use a compliant parser/validator (or an equivalent complete implementation) and add invalid-version fixtures so the metadata gate cannot report these values as valid. See the [npm package version contract](https://docs.npmjs.com/cli/v11/configuring-npm/package-json/#version). [R4 / `scripts/verify-workspace-contracts.mjs:11`]

✅ 3. The update now validates per-workspace TypeScript references and collects both inline import types and multi-argument dynamic imports, closing the first two findings from the previous review.

**Action taken**
- Submitted request changes on head `85ccdb3`.

baixiaohang:
Architecture note (non-blocking for this PR): I do not think this PR needs to become a hard prerequisite for the v0.1 control-plane design or for starting implementation on a separate branch. Its role is to guard the existing workspace boundaries, not to define runtime behavior.

The current four-workspace graph already supports the planned split:

- `apps/cli`: command registration and the daemon process entrypoint
- `packages/client`: daemon runtime, server transport, and provider adapters
- `packages/server`: control-plane APIs and services
- `packages/shared`: runtime schemas and derived types

Please keep this PR scoped to enforcing those dependency and artifact boundaries. Future architecture changes should update `workspace-contracts.json`, manifests, TypeScript references, and affected imports atomically rather than treating the current graph as immutable. I would not add Session, provider, IM adapter, or daemon interfaces to this PR.

For delivery sequencing, I recommend merging this guardrail before the first Phase 1 implementation PR, while allowing Phase 1 detailed design and implementation preparation to proceed in parallel.

yuezengwu:
Addressed both fresh-head findings in 78aaa79.

- Root export targets now reject raw or URL-encoded dot, dot-dot, and node_modules segments, backslashes, encoded path separators, and malformed percent encoding before files-boundary coverage is evaluated.
- Package versions now use the npm node-semver parser through an explicit root dev dependency. The reported invalid versions are covered, while valid SemVer build metadata remains accepted.

Validation: 14/14 workspace-contract fixtures; pnpm check; pnpm build; pnpm typecheck; full pnpm test; git diff --check.

baixiaohang (changes requested):
- Rationale: The two reported metadata cases are addressed, but this head no longer models the repository graph or public package surface on current `main`, and one Node target-rule gap remains.

**Risk level**: B-low
- Path baseline: C — the current head changes root configuration, lockfile, and repository scripts/tests.
- Semantic lift: B-low — the change defines dependency, artifact, and required-CI enforcement.

**PR summary**
- Problem: Contributors need deterministic feedback when a workspace change violates OpenTag's intended dependency graph or package boundary.
- Approach: Add a declarative workspace contract and a verifier in the required `pnpm check` path, covering workspace discovery, manifests, project references, and source imports.
- Impacted modules: root package metadata, workspace contract, verifier and fixtures.

**Review findings**
❌ 1. The branch is now materially stale against `main` (and GitHub reports it as conflicting): current `main` discovers `apps/*` and contains the `apps/web` workspace, while this contract still registers only the original four workspaces. Once the conflict is resolved, the verifier will reject `apps/web/package.json` as unregistered unless the contract, its `@opentag/shared` edge, project reference expectations, and fixtures are updated together. Please synchronize this branch with current `main` and re-establish the contract from that graph before requesting another review. [R5 / `workspace-contracts.json:3`]

❌ 2. Current `main` explicitly publishes `@opentag/shared/browser` through `packages/shared/package.json` and uses it from four web imports, but this verifier rejects every workspace specifier other than the bare package name. An explicit `exports` subpath is a public entry point, not an internal deep import, so the merged gate would fail valid repository code. Please make the target package's `exports` map the source of truth for allowed public specifiers (including patterns if supported) and reject only non-exported subpaths, rather than hard-coding `specifier !== targetName`. [R1 / `scripts/verify-workspace-contracts.mjs:521`]

❌ 3. `isValidPackageExportTarget` still omits the empty path segment that Node 22 rejects. For example, `"./dist//index.mjs"` splits to `"dist", "", "index.mjs"`; the current regex does not match the empty segment and `files: ["dist"]` then marks it covered, although Node's `PACKAGE_TARGET_RESOLVE` rejects it. Please include `""` in the invalid-segment check and add a fixture; the [Node 22 resolver algorithm](https://nodejs.org/download/release/v22.13.1/docs/api/esm.html#resolution-algorithm-specification) lists empty segments alongside `.`, `..`, and `node_modules`. [R4 / `scripts/verify-workspace-contracts.mjs:271`]

✅ 4. The update correctly replaces the hand-written version regex with an explicit `node-semver` dependency and now rejects the previously reported encoded/raw invalid target examples.

**Action taken**
- Submitted request changes on head `78aaa79`.

yuezengwu:
Closing as superseded by #342. That merged PR explicitly adapts the workspace-contract idea from historical PR #9 to the current six-workspace apps/packages/e2e graph, includes current public export/subpath handling, and wires the replacement verifier into pnpm check. This branch still models the obsolete four-workspace graph, is conflicting with current main, and retains unresolved review findings. The branch is preserved; any follow-up should start from the merged #342 implementation.

PR #13, opened 2026-08-18
feat: add Computer runtime connection

baixiaohang:
## Summary

- add user-owned Computer persistence, `/api/v1/me/computers`, and a fenced runtime WebSocket at `/api/v1/computer/ws`
- add stable per-home Computer identity, token-aware reconnecting Client Runtime, foreground `daemon run`, and `computer list`
- migrate the existing auth and user endpoints atomically from `/v1` to `/api/v1`
- add shared dev/staging/prod channel identity for binary, home, package, and future service names
- add private atomic local storage, per-home daemon ownership, safe startup diagnostics, and end-to-end runtime smoke coverage

This PR intentionally excludes Agent, Session, provider, messaging integration, service-manager installation, and local IPC work.

## Correctness boundaries

- Computers belong directly to users and carry no Team authority; register and heartbeat resolve current suspension and active membership state from PostgreSQL
- the current daemon route is fenced by Computer ID, process instance ID, and exact socket, so stale heartbeat or close events cannot overwrite a replacement
- online/offline is derived from the fenced instance and heartbeat freshness rather than a persisted status state machine
- one OpenTag home remains bound to one server, user, and stable Computer ID; connect-code user mismatch is checked before code consumption
- source, staging, and production CLI artifacts use isolated identities and homes (`opentag-dev`, `opentag-staging`, and `opentag`)

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/server test:integration` (22 tests)
- `pnpm --filter @opentag/server db:generate` (no schema drift)
- source CLI tarball installation and execution through `opentag-dev`
- built Server + isolated PostgreSQL + login + daemon online/offline + single-home ownership + stable Computer restart smoke

The local Docker image build reached Corepack dependency download but failed on external DNS resolution for `registry.npmjs.org` (`EAI_AGAIN`). Required CI builds the image and runs the same Computer runtime path against PostgreSQL supplied as a separate service.


yuezengwu (commented):
The Computer persistence/runtime split, instance fencing, and API migration are coherent, and the required CI checks are green. I found one shutdown correctness blocker in the foreground daemon, called out inline, so I am not approving this head yet. This PR also adds the `computers` table and `computer_platform` enum; please include a human schema/migration check before merge.

apps/cli/src/core/daemon-run.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export async function runDaemon(options: DaemonRunOptions = {}): Promise<void> {
+  const home = options.home ?? resolveOpenTagHome();
+  const log = options.log ?? console.log;
+  const signals = options.signals ?? process;
+  const instanceId = randomUUID();
+  const ownership = await acquireDaemonOwner(home, instanceId);
+  let runtime: ClientRuntime | undefined;
+  const stop = () => runtime?.stop();
```
yuezengwu: Blocker: these are one-shot handlers, but `runtime` remains undefined while credential refresh, `/me`, and identity resolution are pending below. A Ctrl+C/SIGTERM during that startup window calls a no-op and consumes the handler; initialization then continues and the daemon connects instead of exiting, while a later signal no longer uses the clean `runtime.stop()` path. Please persist a stop request (and preferably abort the pending startup work), check it before starting the runtime, and cover a signal arriving while one of these awaits is pending.

baixiaohang:
Addressed the shutdown review and the additional runtime correctness findings in `06a60ac`:

- persist and abort daemon startup stop requests, including deferred token refresh;
- use one process-scoped instance ID across reconnects;
- serialize per-Computer DB registration plus registry publication and fence closed replacements;
- reject in-flight frame floods without an unbounded queue;
- recover from registered-token expiry with forced refresh;
- reset reconnect backoff on register success;
- make heartbeat request/result single-flight with request ID matching.

Local validation passes: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, 22 PostgreSQL integration tests, and Drizzle reports no schema drift. Internal re-review found no remaining blockers. Fresh GitHub Actions results are pending.

The `computers` table and `computer_platform` migration remain unchanged from the prior head and still require the requested human schema/migration check before merge.

apps/cli/src/core/daemon-run.ts
baixiaohang: Fixed in `06a60ac`. Shutdown is now persisted before the runtime exists, aborts startup continuation at every async boundary, and releases ownership without connecting. A deferred token-refresh regression test exercises the real `runDaemon()` path. The same follow-up also closes the related runtime concurrency and recovery findings; all required CI checks are green.

yuezengwu (approved):
Re-reviewed the lifecycle hardening on the updated head. The startup signal race is fixed: stop intent is retained across pending startup work, the lifecycle abort is checked after each await, and the new tests cover a signal during token refresh. The additional changes correctly keep one process instance ID across reconnects, make heartbeats single-flight, serialize same-Computer registration persistence, and clean up a registration that completes after its socket closes. I found no remaining code blocker, so approving this head. The database schema/migration still needs the planned human check before merge.

PR #14, opened 2026-08-18
fix: bundle websocket runtime in CLI

baixiaohang:
## Summary

- declare `ws` directly in the CLI build dependencies so tsdown resolves and bundles the WebSocket runtime deterministically
- resolve the `ws` third-party notice from the CLI consumer manifest
- preserve the dependency through staging and production release preparation fixtures

## Problem

After PR #13, a workspace build could emit a CLI chunk with an unresolved external `import "ws"`. Since `apps/cli` did not install `ws`, `opentag-dev daemon run` then failed at module loading with `ERR_MODULE_NOT_FOUND` before reading credentials or connecting to the Server.

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- forced CLI rebuild with no external `from "ws"` import
- source CLI tarball pack/install smoke
- built `daemon run` with an empty home reaches the expected not-logged-in boundary instead of failing module loading

## Non-goals

- no daemon protocol, Server, database, or runtime behavior changes

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -53,6 +53,7 @@
   "devDependencies": {
     "@opentag/client": "workspace:*",
     "@opentag/shared": "workspace:*",
-    "commander": "^13.1.0"
+    "commander": "^13.1.0",
+    "ws": "^8.21.3"
   }
 }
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Reviewed the dependency resolution, release-manifest preservation, lockfile, and third-party notice changes. The direct CLI build dependency addresses the unresolved WebSocket import without changing runtime behavior, protocol, or data structures. No blocking issues found.

PR #17, opened 2026-08-19
feat: add daemon service manager

baixiaohang:
## Summary

- add native user-service lifecycle management through Linux user systemd and macOS LaunchAgents
- replace the public foreground `daemon run` command with `install/start/stop/restart/status/uninstall` and a hidden service-only runtime entry
- make `login` install and start the daemon by default while preserving a credential-only `--no-start` path
- derive service identity from the shared channel contract so dev, staging, and production services can coexist
- add private `daemon.env`, atomic definitions, per-home operation/owner fencing, honest manager status, and typed supervisor exit classification

This PR intentionally excludes Windows service support, foreground fallback, Server/API/database changes, Session/Turn/provider work, Admin Web, messaging integrations, self-update, and local IPC.

## Correctness boundaries

- macOS production uses the exact visible identity `opentag` for Label, plist, wrapper, and status; dev/staging remain isolated
- service definitions pin only the absolute CLI invocation, `OPENTAG_HOME`, service mode, and PATH; credentials and tokens never enter manager definitions
- launchd distinguishes loaded jobs from running processes and uses `kickstart -k` to recover a loaded inactive job
- systemd reports unknown rather than active when MainPID cannot be verified
- a live per-home owner remains authoritative: only a verifiable stop may pass an unknown manager state, and stop succeeds only after the owner is missing or stale
- deterministic configuration and owner conflicts exit cleanly from the hidden runtime so host supervisors do not crash-loop
- successful authentication remains persisted if service installation fails, with `daemon install` as the recovery path

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/server test:integration` (31 tests)
- `pnpm --filter @opentag/server db:generate` (no schema drift)
- `pnpm audit --prod`
- source, staging, and production packed CLI installation/execution smoke
- `git diff --check`

Backend lifecycle behavior is covered with injected systemd/launchd runners and temporary homes. Required CI performs the container vertical smoke through the hidden runtime entry. Real Linux user-systemd and macOS LaunchAgent smoke remain explicit pre-merge manual checks because the latter requires a macOS host and both mutate the user's service manager.


yuezengwu (commented):
This PR adds native Linux user-systemd and macOS LaunchAgent lifecycle management, makes login install/start the daemon by default, moves the runtime behind a hidden service entry, and introduces channel-derived service identity plus per-home owner/operation fencing.

I found two must-fix correctness issues:

1. **[P1] Reject service definitions configured for a different home before mutating them** (`apps/cli/src/core/daemon/service/index.ts:80-84`). The manager target is channel-scoped, but the mutation lease and owner inspection are scoped only to the command's current `home`. If the installed unit/plist points at home A, running `OPENTAG_HOME=B opentag daemon start` sees that global definition, starts A's service, attaches B's missing owner, and returns success because the manager state is active. The same path lets `stop`, `restart`, or `uninstall` from B operate on A. Please compare the canonical `configuredHome` with the requested home for every mutation and fail closed (or implement an explicit, fenced home-transfer path for install).

2. **[P1] Make stale lease replacement conditional on the record that was inspected** (`apps/cli/src/core/daemon/process-lease.ts:61-64`). Two contenders can both read the same stale record. After one renames it and publishes a new lease at the canonical path, the second contender's unconditional `rename(path, ...)` can move that new live lease and then publish its own, leaving both callers believing they hold the operation/daemon authority. The replacement needs an atomic fence/guard that cannot rename a successor's record; PID reuse also needs process-start identity if liveness is meant to be verifiable.

These paths currently break the home-isolation and single-owner guarantees that this change is intended to establish, so I cannot approve the current head.

baixiaohang:
Addressed the service ownership review findings in `686c6f0`:

- serialize every mutation of a channel-scoped systemd/launchd target through a shared per-user target fence, while retaining the per-home operation fence;
- canonicalize requested and configured homes, re-check the definition and manager state inside the target fence, and verify the resulting configured home after mutation;
- treat a missing definition as not installed only when systemd proves `LoadState=not-found` or launchd proves the target is not loaded;
- fence process leases with process-start identity and a guarded stale takeover; unverifiable identity now fails closed without moving the canonical lease;
- make Darwin `ps lstart` identity independent of caller locale and timezone by fixing `LANG=C`, `LC_ALL=C`, and `TZ=UTC`.

Regression coverage includes concurrent first installs from two homes, definition-less loaded manager targets, stale takeover races, gone/reused/unverifiable PIDs, malformed identity records, and Darwin locale/timezone changes. The CLI package test suite (76 cases), typecheck, build, focused Biome checks, and `git diff --check` pass locally. Fresh required CI is pending.

yuezengwu (approved):
Re-reviewed the ownership hardening in `686c6f0` against the two prior blockers.

The channel-scoped target is now serialized independently of the requested home, configured homes are canonicalized and checked before and after mutation, and definition-less manager targets fail closed. Stale lease takeover is protected by an acquisition guard and validates process-start identity, preserving the canonical lease when identity cannot be verified.

Both blockers are resolved, and I found no new correctness issue in the repair diff. Approved.

PR #21, opened 2026-08-19
test: make ciphertext tamper assertion deterministic

baixiaohang:
## Summary

- mutate a decoded AES-GCM authentication tag byte instead of changing Base64URL padding bits
- make the authentication-failure assertion deterministic across Node versions and random tag endings

## Root cause

The final character of a 16-byte tag's unpadded Base64URL encoding contains only two data bits. Replacing canonical `A` with `B` changes the text but not the decoded bytes, so the prior randomized test failed about 25% of runs.

## Validation

- targeted crypto test
- 20 sequential repetitions of the targeted test
- `pnpm check`
- `pnpm typecheck`
- `git diff --check`


yuezengwu (approved):
No blocking issues found. Decoding the authentication tag and flipping a decoded byte makes the AES-GCM tamper assertion deterministic while preserving the intended failure path.

PR #22, opened 2026-08-19
fix: bind Admin Web fetch implementation

baixiaohang:
## Summary

- bind the Admin Web's default `fetch` implementation to the browser global
- add a receiver-sensitive regression test that fails when `fetch` is invoked with the `BrowserApi` instance as its receiver

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- focused Admin Web API test, typecheck, build, and Biome checks
- `git diff --check`

## E2E

Momentic browser validation is not available yet because this repository does not define the required `local` environment in `momentic.config.yaml`. The regression is covered at the browser API boundary with a receiver-sensitive jsdom test.

### DIFF
--- a/apps/web/src/__tests__/api.test.ts
+++ b/apps/web/src/__tests__/api.test.ts
@@ -10,6 +10,27 @@ function setDocumentCookie(value: string): void {
 }
 
 describe("BrowserApi", () => {
+  it("binds the default fetch implementation to the browser global", async () => {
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
No blocking issues found. The default fetch implementation is correctly bound to the browser global, and the receiver-sensitive regression test covers the failure mode.

PR #24, opened 2026-08-19
feat: add loopback development sign-in

baixiaohang:
## Summary

- add an explicitly enabled, loopback-only development browser sign-in for an existing user
- preserve the normal token issuer, suspension checks, and active Team membership authorization
- expose provider availability to Admin Web and route invitations through the unified login page
- document local environment configuration and fail-closed security boundaries

## Security boundaries

- requires raw `OPENTAG_ENV=development`, `OPENTAG_DEV_AUTH_BYPASS_ENABLED=true`, a configured email, and loopback host/public URL
- provider discovery and callback both require a loopback socket and Host; test/production and non-loopback configurations fail during startup
- the configured email must resolve to exactly one existing user; this path creates neither users nor Teams
- the callback validates a same-origin relative redirect before token issuance and stores credentials only in the existing HttpOnly cookies

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- focused PostgreSQL integration: `account-team-foundation` (7 tests)
- `git diff --check`

Momentic E2E is blocked because this repository does not yet define the required `momentic.config.yaml` `local` environment.


yuezengwu (approved):
Reviewed the loopback development sign-in flow, provider discovery, redirect handling, and the shared token issuance path. The implementation remains fail-closed outside explicit loopback development configuration and preserves suspension plus live Team membership authorization. No blocking code issues found.

PR #26, opened 2026-08-19
feat: publish Docker images to GHCR on main and release tags

bestony:
## Summary

Container images are published to GHCR on every push to `main`, not only on release tags, so a deployable image exists for the current trunk at all times. This PR covers building and publishing the image only; it ships no deployment tooling or deployment guide.

| Trigger | Coordinates published |
| --- | --- |
| push to `main` | full commit SHA, then `edge` promoted to it |
| push of `vX.Y.Z` | `X.Y.Z`, `latest` |

Platform is `linux/amd64` only. No QEMU, no arm64.

## Correctness of the mutable and immutable coordinates

Two coordinates are mutable by design (`edge`, `latest`) and the rest must never move. Getting that right drove the job layout.

**Every commit keeps its SHA image.** Workflow concurrency is keyed on the revision so no run is cancelled. A group keyed on the ref alone would not do: with `cancel-in-progress: false` GitHub still cancels the *pending* run when a newer one queues, so a commit landing between two others would silently lose its image.

**`edge` cannot move backwards.** Keying on the revision means builds for consecutive commits overlap, so an older build can finish after a newer one. `edge` is therefore not emitted by the build at all. A separate `promote-edge` job, serialised on its own concurrency group, re-checks the tip of `main` *after* the build completes and retags by digest only when this revision still leads. Checking the tip before the build would not work, because `main` advances during the ten minutes the build takes.

**Release coordinates are immutable.** A tag build now fails if `X.Y.Z` already exists in the registry, so re-running the workflow cannot overwrite a released version with a new digest. The commit-SHA tag is no longer re-emitted on release builds either: a release tag always points at a commit already on `main`, whose build already published that coordinate, so re-emitting it would replace one immutable coordinate with a second digest. Deleting the package version stays the deliberate way to republish.

**The digest is the real pin.** Tags are pointers even when disciplined. The job summary reports `ghcr.io/first-tree-ai/opentag@sha256:...` alongside the tags so an operator can pin byte-for-byte.

## Other workflow changes

- The `Validate production release coordinate` step is gated on `github.ref_type == 'tag'`. Its `run` body and `env` block are unchanged, so the release guard is not weakened; it simply no longer applies to branch pushes, where its repository-visibility, version-match, and main-ancestor checks are meaningless.
- Channel separation uses `docker/metadata-action` `enable=` predicates rather than shell branching, with `flavor: latest=false` so the automatic SemVer flavour cannot attach `latest` outside the tag path.
- Build cache is read from a shared scope and written only from branch builds, because cache entries written from a tag ref are unreadable by anything else and would only consume the repository's cache budget.
- Adds a `github.repository` guard so forks cannot publish, and OCI annotations at manifest and index level so the GHCR package links back to this repository.
- Action SHA pins, permissions, provenance, and SBOM settings are unchanged.

## Verification

- `actionlint` reports zero findings for `docker.yml`
- Both guard commands were exercised against a real registry: `docker buildx imagetools inspect` exits 0 for a published tag and non-zero for an absent one, and the tip check returns the expected SHA
- The job-summary shell was run with representative inputs and renders the intended table
- The image itself builds and runs: an earlier revision of this branch was validated with a real `docker build`, reaching `/readyz`, `/healthz`, and `/admin/` at 200

## Not covered here

- `edge` publishes without waiting for `CI`, so an `edge` image can exist for a commit whose `CI` run later fails. This matches the chosen "publish on every push to `main`" model but differs from `publish-npm-package.yml`, which gates its staging job on `workflow_run`.
- `ci.yml`'s `container-smoke` job builds the same image on every `main` push that this workflow builds, so the image is built twice per commit on two runners. Sharing the cache would require converting that job to Buildx; `ci.yml` is untouched.
- GHCR sets visibility per package and a newly published package starts private, so a maintainer must set the package to Public once after the first successful publish before anonymous `docker pull` works. The workflow never performs that step.


yuezengwu (approved):
Reviewed the publishing workflow, production Compose stack, and English/Chinese deployment documentation.

The branch/tag predicates keep edge separate from release tags, the existing release guard remains tag-only and unchanged, every revision retains its SHA image, and the Compose defaults fail closed on missing secrets while supporting explicit migrations. The documentation matches the implementation and clearly calls out the CI, package-visibility, upgrade, and rollback boundaries.

No blocking issues found.

baixiaohang (changes requested):
- Rationale: the mutable image channels are not ordered or immutable as documented, and the production Compose path silently defaults to an unpinned release.

**Risk level**: B-low
- Path baseline: workflow, Compose, and documentation changes only.
- Semantic lift: changes the release/CI publishing workflow.

**PR summary**
- Problem: operators need a deployable image for current `main` and a documented, production-oriented way to run and maintain OpenTag.
- Approach: publish `edge` plus a commit-SHA tag on `main`, retain SemVer/`latest` publishing for release tags, and add a PostgreSQL-backed Compose stack with bilingual deployment guidance.
- Impacted modules: `.github/workflows/docker.yml`, `docker-compose.prod.yml`, deployment/release docs, and README/development entry points.

**Review findings**

❌ 1. Back-to-back `main` builds can move `edge` backwards. Each run gets a unique concurrency group because the group includes `github.sha`, so builds for commits A and B run concurrently; both publish `edge`, and if A finishes after B, `edge` ends on the older commit. This breaks the documented "newest main build" contract. Keep every SHA build, but serialize or conditionally promote the mutable `edge` tag so an older run cannot overwrite a newer one. [R4 / `.github/workflows/docker.yml:19` and `.github/workflows/docker.yml:73`]

❌ 2. The guide promises an immutable release pin that the workflow does not enforce. Re-running the tag workflow rebuilds and pushes the same SemVer and commit-SHA tags again; with attestations and mutable build inputs, the resulting digest can change. Therefore the statement that `X.Y.Z` "never moves" is not guaranteed. Either enforce immutable release coordinates or make the deployment contract use an image digest (`@sha256:...`) for byte-for-byte pinning and document how operators obtain it. [R3 / `.github/workflows/docker.yml:72` and `docs/deploying.md:38`]

❌ 3. `docker-compose.prod.yml` defaults `OPENTAG_IMAGE` to mutable `latest`, even though the guide warns that `latest` can replace the running version without an explicit upgrade decision. On a first deploy or the documented `docker compose pull`, omitting this variable can select an unintended release and then auto-apply its migrations. Make the production image coordinate required (or otherwise immutable) so the Compose validation fails closed just like the required secrets. [R4 / `docker-compose.prod.yml:31` and `docs/deploying.md:41`]

✅ The branch/tag predicates, fork guard, secret validation, explicit migration service, and English/Chinese documentation structure are otherwise coherent, and the current CI checks are green.

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Updating my prior approval after re-evaluating the mutable-tag behavior. These are blocking release-contract issues:

1. The concurrency key is unique per SHA, so main builds can finish out of order and an older run can overwrite edge after a newer run. Preserve each SHA build, but serialize or conditionally promote edge.
2. Tag workflow reruns can rebuild and repush the same SemVer and commit-SHA tags, so the guide cannot promise that X.Y.Z never moves. Enforce immutability or make digest pinning the byte-for-byte deployment contract.
3. The production Compose file defaults to mutable latest even though the guide warns that pull may silently select a new release and auto-apply migrations. Require OPENTAG_IMAGE or another immutable coordinate.

Please address these before merge.

yuezengwu (changes requested):
One blocking immutability issue remains after the rewrite:

The main path still always emits the full commit-SHA tag and pushes it without an existence guard. Re-running the workflow for the same main commit therefore rebuilds and can move ghcr.io/first-tree-ai/opentag:<full-sha> to a different digest. That contradicts the stated contract that only edge and latest are mutable. Serialising equal-SHA runs prevents concurrent writes, but it does not prevent a later rerun from overwriting the coordinate.

Please make the SHA path idempotent/immutable as well (for example, reject or reuse an already-published SHA coordinate and report its existing digest). Any registry existence check used for this contract should also distinguish a definitive not-found response from network, authentication, or registry errors; the current release check treats every imagetools inspect failure as absence, so it is not fail-closed.

The separate edge promotion fixes the backwards-tag race, and removing the Compose/deployment artifacts resolves the unpinned production default concern.

baixiaohang (changes requested):
- Rationale: the rewrite fixes `edge` ordering and removes the unsafe Compose default, but the remaining immutable/mutable image-coordinate contracts are still not enforced atomically.

**Risk level**: B-low
- Path baseline: GitHub workflow change.
- Semantic lift: changes production image publishing and release coordinates.

**PR summary**
- Problem: operators need a GHCR image for every `main` revision and stable release coordinates without mutable tags regressing under overlapping builds.
- Approach: publish SHA images in overlapping build jobs, promote `edge` separately after re-checking the `main` tip, reject reused release versions, and report the published digest; deployment artifacts were removed from scope.
- Impacted modules: `.github/workflows/docker.yml`, with an existing release-guide contract affected by the changed tag outputs.

**Review findings**

❌ 1. The full-SHA coordinate is still mutable on a rerun. The `main` path always emits `${{ github.sha }}` and pushes it; the per-SHA concurrency group only serializes equal-revision runs and does not stop a later rerun from rebuilding and replacing that tag with another digest. This contradicts the updated PR contract that only `edge` and `latest` may move. Make the SHA path idempotent: reject or reuse an existing SHA coordinate, and let `edge` promote the resolved existing digest when appropriate. [R4 / `.github/workflows/docker.yml:95`]

❌ 2. The release-coordinate guard fails open. `docker buildx imagetools inspect` returns non-zero for a missing tag, but also for authentication, network, and registry failures; line 77 suppresses and treats every one of those cases as permission to publish. A transient inspection failure can therefore overwrite an existing SemVer coordinate. Distinguish a definitive not-found response from all other failures, and abort on anything that is not confirmed absence. [R4 / `.github/workflows/docker.yml:77`]

❌ 3. Concurrent release builds can still move `latest` backwards. Different release tags have different workflow concurrency groups, and both publish `latest` directly in the build step. If a newer SemVer build finishes before an older one, the older run wins last-write and leaves `latest` pointing backward. Apply the same single-writer/conditional-promotion discipline used for `edge`, with an ordering check appropriate for releases. [R1 / `.github/workflows/docker.yml:19` and `.github/workflows/docker.yml:98`]

❌ 4. The repository release guide still says a release tag publishes a full-commit coordinate, while the tag path now disables that output. Being in `main` history does not prove the earlier `main` image job succeeded or even ran (for example, an older commit can predate this workflow), so a release can complete with no SHA tag at all. Either update the documented contract and remove that assumption, or explicitly verify/reuse the SHA image as part of release publication. [R5 / `docs/releasing.md:64` and `.github/workflows/docker.yml:92`]

✅ The separate serialized `edge` promotion addresses the original backwards-tag race, and removing the Compose/deployment artifacts resolves the unpinned production default concern. The current CI checks are green.

**Action taken**
- Submitted request changes on the updated head.

yuezengwu (changes requested):
Two additional blocking issues are confirmed on the same head:

1. Release runs for different vX.Y.Z refs have different top-level concurrency groups and both publish latest directly. If a newer release build finishes first and an older release build finishes later, the older run overwrites latest and moves it backwards. Make latest a serialized, conditionally ordered promotion rather than a tag emitted by each independent build.
2. docs/releasing.md still says a production tag publishes full-commit, SemVer, and latest image tags, but the tag path now explicitly disables the full-commit output. Membership in main history does not guarantee that the corresponding main image job succeeded or even existed for an older commit. Update the English and Chinese release contract, or explicitly verify/reuse the SHA image as part of the release path.

The previously reported full-SHA rerun mutability and fail-open registry probe also remain unresolved.

bestony:
Addressed the four findings in `facad32`. The first and fourth shared a root cause, so the fix restructures how coordinates are produced rather than patching each site.

**One build per commit; every other coordinate is a retag.** The publish job now looks up the commit coordinate before building. If it is already published the build is skipped and the existing digest is reused, so a workflow re-run can no longer replace it with a second image (finding 1). A release no longer rebuilds at all: it retags that digest as `X.Y.Z`. That also settles finding 4 — the publish job guarantees the commit coordinate exists, including for a commit that predates this workflow, so a release can never complete without one. The release image is now the exact bytes CI exercised on `main`.

I verified the retag preserves attestations rather than assuming it: building with `--provenance mode=max` into a local registry and retagging by digest yields an index with the same manifest count, the same `vnd.docker.reference.type=attestation-manifest` entry, and an identical index digest.

**Registry lookups fail closed** (finding 2). `imagetools inspect` exits non-zero for authentication, throttling, and registry faults as well as for a missing tag, so the guard is now `scripts/registry-coordinate.mjs`, which queries the registry API and treats only `200` as present and only `404` as absent. Anything else — including a transport error or a `200` without a usable `Docker-Content-Digest` — throws. This follows the existing `classifyRegistryLookup` pattern in `scripts/check-npm-version.mjs`.

**`latest` moves under the same discipline as `edge`** (finding 3). It is no longer emitted by the build. `scripts/release-latest-pointer.mjs` promotes it only when no higher release is already published, so an older release build that finishes last cannot drag the pointer backwards. A higher tag whose own build failed does not block promotion, which a plain "is this the highest git tag" check would have got wrong.

One subtlety worth stating: the release coordinate is deliberately **not** on the pointer job's shared concurrency group. GitHub cancels a *pending* run when a newer one queues into the same group, which is harmless for a mutable pointer but would silently drop an immutable release coordinate. Mutable pointers are serialised; the release coordinate is not.

Both scripts are covered by 19 new unit tests in `scripts/__tests__/registry-coordinate.test.mjs`, including every non-404 failure status, a transport failure, the bearer-challenge exchange, and the ordering cases. `pnpm check` is clean, `actionlint` reports nothing for `docker.yml`, and the guards were exercised against real GHCR.

Two consequences to flag rather than bury:

- Because the release image is a retag, its `org.opencontainers.image.version` label carries the commit SHA, not `X.Y.Z`. One image with several coordinates can only label one of them, and labelling the build it actually is seemed more honest than relabelling a copy.
- `docs/releasing.md` claimed a release unconditionally carries `latest`. That is no longer true when a higher release exists, so the sentence is corrected in both mirrors.

yuezengwu (changes requested):
Two blocking races remain in the restructured workflow:

1. edge and latest share the same GitHub concurrency group. GitHub replaces an existing pending job even with cancel-in-progress false, so a main promotion can cancel a pending release promotion (or vice versa) and then update only its own pointer. There is also a same-pointer failure: if v1.2.3 promotion is pending and a later-queued v1.1.0 promotion replaces it, v1.1.0 will correctly see published v1.2.3 and skip, but nothing moves latest to v1.2.3. Separate pointer writers are necessary, and the latest writer must converge to the highest published release rather than merely decide whether the current run may promote itself.
2. The workflow-level group still includes github.ref. A main push and a release-tag push for the same SHA can therefore run concurrently, both observe the commit coordinate as absent, build different attested images, and push the same SHA tag. The release may then retag one digest while the SHA coordinate ends on the other. Ensure creation of the commit coordinate is single-writer/idempotent across branch and tag events; checking before building is not atomic.

The registry status classification, release-guide correction, and retag-only release path otherwise address the earlier findings.

baixiaohang (changes requested):
- Rationale: the new lookup/retag design resolves the prior single-run issues, but cross-trigger concurrency can still overwrite an immutable SHA coordinate or permanently drop a mutable-pointer update.

**Risk level**: B-low
- Path baseline: workflow, release scripts/tests, and release documentation.
- Semantic lift: changes production image publishing and release-coordinate ownership.

**PR summary**
- Problem: operators need one immutable image per commit plus `edge` and `latest` pointers that converge correctly despite retries and overlapping main/release events.
- Approach: query GHCR with explicit failure classification, build only when a commit coordinate is absent, derive release tags by digest retagging, and serialize mutable-pointer promotion with ordering checks.
- Impacted modules: Docker publishing workflow, registry/release-coordinate scripts and tests, Turbo environment config, and bilingual release guides.

**Review findings**

❌ 1. Commit-coordinate creation is not single-writer across event types. The workflow-level group includes `github.ref`, so a `main` run and a release-tag run for the same SHA use different groups. If the tag is pushed while the main image is still building, both runs can observe the SHA coordinate as absent, build different attested digests, and push the same supposedly immutable SHA tag; the release may then point to one digest while the SHA tag ends on the other. Serialize ownership by commit across branch/tag events, or make the tag path wait for/reuse a definitively published commit coordinate instead of participating in the check-then-build race. [R4 / `.github/workflows/docker.yml:23` and `.github/workflows/docker.yml:77`]

❌ 2. The shared pointer queue can drop updates, and `latest` does not converge after such a drop. Because `edge` and `latest` use one concurrency group, a main promotion can replace a pending release promotion (or vice versa) and then update only its own pointer. Even among releases, a later-queued lower version can replace a pending higher-version promotion; it then sees that higher release, skips itself, and leaves `latest` unchanged rather than moving it to the highest published digest. Give the pointers independent writers, and make the `latest` writer reconcile to the highest published release rather than only decide whether the current release may promote itself. [R1 / `.github/workflows/docker.yml:191` and `scripts/release-latest-pointer.mjs:43`]

✅ The 200/404 registry classification now fails closed, same-event SHA reruns reuse the existing digest, release coordinates are retags rather than rebuilds, and the English/Chinese release contract is synchronized.

**Action taken**
- Submitted request changes on the updated head.

bestony:
Both blocking races are addressed in `7eb6cce`.

**Commit coordinate now has one writer.** The tag path no longer builds at all. The branch path is the sole producer of the commit coordinate; a release waits for that image to be published and retags it. This removes the check-then-build race outright rather than trying to make it atomic — there is no second builder left to race. A release for a commit that never published an image now fails with a named error after the wait elapses, instead of quietly building a competing digest.

**Pointers have independent writers and reconcile rather than self-decide.** `edge` and `latest` are on separate concurrency groups, so a branch promotion can no longer displace a pending release promotion. More importantly, neither writer asks "may I promote myself?" any more:

- `latest` resolves the highest release that is actually published and points at that.
- `edge` resolves the newest `main` commit that has a published image and points at that.

Every run computes the same target, so whichever run survives the queue lands the pointer correctly. This fixes the case you described — a later-queued `v1.1.0` displacing a pending `v1.2.3` promotion now moves `latest` to `v1.2.3` rather than skipping and stranding it. A release tag whose build never published is skipped, so a failed release cannot pin the pointer.

I applied the same reconciliation to `edge` even though only `latest` was raised. The self-deciding version had a matching flaw: when the build for the tip failed, a superseded run would see it was not the tip and skip, leaving `edge` on an older image. That one self-healed on the next push rather than stranding permanently, but the asymmetry was not worth keeping.

`scripts/edge-pointer.mjs` is new and mirrors `release-latest-pointer.mjs`. Script tests went from 19 to 36, covering the bounded wait (including that an inconclusive lookup propagates instead of being retried away), the convergence of both reconcilers from different starting points, and that a higher release whose build failed does not block the pointer.

`actionlint` is clean on `docker.yml`, `pnpm check` is clean, and all 36 script tests pass.

One note on the earlier CI failure, since it was not caused by this branch: `Node 22.13.0 Compatibility` failed on `facad32` in `packages/client/src/__tests__/session-binding-store.test.ts` with a 5s test timeout. Re-running it with no code change passed. It surfaced here because adding entries to `globalPassThroughEnv` in `turbo.json` invalidates the Turbo cache, so that test actually executed instead of being restored from cache.

yuezengwu (approved):
Re-reviewed the coordinate workflow at 7eb6cce. The two remaining races are resolved:

- The branch path is now the sole commit-image writer; the tag path waits for and reuses that published digest, so branch/tag overlap cannot create competing SHA images.
- edge and latest have independent concurrency groups, and each surviving queued run reconciles to the newest published main commit or highest published release instead of deciding only for itself.

The bounded wait fails clearly when no branch image arrives, registry errors remain fail-closed, and the English/Chinese release contract matches the implementation. No blocking issues found.

PR #31, opened 2026-08-19
feat: add structured client logging

baixiaohang:
## Summary

- add a Pino-based `ClientLogger` with structured fields, credential redaction, configurable levels, and a private rotating NDJSON sink at `$OPENTAG_HOME/logs/client.log`
- migrate daemon, connection, client runtime, Turn, and report-recovery diagnostics from string callbacks to structured logger children with minimal correlation IDs
- make logging failures observational only through synchronous stderr fallback, short-write handling, rotation failure isolation, and daemon ownership-scoped logger gating
- update launchd/systemd status output to identify `client.log` as the primary service log while retaining supervisor output as a startup-crash fallback

## Identity and safety boundaries

- `instanceId` remains the per-daemon connection instance and is not treated as `agentId`
- events add only the IDs available at their layer (`computerId`, `agentId`, `sessionId`, `turnId`, `requestId`, `deliveryId`, and `placementGeneration` as applicable)
- no message payloads, prompts, raw frames, provider stderr, credentials, external IM identifiers, database changes, or protocol/trace schema changes are included
- only the daemon holding the canonical-home ownership lease can configure or write the service sink; the entire runtime logger tree is disabled before lease release

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
  - scripts: 12 tests
  - shared: 42 tests
  - client: 103 tests
  - server: 79 tests
  - web: 13 tests
  - CLI: 99 tests
- `git diff --check`

### DIFF
--- a/apps/cli/THIRD_PARTY_NOTICES
+++ b/apps/cli/THIRD_PARTY_NOTICES
@@ -32,6 +32,253 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 ---
 
+## pino@10.3.1
+
+Source: https://github.com/pinojs/pino
+License: MIT
+
+The MIT License (MIT)
+
+Copyright (c) 2016-2025 Matteo Collina, David Mark Clements and the Pino contributors listed at <https://github.com/pinojs/pino#the-team> and in the README file.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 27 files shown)


yuezengwu (changes requested):
Blocking finding in `packages/client/src/observability/rotating-file-stream.ts:103-108` (and the reopen at line 122): the private-path invariant is only applied to newly created paths. `existsSync` follows symlinks, `openSync(path, "a", ...)` follows a pre-existing `client.log` symlink, and existing directories/files are never chmod'd. A pre-existing `$OPENTAG_HOME/logs` or `client.log` can therefore stay group/world-readable; a planted symlink can also make the daemon append NDJSON to another file writable by the user. This contradicts the advertised private sink and the repository's existing secure-storage posture.

Please validate that `logs` is a real directory, reject symlink/non-regular log files, open with `O_NOFOLLOW` plus an `fstat` check, and normalize existing directory/file modes to `0700`/`0600`. The same safe open needs to be used after rotation. Failure can still fall back to stderr to preserve the observational-only behavior. Please add regression coverage for pre-existing loose permissions and symlinked directory/file paths.

baixiaohang:
Addressed the security review and the failing CI jobs in commit `7df2b0d`:

- hardened `client.log` setup and reopen with `lstat`, `O_NOFOLLOW`, `fstat`, and fd-based permission normalization (`0700` directory / `0600` file)
- rejected symlinked directories/files and non-regular log paths while preserving stderr failover
- added regression coverage for pre-existing loose permissions, symlink targets, non-regular paths, and rotation reopen behavior
- included Pino and its bundled transitive dependencies in generated `THIRD_PARTY_NOTICES`, fixing Node compatibility and CLI pack smoke failures
- restored the stable `already running` busy-daemon diagnostic required by the container smoke test without exposing dynamic error details

Local verification passed: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, `node scripts/node-compat-smoke.mjs`, and `git diff --check`. Independent code review found no remaining blocker.

yuezengwu (approved):
Re-reviewed the current head. The previous blocking finding is resolved: the log directory and file are type-checked with `lstat`, opened with `O_NOFOLLOW`, verified with `fstat`, and normalized through the opened descriptors to `0700`/`0600`; rotation reuses the same safe open path, while failures still degrade to stderr without affecting runtime behavior.

I also reviewed the related stable busy-daemon diagnostic and generated third-party notice changes. No remaining code issue found. Approved.

PR #34, opened 2026-08-19
fix: stabilize Agent runtime snapshots and add CLI settings

baixiaohang:
## Summary

- add the single server-side `EffectiveRuntimeSnapshotAssembler` with stable versioned revision hashes and effective Agent/Session revision sequencing
- pin accepted IM dispatch payloads for recovery, clean them up after atomic Turn reporting, and serialize same-Agent custody with renewable owner-token claims
- add Agent runtime settings to CLI create/update, including file-based instructions, repeatable tools, timeout, CAS, and clear semantics
- replace the IM worker's inline snapshot builder and validate persisted payload/hash correlation

## Correctness

- Agent configuration changes advance both effective revisions for existing Sessions without causing Client configuration conflicts
- accepted Turns recover with their original immutable snapshot before a newer Agent revision can be applied
- active claims are fenced across workers with conditional leases; inactive Session/Integration/Agent authority cannot leave a global routing fence
- unsupported providers and invalid stored configuration fail closed

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/server test:integration` (6 files, 96 tests)
- `git diff --check`

No database migration is included.

### DIFF
--- a/apps/cli/src/__tests__/agent.test.ts
+++ b/apps/cli/src/__tests__/agent.test.ts
@@ -1,5 +1,9 @@
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { resolve } from "node:path";
 import type { Agent } from "@opentag/shared";
 import { describe, expect, it, vi } from "vitest";
+import { createProgram } from "../cli/program.js";
 import { formatAgent, formatAgentCreated, formatAgentList } from "../core/agent/formatting.js";
 import { runAgentCreate, runAgentDelete, runAgentUpdate, selectComputer } from "../core/agent/mutations.js";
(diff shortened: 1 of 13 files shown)


yuezengwu (approved):
Reviewed the exact head `5fe6152`.

No blocking correctness issues found. The centralized snapshot assembler keeps Agent- and Session-layer revisions deterministic, accepted delivery recovery remains pinned to the persisted immutable payload until atomic Turn reporting, and same-Agent dispatch custody is fenced with renewable owner tokens. The CLI runtime-setting paths preserve the existing Agent revision CAS behavior and validate clear/conflict semantics before API access.

No database migration is included. CI is green; I did not rerun tests as part of this diff review.

PR #36, opened 2026-08-20
fix(server): accept Google OAuth callback fields

baixiaohang:
## Summary

- accept provider-owned query fields on the Google OAuth callback while preserving validation for `state` and exactly one of `code` or `error`
- verify OAuth state before mapping provider errors, returning stable `AUTH_OAUTH_FAILED` responses for cancellations and other failures without reflecting provider text
- clear the one-time OAuth context cookie before code exchange or provider error handling so terminal callbacks cannot reuse the flow context
- add route and service regression coverage for real Google callback fields, invalid callback shapes, provider errors, invalid flows, and cookie cleanup

The root cause was a strict callback query schema that rejected valid Google-owned fields such as `scope`, `authuser`, and `prompt` before OAuth state verification or code exchange.

## Testing

- `pnpm --filter @opentag/server exec vitest run --maxWorkers=1 src/__tests__/browser-auth.test.ts src/__tests__/google-auth.test.ts`
- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`

## Breaking changes

None.

## Checklist

### DIFF
--- a/packages/server/src/__tests__/browser-auth.test.ts
+++ b/packages/server/src/__tests__/browser-auth.test.ts
@@ -36,10 +36,25 @@ function google() {
     start: vi
       .fn()
       .mockResolvedValue({ authorizationUrl: "https://accounts.google.com/auth", context: "signed-context" }),
-    callback: vi.fn().mockResolvedValue({
-      next: "/admin",
-      tokens: { accessToken: "access-secret", refreshToken: "refresh-secret", tokenType: "Bearer", expiresIn: 900 },
+    callback: vi.fn().mockImplementation(async (_input: unknown, _context: string, options: { onVerified(): void }) => {
+      options.onVerified();
+      return {
+        next: "/admin",
+        tokens: {
+          accessToken: "access-secret",
+          refreshToken: "refresh-secret",
+          tokenType: "Bearer" as const,
+          expiresIn: 900,
+        },
+      };
     }),
+  };
+}
+
+function failAfterVerification(error: Error) {
+  return async (_input: unknown, _context: string, options: { onVerified(): void }) => {
+    options.onVerified();
+    throw error;
   };
 }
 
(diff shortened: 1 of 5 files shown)


yuezengwu (changes requested):
Blocker (`packages/server/src/api/browser-auth.ts:158`): the route clears the OAuth context cookie before `GoogleBrowserAuthService.callback()` verifies `state`. Because this callback cookie is `SameSite=Lax`, a cross-site top-level GET with a syntactically valid fake `code`/`state` can carry the victim’s cookie and make the response delete it, aborting an in-progress sign-in even though the callback is unauthenticated. Please clear the context only after the state/context pair has been verified, while still doing so before provider-error handling or code exchange, and add a route regression test showing that an invalid or mismatched state does not clear the cookie.

baixiaohang:
Addressed the state-verification blocker in `fc90efd`.

- The route now passes cookie deletion as an `onVerified` lifecycle callback instead of clearing the OAuth context before service validation.
- `GoogleBrowserAuthService.callback()` invokes `onVerified` only after `OAuthFlowService.verify(state, context)` succeeds, and before provider-error mapping or code exchange.
- A new route regression test confirms that an invalid or mismatched state returns the existing OAuth failure without a `Set-Cookie` header.
- Service tests lock the ordering for provider errors and code exchange, and confirm that invalid flow verification never invokes `onVerified`.

Validation after the fix:

- targeted browser/Google auth tests: 16 passed
- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`

yuezengwu (approved):
The blocker is resolved in `fc90efd`: the OAuth context is now cleared only after state/context verification succeeds, while still being cleared before provider-error handling or authorization-code exchange. The added route and service tests cover the invalid-state behavior and lifecycle ordering. No further code issues found.

PR #37, opened 2026-08-20
feat(client): add Pi agent runtime

yuezengwu:
## Summary

- add a provider-neutral Pi Agent Runtime implementation over the user-installed pi RPC CLI
- create and resume exact local sessions with opaque session-file fingerprint bindings and fail-closed policy enforcement
- add strict correlated JSONL process handling, process-tree cleanup, exhaustive protocol tests, and a live create/resume smoke test
- reject common hosted-tool allow-lists because the documented Pi RPC cannot register Client-hosted tools
- preserve the local artifact boundary: no Pi package, SDK, CLI, binary, installer, or download fallback is added

## Validation

- pnpm --filter @opentag/client test:agent-runtime:coverage (207 tests; 100% statements, branches, functions, and lines)
- pnpm --filter @opentag/client test (304 tests)
- pnpm --filter @opentag/client test:e2e:pi-agent-runtime (pi 0.83.0; create and exact resume completed)
- pnpm --filter @opentag/server test:integration (96 tests)
- pnpm check
- pnpm build
- pnpm typecheck

The full pnpm test command reaches two existing macOS CLI assertions that compare /tmp with its canonical /private/tmp path. All other package tests pass; GitHub CI runs on Linux.


baixiaohang (changes requested):
- Rationale: the readiness probe can accept Pi artifacts that cannot satisfy the Runtime terminal protocol, and the event translator accepts cross-boundary sequences that publish invalid lifecycle ordering.

**Risk level**: A
- Semantic lift: none; the path baseline is already A.

**PR Summary**
- Problem: OpenTag needs to run and exactly resume local Pi conversations through the provider-neutral Agent Runtime without bundling or installing Pi, weakening common safety policy, or leaking local session paths.
- Approach: launch the user-installed Pi CLI in RPC mode once per Run, bind the persistent session by UUID plus an opaque session-file-path fingerprint, translate Pi JSONL events into the common Runtime stream, reject unsupported hosted-tool and sandbox policies, and cover the wire and live create/resume boundaries.
- Impacted modules: Client Agent Runtime provider exports, Pi RPC process ownership, provider tests and live smoke coverage, and the Agent Runtime test plan.

**Review findings**
❌ 1. The readiness check does not establish the minimum Pi protocol this Runtime actually needs. Pi v0.80.3 exposes every help token tested here, including RPC and `--session-id`, so it can pass this probe when credentials are present; however, it predates the RPC `agent_settled` event added in v0.80.4. `executeRun` waits exclusively on that event, and even an abort only sends the RPC command without rejecting the terminal waiter, so a supposedly ready v0.80.3 installation can leave every Run stuck indefinitely. Parse and enforce a compatible version/capability floor (also accounting for the accepted `max` thinking level), and lock the negative case with a probe fixture. Upstream evidence: [v0.80.3 CLI surface](https://github.com/badlogic/pi-mono/blob/a23abe4a695df8b69b613f73e9fdda2a8af894d4/packages/coding-agent/src/cli/args.ts#L238-L280), [`agent_settled` introduction](https://github.com/badlogic/pi-mono/blob/845d6ff1f6643aba440341cce877ce1c43ebbc39/packages/coding-agent/CHANGELOG.md#L318-L328). [R4, R5 / `packages/client/src/providers/pi/agent-runtime.ts:792`]

❌ 2. `turn_end` publishes `model_turn_completed` after checking only that a turn ID exists; it does not require the active assistant message and tool executions to be closed, and the message/tool handlers do not require an active turn. Consequently, `turn_start -> assistant message_start -> turn_end -> message_end -> agent_settled` is accepted as a completed Run and emits message events after `model_turn_completed`. That violates the ordered stateful Runtime boundary and can make downstream trace consumers close a model turn before its content. Fence child events to the active turn, require no open message/tools before `turn_end`, and add crossed-boundary transition cases. [R4, R5 / `packages/client/src/providers/pi/agent-runtime.ts:322`]

❌ 3. The credential probe runs bare `--offline --list-models`, while production execution explicitly disables extensions, skills, templates, themes, context files, and project trust. Pi extension factories can contribute models to `--list-models`, so the probe can execute a user-installed extension and report `ready: true` from a model that is unavailable when the real RPC process starts with `--no-extensions`; it is both a readiness false positive and a different executable surface from production. Run model discovery with the same resource-disable/trust flags as the Runtime (and cover an extension-only model case). [R5 / `packages/client/src/providers/pi/agent-runtime.ts:795`]

**Action taken**
- Submitted request changes on head `e813a266cd7c75096dbaf2fffef0f52a1ee96e5b`.

yuezengwu:
The branch is current with main at 557f095 and all six GitHub checks pass. Local verification on the merged head also passed 207 scoped tests at 100% coverage, 304 Client tests, 96 Server integration tests, check/build/typecheck, and a real pi 0.83.0 create plus exact-resume E2E. The dependency diff adds no Pi package or binary; runtime execution remains the literal user-PATH pi artifact. There are no inline review threads. Please review the current head e813a26.

yuezengwu:
Addressed all three review blockers in e01cbff:

1. The default probe now requires Pi >= 0.80.6, covering both RPC agent_settled (introduced in 0.80.4) and the accepted max thinking level (introduced in 0.80.6). Controlled artifacts lock 0.80.3 and 0.80.5 as version_incompatible.
2. Every message and tool child event now requires an active turn, and turn_end fails closed while an assistant message or tool remains open. Regression sequences cover all six child event types outside a turn and both unfinished child families at turn_end.
3. Model discovery now reuses the same offline, extension, skill, prompt-template, theme, context-file, and project-trust disabling arguments as production. An extension-only model fixture now reports credential_missing.

Verification on the updated head:
- Agent Runtime coverage: 207/207, 100% statements, branches, functions, and lines
- Client tests: 304/304
- Real pi 0.83.0 create and exact-resume E2E: passed
- Client build/typecheck and repository check: passed

Please re-review head e01cbff.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: head `e01cbff` closes all three protocol blockers with fail-closed implementation changes and focused regression coverage; no new blocking issue remains.

**Risk level**: A
- Semantic lift: none; the path baseline is already A.

**Review findings**
✅ Pi now requires version 0.80.6 or newer, covering both `agent_settled` and the accepted `max` thinking level.
✅ Message and tool child events are fenced to an active turn, and `turn_end` rejects unfinished children.
✅ Credential discovery now uses the same resource-disable arguments as production execution.
✅ All six GitHub checks pass on the reviewed head.

**Action taken**
- Approved exact head `e01cbffc1e4373c26ad94438c2286d7664a01d06`.

PR #43, opened 2026-08-20
feat: complete Team invitation flow

baixiaohang:
## Summary

- add admin-facing invitation link creation, copying, and rotation to Settings → Members
- resume pending invitation joins after sign-in, select the joined Team only after fresh membership verification, and hand off the Team selected during Google OAuth without replaying the bearer token
- make repeated invitation redemption audit-safe by reporting membership lifecycle transitions from the owning service
- document the completed Web invitation workflow in the canonical English docs and Chinese mirrors

## Testing

- `pnpm check`
- `pnpm typecheck`
- `pnpm test`
- `pnpm build`
- `pnpm --filter @opentag/server exec vitest run src/__tests__/integration/account-team-foundation.test.ts --maxWorkers=1`
- Browser E2E was intentionally skipped at the requester's direction.

## Breaking changes

None.

## Checklist

- [x] The change is focused and contains no unrelated work.
- [x] Tests and documentation cover the changed behavior.
- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass.
- [x] No credentials, private data, or generated build output are included.
- [x] Canonical English docs and Chinese mirrors are synchronized when applicable.

PR #44, opened 2026-08-20
feat(web): create Teams explicitly instead of provisioning them at sign-in

Gandy2025:
## Summary

Adds the Web Team creation flow that [OpenTag Web product plan v0.1](https://github.com/first-tree-ai/first-tree-context/pull/962) puts at the start of onboarding, and removes the auto-provisioning that made that step unreachable.

The proposal is explicit that Team creation is a user action: the first member to create a Team becomes its admin (§2), onboarding begins with 创建或加入 Team (§6.1, §7), `/teams/new` is its own page in the page map (§5), and §10 lists Team 创建 as its own v0.1 deliverable. Sign-in previously created a personal Team behind the user's back, so that step never happened and the canonical name was machine-generated (`ada-3f2a1b0c`).

That first onboarding step also requires an authenticated session to exist before any Team does, so **a session with no membership is now legal**. Team authority was already re-checked per resource on every request, with one exception: `ComputerService` treated holding a session as proof of membership, so it now asserts membership explicitly and keeps rejecting Computers owned by users who belong to no Team.

Zero-membership was already reachable before this change — a member who left their only Team was left with a 403 on `/me` and a permanently stuck Web session. Such a session now lands on Team creation like any other.

- **shared** — `CreateTeamRequest` / `CreateTeamResponse` schemas and the `teams` path
- **server** — `TeamMembershipService.createTeam`, reusing `bootstrapAdminInTransaction` so creator-becomes-admin has one implementation, and `POST /api/v1/teams`
- **web** — a real `/teams/new` with an auto-slugged, editable canonical name, and a Team-less session routed there instead of a dead end

A duplicate canonical name is reported as `409 TEAM_NAME_CONFLICT` and rolls the whole transaction back, leaving no half-created Team.

The second commit applies three fixes from a review pass over the first:

- Team creation had no ceiling. Previously a user could self-serve exactly one Team and needed an invitation for more, so nothing had to enforce a limit. Creation now refuses past `TEAM_MEMBERSHIP_LIMIT` (50) active memberships, counted under the caller's existing user-row lock so concurrent creations cannot both pass. Capping memberships rather than authored Teams is deliberate: memberships are what the per-request read in `#resolveActiveUser` walks, and a sole admin cannot leave their own Team, so the cap does not reset.
- Canonical-name derivation is ASCII-only. `示例团队` derived an empty name and `OpenTag 中文团队` silently derived `opentag`, keeping only the fragment. The page now says which characters were left out, and only when derivation actually drops something (`Café Ops` → `cafe-ops` stays quiet).
- Emptying the canonical name latched auto-derivation off for good. Clearing now returns control to the display name.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck` — pass
- `pnpm test` — shared 54, server 131, web 28 pass
- `pnpm --filter @opentag/server test:integration` — 108 pass
- `pnpm --filter @opentag/server db:generate` — no schema drift

Covering the acceptance criteria: a Team-less session is routed to `/teams/new` and lands on `/agents` with the new Team selected; a case-insensitive name collision returns 409 and keeps the form filled with no partial rows written; an existing member creates a second Team and sees both in the picker; the 51st creation is refused and its name stays free.

Two pre-existing CLI failures are unrelated to this branch — `daemon-service-shared` and `daemon-service-renderers` do not resolve macOS's `/var` → `/private/var` symlink. Both fail identically on a clean checkout of `main` at 12b574d with no changes; CI runs on `ubuntu-latest`, where `/tmp` is not a symlink.

## Breaking changes

**A session can now exist without any Team membership.** `AuthService.#resolveActiveUser` no longer throws `AUTH_MEMBERSHIP_REQUIRED`; `/me` may return `memberships: []`. Any future caller that treats holding a session as proof of Team membership must check explicitly — `ComputerService` is updated here, and the CLI's `selectTeam` already handled the empty case.

**Self-serve sign-in no longer yields a Team.** A new user must create one before they can generate a connect code, since `ConnectCodeService.issueForUser` still requires an active membership. Invited users are unaffected — that path redeems the invitation and never created a personal Team.

Four existing assertions changed because this PR removed their premise, rather than because they regressed: `auth-migrations.test.ts` now asserts the session survives membership loss (and re-covers the 403 branch via account suspension), and `account-team-foundation.test.ts` drops the personal-Team test in favour of "solo sign-in creates no Team". `computers.test.ts` is untouched and stays green — which is why `ComputerService` needed its own gate.

No schema change, no migration.

## Non-goals

Onboarding entry beyond the redirect is PR-A2's; `/teams/new` is a plain redirect target here rather than a wizard step.

**The Team-less redirect only serves half of §6.1's first step, deliberately.** That step reads 创建**或加入** Team, and this PR sends a Team-less session straight to creation with no visible path to joining. The proposal specifies that `/teams/new` exists (§5) and that creation comes first (§6.1, §7), but says nothing about how a Team-less session reaches it — the redirect is this PR's choice, not something the document mandates.

Who this actually affects is narrow. Anyone holding an invite link goes through `/invites/:token` and, thanks to #43's pending-invitation handling, is carried back there after signing in without ever passing the gate. The exposed case is someone who signs in cold expecting to have been invited: they land on Create Team and may create a junk Team instead of going to find their invitation. Giving that step both options belongs with the rest of the onboarding entry work in PR-A2.

§6.12 wants Settings → Team to show 创建者, but `teams` has no `created_by` column. Adding one is a DB change needing explicit approval, so it is left out.

## Checklist


baixiaohang (changes requested):
- Rationale: the new Team-creation ceiling and field contracts do not yet enforce the invariants they claim, and the branch must be reconciled with the newly merged invitation flow before the combined onboarding behavior is reviewable.

**Risk level**: B-high
- Path baseline: `packages/server/**` is changed -> B-low.
- Semantic lift: the Web API/data layer and live authentication/membership semantics change -> B-high.

**PR summary**
- Problem: new or membership-less users currently cannot explicitly create a Team and can become stuck after sign-in or after losing their last membership.
- Approach: allow an authenticated session with zero memberships, add an explicit transactional Team-creation API that makes the creator admin, route Team-less Web sessions to `/teams/new`, and keep Computer operations gated on a current membership.
- Impacted modules: Web routing/API, shared Team contracts, server authentication, Team membership, Computer authorization, and integration tests.

**Review findings**

❌ 1. The new limit is neither a durable Team-creation ceiling nor a bound on the membership fan-out it is meant to protect. `createTeam` counts only active memberships, but a creator can add another admin and then `leave`, which changes that row to `left`; the same account can then repeat the next batch of 50 creations. Invitation redemption and admin restore can also add active memberships without consulting this limit, so `/me` can still exceed 50 rows. Please enforce one clearly named invariant at every writer under compatible locking: either a durable self-serve creation quota, or a true active-membership ceiling across create/join/restore, and cover transfer/leave plus concurrent join/create behavior. [R3, R5 / `packages/server/src/services/teams/team-membership-service.ts:197`]

❌ 2. Team field limits now have two sources of truth. Creation caps `name` at 64 and `displayName` at 120, while `UpdateTeamProfileRequestSchema` and the response projections accept unbounded values. An admin can therefore create a valid Team and immediately rename it to values the creation API rejects. Put the field constraints in reusable canonical schemas and apply them consistently to create, update, and projections, with update-boundary coverage. [R1 / `packages/shared/src/team.ts:14`]

❌ 3. The response contract does not encode the creator-becomes-admin guarantee: `CreateTeamResponseSchema.role` accepts both `admin` and `member`. Because the server reparses the service result through this schema, a regression returning a member would pass the API boundary. Make this field the literal `admin` and test that `member` is rejected. [R1 / `packages/shared/src/team.ts:21`]

❌ 4. The current head conflicts with `main` in `router.tsx` and its app tests, exactly where #43 added pending-invitation replay and server-selected-Team handling after OAuth. Resolve against current `main` while preserving both behaviors, and add/retain a combined case showing that a Team-less invitee resumes the invitation instead of being diverted to Team creation. [R5 / `apps/web/src/router.tsx:134`]

**Action taken**
- Submitted request changes. I reviewed the supplied test evidence and source assertions without running checks or tests.

baixiaohang (changes requested):
- Rationale: the merge conflict and invitation-flow integration are resolved, but the first three contract/invariant blockers from the prior review are unchanged on the new head.

**Risk level**: B-high
- Path baseline: `packages/server/**` is changed -> B-low.
- Semantic lift: the Web API/data layer and live authentication/membership semantics change -> B-high.

**PR summary**
- Problem: new or membership-less users need to authenticate before explicitly creating or joining a Team instead of receiving an implicit personal Team or becoming stuck.
- Approach: preserve zero-membership sessions, add transactional explicit Team creation with creator-as-admin, route Team-less Web sessions to `/teams/new`, and retain explicit membership gates for Computer operations.
- Impacted modules: Web routing/API, shared Team contracts, server authentication, Team membership, Computer authorization, and integration tests.

**Review findings**

❌ 1. Still unresolved: the limit counts only active memberships at creation time. A creator can add another admin, leave the Team, and reset the counter; invitation redemption and admin restore also add active memberships without enforcing this limit. The implementation therefore neither durably bounds self-serve Team creation nor bounds the `/me` membership fan-out named in its rationale. Enforce one shared, clearly named invariant across every applicable writer and cover transfer/leave plus join/create concurrency. [R3, R5 / `packages/server/src/services/teams/team-membership-service.ts:197`]

❌ 2. Still unresolved: creation caps `name` at 64 and `displayName` at 120, while `UpdateTeamProfileRequestSchema` and the response projections accept unbounded values. Reuse canonical field schemas across create, update, and projections so the same Team cannot immediately be renamed to values its creation contract rejects. [R1 / `packages/shared/src/team.ts:14`]

❌ 3. Still unresolved: `CreateTeamResponseSchema.role` accepts both membership roles even though creator-becomes-admin is a hard contract. Use the literal `admin` and add a rejection assertion for `member`. [R1 / `packages/shared/src/team.ts:21`]

✅ 4. Resolved: head `77c685c158f3b037f53f488a99efc05dd9d7811b` is mergeable and preserves #43's pending-invitation replay, OAuth-selected Team handling, and the new Team-less creation redirect.

**Action taken**
- Submitted request changes on the updated head. I inspected the incremental merge resolution and combined source assertions without running checks or tests.

Gandy2025:
Thanks — all three were real. Fixed in 8ff7b0d.

**1. Membership ceiling.** You were right on both halves, and the first one was worse than a gap: the comment asserted an invariant the code did not have. `leave` requires only that *another* admin exists, not that the caller is the sole one, so a creator could promote an admin, leave, and free the slot — my claim that "a sole admin cannot leave their own Team, so the cap does not reset" was simply wrong. Redemption and restore also raised the count without consulting the limit, so it never bounded the `/me` fan-out it named as its own rationale.

One helper now enforces it at all three writers that can raise the count, under a lock on the affected user's row, and the comment states only what holds. Integration coverage asserts redemption and restore are both refused at the ceiling.

I took the second of the two options you offered — a true active-membership ceiling rather than a durable creation quota. A durable quota needs a record of authorship the schema does not carry (`teams` has no `created_by`), which is a DB change the repo requires explicit human approval for, and it was declined for this PR. So the leftover exposure is name squatting via leave-and-repeat. Rather than half-claim it, the comment now says plainly that this is not a creation quota, and squatting is tracked separately.

**2. Two sources of truth.** Correct, and I had noticed the unbounded update path while implementing and wrongly judged it out of scope — a half-bounded contract is worse than an unbounded one. `TeamNameInputSchema` and `TeamDisplayNameSchema` are now the single contract across creation, update and projections.

One addition beyond what you listed: bounding the projections is only safe if no writer can persist an out-of-bounds value, and `bootstrapInitialAdmin` still accepted an unbounded team display name from the CLI. It uses the canonical schemas too now, otherwise a bootstrap could seed a row that later 500s every `/me`.

**3. Literal `admin`.** Correct — the server reparses the service result through that schema, so it was the one place the guarantee could regress silently. Now `z.literal("admin")` with a rejection assertion for `member`.

**Verification:** `pnpm check`, `pnpm build`, `pnpm typecheck` pass; shared 56, server 133, web 35; integration 109. The two `daemon-service-*` CLI failures are pre-existing and unrelated — they do not resolve macOS's `/var` → `/private/var` symlink and fail identically on a clean `main` at 12b574d; CI is green on this branch.

yuezengwu (changes requested):

The updated head correctly makes creator-as-admin a literal response contract, applies the input limits to create/update/bootstrap, and enforces membership headroom in create, invitation redemption, and restore. Two blockers remain in the implementation.

1. The new user-row lock introduces an inverse lock order in invitation redemption. `PostAuthenticationService.completeInTransaction` locks the user first and then calls redemption, which locks the Team. The explicit redeem path locks the Team first and then `#requireMembershipHeadroom` locks the user. Two concurrent redemptions for the same user and Team can therefore deadlock (`user -> team` versus `team -> user`), and PostgreSQL will abort one request. Please establish one global user/Team lock order across post-auth redemption, explicit redemption, and restore, and add the requested concurrent join/create or equivalent cross-path coverage. (`packages/server/src/services/auth/post-authentication.ts:37`, `packages/server/src/services/invitations/invitation-service.ts:123`, `packages/server/src/services/teams/team-membership-service.ts:168`)

2. The canonical-name projection contract is still unbounded. `TeamNameInputSchema` caps writer input at 64 characters, but `TeamNameSchema` has no maximum and remains the schema used by `MeMembershipSchema.teamName`, `TeamProfileSchema.name`, and `CreateTeamResponseSchema.name`. Those response projections still accept a 65+ character slug, so the new comment that every projection shares one contract is not true. Put the 64-character invariant in a non-transforming canonical Team-name schema used by projections (with the trimming/lowercasing input schema piping into it), and add overlong-name projection assertions alongside the display-name coverage. (`packages/shared/src/auth.ts:40`, `packages/shared/src/team.ts:14`)

I inspected the incremental fix, related call paths, mergeability, and source assertions without running checks or tests.

baixiaohang (changes requested):
- Rationale: the membership writers and creator-role contract are now aligned, but the new head introduces an inverse user/Team lock order and the Team field rollout is both incomplete and incompatible with values older valid writers could persist.

**Risk level**: B-high
- Path baseline: `packages/server/**` is changed -> B-low.
- Semantic lift: the Web API/data layer and live authentication/membership semantics change -> B-high.

**PR summary**
- Problem: new or membership-less users need to authenticate before explicitly creating or joining a Team instead of receiving an implicit personal Team or becoming stuck.
- Approach: preserve zero-membership sessions, add transactional explicit Team creation with creator-as-admin, route Team-less Web sessions to `/teams/new`, and enforce a shared active-membership ceiling across creation, invitation redemption, and admin restore.
- Impacted modules: Web routing/API, shared Team contracts, server authentication, Team membership, Computer authorization, and integration tests.

**Review findings**

❌ 1. Tightening `TeamDisplayNameSchema` on projections has no compatibility path for values that the old API validly stored. Before this commit, Team profile updates and bootstrap accepted any non-empty display-name length, and the database column is unbounded `text`; now `MeMembershipSchema` and `TeamProfileSchema` reject values over 120. An existing Team with such a value will therefore make `/me` or Team-profile response parsing fail even though this PR has no migration or backfill. Separate normalized write schemas from backward-compatible projection schemas, or include an approved migration/compatibility plan before narrowing responses. [R5 / `packages/shared/src/auth.ts:47`, `packages/server/src/db/schema/auth.ts:25`]

❌ 2. The claimed single Team-name contract is still split: `TeamNameInputSchema` adds the 64-character limit, while `TeamNameSchema` remains regex-only and is still used by `MeMembershipSchema`, `TeamProfileSchema`, and `CreateTeamResponseSchema`. Thus overlong canonical names pass every projection, and the new projection test checks only `displayName`. Define a canonical stored/output name schema that includes the bound and layer input normalization on top of it, while accounting for the legacy-row issue above. [R1 / `packages/shared/src/auth.ts:40`, `packages/shared/src/team.ts:14`]

❌ 3. The new user-row lock creates an inverse lock order across invitation paths. Post-authentication locks the user and then calls redemption, which locks the Team; explicit redemption locks the Team first and then `#requireMembershipHeadroom` locks the user. Concurrent post-auth and explicit redemption for the same user/Team can therefore deadlock (`user -> Team` versus `Team -> user`) and PostgreSQL will abort one request. Establish one user/Team lock order across post-auth redemption, explicit redemption, and restore, and cover the cross-path race with two connections. [R4, R5 / `packages/server/src/services/auth/post-authentication.ts:37`, `packages/server/src/services/invitations/invitation-service.ts:123`, `packages/server/src/services/teams/team-membership-service.ts:168`]

✅ 4. Resolved: the active-membership ceiling now covers creation, invitation redemption, and admin restore through one helper, and `CreateTeamResponseSchema.role` is the literal `admin` with a `member` rejection assertion.

**Action taken**
- Submitted request changes on head `8ff7b0d46645d20ca5e533fbe8dbf34ff2f76fdf`. I inspected the incremental source and test assertions without running checks or tests.

baixiaohang (changes requested):
- Rationale: the new head only extracts the Web Team-creation form for PR-A2 reuse; that refactor preserves the existing behavior, but all three server/shared blockers from the immediately prior review remain unchanged.

**Risk level**: B-high
- Path baseline: `packages/server/**` remains in the PR -> B-low.
- Semantic lift: the Web API/data layer and live authentication/membership semantics change -> B-high.

**PR summary**
- Problem: new or membership-less users need to authenticate before explicitly creating or joining a Team instead of receiving an implicit personal Team or becoming stuck.
- Approach: preserve zero-membership sessions, add transactional explicit Team creation with creator-as-admin, route Team-less Web sessions to `/teams/new`, and expose one reusable creation form for both the standalone page and onboarding.
- Impacted modules: Web routing/API, shared Team contracts, server authentication, Team membership, Computer authorization, and integration tests.

**Review findings**

❌ 1. Still unresolved: the inverse user/Team lock order remains unchanged. Post-authentication locks the user before redemption locks the Team, while explicit redemption locks the Team before membership headroom locks the user, so concurrent paths for the same user/Team can deadlock. Establish one order across post-auth redemption, explicit redemption, and restore, with two-connection coverage. [R4, R5 / `packages/server/src/services/auth/post-authentication.ts:37`, `packages/server/src/services/invitations/invitation-service.ts:123`, `packages/server/src/services/teams/team-membership-service.ts:168`]

❌ 2. Still unresolved: `TeamDisplayNameSchema` now narrows `/me` and Team projections to 120 characters without a migration or compatibility path for longer values that the previous update/bootstrap writers validly stored in the unbounded `text` column. Keep projections backward-compatible or provide the approved data rollout before narrowing them. [R5 / `packages/shared/src/auth.ts:47`, `packages/server/src/db/schema/auth.ts:25`]

❌ 3. Still unresolved: `TeamNameInputSchema` caps writes at 64 characters, but the projection-facing `TeamNameSchema` remains regex-only, so `/me`, Team profile, and Team creation responses still accept overlong canonical names. Put the bound in a non-transforming stored/output schema and layer input normalization on top, accounting for legacy rows as above. [R1 / `packages/shared/src/auth.ts:40`, `packages/shared/src/team.ts:14`]

✅ 4. The incremental `CreateTeamForm` extraction keeps the standalone page's field state, derivation warning, error surface, selected-Team write, and unauthenticated redirect intact; I found no new blocker in that refactor.

**Action taken**
- Submitted request changes on head `9f0b7f7d70cfa281e599c8638737341853562b26`. I inspected the incremental refactor and unchanged blocker paths without running checks or tests.

Gandy2025:
Both blockers fixed in 7cbd091. Thanks — the deadlock was real and mine, introduced by the membership ceiling in the previous head.

**Lock order.** Confirmed exactly as described: post-authentication locks the redeeming user then redeems, the explicit path locked the Team first. One correction worth recording, because it changes where the fix goes: putting the user lock at the top of `joinWithInvitationInTransaction` is **not** sufficient. `redeemInTransaction` calls `lockTeamForMutation` itself, before it ever reaches the membership service, so a lock taken inside `join` never runs early enough. The fix is in `redeemInTransaction`. There is now one order — user, then Team — across post-auth redemption, explicit redemption, creation and restore.

I found that second inversion because the first regression test I wrote was wrong: it drove the explicit path on both sides, so both transactions took the same order and it passed against the bug. The test now mirrors post-auth by holding a transaction between its user lock and its redemption while the explicit path runs on a second connection. Reverting the ordering makes it fail with `PostgresError: deadlock detected` (40P01, `while locking tuple in relation "users"`), so it is a genuine regression test rather than one that always passes.

**Field contracts.** Taking @baixiaohang's first option — separate normalized write schemas from backward-compatible projections. Length now lives on the writers alone: creation, profile update, and the bootstrap CLI, which together are every path that can persist a value. `MeMembershipSchema`, `TeamProfileSchema` and `CreateTeamResponseSchema` stay readable.

@yuezengwu — this does not put the 64-character bound on the name projections as you asked, so here is the reasoning rather than a silent decline. The bound has the same legacy exposure for `name` as for `displayName`: the old `UpdateTeamProfileRequestSchema` accepted unbounded names into an unbounded `text` column, so bounding `TeamNameSchema` would make `/me` unparseable for exactly the rows a migration would be needed to fix. Between the two, a bound on a writer prevents bad data while a bound on a read path only turns a value our own older code accepted into an error on a page the user cannot get past. So the rule is now statable in one line — bounds at every writer, projections validate shape and not length — which also resolves the split contract you flagged, since no writer can produce an out-of-range value any more. If you would rather have the bound on projections, that needs an approved backfill migration and I will open it separately.

The previous head's inaccurate comment claiming projections shared one contract is gone.

**Verification:** `pnpm check`, `pnpm build`, `pnpm typecheck` pass; shared 56, server 133, web 35; integration 110. The two `daemon-service-*` CLI failures remain pre-existing and unrelated (macOS `/var` → `/private/var` symlink; identical on a clean `main` at 12b574d).

baixiaohang (changes requested):
- Rationale: the three substantive server/shared blockers are resolved, but the current head conflicts with newly merged #48 in the Web API, router, and app-test surfaces, so the combined behavior is not yet reviewable.

**Risk level**: B-high
- Path baseline: `packages/server/**` is changed -> B-low.
- Semantic lift: the Web API/data layer and live authentication/membership semantics change -> B-high.

**PR summary**
- Problem: new or membership-less users need to authenticate before explicitly creating or joining a Team instead of receiving an implicit personal Team or becoming stuck.
- Approach: preserve zero-membership sessions, add explicit transactional Team creation with creator-as-admin, enforce a shared active-membership ceiling, route Team-less Web sessions to creation, and expose one creation form for standalone and onboarding use.
- Impacted modules: Web routing/API, shared Team contracts, server authentication, Team membership, Computer authorization, and integration tests.

**Review findings**

❌ 1. Head `7cbd091fb7ef09f1d7bfb2c7ab4187b5b5b3febf` is `CONFLICTING` with current `main` after #48. The conflicts are exactly in `apps/web/src/api.ts`, `apps/web/src/router.tsx`, and `apps/web/src/__tests__/app.test.tsx`: this PR adds Team creation/reusable-form behavior while #48 adds member-role updates, self-authority refresh, and the corresponding mutable test fixture. Reconcile against current main while retaining both API methods, both Settings/creation flows, and both sets of assertions. [R5 / `apps/web/src/api.ts`, `apps/web/src/router.tsx`, `apps/web/src/__tests__/app.test.tsx`]

✅ 2. Resolved: explicit invitation redemption now locks the user before the Team, matching post-authentication and restore; the two-connection regression drives the prior inverse order and checks both redemptions complete without duplicating membership.

✅ 3. Resolved: Team field bounds now consistently protect every writer (creation, profile update, bootstrap) while projections deliberately remain compatible with longer rows accepted by older code; the contract test covers both legacy reads and bounded updates.

✅ 4. Previously resolved items remain intact: the active-membership ceiling covers create/join/restore, creator role is the literal `admin`, Computer access rechecks membership, and the reusable Web form preserves standalone behavior.

**Action taken**
- Submitted request changes on the updated head. I inspected the incremental fix, lock paths, schema usage, legacy assertions, and merge conflict without running checks or tests.

Gandy2025:
Reconciled with current `main` in a7452df — `MERGEABLE` again.

Both conflicts were additive, and nothing from #48 was dropped:

- `api.ts`: `createTeam` and `updateTeamMember` only collided because both were added at the same offset; both are present.
- `app.test.tsx`: the seeded `/me` membership now carries #48's mutable `currentRole` while keeping this branch's Team-less case, so a demotion still reaches the fixture and a zero-Team session still starts with no membership. The invited (#43) and newly created (this PR) memberships continue to append after it. All three PRs' assertions run together — 40 web tests.

**Verification on the merged tree:** `pnpm check`, `pnpm build`, `pnpm typecheck` pass; shared 56, server 133, web 40; integration 110. The two `daemon-service-*` CLI failures remain pre-existing and unrelated (macOS `/var` → `/private/var` symlink; identical on a clean `main`).

yuezengwu (approved):
The merge resolution preserves both sides cleanly: Team creation and zero-membership routing remain intact, while #48's member-role mutation, self-authority refresh, and mutable Web fixtures are retained. The previous lock-order, membership-ceiling, legacy-projection, and creator-role blockers are resolved, and the head merges cleanly with current `main`.

Approved on exact head `a7452df6d7523f8e93168df05d18c79f955c4274`. I inspected the incremental merge resolution and related assertions without running tests.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: all previously identified contract, locking, compatibility, and integration blockers are resolved on the exact reviewed head.

**Risk level**: B-high
- Path baseline: `packages/server/**` is changed -> B-low.
- Semantic lift: the Web API/data layer and live authentication/membership semantics change -> B-high.

**PR summary**
- Problem: new or membership-less users need to authenticate before explicitly creating or joining a Team instead of receiving an implicit personal Team or becoming stuck.
- Approach: preserve zero-membership sessions, add atomic explicit Team creation with creator-as-admin, enforce active-membership headroom under one user-before-Team lock order, route Team-less Web sessions to creation, and reuse the form in onboarding.
- Impacted modules: Web routing/API, shared Team contracts, server authentication, Team membership, Computer authorization, and integration tests.

**Review findings**

✅ 1. The active-membership ceiling now covers create, invitation redemption, and restore under a consistent user-before-Team lock order, with a two-connection deadlock regression.

✅ 2. Team field bounds protect every writer while projections remain compatible with legacy longer rows.

✅ 3. Creator role is the literal `admin`; zero-Team sessions retain resource-level membership gates.

✅ 4. The #43 invitation and #48 member-role flows are preserved alongside Team creation, and the head merges cleanly.

**Action taken**
- Approved exact head `a7452df6d7523f8e93168df05d18c79f955c4274` after explicit B-high authorization. I inspected source and assertions without running checks or tests.

PR #47, opened 2026-08-20
fix: handle leading-hyphen connect codes

baixiaohang:
## Summary

- place login options before Commander's option terminator so one-time connect codes beginning with `-` remain positional arguments
- update server-generated bootstrap commands, Container Smoke, and user-facing command examples
- add parser-level regression coverage and align existing command fixtures with the supported CLI contract

## Testing

- `pnpm build`
- `pnpm check`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`
- targeted CLI, server, shared, and Web tests
- `git diff --check`

Browser E2E was intentionally skipped per the task owner's instruction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,7 +308,7 @@ jobs:
           }
           trap cleanup_runtime EXIT
           OPENTAG_HOME="$runtime_home" node apps/cli/dist/cli/index.mjs \
-            login "$connect_code" --server http://127.0.0.1:18000 --no-start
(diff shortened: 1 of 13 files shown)


yuezengwu (approved):
Reviewed the exact PR head. Moving login options before the option terminator correctly preserves leading-hyphen connect codes as positional arguments, while retaining shell quoting. Parser regression coverage and generated command fixtures are aligned. All current CI checks are green.

PR #48, opened 2026-08-20
feat(web): manage Team member roles

baixiaohang:
## Summary

- add the CSRF-protected browser API adapter for the existing Team member role update endpoint
- show `admin` / `member` roles in Settings → Members, with editable controls for Team admins and read-only roles for members
- reload server-confirmed member roles, refresh `/me` after self-role changes, and preserve the confirmed role when the server rejects an update

## Testing

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

Context Tree: https://github.com/first-tree-ai/first-tree-context/pull/965

### DIFF
--- a/apps/web/src/__tests__/api.test.ts
+++ b/apps/web/src/__tests__/api.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it, vi } from "vitest";
 import { ApiError, BrowserApi } from "../api.js";
 
 const teamId = "d3fda800-7ce2-4338-aae8-3d2120401ed6";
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Reviewed the Team member role management flow. The browser API uses the existing CSRF-protected endpoint, admin/member rendering stays presentation-only while the server remains authoritative, successful changes reload server-confirmed roles, self-role changes refresh /me, and rejected mutations preserve the confirmed value. No blocking code issues found. No database changes in this PR. Tests were not rerun as part of this diff-only review.

PR #39, opened 2026-08-20
Upgrade Web permissions and IM binding domain

baixiaohang:
## Summary

- replace the IM-only `Integration` domain with `ImBinding` across the database, shared contracts, Server, runtime, Client, CLI, and Web, and make new Agents default to `mention_only`
- enforce Team Admin-only Agent and IM mutations while exposing separate member-safe and admin configuration projections
- replace the temporary `/admin` UI with a shared Team App Shell, explicit Computer/Feishu setup flows, and member read-only views
- allow Team Admins to hard-rename the canonical Team handle and display name while keeping the Team UUID stable

## Database and compatibility

- preserves immutable migrations `0000` through `0005` and adds forward migration `0006` for deployed databases whose Drizzle journal already records the historical chain
- renames the existing IM table, enum, references, indexes, and constraints to the final `im_bindings` / `im_binding_id` vocabulary
- does not add dual reads, aliases, row backfills, or compatibility endpoints; staging has no application rows, but its existing migration history remains deployable
- does not add a migration for Team profile edits because the existing `teams.name`, `teams.display_name`, UUID, and uniqueness index already provide the required storage contract

## Logical commits

1. Establish the final `ImBinding` schema and full-stack vocabulary, including `mention_only` defaults.
2. Enforce realtime Team Admin authorization and safe response projections.
3. Add the shared Web App Shell, explicit onboarding/setup actions, and role-aware pages.
4. Add Admin-only Team canonical/display-name edits with UUID-stable hard rename semantics.

## Product boundaries

- Task/thread read models and Agent lifecycle/aggregate status remain deferred.
- No generic Agent Integration marketplace, Resource model, Web Chat, manager transfer, Computer rebind, provider switch, or multi-Team creation is introduced.
- Team canonical-name changes do not retain an alias: the old CLI selector stops resolving immediately; the new name and stable UUID resolve the Team.
- `/admin` and old IM `integration` routes/names are not retained as redirects or aliases.

## Validation

- `pnpm check`
- `pnpm typecheck`
- `pnpm build`
- shared, Client, CLI, and Web package test suites
- Server Agent/Team/Computer/API unit suites and PostgreSQL integration suites for migrations, Agent authorization, Team foundations, and ImBinding behavior
- migration tests cover both a fresh database and a real `0000`-`0005` Drizzle history upgrading through `0006`, including a second idempotent migration run
- concurrency tests cover membership revocation against Agent/IM/Team mutations and consistent Agent-to-ImBinding lock ordering
[...]

PR #32, opened 2026-08-19
feat(client): add Claude Code agent runtime

yuezengwu:
## Summary

- Add a provider-neutral Claude Code Agent Runtime implementation over the documented local CLI stream-json protocol.
- Preserve opaque UUID session bindings across exact create and resume processes.
- Translate streamed model, message, tool, usage, warning, extension, failure, abort, and terminal events into the Agent Runtime contract.
- Enforce fail-closed filesystem, network, approval, and tool policy mappings.
- Reuse the process-tree owner shared with Codex while keeping Provider wire protocols independent.
- Use only the user-installed claude executable resolved from PATH; no Claude or Anthropic package, bundled binary, installer, or download fallback is added.
- Add deterministic process fixtures, exhaustive Provider tests, a live local create/resume smoke test, and update the Agent Runtime test plan.

## Testing

- [x] pnpm check
- [x] pnpm build
- [x] pnpm typecheck
- [x] pnpm --filter @opentag/client test — 243 passed
- [x] pnpm --filter @opentag/client test:agent-runtime:coverage — 139 passed; 100% statements, branches, functions, and lines
- [x] pnpm test — GitHub CI passed the full suite and PostgreSQL integration; local macOS retains two unrelated CLI assertions comparing /tmp with its canonical /private/tmp spelling
- [x] GitHub CI — Checks, Node 22.13.0, Node 26, CLI Pack Smoke, Container Smoke, and aggregate CI passed
- [ ] pnpm --filter @opentag/client test:e2e:claude-code-agent-runtime — installed Claude Code 2.1.210 is present, but the available vendor account is suspended; readiness fails closed with credential_missing and no live model request is made

## Breaking changes

None.

## Checklist

- [x] The change is focused and contains no unrelated work.
- [x] Tests and documentation cover the changed behavior.
- [x] pnpm check, pnpm build, pnpm typecheck, and pnpm test pass in GitHub CI.
- [x] No credentials, private data, or generated build output are included.
- [x] Canonical English docs and Chinese mirrors are synchronized when applicable.

PR #49, opened 2026-08-20
feat(web): make onboarding a real flow

Gandy2025:
## Summary

- derive the current onboarding step from Server facts on every entry
  instead of rendering a fixed checklist, so a refresh, a lost response,
  or a second tab all resume at the same step
- recover an Agent by name when a retried creation hits
  `agents_team_name_active_unique`, so a lost create response cannot
  produce a second Agent
- share one Feishu setup-attempt lifecycle between the Agent IM tab and
  onboarding instead of adding a second implementation
- send a signed-in user with no Team to onboarding instead of the
  "No active Team" dead end, and point the `/teams/new` stub at the same
  destination so Team creation has one entry
- extract the shared page primitives and Team session context out of
  `router.tsx` so onboarding reuses them
- report the read-only member view from Team-scoped Computers and
  Server-observable steps only, rather than from the viewer's own
  Computers and the Admin's browser-local provider confirmation

## Product boundaries

- Team creation needs `POST /api/v1/teams`, which is not on `main` yet.
  The first step explains the gap instead of offering a control that
  cannot work; `TEAM_CREATION_AVAILABLE` in `onboarding-page.tsx` is the
  whole seam.
- The completion state does not deep-link into the IM client. The Server
  serves both Feishu and Lark tenants via `feishuDomainForTeamBrand`, and
  the brand lives on `ImBindingAdminDetail` rather than the summary this
  page reads, so any hardcoded host would be wrong for half the tenants.
- v0.1 covers the Feishu branch only. Slack needs a public callback and
  an app-provisioning decision, so a degraded Slack binding is reported
  and handed to the Agent IM tab rather than retried here.
- No provider selector, because only one provider is runnable **today** --
  this reflects a temporary implementation gap, not a decision that
  OpenTag is Codex-only. `AgentRuntimeProviderSchema` already allows
  `claude-code`, and a complete Claude Code runtime with its own test
  suites landed in #32, but three layers still hardcode Codex:
  `EffectiveRuntimeSnapshotSchema.provider` is `z.literal("codex")`,
  `effective-runtime-snapshot-assembler.ts` throws `UNSUPPORTED_PROVIDER`
  for anything else, and `createClientRuntime` registers a single factory.
  A claude-code Agent is therefore rejected while the Server assembles its
  runtime snapshot, so offering the choice here would be a trap. When those
  layers are widened, this step should offer the selector again -- the
  derivation already models the multi-Computer case the same way and shows
  a chooser only for genuine ambiguity.
- The Codex CLI step is self-attested. `AgentRuntimeProbeResult` already
  separates `artifact_missing` from `credential_missing`, but
  `client-runtime-composition.ts` collapses that into the in-memory
  `imMessageTool` capability bit and no HTTP surface exposes it. The step
  never gates Agent readiness; a missing sign-in surfaces as a failed
  Turn instead. Exposing real readiness is deliberately left to a
  separate change.
- New Agents keep the `mention_only` default from #39: onboarding omits
  `receiveMode` on create.
- The provider confirmation is the one onboarding fact with no Server
  home, so it is remembered per Team in the browser. Every other step is
  re-derived from the Server; losing it only repeats a step that creates
  nothing.

## Attribution

The step-derivation and create-recovery approaches are reimplemented from
equivalent logic in the First Tree Web package. No code, data model, or
product constraint is inherited.

## Coordination with PR-A1

PR-A1 adds `POST /api/v1/teams`. Three things this change touches are likely
to collide with it:

1. `/teams/new` was a placeholder page and is now a redirect to
   `/onboarding`, so a Team-less user has a single creation entry.
   Proposal: onboarding owns the Team-creation UI and A1 supplies only the
   API. If A1 intends to build a real page there, we should settle which
   one stays before either merges.
2. `router.tsx` is 246 lines smaller. `useResource`, `AsyncState`,
   `LoadState`, `Page`, `DefinitionList`, `EmptyState`, `UnavailablePage`,
   `NotFoundPage`, `titleCase`, and `formatDate` moved to `ui.tsx`;
   `TeamSession`, `TeamContext`, `useTeam`, `useTeamSession`,
   `readTeamPreference`, and `SELECTED_TEAM_STORAGE_KEY` moved to
   `team-context.tsx`. Any A1 edit to `router.tsx` will conflict.
3. `AuthenticatedTeamGate` no longer dead-ends a Team-less user on
   "No active Team"; it redirects to `/onboarding`. A1 should not add a
   second redirect for the same state. `useTeam()` now asserts an active
   membership, so Team-less rendering must use `useTeamSession()`.

To finish the first step once A1 lands: add the `TEAMS_TEMPLATE` path and a
`browserApi.createTeam`, confirm `/me` reflects the new membership
immediately, then flip `TEAM_CREATION_AVAILABLE` and fill the matching
branch of `TeamStep`.

Suggested order: merge A2 first, since it does not depend on A1, and let A1
build on the reorganized `router.tsx`. Merging A1 first also works if this
branch rebases. Editing `router.tsx` in parallel does not.

## Testing

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- Web: 90 tests covering every derivation branch, the name-conflict
  recovery path, single-Computer auto-selection, member read-only state,
  and the flow's DOM
- an independent review of this diff found four defects, all fixed here
  and each pinned by a test that fails without its fix: the extracted
  Feishu controller cleared a pending reauthorization even when the
  attempt failed to start, `ImTab` stopped clearing a stale receive-mode
  error so it masked later setup errors, the completion panel hardcoded
  the `mention_only` policy instead of reading `receiveMode`, and the
  member view derived its step from Computers a member cannot see
- shared, Server, and Client suites pass unchanged
- `apps/cli` `daemon-service-shared.test.ts` has 2 pre-existing failures
  on macOS (`/tmp` vs `/private/tmp`) that also fail on an unmodified
  `main`

## Breaking changes

None.

## Checklist

- [x] `pnpm check`, `pnpm build`, and `pnpm typecheck` pass. `pnpm test`
      passes for shared, Server, Client, and Web; the only failures are 2
      pre-existing `apps/cli` cases that also fail on an unmodified `main`.
- [x] No credentials, private data, or generated build output are
      included.
- [x] Canonical English docs and Chinese mirrors are synchronized when
      applicable.


Gandy2025:
From PR-A1 (#44), on `/teams/new` ownership. The repo owner has settled it: **one shared form component, two mount points.** `/teams/new` stays a real page in A1, and onboarding's `team` step mounts the same component rather than growing its own.

I've done A1's half already — [`apps/web/src/create-team-form.tsx`](https://github.com/first-tree-ai/opentag/blob/feat/web-team-creation/apps/web/src/create-team-form.tsx) exports `CreateTeamForm`:

```tsx
<CreateTeamForm
  onCreated={(team) => { /* advance the wizard; team.id / team.name / team.role are yours */ }}
  onUnauthenticated={() => { /* optional */ }}
  submitLabel="Create Team"   // optional
/>
```

It owns the fields, the display-name → canonical-name derivation, the warning when derivation drops non-ASCII characters, and the error surface (including the 409 name collision). The caller owns only what happens next, so `/teams/new` selects the Team and goes to `/agents` while your step can advance its own progress instead. That should let `TeamStep` drop the "not available in this build" placeholder — the `POST /api/v1/teams` your comment is waiting on is in A1.

**Two things to change on this branch:**

1. Drop `<Route path="/teams/new" element={<Navigate replace to="/onboarding" />} />` and let A1's page stand.
2. Your gate line and A1's are the same line, so whoever merges second resolves it by hand. Worth flagging what must not come out of that resolution: A1 currently has `if (!membership) return <Navigate replace to="/teams/new" />`, and combining *that* with *this branch's* `/teams/new → /onboarding` gives an infinite redirect, because `/onboarding` sits inside the gate. Your `location.pathname !== "/onboarding"` guard means this branch alone is fine — the hazard is only in a careless merge. Once (1) lands, either gate destination is safe.

**On the reasoning**, since "零 Team 用户不该有两个建 Team 的入口" is the right instinct but the premise doesn't hold today: a zero-Team user only ever sees one entry. A1's sidebar "Create Team" link lives in `AppShell`, which is inside the gate and never renders while `!membership`. There is exactly one Team-creation UI in the tree, and after this change there still will be — just mounted twice.

The reason it shouldn't move wholesale into onboarding is §5 of the product plan: the page map lists `/teams/new` (创建 Team) and `/onboarding` (首次接入 **Computer、Provider、Agent 和 IM**) as separate pages, and the structure tree puts 创建 Team in the 登录与 Team group next to 切换 Team — i.e. alongside things that serve users who already have a Team. Concretely, A1's acceptance criteria include an existing member creating a second Team; if that UI lives only inside the wizard, they get dragged through "connect a Computer → prepare a Provider → create an Agent" to add a Team. §6.1's 创建或加入 Team → ... is the journey narrative; §5 is where page ownership is stated.

Your point about the progress rail breaking if step 1 lives outside the wizard is the one I couldn't argue with, and mounting the shared component inside your step is what answers it.

baixiaohang (changes requested):
- Rationale: The server-derived flow is a strong direction, but the current head can announce an unusable IM handoff as ready, associates provider confirmation with the wrong resource lifetime, and has unsafe recovery/provider branches.

**Risk level**: B-low
- Path baseline: only `apps/web/**` changes -> C.
- Semantic lift: the Web change composes Team, Computer, Agent, and IM API state and changes protected routing/data flow -> B-low.

**PR summary**
- Problem: a signed-in user should be able to enter or resume setup after a refresh, lost response, or second tab, while members see the same Team progress without mutation controls.
- Approach: derive a conditional onboarding step from current API facts, recover retried Agent creation, share the Feishu attempt controller with the Agent IM tab, and route Team-less sessions into onboarding.
- Impacted modules: Web routing and Team session context, onboarding state derivation/UI, Agent creation recovery, and Feishu setup presentation.

**Review findings**

❌ 1. `bindingState === "active"` is treated as sufficient for the `ready` step, and `ReadyStep` then tells the user to start sending work in IM. That status proves the stored binding is active, not that the handoff is usable: the existing `imBindingDiagnostics` response has a separate `ready` fact that also requires the runtime message tool and, for Feishu, a live connection. With an offline daemon/missing message tool or an expired Feishu connection lease, this flow still announces “Reviewer is ready” and sends the user to `@` a bot that cannot complete the handoff. Derive the completion claim from the Server readiness fact (or remain explicitly non-ready where that fact is not readable) rather than reinterpreting binding status. [R3, R5 / `apps/web/src/onboarding/flow.ts:94`, `apps/web/src/onboarding/flow.ts:136`, `apps/web/src/onboarding/onboarding-page.tsx:384`]

❌ 2. The provider acknowledgement is stored per Team even though it describes one Computer, and the flow asks for it before resolving a multi-Computer choice. With two online Computers, an admin can confirm Codex on Computer A and then select Computer B for the Agent; a later change in the online set also reuses the same Team-wide flag. The new Agent can therefore be bound to a Computer on which the acknowledged prerequisite was never checked. Select the Computer before this step and scope the acknowledgement to that concrete Computer (or combine the confirmation with the selected-Computer action). [R3 / `apps/web/src/onboarding/onboarding-page.tsx:240`, `apps/web/src/onboarding/onboarding-page.tsx:281`, `apps/web/src/onboarding/onboarding-page.tsx:424`]

❌ 3. A 409 plus matching name/manager does not prove that the existing Agent came from a lost response to this intent. A stale or parallel tab can create the same slug on a different Computer (or with a different display/config), after which this submission silently “recovers” that unrelated configuration and advances. Use a server idempotency identity for the creation intent, or at minimum verify every intent-defining field and only recover after this client has an ambiguous prior submission; an ordinary first-attempt name conflict must remain a conflict. [R4 / `apps/web/src/onboarding/recover-created-agent.ts:33`, `apps/web/src/onboarding/recover-created-agent.ts:41`]

❌ 4. Every active binding reaches the same Feishu completion UI. An already-active Slack binding is classified as `ready` and told to open Feishu, while a Feishu-provider binding may belong to a Lark tenant whose brand is unavailable in this summary. The degraded-Slack branch handles provider asymmetry, but the active branch drops it. Branch completion/presentation on authoritative provider and brand facts, or hand unsupported/unknown presentation back to the Agent IM surface instead of issuing provider-wrong instructions. [R5 / `apps/web/src/onboarding/flow.ts:92`, `apps/web/src/onboarding/onboarding-page.tsx:384`]

❌ 5. The settled A1/A2 integration contract is not represented in this head: `/teams/new` is still redirected away and the onboarding Team step remains a placeholder. Per the current coordination comment, `/teams/new` must remain A1's real page while onboarding mounts the same exported `CreateTeamForm`; resolving the overlapping gate without that shape can also produce the documented `/teams/new` ↔ `/onboarding` redirect loop. Rebase/integrate the shared form and preserve both mount points before merge. [R2, R5 / `apps/web/src/router.tsx:37`, `apps/web/src/onboarding/onboarding-page.tsx:166`]

**Action taken**
- Submitted request changes on head `d12123bf9145ca104242c3371a4bf921d2e486f2`. I inspected the diff, related Server contracts, Context Tree decisions, current PR state, and A1 coordination without running checks or tests.

yuezengwu (changes requested):
Requesting changes for two onboarding blockers:

1. `apps/web/src/onboarding/flow.ts:136` treats `bindingState === "active"` as sufficient to enter the final “Start working in Feishu” state. That field is only the stored binding status; it does not prove either a fresh Feishu connection or the current Computer's OpenTag message-tool capability. The existing `imBindingDiagnostics` API already computes `ready` from all three facts. As written, an active binding with a disconnected Channel or missing runtime tool is presented as ready and directs the user to send work that cannot be handled. Please derive the completion state from actual readiness (with a safe server projection for read-only members if necessary) and cover the active-but-not-ready cases.

2. `apps/web/src/onboarding/onboarding-page.tsx:240-263,424-433` records the Codex self-attestation once per Team before the Computer is selected. In the multi-Computer branch, the user confirms Codex for an unspecified/other machine and can then choose any online Computer in `AgentStep`; later visits also skip the provider step for every Computer in that Team. This allows onboarding to create an Agent on a Computer where Codex is not installed or signed in. Please select/bind the Computer before this confirmation and scope the acknowledgement to that Computer (or otherwise carry the confirmed Computer into Agent creation), with coverage for two online Computers.

Separately, GitHub currently reports this PR as conflicting with `main`, so the branch also needs its conflicts resolved before merge.

baixiaohang (changes requested):
- Rationale: The shared Team form integration is resolved, but three original blockers remain and the multi-Computer acknowledgement still allows a confirmation made for one selection to be recorded against another.

**Risk level**: B-low
- Path baseline: the PR changes `apps/web/**` -> C.
- Semantic lift: onboarding composes Team, Computer, Agent, and IM API state and changes protected routing/data flow -> B-low.

**PR summary**
- Problem: users need one resumable first-run flow derived from current Server facts, while Team members can observe progress without mutation controls.
- Approach: compose Team creation, Computer/provider preparation, Agent creation, and IM setup in onboarding; reuse the standalone Team form and Feishu controller; recover selected mutations after uncertain responses.
- Impacted modules: Web routing/session context, Team creation, onboarding derivation/UI, Agent recovery, and Feishu setup.

**Review findings**

❌ 1. Still blocking: an `active` binding still enters `ready` without consulting `imBindingDiagnostics.ready`, so an unavailable runtime message tool or disconnected Feishu Channel is presented as a usable handoff. The Server already owns the combined readiness fact; completion must use it or remain conservative where it is not readable. [R3, R5 / `apps/web/src/onboarding/flow.ts:94`, `apps/web/src/onboarding/flow.ts:143`, `apps/web/src/onboarding/onboarding-page.tsx:394`]

❌ 2. Partially resolved, but the multi-Computer confirmation is not bound to the selection at the time it is made. The checkbox is uncontrolled and is not reset when the `<select>` changes. A user can check it while Computer A is selected, switch to Computer B, and submit; `submit` then records B as acknowledged even though the confirmation was made for A. The new test performs exactly that order (check first, switch second) and treats the false acknowledgement as success. Track the confirmed Computer id and invalidate/reconfirm on selection change, or make selection plus confirmation atomic. [R3 / `apps/web/src/onboarding/onboarding-page.tsx:284`, `apps/web/src/onboarding/onboarding-page.tsx:326`, `apps/web/src/onboarding/__tests__/onboarding-page.test.tsx:426`]

❌ 3. Still blocking: any 409 with matching name/manager is silently treated as this browser's lost create response. A stale or parallel intent for the same slug but a different Computer/display/config can therefore be adopted and advanced. Use a Server idempotency identity, or only recover an exact intended result after this client has an ambiguous prior submission. [R4 / `apps/web/src/onboarding/recover-created-agent.ts:33`, `apps/web/src/onboarding/recover-created-agent.ts:41`]

❌ 4. Still blocking: every active binding reaches Feishu-specific completion. An active Slack binding, or a Lark tenant represented by provider `feishu`, receives instructions to open Feishu. Completion copy/navigation needs authoritative provider and brand facts, or must hand unsupported presentation back to the Agent IM surface. [R5 / `apps/web/src/onboarding/flow.ts:94`, `apps/web/src/onboarding/onboarding-page.tsx:394`]

✅ 5. Resolved: Team creation now has one `CreateTeamForm` implementation mounted by both `/teams/new` and onboarding, with the standalone route preserved and Team-less onboarding advancing after `/me` refresh.

**Action taken**
- Submitted request changes on head `b37f2ed8123dbaa27ed1cc8dbfe54bd65c12b6f5`. I inspected the incremental merge resolution, affected tests and source contracts without running the test suite. GitHub still reports this head as conflicting with the current `main` (`4c9577f`).

baixiaohang (changes requested):
- Rationale: The branch now integrates cleanly with `main`, but the latest merge does not address the four remaining onboarding correctness blockers.

**Risk level**: B-low
- Path baseline: the PR changes `apps/web/**` -> C.
- Semantic lift: onboarding composes Team, Computer, Agent, and IM API state and changes protected routing/data flow -> B-low.

**PR summary**
- Problem: users need one resumable first-run flow derived from current Server facts, while Team members can observe progress without mutation controls.
- Approach: compose Team creation, Computer/provider preparation, Agent creation, and IM setup in onboarding; reuse the standalone Team form and Feishu controller; recover selected mutations after uncertain responses.
- Impacted modules: Web routing/session context, Team creation, onboarding derivation/UI, Agent recovery, and Feishu setup.

**Review findings**

❌ 1. Still blocking: an `active` binding enters `ready` without consulting `imBindingDiagnostics.ready`, so an unavailable runtime message tool or disconnected Feishu Channel is presented as a usable handoff. Completion must use the Server-owned combined readiness fact or remain conservative where it is not readable. [R3, R5 / `apps/web/src/onboarding/flow.ts:94`, `apps/web/src/onboarding/flow.ts:143`, `apps/web/src/onboarding/onboarding-page.tsx:398`]

❌ 2. Still blocking: the multi-Computer checkbox remains bound to no particular selection. A user can confirm while Computer A is selected, switch to B, and submit; the code then records B as acknowledged. The test at lines 426-440 still encodes that false-positive sequence. Reset/reconfirm when the selection changes, or track the Computer id that was confirmed. [R3 / `apps/web/src/onboarding/onboarding-page.tsx:288`, `apps/web/src/onboarding/onboarding-page.tsx:330`, `apps/web/src/onboarding/__tests__/onboarding-page.test.tsx:426`]

❌ 3. Still blocking: any 409 with matching name/manager is silently treated as this browser's lost create response. A stale or parallel intent for the same slug but a different Computer/display/config can therefore be adopted and advanced. Use a Server idempotency identity, or only recover an exact intended result after this client has an ambiguous prior submission. [R4 / `apps/web/src/onboarding/recover-created-agent.ts:33`, `apps/web/src/onboarding/recover-created-agent.ts:41`]

❌ 4. Still blocking: every active binding reaches Feishu-specific completion. An active Slack binding, or a Lark tenant represented by provider `feishu`, receives instructions to open Feishu. Completion copy/navigation needs authoritative provider and brand facts, or must hand unsupported presentation back to the Agent IM surface. [R5 / `apps/web/src/onboarding/flow.ts:94`, `apps/web/src/onboarding/onboarding-page.tsx:398`]

✅ 5. Resolved: head `844fad2` incorporates the current `main`, and GitHub now reports the PR as mergeable rather than conflicting. The shared Team form integration remains intact.

**Action taken**
- Submitted request changes on head `844fad263466dba97aa8ec4b977f593acecf3c14`. I inspected the incremental merge and the affected source/tests without running checks or tests.

Gandy2025:
Superseded by a smaller, contract-first onboarding PR series.

The first replacement is #58, which moves lost-response and parallel-tab recovery into Server-owned Agent creation idempotency instead of recovering by name in Web code. Separate focused PRs will cover Provider readiness, the pure onboarding state model, authoritative IM handoff readiness, and the final Web flow.

The useful product cases and review findings from this PR remain reference material; its branch will not be merged.

PR #51, opened 2026-08-20
fix: repair IM runtime bootstrap contracts

baixiaohang:
## Summary

- preserve the manager's absolute user `PATH` entries in systemd and launchd service definitions while keeping the resolved OpenTag and Node entry points first; existing service definitions are reconciled and restarted when they drift
- move managed `AGENTS.md` instructions into the actual Codex working directory and migrate workspace state from v1 to v2 with hash validation, crash-safe re-entry, trusted legacy cleanup, and fail-closed conflict handling
- make the Client Runtime own outbound message request UUIDs, expose only `retryRequestId` for explicit retries of `unknown` results, and preserve request identity across timeout and transport uncertainty
- reuse the shared 16 KiB outbound text limit so invalid tool input is rejected before dispatch

## Compatibility and boundaries

- Client-to-Server wire schemas, Server APIs, database schemas, provider adapters, Feishu setup, and Web binding behavior are unchanged.
- Existing workspace state migrates automatically from v1 to v2. A user-modified or conflicting managed instruction file fails closed instead of being overwritten.
- Model-facing `opentag_message_send`, `opentag_message_reply`, and `opentag_message_react` calls no longer accept `requestId`; new request identities are generated by the Runtime. `retryRequestId` is optional and reserved for an explicit retry of the same payload after an `unknown` result.

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

## Post-release verification

The staging incident currently has a local systemd PATH drop-in and a temporary managed instruction compatibility file. They must remain until this change is released to staging. After release, remove both workarounds and run one real Feishu inbound/outbound acceptance check.

### DIFF
--- a/apps/cli/src/__tests__/daemon-service-renderers.test.ts
+++ b/apps/cli/src/__tests__/daemon-service-renderers.test.ts
@@ -161,6 +161,45 @@ describe("systemd service backend", () => {
     );
     expect(unit).toContain(`ExecStart="${installedCli}" "daemon" "service-run"`);
     expect(unit).not.toContain(staleCli);
+    expect(unit).toContain(staleBin);
+    expect(unit.indexOf(installedBin)).toBeLessThan(unit.indexOf(staleBin));
+  });
+
+  it("detects and reconciles an active service whose PATH predates user-directory inheritance", async () => {
+    const userHome = await temporaryDirectory("opentag-systemd-path-drift-");
+    const home = await temporaryDirectory("opentag-systemd-home-");
+    const invocation = { args: [], program: "/opt/opentag/bin/opentag" };
+    const unitPath = join(userHome, ".config", "systemd", "user", "opentag.service");
+    await writeFileWithParents(
+      unitPath,
+      renderSystemdUnit({
+        home,
+        invocation,
+        path: buildServicePath(invocation, "linux"),
+        serviceId: "opentag",
+      }),
+    );
+    const runner = fakeRunner((program, args) => {
+      if (program === "loginctl" || args.includes("show-environment")) return result(0, "", "");
+      if (args.includes("is-active")) return result(0, "active", "");
+      if (args.includes("MainPID")) return result(0, "321", "");
+      return result(0, "", "");
+    });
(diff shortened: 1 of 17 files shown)


yuezengwu (commented):
Blocker — `packages/client/src/runtime/runtime-tool-host.ts:300-333`

`retryRequestId` is currently accepted whenever it is a syntactically valid UUID. The Client does not verify that the ID came from a prior `unknown` result, and once the pending request settles or times out its payload hash is discarded. As a result, a model can still choose an arbitrary outbound request identity, or reuse an unknown request ID with a different payload. In the important case where the first send never reached the Server, that changed retry becomes the first request stored under the ID and can execute a different side effect.

This violates the PR's core contract that the Client Runtime owns new UUIDs and exposes an ID only for an explicit retry of the exact same payload after `unknown`. Please retain a bounded Run-scoped record of retryable unknown identities and their canonical intent hash, reject `retryRequestId` values not in that record, and reject any payload mismatch before dispatch. Add coverage for an arbitrary retry UUID and for a changed payload after timeout/`unknown`.

baixiaohang:
Fixed in `6354165`.

- Added a bounded (64-entry), Run-scoped ledger of retryable `unknown` request identities keyed by the canonical wire intent hash.
- `retryRequestId` now fails before dispatch unless it belongs to the active Run's unknown ledger and the complete intent matches exactly.
- Consolidated ledger ownership into request settlement: definite Server results synchronously revoke retryability and win same-tick caller-abort races; abort-first paths remain unknown until a late definite result revokes them.
- Added coverage for arbitrary retry UUIDs, changed payloads after Server unknown and timeout, bounded eviction, shared pending requests, and both result-before-abort and abort-before-result races for succeeded/deterministic results.

Validation passed: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm test:coverage`; Client agent-runtime coverage is 210/210 tests with 100% statements, branches, functions, and lines.

PR #56, opened 2026-08-20
feat: support existing Feishu bot authorization

baixiaohang:
## Summary

- allow Feishu create and replace setup flows to select an existing Bot/App or create a new one, while reauthorization remains pinned to the current App
- define one shared 66-scope tenant permission contract and use it for registration, activation, readiness, diagnostics, and reauthorization decisions
- keep legacy Bots online while expanded permissions are pending, preserve same-App bindings and Sessions, and expose stable errors for occupied Apps or identity mismatches
- update the Web setup and recovery copy with an allowlisted, redacted domain-error mapping

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/server exec vitest run src/__tests__/integration/im-binding.test.ts --maxWorkers=1` (59/59)
- `pnpm --filter @opentag/web exec vitest run --maxWorkers=1` (34/34)
- `pnpm --filter @opentag/client test:agent-runtime:coverage` (251/251, 100%)
- `pnpm test:coverage` (84 files, 692 tests)
- `git diff --check`

No database schema, migration, backfill, manual secret input, or non-message Feishu tool-chain changes are included.


yuezengwu (changes requested):
P1 — Preserve the mention_only ingress boundary. packages/shared/src/im-binding.ts:12 now puts im:message.group_msg in the unconditional tenant scope contract, and DefaultFeishuRegistrationGateway no longer uses receiveMode when constructing requested scopes. Before this change, that scope was added only for all_message. With the receive event subscribed, the broader grant allows ordinary group callbacks to reach OpenTag; ImMessageInbox persists each inbound event before it checks receiveMode to decide delivery, so a mention_only Agent can now store ambient group messages it was configured not to receive. Please split always-required tool/message scopes from receive-mode ingress scopes, omit im:message.group_msg for mention_only, and make activation/readiness checks mode-aware. Add a regression test proving the mention_only registration grant excludes the all-group-message scope.

baixiaohang:
@yuezengwu Thanks for calling out the boundary. The product owner has now explicitly clarified that `im:message.group_msg` must be requested unconditionally, including for `mention_only` Agents.

In the approved model, `receiveMode` gates Agent delivery/Turn triggering; it is not a provider-authorization or durable-history boundary. Persisting ambient group callbacks as server-authoritative `ImMessage` history before suppressing Agent delivery is therefore intentional. The existing integration coverage (`includes unmentioned messages in mention-only history and stale-reply fencing`) exercises that behavior.

I kept the runtime behavior unchanged and pushed `36803fe` to make this contract explicit:

- the registration code now documents that `receiveMode` does not narrow the Bot authority ceiling or durable `ImMessage` history;
- the `mention_only` registration regression now explicitly asserts that `im:message.group_msg` is included.

Could you please reassess the change request against this clarified product contract?

yuezengwu (approved):
Reassessed at exact head 36803fe01f7167e6ba16411a9e8a3903cce129e3. The clarified product contract establishes receiveMode as the Agent delivery/Turn gate rather than a provider-authorization or durable-history boundary, so the unconditional group-message scope is intentional. The follow-up change documents that boundary and locks it with a focused assertion; all checks are green. My previous change request is resolved.

PR #61, opened 2026-08-20
feat(runtime): project Codex readiness by Computer

Gandy2025:
## Summary

- publish a bounded Codex readiness snapshot on owned and Team Computer projections
- map the existing production Codex probe to checking, install, sign-in, ready, or unavailable
- negotiate the additive HTTP and runtime fields so old/new Client and Server versions remain mutually compatible
- keep freshness and current-instance authority in the in-memory ConnectionRegistry; offline Computers fail closed without a migration

## Semantics

- `ready` is the only runnable status
- an online Computer with no fresh observation is `checking`
- an offline Computer is `unavailable`, regardless of its last observation
- an established `ready` result is leased while its periodic reprobe remains in flight, matching local admission; the completed probe atomically replaces it
- an accepted heartbeat that omits an expired observation clears the Server snapshot

## Current provider boundary

This PR intentionally reports Codex only because it is the only Provider currently registered and admissible in the production Client Runtime. Although a Claude Code factory probe exists, `createClientRuntime` does not register it and `EffectiveRuntimeSnapshotAssembler` currently rejects `claude-code`; reporting it as runnable here would be false. Wiring Claude Code through the production execution policy/snapshot boundary is a separate follow-up.

## Verification

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- focused shared/client/server/web tests, including mixed-version rollout, negotiated heartbeat beyond TTL, rejected heartbeat visibility, and stale/offline projection
- `pnpm test` was also run; readiness suites pass, while the existing macOS `/tmp` vs `/private/tmp` path assertions fail in unrelated daemon/workspace tests


baixiaohang (changes requested):
- Rationale: The readiness model is coherent inside the new code, but the unchanged strict v1 boundaries make the rollout incompatible and the local admission guard can disagree with the published runnable state.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` -> A
- Semantic lift: none; A is already the highest level

**PR summary**
- Problem: Operators and Team members cannot currently tell whether an online Computer can run Codex, needs installation or sign-in, is still checking, or is unavailable.
- Approach: Probe Codex on the Client, send a bounded observation during registration and heartbeats, retain only the current instance's fresh observation in the Server registry, and project one normalized status on owned and Team Computer APIs.
- Impacted modules: shared v1 contracts, Client provider registry and connection, Server runtime registry/session, Computer and Team projections, CLI/Web API consumers

**Review findings**
❌ 1. The unchanged strict v1 contracts cannot be rolled between the existing `main` client and server. A new Server adds `providerReadiness` to every `/api/v1/me/computers` item, which the existing strict `ComputerSchema` rejects as an unknown key; in the other order, the new daemon unconditionally sends the field in v1 register/heartbeat frames, which the existing strict Server schemas reject. Please provide an explicit compatibility/version-negotiation path and cover both old-client/new-server and new-client/old-server pairs. [R5 / `packages/shared/src/computer.ts:24`, `packages/client/src/runtime/runtime-connection.ts:520`]
❌ 2. A periodic refresh publishes `checking` before probing, but `AgentRuntimeProviderRegistry` retains the Provider in `#ready` until a failed probe settles. During a slow reprobe, `ensureReady()` returns immediately and new work remains admissible even though the published contract says only `ready` is runnable. Revoke local readiness when entering `checking` (or keep the projection `ready` until admission is actually revoked), and add an in-flight reprobe regression case. [R1, R4 / `packages/client/src/runtime/client-runtime-composition.ts:191`, `packages/client/src/runtime/agent-runtime-provider-registry.ts:68`]
⚠️ 3. The heartbeat mutates the in-memory authority/readiness snapshot before the durable heartbeat guard accepts the current instance. If the DB update returns false or throws, concurrent API reads can briefly expose a rejected observation until socket close cleanup runs. Apply the registry mutation only after the durable guard succeeds, while still fencing the exact socket/instance. [R4 / `packages/server/src/runtime/runtime-session.ts:290`]

**Action taken**
- Submitted request changes on head `878c84a92b129586575734bba20ad8e851d631a9`.

yuezengwu (changes requested):
The readiness projection does not revoke an observation when the client stops sending it. RuntimeConnection intentionally omits providerReadiness after its local TTL expires, but ConnectionRegistry.touch() only updates the stored snapshot when the optional field is present; an omitted heartbeat therefore leaves the previous ready/install/sign-in observation in place until the server-side TTL elapses from the last heartbeat that carried it. Because every preceding heartbeat re-stamps the same observation, the API can keep presenting a locally expired result for nearly another full TTL while the Computer is still online. Please clear the stored readiness when an accepted heartbeat omits providerReadiness (or carry an explicit observation timestamp/expiry and honor it), and add a regression test that advances past the client validity window, sends an omission, and projects checking rather than the old status.

Gandy2025:
Follow-up `9633d4f` addresses all requested changes:

- Adds explicit `x-opentag-provider-readiness: 1` opt-in for HTTP and WebSocket upgrade. HTTP projections include readiness only for opted-in consumers; new response schemas accept its absence from old Servers. WebSocket welcome acknowledges the opt-in before the daemon includes readiness in register/heartbeat frames.
- Keeps last-known `ready` during an in-flight periodic probe, matching local admission and avoiding probe-induced work churn; the completed probe updates both atomically.
- Applies the durable heartbeat guard before mutating `ConnectionRegistry`, with current socket/instance fences on both sides of the await.
- Clears stored readiness when an accepted negotiated heartbeat omits an expired local observation.

Regression coverage includes old-client/new-server, new-client/old-server, in-flight reprobe, rejected heartbeat visibility, and expired observation omission.

baixiaohang (changes requested):
- Rationale: The rollout negotiation and Server-side authority fixes are sound, but a slow periodic probe can still make the published status `checking` while local admission remains runnable.

**Risk level**: A
- Path baseline: includes `packages/client/**` and `apps/cli/**` -> A
- Semantic lift: none; A is already the highest level

**PR summary**
- Problem: Operators and Team members need a trustworthy per-Computer view of whether Codex is runnable, checking, needs installation/sign-in, or is unavailable.
- Approach: Negotiate the new projection independently on HTTP and WebSocket boundaries, report the Client probe result, retain only current-instance fresh observations, and fail closed for offline or missing observations.
- Impacted modules: shared v1 contracts, Client HTTP/runtime connections, Provider readiness composition, Server runtime registry/session, Computer and Team API projections

**Review findings**
❌ 1. The in-flight reprobe alignment still expires for a sufficiently slow probe. `refreshCapability()` renews the published `ready` observation only once at probe start, with the fixed 60-second Client TTL, while `AgentRuntimeProviderRegistry` keeps `#ready` true until the probe settles. The production Codex probe can exceed that TTL because local checks precede an App Server request whose timeout is itself 60 seconds, and a custom probe may wait longer. Once a negotiated heartbeat runs after the observation expires, it omits readiness, the Server now correctly clears the snapshot, and the API projects `checking` even though `ensureReady()` still admits work. The added regression asserts the immediate `ready` update but neither advances beyond the TTL nor observes a negotiated heartbeat. Please make the published readiness and local admission share the same lifetime (or renew the last-known observation while the probe owns readiness), and cover a negotiated reprobe that remains in flight past the observation TTL. [R1, R4 / `packages/client/src/runtime/client-runtime-composition.ts:191`, `packages/client/src/runtime/runtime-connection.ts:212`, `packages/client/src/providers/codex/app-server-wire.ts:5`]
✅ 2. HTTP opt-in plus WebSocket welcome acknowledgement now preserves both old-client/new-server and new-client/old-server compatibility.
✅ 3. The durable heartbeat guard now precedes the exact-instance registry mutation, and an accepted omission clears the expired Server observation.

**Action taken**
- Submitted request changes on head `9633d4f4a3fcda5c03e2bbc331cc5c85309b142f`.

Gandy2025:
Follow-up `170e96d` addresses the remaining slow-reprobe finding. `RuntimeConnection` now supports an explicitly scoped readiness lease; `createClientRuntime` owns that lease only while a periodic probe starts from an admitted `ready` state, then replaces the observation with the completed probe result before releasing it. This keeps negotiated heartbeats publishing `ready` beyond the ordinary TTL exactly while local admission remains ready, without lengthening the TTL or introducing a timer.

The regression advances the Client clock beyond `RUNTIME_CLIENT_CAPABILITY_TTL_MS`, observes a negotiated heartbeat still carrying `ready` under the lease, releases the lease, and verifies the next heartbeat omits the expired observation. Focused tests, `pnpm check`, `pnpm typecheck`, and `pnpm build` pass.

Scope note: the PR remains truthfully Codex-only. Claude Code has a probe but is not production-admissible (`createClientRuntime` does not register it and the Server snapshot assembler rejects it), so production Claude composition is tracked as separate follow-up work rather than being mislabeled runnable here.

Gandy2025:
Tracking the runtime contract and the keep/split/replace decision in #64, assigned to @yuezengwu. The issue captures the truthful Codex/Claude execution boundary, rollout compatibility, current-instance authority, omitted-observation expiry, and the slow-reprobe admission/readiness invariant. Please treat #64 as the owning issue; this PR remains open but should not merge until that direction is settled.

yuezengwu (approved):
Re-reviewed exact head 170e96d865427fd3afa3d0e09b49a920c23d24a8. The negotiated HTTP/WebSocket rollout, accepted-heartbeat fencing, omitted-readiness revocation, and cross-TTL in-flight reprobe lease resolve my prior finding and preserve current-instance fresh authority. Local check/build/typecheck and affected client/server/Web/CLI tests pass; the only full-suite failures are the pre-existing macOS /tmp canonicalization assertions disclosed in the PR.

yuezengwu (changes requested):
Withdrawing my approval on exact head 170e96d865427fd3afa3d0e09b49a920c23d24a8 after the new owning contract issue #64 explicitly made keep/split/replace a merge gate. Recommendation: Keep #61. The Codex-only projection is truthful because Codex is the only production-admissible Provider; Claude Code production enablement should remain separate, consistent with the runtime contract that conformance alone does not enable a Provider. The current seam and prior code findings remain sound, but #64 acceptance is not yet fully demonstrated: (1) the shutdown test proves the in-flight probe aborts but does not directly assert the readiness lease is released on cancellation/stop; (2) replacement tests do not carry an old-instance readiness observation and prove it becomes unobservable after the new instance wins; and (3) the monitor guard test does not assert that concurrent timer/run pressure starts only one refresh and one readiness lease. Please add deterministic lifecycle assertions for cancellation/stop, connection replacement, and concurrent refresh/lease ownership, document the Keep decision in #64, then request exact-final-head review. Do not merge before that contract gate is settled.

PR #63, opened 2026-08-20
feat(account): support self-service display names

baixiaohang:
## Summary

- add a shared, strict user display-name contract and an authenticated, CSRF-protected `PATCH /api/v1/me` self-profile mutation
- preserve user-maintained display names on returning provider sign-ins while continuing to synchronize provider email addresses
- add `Settings → Account` with a read-only email, editable display name, duplicate-submit protection, server-value rollback, and `/me` refresh after save
- align initial-admin bootstrap with the same 255-character display-name boundary and cover global projections across Teams, Computers, and Agent managers

No database schema or migration changes are included.

## Testing

- `pnpm check`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test`
- `pnpm test:coverage`
- `pnpm --filter @opentag/server exec vitest run src/__tests__/integration/account-team-foundation.test.ts --maxWorkers=1`

## Context Tree

- [Self-service display identity design (Draft PR #966)](https://github.com/first-tree-ai/first-tree-context/pull/966)


yuezengwu (changes requested):
Requesting changes on 32d135bb195a83d6b520cb99ffca67c222da6943.

[P1] Keep legacy display names readable while enforcing the new writer bound. MeResponseSchema now uses UserProfileSchema, whose displayName is capped at 255 characters. Existing users can already have longer values because the users column is text and the previous bootstrap writer had no upper bound; this PR includes no migration. For such an account, GET /api/v1/me fails response parsing before the Web app can render the new self-service editor. I reproduced this directly: MeResponseSchema.safeParse returns too_big at user.displayName for a 256-character legacy value. Please keep the read projection tolerant (for example, min(1) only) while applying the 255-character rule to UpdateUserProfileRequestSchema and other writers, and add a legacy-value regression test. A data repair/migration would require the explicit database-change approval required by repository policy.

The current head is also merge-conflicting with main in the Web API/router test surface. Please rebase or merge current main and resolve those conflicts before the next review.

baixiaohang:
Updated the branch by merging current `origin/main` (`a45bcbb`) and resolving the Web router conflict. The Account editor remains under `Settings → Account`, while the newer account-menu and onboarding/navigation behavior from `main` is preserved.

The product owner has explicitly confirmed that this deployment has no legacy display names over 255 characters. Accordingly, this PR intentionally keeps the shared profile read and all writers on the same trimmed, non-empty, 255-character boundary; it does not add legacy read tolerance or a database migration.

Validation on the merged head:

- Web API/app/onboarding tests: 78 passed
- Shared tests: 61 passed
- Server auth/bootstrap tests: 10 passed
- Account/Team integration tests: 18 passed
- Merge-sensitive Agent API and Feishu registration tests: 16 passed
- `pnpm check`, `pnpm build`, and `pnpm typecheck` passed

Please re-review the updated head.

yuezengwu (approved):
Approved fresh head 74041c9daf4e331518170fb6eda645b7a20a1214 after a full re-review.

The previous legacy-read concern is superseded by the explicit product-owner confirmation that this deployment has no display names over 255 characters. Under that confirmed product premise, keeping the profile read and every writer on the same trimmed, non-empty, 255-character contract is acceptable without a migration.

I verified that a45bcbb is an ancestor, the merge conflicts are resolved, and the final diff preserves the current Account menu, Team picker, Team-less onboarding, navigation, and Agent behavior while adding the Settings to Account editor. Local validation passed: pnpm check, pnpm build, pnpm typecheck, Web 78/78, shared 61/61, and merge-sensitive server 26/26. All six GitHub checks are green on this exact head.

PR #66, opened 2026-08-20
ci: deploy main to CapRover staging from the GHCR commit image

bestony:
## Summary

Staging now follows `main` automatically, deployed to CapRover from the container image the `Docker` workflow already
publishes to GHCR. Nothing is built on the deployment path and no source tarball is uploaded; a deployment is a pointer
change to an immutable image.

`.github/workflows/deploy-staging.yml`:

- **Triggers on successful `main` CI**, not on the `Docker` workflow. CI is the gate that proves a revision passed its
  checks; the image it needs is produced by the `Docker` workflow running alongside it, so this workflow waits for that
  coordinate via the existing `scripts/registry-coordinate.mjs --wait` instead of building anything.
- **Deploys the immutable per-commit coordinate** `ghcr.io/first-tree-ai/opentag:<commit-sha>`, never `edge` or
  `latest`. A moving tag would leave CapRover with an unchanged image reference and nothing to roll forward to, and it
  would make the running revision unidentifiable.
- **Never moves staging backwards.** Runs for consecutive commits can overlap, so an automatic run whose revision is no
  longer the tip of `main` skips and records that in its job summary rather than overwriting a newer revision. Every
  revision must also be proven to belong to `main` history.
- **`workflow_dispatch` with an optional `revision` input** is the manual path, including rollback. A manual run is an
  explicit decision and is never skipped as stale.
- Missing secrets are reported by name before CapRover is called, so an unconfigured deployment fails legibly instead of
  as an opaque CLI error.
- `caprover/deploy-from-github` is pinned to the `v1.2.0` commit, matching how every other action in this repository is
  pinned, and covered by the existing Dependabot `github-actions` ecosystem.
- Declares a `staging` GitHub Environment so deployments appear in the repository's Deployments view and can later be
  given protection rules.

Docs: `docs/deploying.md` with the Chinese mirror `docs/zh-CN/deploying.md`, cross-linked from both READMEs and the
release guide.

## Follow-up required before this works

This PR adds no credentials. A maintainer must configure three repository secrets — `CAPROVER_STAGING_SERVER`,
`CAPROVER_STAGING_APP`, `CAPROVER_STAGING_APP_TOKEN` — and the CapRover App prerequisites listed in
`docs/deploying.md` (container HTTP port `8000`, HTTPS, and the `OPENTAG_*` environment variables). `workflow_run` and
`workflow_dispatch` only take effect once the workflow file is on the default branch, so the first automatic deployment
happens on the first successful `main` CI after merge.

The GHCR package is currently public, so CapRover pulls the image anonymously and no Docker registry credential is
needed. `docs/deploying.md` records what to add if that ever changes.

## Testing

- `actionlint .github/workflows/deploy-staging.yml` — clean, no findings.
- `pnpm check` — passes.
- `pnpm build` — passes.
- `pnpm typecheck` — passes.
- `pnpm test` — 2 pre-existing failures in `apps/cli` daemon service tests, both caused by macOS resolving
  `/var/folders/...` to `/private/var/folders/...`; unrelated to this change, which touches only YAML and Markdown.
  All 38 `scripts/__tests__` tests pass.
- `node scripts/registry-coordinate.mjs --image ghcr.io/first-tree-ai/opentag --reference <main-sha>` resolves a real
  published commit coordinate without a token, confirming both the wait step and anonymous pullability.

## Breaking changes

None. No existing workflow, package, or runtime behavior is modified.

## Checklist

- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass, except the pre-existing macOS-only `apps/cli`
      path failures noted above.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -1,0 +1,182 @@
+name: Deploy Staging
+
+on:
+  # CI, not Docker: staging must only ever run a revision whose checks passed, and `main` CI is the
+  # gate that proves that. The image the revision needs is published by the Docker workflow running
+  # alongside CI, so this workflow waits for that coordinate instead of building anything itself.
+  workflow_run:
+    workflows:
+      - CI
+    branches:
+      - main
+    types:
+      - completed
+  workflow_dispatch:
+    inputs:
+      revision:
+        description: "Commit in main history to deploy; defaults to the current main tip"
+        required: false
+        type: string
+
+permissions:
+  contents: read
+  packages: read
+
+env:
+  PRIMARY_NODE_VERSION: "24"
+  IMAGE: ghcr.io/${{ github.repository }}
+
+concurrency:
+  # Staging is one mutable target where only the newest revision is worth shipping, so a superseded
+  # pending run should be dropped rather than deployed after the run that replaced it. This is the
+  # opposite of the image workflow, whose coordinates are immutable and must never be dropped.
+  group: deploy-staging
+  cancel-in-progress: false
+
+jobs:
+  staging:
+    name: Deploy Staging
+    if: >-
+      github.repository == 'first-tree-ai/opentag' &&
+      ((github.event_name == 'workflow_run' &&
+        github.event.workflow_run.conclusion == 'success' &&
+        github.event.workflow_run.event == 'push' &&
+        github.event.workflow_run.head_repository.full_name == github.repository &&
+        github.event.workflow_run.head_branch == 'main') ||
+       (github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main'))
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    environment: staging
+    steps:
+      - name: Check out main
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          # Full history so main membership of the target revision can be proven locally.
+          fetch-depth: 0
+
+      - name: Set up Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: ${{ env.PRIMARY_NODE_VERSION }}
+
+      - name: Resolve the deployment revision
+        id: target
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          RUN_REVISION: ${{ github.event.workflow_run.head_sha }}
+          REQUESTED_REVISION: ${{ inputs.revision }}
+        run: |
+          set -euo pipefail
+          git fetch origin main:refs/remotes/origin/main
+
+          if [ "$EVENT_NAME" = "workflow_run" ]; then
+            REVISION="$RUN_REVISION"
+          elif [ -n "$REQUESTED_REVISION" ]; then
+            REVISION=$(git rev-parse --verify "${REQUESTED_REVISION}^{commit}")
+          else
+            REVISION=$(git rev-parse --verify refs/remotes/origin/main)
+          fi
+
+          if ! printf '%s' "$REVISION" | grep -Eq '^[0-9a-f]{40}$'; then
+            echo "Could not resolve a full commit SHA to deploy (got \"$REVISION\")" >&2
+            exit 1
+          fi
+
+          # A revision outside main history has no reviewed provenance, and the image workflow only
+          # guarantees a commit coordinate for branch pushes it built.
+          if ! git merge-base --is-ancestor "$REVISION" refs/remotes/origin/main; then
+            echo "Revision $REVISION is not contained in main history" >&2
+            exit 1
+          fi
+
+          echo "revision=$REVISION" >> "$GITHUB_OUTPUT"
+          echo "Deployment target is $REVISION"
+
+      - name: Wait for the commit image
+        id: image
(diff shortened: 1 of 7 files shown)


baixiaohang (changes requested):
- Rationale: The deployment itself is commit-addressed, but the GitHub deployment record can claim the wrong revision or a deployment that never ran, and the deployment executable is not actually reproducible.

**Risk level**: B-low
- Path baseline: workflow and documentation changes only -> C
- Semantic lift: changes the staging release/deployment workflow -> B-low

**PR summary**
- Problem: Maintainers need Staging to follow CI-green `main` revisions automatically while retaining an explicit rollback path and an identifiable deployed image.
- Approach: Trigger after successful `main` CI, resolve and wait for the immutable per-commit GHCR coordinate, reject non-`main` revisions, skip stale automatic runs, and point the CapRover App at that exact image; document setup, rollback, and verification in synchronized English and Chinese guides.
- Impacted modules: GitHub Actions staging deployment, GHCR/CapRover release operations, repository deployment and release documentation.

**Review findings**
❌ 1. The `staging` Environment is attached to the whole job before `target` is resolved, so GitHub's deployment record is neither gated by an actual CapRover call nor keyed to the resolved revision. GitHub documents that `workflow_run` uses the last default-branch commit as `GITHUB_SHA`, while a dispatch on `main` uses the current `main` tip; it also creates an Environment deployment automatically and mirrors the job status into that deployment. Consequently, a manual rollback to an older `revision` is recorded against current `main`, and the stale path at lines 128-131 completes successfully and records a successful deployment even though line 155 never ran. The documented A-passes/B-fails case can therefore show B as successfully deployed while Staging stayed on the prior image. Resolve and gate outside the deployment-recording job, then create/update a deployment for the exact target SHA only around the real CapRover operation (or disable the automatic record and write an exact-SHA deployment/status explicitly). [R4/R5 / `.github/workflows/deploy-staging.yml:49`, `.github/workflows/deploy-staging.yml:120`, `.github/workflows/deploy-staging.yml:153`; [GitHub event SHA semantics](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run), [Environment deployment semantics](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#tracking-deployments-through-apps)]

❌ 2. Pinning `caprover/deploy-from-github` to `c221bd9d...` does not pin the code that actually handles the App token and deploys the image. That commit declares a Dockerfile action, so GitHub builds it for each use; its Dockerfile starts from mutable `node:22-alpine` and runs unversioned `npm i -g caprover`. A future base-image or npm release can therefore change this privileged deployment path without any OpenTag commit or Dependabot update. Use a deployment implementation whose complete executable closure is fixed—for example, an audited image by digest or an exact CapRover CLI version in a digest-pinned container—instead of treating this action SHA as a reproducible pin. [R5 / `.github/workflows/deploy-staging.yml:155`; [upstream Dockerfile at the pinned commit](https://github.com/caprover/deploy-from-github/blob/c221bd9d1c3a55de4c2b66c0dc4a3972d04c027c/Dockerfile), [GitHub Docker action build semantics](https://docs.github.com/en/actions/tutorials/use-containerized-services/create-a-docker-container-action#creating-an-action-metadata-file)]

⚠️ 3. The always-run summary treats every pre-gate failure as a stale skip. If the image wait fails or times out, `steps.gate.outputs.deploy` is empty, so lines 179-180 still write "Skipped: this revision is no longer the tip of main." Distinguish an explicit `deploy == false` from an unset gate output so missing images and registry failures retain their real operational diagnosis. [R4 / `.github/workflows/deploy-staging.yml:164`]

**Action taken**
- Submitted request changes.

yuezengwu (changes requested):
Blocking on three deployment-integrity and rollback issues at exact head `4188fa695b355d1961a8105fdcc514d90b7249a0`.

1. **[P1] Pin the code that receives the CapRover token** (`deploy-staging.yml:155`). The pinned action SHA still builds from mutable `node:22-alpine` and runs unversioned, lockfile-free `npm i -g caprover`. Every deployment therefore executes whatever base image and latest npm package resolve that day while `INPUT_TOKEN` is present. Use a path whose complete runtime dependency graph is immutable and reviewed, such as a lockfile-pinned CLI or container digest. If a third-party container continues to receive the workspace, also set checkout `persist-credentials: false`.

2. **[P1] Deploy the resolved digest, not the mutable tag** (`deploy-staging.yml:162`). The wait step proves and exports a digest, but the deploy action discards it and receives `:<commit-sha>`. A tag retargeted after lookup changes the bytes CapRover pulls without changing the reviewed revision, breaking the CI-to-image correspondence and immutable-image guarantee. Pass `IMAGE@sha256:...` after confirming CapRover accepts it and report that same coordinate.

3. **[P2] Do not let automatic runs displace an explicit rollback** (`deploy-staging.yml:33`). GitHub concurrency retains only one pending run per group and cancels the previous pending run when another arrives. A manual rollback queued behind a running deploy can therefore be canceled by the next successful-main run, despite the documented deliberate rollback contract. Use a serialization design that preserves manual rollback, or explicitly change the contract and provide a safe retry path.

The trigger filter, repository and branch provenance checks, least-privilege workflow permissions, bounded image wait, and no-backward automatic gate otherwise look sound.

Validation: `git diff --check`, `pnpm check`, `pnpm build`, `pnpm typecheck`, and registry-coordinate tests (26/26) passed locally. Current CI is red only because Node 22 timed out in an unchanged SessionBindingStore test.

PR #77, opened 2026-08-20
feat(web): open new agent form in dialog

Gandy2025:
## Summary

- replace the Agents page creation link with a New Agent dialog that keeps users in list context
- share the complete creation form with the direct /agents/new route
- add focus management, cancellation, duplicate-submit protection, responsive styling, and creation-flow coverage

## Testing

- pnpm build
- pnpm typecheck
- pnpm --filter @opentag/web test (115 tests passed)
- pnpm exec biome check apps/web/src/router.tsx apps/web/src/styles.css apps/web/src/__tests__/app.test.tsx
- node scripts/third-party-notices.mjs --check

Local full-suite exceptions unrelated to this change:

- pnpm check is blocked by nested root Biome configs in local .claude/worktrees directories
- pnpm test reaches a pre-existing macOS /tmp versus /private/tmp assertion mismatch in client tests; after installing locked dependencies, server tests passed 189/190 with one existing observability timing assertion

## Breaking changes

None.

## Important non-goals

- the direct /agents/new route remains available as a fallback and deep link
- no Agent API or creation contract changes

## Checklist

- [x] Tests cover the changed behavior.
- [ ] All repository-wide commands pass locally; see environment exceptions above.
- [x] No canonical documentation mirrors require updates.

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -522,7 +522,11 @@ describe("OpenTag Web App Shell", () => {
           ),
       ).toHaveLength(1),
     );
-    await waitFor(() => expect(within(dialog).getByLabelText("Display name")).toBe(document.activeElement));
+    await waitFor(() => expect(dialog).toBe(document.activeElement));
+    expect((within(dialog).getByLabelText("Display name") as HTMLInputElement).disabled).toBe(true);
+    expect((within(dialog).getByLabelText("Agent name") as HTMLInputElement).disabled).toBe(true);
+    expect((within(dialog).getByLabelText("Provider") as HTMLSelectElement).disabled).toBe(true);
+    expect((within(dialog).getByLabelText("Computer") as HTMLSelectElement).disabled).toBe(true);
     expect(within(dialog).getByRole("button", { name: "Creating…" }).hasAttribute("disabled")).toBe(true);
     expect(within(dialog).getByRole("button", { name: "Cancel" }).hasAttribute("disabled")).toBe(true);
     expect(within(dialog).getByRole("button", { name: "Close new Agent dialog" }).hasAttribute("disabled")).toBe(true);
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -36,6 +36,7 @@ function installApi(
     agentRead?: () => Promise<void> | void;
     agentReadStatus?: () => number | undefined;
     agentListStatus?: () => number | undefined;
+    agentCreate?: (input: Record<string, unknown>) => Promise<void> | void;
     alreadyJoinedInvitation?: boolean;
     bindingReauth?: boolean;
     bindingEvidenceFails?: boolean;
@@ -185,6 +186,7 @@ function installApi(
       });
     }
     if (path === `/api/v1/teams/${teamId}/agents` && init?.method === "POST") {
+      await options.agentCreate?.(JSON.parse(String(init.body)) as Record<string, unknown>);
       return json(adminConfig(), 201);
     }
     // Any Team id, so a Team created during the test is served like the seeded and invited ones.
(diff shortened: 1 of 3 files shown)


baixiaohang (changes requested):
- Rationale: the dialog flow does not yet make creation retries idempotent or keep focus reliably contained while creation is pending.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed.
- Semantic lift: none; this does not touch subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: Team admins currently leave the Agent list for a separate creation page, which loses list context for a short setup task.
- Approach: open the full creation form in a modal from the Agents page, retain `/agents/new` as a deep-link fallback, and share the form implementation between both surfaces.
- Impacted modules: Web routing/UI, dialog styling, and app-shell interaction tests.

**Review findings**
❌ 1. The duplicate-submission boundary is not atomic or retry-safe. `submitting` is render state, so it is not a synchronous guard before React commits the update, and the POST at `apps/web/src/router.tsx:1093` omits the existing `CreateAgentRequest.creationIntentId`. If the create succeeds but its response is lost, retrying the same form sends a fresh non-idempotent request and surfaces a name conflict instead of replaying the created Agent. Please use a synchronous in-flight guard and retain one creation-intent UUID across retries of the unchanged request, then cover both repeated submission and retry behavior. [R1/R4; `apps/web/src/router.tsx:1085`]

❌ 2. Focus is not repaired when submission disables the focused controls. Clicking Create focuses the submit button, then this render disables Create, Cancel, and Close; unlike `InvitationDialog`, `NewAgentDialog` has no pending-transition effect that moves focus back to an enabled element or the dialog. The keydown handler only intervenes on a later Tab and does not handle a disabled active element that is still inside the dialog, so focus can be dropped outside the modal while Escape/backdrop cancellation is intentionally blocked. Please restore focus containment when `submitting` changes and add pending-state focus coverage. [R4; `apps/web/src/router.tsx:966`]

**Action taken**
- Submitted request changes.

Gandy2025:
Addressed both requested changes in d57892f:

- added a synchronous in-flight guard and a stable creationIntentId keyed to the unchanged request payload, with duplicate-submit and retry coverage
- repaired focus when pending state disables the active control, and covered pending focus containment plus Escape blocking

Validation: Web typecheck, production build, Biome, and all 117 Web tests pass.

baixiaohang (changes requested):
- Rationale: the original blockers are substantially addressed, but the form remains mutable while the protected creation request is unresolved, which can still defeat retry safety.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed.
- Semantic lift: none.

**PR summary**
- Problem: Team admins should be able to create an Agent without leaving the Agents list context.
- Approach: share one creation form between a new modal and the deep-link page, with modal focus management and retry-safe creation intent handling.
- Impacted modules: Web routing/UI, dialog styling, and app-shell interaction tests.

**Review findings**
❌ 1. The request-bearing fields remain editable while creation is pending, and the new focus-repair effect intentionally moves focus to the first of those enabled fields. If the first POST commits but its response is lost, a user can change the Agent name while that request is unresolved; the retry then receives a different fingerprint and a new `creationIntentId`, so it can create a second Agent instead of replaying the first one. This leaves the idempotency boundary dependent on the user not editing the focused form during the mutation. Please freeze all request-bearing inputs/selects while `submitting` is true (and focus the dialog or another non-editable target), with coverage that pending input cannot drift before the request settles. [R4; `apps/web/src/router.tsx:1144`]

✅ 2. The synchronous in-flight guard, stable intent reuse for an unchanged request, and pending-state focus repair resolve the mechanics raised in the previous review.

**Action taken**
- Submitted request changes on the revised head.

Gandy2025:
Addressed the remaining retry-boundary finding in 22099b4: all request-bearing fields are disabled while creation is pending, so the payload cannot drift before the protected request settles. With every interactive control disabled, focus is repaired to the dialog container. The pending-state test now covers field freezing, focus containment, duplicate submission, and Escape blocking. All 117 Web tests and Web typecheck pass.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: all creation retry, atomic submission, and pending-focus boundaries raised in the prior reviews are now closed without changing the Agent API contract or regressing the deep-link flow.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed.
- Semantic lift: none; subscriptions, WebSocket parsing, and the data layer are untouched.

**PR summary**
- Problem: Team admins currently leave the Agents list for a separate page during a short creation task.
- Approach: open the complete creation form in a modal while retaining `/agents/new`, reuse one form implementation, and make creation atomic and retry-safe with a stable server-supported intent ID.
- Impacted modules: Web routing/UI, dialog styling, and app-shell interaction tests.

**Review findings**
✅ 1. A synchronous in-flight guard prevents same-tick duplicate POSTs, while an unchanged request retains its `creationIntentId` across a failed response and retry. [R1/R4]
✅ 2. All request-bearing fields and dismissal controls are frozen while creation is pending, so the protected payload cannot drift before the mutation settles. [R4]
✅ 3. Focus moves to the dialog container when every interactive control is disabled, Escape remains blocked during the mutation, and cancellation restores focus to the trigger afterward. [R4]
✅ 4. The direct `/agents/new` route and no-Computer fallback continue to share the same creation content rather than introducing a second write path. [R1/R2]

**Action taken**
- Approved head `22099b496d406f0be3b74a2907cd932a78598f5a`.

PR #84, opened 2026-08-20
fix(web): scope decorative backgrounds

Gandy2025:
## Summary
- explicitly mark entry and recovery pages that use the branded grid background
- keep decorative grid and subtle grain out of the authenticated product workspace
- consolidate the duplicate empty-state background rules
- add coverage for decorative-page scoping

## Verification
- pnpm exec biome check apps/web/src/router.tsx apps/web/src/styles.css apps/web/src/__tests__/app.test.tsx
- pnpm --filter @opentag/shared build
- pnpm --filter @opentag/web test
- pnpm --filter @opentag/web typecheck
- pnpm --filter @opentag/web build

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -453,6 +453,24 @@ describe("OpenTag Web App Shell", () => {
     expect(window.location.search).toBe(`?next=${encodeURIComponent(expectedNext)}`);
   });
 
+  it("keeps authenticated invalid Agent tabs on the plain workspace canvas", async () => {
+    installApi("admin");
+    window.history.replaceState({}, "", `/agents/${agentId}/unknown`);
+    render(<App />);
+
+    const heading = await screen.findByRole("heading", { name: "Page not found" });
+    expect(heading.closest(".center-card")?.classList.contains("decorative-page")).toBe(false);
+    expect(screen.getByRole("main").classList.contains("decorative-page")).toBe(false);
+  });
+
+  it("keeps the standalone not-found route on the decorative canvas", async () => {
+    window.history.replaceState({}, "", "/unknown");
+    render(<App />);
+
+    const heading = await screen.findByRole("heading", { name: "Page not found" });
+    expect(heading.closest("main")?.classList.contains("decorative-page")).toBe(true);
+  });
+
   it("lets members enter the same shell without admin controls", async () => {
     installApi("member");
     render(<App />);
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -427,6 +427,7 @@ describe("OpenTag Web App Shell", () => {
     installApi("admin");
     render(<App />);
     expect(await screen.findByRole("heading", { name: "Agents" })).toBeTruthy();
+    expect(screen.getByRole("main").classList.contains("decorative-page")).toBe(false);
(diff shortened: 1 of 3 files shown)


baixiaohang (changes requested):
- Rationale: The new marker still reaches authenticated shell states through shared recovery components, so the decorative background is not actually excluded from the product workspace.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed, which falls under the default C tier.
- Semantic lift: none; this is route presentation, CSS, and focused Web coverage without subscription, WebSocket, or data-layer changes.

**PR summary**
- Problem: Authenticated users can see the branded grid and grain in the product workspace because the decoration was coupled to a generic card style rather than to intentional entry and recovery screens.
- Approach: Gate the body decorations on an explicit `decorative-page` marker, apply that marker to selected entry/recovery components, and simplify authenticated empty states to a plain soft surface.
- Impacted modules: Web routing, global styles, and app-shell tests.

**Review findings**
❌ 1. `NotFoundPage` and `UnavailablePage` are not exclusively standalone recovery pages: they are also returned inside `AppShell` for invalid Agent tabs, invalid Settings sections, the fallback Agent tab, and the non-admin `/agents/new` path. Because both shared components now unconditionally carry `decorative-page`, `body:has(.decorative-page)` still turns the grid and grain on for those authenticated workspace states, which breaks this PR's central scoping guarantee. Separate standalone recovery presentation from shell-embedded errors (or make the marker explicit at the standalone call sites), and cover at least one authenticated invalid/restricted route that asserts no decorative marker is present. [R5 / `apps/web/src/router.tsx:1341`, `apps/web/src/router.tsx:1877`, `apps/web/src/router.tsx:2794`, `apps/web/src/router.tsx:2804`; `apps/web/src/__tests__/app.test.tsx:426`]

**Action taken**
- Submitted request changes.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The new head closes the authenticated-shell leak by separating standalone recovery presentation from embedded shell errors and adds regression coverage for both sides of that boundary.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed, which falls under the default C tier.
- Semantic lift: none; the change remains limited to route presentation, CSS, and focused Web coverage.

**PR summary**
- Problem: Branded grid and grain decoration was coupled to a generic card style, so it could appear inside the authenticated product workspace instead of only on intentional entry and recovery screens.
- Approach: Gate decoration on an explicit marker, keep shell-embedded empty and recovery states plain, and give the top-level catch-all its own decorative standalone page.
- Impacted modules: Web routing, global styles, and app-shell tests.

**Review findings**
✅ 1. The previous blocker is closed: `NotFoundPage` and `UnavailablePage` now render marker-free `<section>` elements inside `AppShell`, while only the top-level catch-all uses `StandaloneNotFoundPage` with `decorative-page`. [R5 / `apps/web/src/router.tsx:351`, `apps/web/src/router.tsx:2794`, `apps/web/src/router.tsx:2804`, `apps/web/src/router.tsx:2814`]
✅ 2. The regression coverage now exercises an authenticated invalid Agent tab and the standalone not-found route in addition to the normal workspace and login cases, protecting both sides of the scoping boundary. [`apps/web/src/__tests__/app.test.tsx:426`, `apps/web/src/__tests__/app.test.tsx:446`, `apps/web/src/__tests__/app.test.tsx:456`, `apps/web/src/__tests__/app.test.tsx:466`]

**Action taken**
- Approved.

PR #85, opened 2026-08-20
fix(web): simplify Team identity settings

Gandy2025:
## Summary
- expose a single editable Team name and keep the generated CLI identifier read-only
- preserve the CLI identifier when renaming a Team
- group invitation-link management inside Team members
- update accessible structure and regression coverage

## Validation
- `pnpm --filter @opentag/web test` (137 tests)
- `pnpm --filter @opentag/web typecheck`
- `pnpm --filter @opentag/web build`
- `pnpm check`

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -639,7 +639,7 @@ describe("OpenTag Web App Shell", () => {
     window.history.replaceState({}, "", "/settings/team");
     render(<App />);
     expect(await screen.findByRole("heading", { name: "Team members" })).toBeTruthy();
-    expect(screen.getAllByLabelText("Display name")).toHaveLength(1);
+    expect(screen.getAllByLabelText("Team name")).toHaveLength(1);
     expect(screen.queryByRole("button", { name: "Save account profile" })).toBeNull();
   });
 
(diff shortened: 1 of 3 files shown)


baixiaohang (changes requested):
- Rationale: The Web now presents the generated CLI identifier as stable and read-only, but the authoritative update contract still lets any Team-admin client change it.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed -> C
- Semantic lift: none; no Web subscription, WebSocket parsing, or data-layer code is changed

**PR summary**
- Problem: Team admins currently see two editable names, making it unclear which identity people should rename, while invitation-link controls sit apart from the members they affect.
- Approach: expose only the display name as editable, render the generated CLI selector as read-only, send a display-name-only rename request, and nest invite-link management under Team members with corrected heading levels.
- Impacted modules: `apps/web/src/router.tsx`, Web app regression tests, and Team Settings styles

**Review findings**
❌ 1. Enforce the read-only CLI identifier at the authoritative update boundary, or document and implement a bounded compatibility/deprecation path. The new copy says the identifier "stays the same when you rename the Team," but `UpdateTeamProfileRequestSchema` still accepts `name`, the public Client still exposes that request, and `TeamMembershipService.updateTeamProfile` still writes it. Another client can therefore perform the old canonical-name rename and invalidate the invariant this UI now promises. The member view also says admins can change both displayed fields, even though this PR intends the identifier to be immutable. Remove `name` from the update contract/service (and update affected tests/callers), or make the remaining legacy authority explicit and temporary. [R2/R5 / `apps/web/src/router.tsx:2248`; `packages/shared/src/team.ts:43`; `packages/server/src/services/teams/team-membership-service.ts:266`]
✅ 2. The focused Web regression correctly verifies that a rename sends only `displayName`, retains the CLI identifier after refresh, and preserves UUID-based Team selection.

**Action taken**
- Submitted request changes.

PR #87, opened 2026-08-20
refactor(web): drive typography from a token layer

Gandy2025:
## Summary

`apps/web/src/styles.css` carried 121 `font-size` declarations across 35 distinct values, 19 distinct font weights between 430 and 800, and eight scattered line heights — all literals written at the point of use, with no token layer and nothing preventing the next component from adding a 36th size. Twelve selectors declared type twice, with a later override layer silently winning the cascade, so editing the first declaration had no effect on the rendered page.

Five independently revertible commits:

| Commit | Change |
| --- | --- |
| `45013fe` | Hoist the winning value of 12 shadowed selectors and drop the override; computed styles unchanged |
| `e6ba060` | Introduce `--fs-*` steps and `--text-*` roles; map every size, weight, tracking and font stack onto them |
| `6d6cc48` | Rebind the role tokens on `.settings-page` / `.onboarding-shell` / `.decorative-page` so density follows the surface |
| `d43aabe` | Name CJK fallbacks in each stack, opt tables into tabular numerals, stop advertising an unimplemented dark scheme |
| `3b887d1` | Assert the token layer holds, so the scale cannot drift back apart |

Result: 35 sizes to 9 whole-pixel steps behind 9 semantic roles, 19 weights to 4, 8 line heights to 3 tokens, 31 fractional-pixel values to 0.

Values follow a dense-console scale — 13px interface text, 14px prose, 12px floor for labels — with the reading surfaces stepping each role up one, because the product is used both by operators living in the agent and computer tables and by team admins who mostly touch settings and onboarding.

Notable rendered changes: page `h1` 32–36px to 24px, section `h2` 24.8px to 20px, sidebar nav and inputs to 13px, buttons 16px to 13px (they previously inherited the user-agent default), micro labels up from 10.5–11.5px to a 12px floor. `h4` no longer renders larger than `h3`, `h2` no longer has three sizes, and `body`/`p` carry explicit sizes instead of falling back to 16px wherever nothing overrode them.

Three findings from the audit are deliberately left alone: Bricolage Grotesque stays (dropping it would save ~70KB but removes the brand voice from entry pages — a product call, not a cleanup); `pnpm dev` is broken by a `@babel/generator@7` versus pinned `@babel/parser@8.0.0-rc.2` mismatch that predates this branch; and two `@opentag/client` process-group tests fail on `e1d2cd1` as well.

## Testing

- `pnpm check` — pass
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm --filter @opentag/web test` — 189 passed, including 7 new token assertions
- `pnpm test` — the workspace run fails on two `@opentag/client` tests (`codex-app-server` D-11 process-group teardown and `client-runtime-composition` Codex deferral). Both fail identically when checked out at the base commit `e1d2cd1`; nothing in this branch touches `packages/client`.
- Guard verified by injecting `font-size: 0.83rem` and `font-weight: 615`, which fails the size and weight assertions as intended.
- Visual check on the production build via `vite preview` with stubbed API responses: sign-in, Agents list, Team settings and Computer settings compared before and after, including Chinese team, agent and member names.

## Breaking changes

None for the API or build output. The change is visual: interface text renders smaller across product surfaces and larger on configuration and entry surfaces, as described above.

## Checklist

- [ ] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. — the first three pass; `pnpm test` is red only on the two pre-existing `@opentag/client` failures documented above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PoLjTcXgmc3qGBNFekv9nL)_

### DIFF
--- a/apps/web/index.html
+++ b/apps/web/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta name="color-scheme" content="light dark" />
+    <meta name="color-scheme" content="light" />
     <title>OpenTag</title>
   </head>
   <body>
--- a/apps/web/src/__tests__/typography-tokens.test.ts
+++ b/apps/web/src/__tests__/typography-tokens.test.ts
@@ -1,0 +1,85 @@
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+import { describe, expect, it } from "vitest";
+
+/**
+ * The stylesheet once carried 35 distinct font sizes and 19 font weights as
+ * literals written at the point of use, most of them within a pixel of each
+ * other. These tests keep type flowing through the token layer so the scale
+ * cannot drift back apart one component at a time.
+ */
+
+const stylesheet = readFileSync(resolve(process.cwd(), "src/styles.css"), "utf8");
+
+function captures(pattern: RegExp, group = 1): string[] {
+  return [...stylesheet.matchAll(pattern)]
+    .map((match) => match[group])
+    .filter((value): value is string => value !== undefined)
+    .map((value) => value.trim());
+}
+
+function declarations(property: string): string[] {
+  return captures(new RegExp(`(?<![\\w-])${property}:\\s*([^;]+);`, "g"));
+}
+
+function definedTokens(prefix: string): Set<string> {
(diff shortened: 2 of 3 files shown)


Gandy2025:
Closing in favour of #88, which carries the same five commits from `refactor/web-typography-tokens`. The branch-name check rejects the `claude/` prefix before lint, build or tests run, so this PR could not produce a CI signal.


---

PR #88, opened 2026-08-20
refactor(web): drive typography from a token layer

Gandy2025:
## Summary

`apps/web/src/styles.css` carried 121 `font-size` declarations across 35 distinct values, 19 distinct font weights between 430 and 800, and eight scattered line heights — all literals written at the point of use, with no token layer and nothing preventing the next component from adding a 36th size. Twelve selectors declared type twice, with a later override layer silently winning the cascade, so editing the first declaration had no effect on the rendered page.

The branch is in two parts.

**The migration**, in five independently revertible commits:

| Commit | Change |
| --- | --- |
| `45013fe` | Hoist the winning value of 12 shadowed selectors and drop the override; computed styles unchanged |
| `e6ba060` | Introduce `--fs-*` steps and `--text-*` roles; map every size, weight, tracking and font stack onto them |
| `6d6cc48` | Rebind the role tokens on `.settings-page` / `.onboarding-shell` / `.decorative-page` so density follows the surface |
| `d43aabe` | Name CJK fallbacks in each stack, opt tables into tabular numerals, stop advertising an unimplemented dark scheme |
| `3b887d1` | Assert the token layer holds, so the scale cannot drift back apart |

Result: 35 sizes to 9 whole-pixel steps behind 9 semantic roles, 19 weights to 4, 8 line heights to 3 tokens, 31 fractional-pixel values to 0.

Values follow a dense-console scale — 13px interface text, 14px prose, 12px floor for labels — with the reading surfaces stepping each role up one, because the product is used both by operators living in the agent and computer tables and by team admins who mostly touch settings and onboarding.

Notable rendered changes: page `h1` 32–36px to 24px, section `h2` 24.8px to 20px, sidebar nav and inputs to 13px, buttons 16px to 13px (they previously inherited the user-agent default), micro labels up from 10.5–11.5px to a 12px floor. `h4` no longer renders larger than `h3`, `h2` no longer has three sizes, and `body`/`p` carry explicit sizes instead of falling back to 16px wherever nothing overrode them.

**The guard**, in the remaining commits. It began as regular expressions over the stylesheet text and was rewritten across review into a PostCSS-based check with a canonical-source policy. Thirteen rounds of review found thirteen ways a declaration could reach the browser while the guard read it as something else — the `font` shorthand read as a longhand, tokens defined only inside a comment or a string, case-insensitive property names, escaped identifiers, definitions hidden under `@media` or an unrelated selector, exceptions allowed by value rather than by selector. Every one was real. The guard now:

- reads parsed declarations, never stylesheet text, and never the name as written — identifier escapes are decoded before any comparison, and PostCSS nodes do not leave the parse layer
- refuses escapes in the non-string portion of any value rather than reimplementing CSS value tokenization, since the stylesheet has never contained one
- resolves the baseline only from an unconditional `:root`, limits role rebinding to the four reading surfaces, and binds each literal exception to the exact selectors that earn it
- tests itself: 22 of its 40 cases assert that it rejects what it exists to catch, including both the escaped and quoted forms of each bypass

It has already earned its place twice: it caught `.connect-command-meta { font-size: 0.84rem }` arriving from `main` mid-review, and it caught a regression I introduced while refactoring it.

Two findings from the audit are deliberately left alone: Bricolage Grotesque stays (dropping it would save ~70KB but removes the brand voice from entry pages — a product call, not a cleanup), and `pnpm dev` is broken by a `@babel/generator@7` versus pinned `@babel/parser@8.0.0-rc.2` mismatch that predates this branch and needs its own fix.

Supersedes #87, which was opened from a branch prefix the branch-name check rejects.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — pass in CI; all six checks are green on `f17e2f0`.
- `pnpm --filter @opentag/web test` — 238 passed, 40 of them the typography guard.
- Thirteen injections re-run against the real stylesheet after every change, each rejected: literal size, literal `font-family`, `font` shorthand, uppercase property, escaped property, escaped custom property, escaped parenthesis forging a `var()` call, escaped reference under an unowned property, poisoned string in a value, token defined only in a comment, only in a string, only under `@media`, and a role retuned by an arbitrary component. Two negative cases confirm the guard has not simply been tightened: `content: "var(--fs-99)"` raises nothing, and a backslash inside a string is left alone.
- Visual check on the production build via `vite preview` with stubbed API responses: sign-in, Agents list, Team settings and Computer settings compared before and after, including Chinese team, agent and member names.

Locally, two `@opentag/client` process-group tests fail in the sandbox this branch was developed in. They fail identically at the base commit and pass on the runner, so they are an artifact of the sandbox rather than of this change.

## Breaking changes

None for the API or build output. The change is visual: interface text renders smaller across product surfaces and larger on configuration and entry surfaces, as described above.

## Checklist

### DIFF
--- a/apps/web/index.html
+++ b/apps/web/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta name="color-scheme" content="light dark" />
+    <meta name="color-scheme" content="light" />
     <title>OpenTag</title>
   </head>
   <body>
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -26,6 +26,7 @@
     "@types/react-dom": "^19.1.7",
     "@vitejs/plugin-react": "^4.7.0",
     "jsdom": "^26.1.0",
+    "postcss": "^8.5.26",
     "vite": "^7.1.3"
   }
 }
--- a/apps/web/src/__tests__/typography-tokens.test.ts
+++ b/apps/web/src/__tests__/typography-tokens.test.ts
@@ -101,16 +101,6 @@ function referenceProjection(value: string): string {
   return withoutStrings(value);
 }
 
-/** The properties this guard owns, alongside the token families themselves. */
-const TYPOGRAPHY_PROPERTIES = new Set([
-  "font",
-  "font-size",
-  "font-weight",
-  "font-family",
-  "line-height",
-  "letter-spacing",
-]);
-
 /**
  * What the browser sees: the decoded name, the decoded value exactly as the
  * browser validates it, where it applies, and whether it applies
@@ -193,21 +183,17 @@ function offTokenValues(css: string, property: string, allowed: RegExp, literals
   return valuesOf(css, property).filter((value) => !allowed.test(value) && !(literals?.has(value) ?? false));
 }
 
-/** A typography declaration is one this guard reads: an owned property, or a token definition. */
-function ownedByGuard(declaration: Declared): boolean {
-  return (
-    TYPOGRAPHY_PROPERTIES.has(propertyName(declaration)) ||
-    FAMILIES.some((family) => declaration.name.startsWith(`--${family}-`))
-  );
-}
-
 /**
  * An escape inside a value changes which tokens the browser consumes, so the
- * text is no longer what it looks like. The stylesheet has never needed one.
+ * text stops meaning what it looks like. References are read from every
+ * declaration, not only the ones this guard owns, so every declaration has to
+ * be canonical for that scan to be worth anything -- outside strings, where a
+ * backslash is ordinary content and no reference is read. The stylesheet has
+ * never contained one.
  */
 function escapedValues(css: string): string[] {
   return declarations(css)
-    .filter((declaration) => ownedByGuard(declaration) && declaration.value.includes("\\"))
+    .filter((declaration) => declaration.references.includes("\\"))
     .map((declaration) => `${declaration.name}: ${declaration.value}`);
 }
 
@@ -327,7 +313,7 @@ describe("typography tokens", () => {
     expect(unresolvedRoles(stylesheet)).toEqual([]);
   });
 
-  it("spells every value it owns without escapes", () => {
+  it("spells every value outside a string without escapes", () => {
     expect(escapedValues(stylesheet)).toEqual([]);
   });
 
@@ -502,6 +488,16 @@ describe("the guard itself", () => {
     const role = String.raw`:root { --text-ui: var\28--fs-13\29; }`;
     expect(escapedValues(role)).toEqual([String.raw`--text-ui: var\28--fs-13\29`]);
     expect(malformedRoles(role)).toEqual([String.raw`--text-ui: var\28--fs-13\29`]);
+  });
+
+  it("refuses an escaped reference under a property it does not otherwise read", () => {
+    const css = String.raw`.row { padding: v\61 r(--fs-99); }`;
+    expect(escapedValues(css)).toEqual([String.raw`padding: v\61 r(--fs-99)`]);
+  });
+
+  it("leaves a backslash inside a string alone, where it is content", () => {
+    const css = String.raw`.row::after { content: "\201C"; }`;
+    expect(escapedValues(css)).toEqual([]);
   });
 
   it("does not read a commented-out declaration as a violation", () => {
--- a/apps/web/src/__tests__/typography-tokens.test.ts
+++ b/apps/web/src/__tests__/typography-tokens.test.ts
@@ -43,8 +43,17 @@ const LINE_HEIGHT_LITERALS = new Set(["1", "38px"]);
  */
 const FONT_SHORTHAND_LITERALS = new Set(["inherit"]);
 
+/**
+ * A commented-out declaration is not a declaration. The browser never sees it,
+ * so neither should any check below: a token defined only inside a comment
+ * does not exist, and a rule parked inside one cannot violate anything.
+ */
+function withoutComments(css: string): string {
+  return css.replace(/\/\*[\s\S]*?\*\//g, "");
+}
+
 function captures(css: string, pattern: RegExp, group = 1): string[] {
-  return [...css.matchAll(pattern)]
+  return [...withoutComments(css).matchAll(pattern)]
     .map((match) => match[group])
     .filter((value): value is string => value !== undefined)
     .map((value) => value.trim());
@@ -157,6 +166,17 @@ describe("the guard itself", () => {
     const css = ":root { --text-ui: var(--fs-15); }\n.row { line-height: var(--lh-prsoe); }";
     expect(danglingReferences(css)).toEqual(["fs-15", "lh-prsoe"]);
     expect(unresolvedRoles(css)).toEqual(["fs-15"]);
+  });
+
+  it("ignores a step that exists only inside a comment", () => {
+    const css = ":root {\n  --text-ui: var(--fs-15);\n  /* --fs-15: 0.9375rem; */\n}";
+    expect(unresolvedRoles(css)).toEqual(["fs-15"]);
+    expect(danglingReferences(css)).toEqual(["fs-15"]);
+  });
+
+  it("does not read a commented-out declaration as a violation", () => {
+    const css = ".row { /* font-size: 0.83rem; */ font-size: var(--text-ui); }";
+    expect(offTokenValues(css, "font-size", /^var\(--text-[a-z]+\)$/)).toEqual([]);
   });
 
   it("does not mistake a longhand for the shorthand it starts with", () => {
--- a/apps/web/src/__tests__/typography-tokens.test.ts
+++ b/apps/web/src/__tests__/typography-tokens.test.ts
@@ -1,6 +1,6 @@
 import { existsSync, readFileSync } from "node:fs";
 import { resolve } from "node:path";
-import postcss, { type Declaration } from "postcss";
+import postcss, { type Declaration, type Rule } from "postcss";
 import { describe, expect, it } from "vitest";
 
 /**
@@ -55,6 +55,22 @@ function declarations(css: string): Declaration[] {
   return found;
 }
 
+/** The selectors a declaration applies under, or none when it sits outside a rule. */
+function selectorsOf(declaration: Declaration): string[] {
+  const parent = declaration.parent;
+  if (parent === undefined || parent.type !== "rule") return [];
+  return (parent as Rule).selector.split(",").map((selector) => selector.trim());
+}
+
+/**
+ * A custom property exists only where its selector matches. A step parked under
+ * some other rule is invisible to a :root role, so the baseline scale has to be
+ * declared at :root to count as declared at all.
+ */
+function atRoot(declaration: Declaration): boolean {
+  return selectorsOf(declaration).includes(":root");
+}
+
 /**
  * A quoted run is content, not code: `content: "--fs-15: 0.9375rem"` declares
  * no token and references none, so strings drop out before a value is read.
@@ -73,7 +89,7 @@ function valuesOf(css: string, property: string): string[] {
 function definedTokens(css: string, prefix: string): Set<string> {
   const pattern = new RegExp(`^--(${prefix}-[a-z0-9-]+)$`);
   const names = new Set<string>();
-  for (const declaration of declarations(css)) {
+  for (const declaration of declarations(css).filter(atRoot)) {
     const name = declaration.prop.match(pattern)?.[1];
     if (name !== undefined) names.add(name);
   }
@@ -117,12 +133,30 @@ function unresolvedRoles(css: string): string[] {
 
 /** Raw steps as declared: the number in the name, and the pixel value it resolves to. */
 function rawSteps(css: string): { name: number; px: number }[] {
-  return declarations(css).flatMap((declaration) => {
-    const name = declaration.prop.match(/^--fs-([0-9]+)$/)?.[1];
-    const rem = withoutStrings(declaration.value).match(/^([0-9.]+)rem$/)?.[1];
-    if (name === undefined || rem === undefined) return [];
-    return [{ name: Number(name), px: Number(rem) * 16 }];
-  });
+  return declarations(css)
+    .filter(atRoot)
+    .flatMap((declaration) => {
+      const name = declaration.prop.match(/^--fs-([0-9]+)$/)?.[1];
+      const rem = withoutStrings(declaration.value).match(/^([0-9.]+)rem$/)?.[1];
+      if (name === undefined || rem === undefined) return [];
+      return [{ name: Number(name), px: Number(rem) * 16 }];
+    });
+}
+
+/**
+ * Typography tokens declared somewhere other than :root. A surface may retune a
+ * role it inherits; it may not invent a token, because nothing outside that
+ * surface could resolve it.
+ */
+function scopedDefinitions(css: string): { token: string; selector: string }[] {
+  const pattern = new RegExp(`^--((?:${FAMILIES.join("|")})-[a-z0-9-]+)$`);
+  return declarations(css)
+    .filter((declaration) => !atRoot(declaration))
+    .flatMap((declaration) => {
+      const token = declaration.prop.match(pattern)?.[1];
+      if (token === undefined) return [];
+      return [{ token, selector: selectorsOf(declaration).join(", ") }];
+    });
 }
 
 describe("typography tokens", () => {
@@ -162,6 +196,14 @@ describe("typography tokens", () => {
     const roles = declarations(stylesheet).filter((declaration) => /^--text-[a-z]+$/.test(declaration.prop));
     expect(roles.length).toBeGreaterThan(0);
     expect(roleEdges(stylesheet).length).toBe(roles.length);
+  });
+
+  it("rebinds only roles that :root already declares, and never a raw step", () => {
+    const roles = definedTokens(stylesheet, "text");
+    const introduced = scopedDefinitions(stylesheet).filter(
+      (definition) => !definition.token.startsWith("text-") || !roles.has(definition.token),
+    );
+    expect(introduced).toEqual([]);
   });
 
   it("keeps every raw step on a whole pixel, named after the size it produces", () => {
@@ -218,6 +260,21 @@ describe("the guard itself", () => {
   it("does not read a token reference quoted inside a value", () => {
     const css = ':root { --fs-13: 0.8125rem; }\n.row::after { content: "var(--fs-99)"; }';
     expect(danglingReferences(css)).toEqual([]);
+  });
+
+  it("ignores a step declared under some other selector", () => {
+    const css = ":root { --text-ui: var(--fs-13); }\n.dead { --fs-13: 0.8125rem; }";
+    expect(unresolvedRoles(css)).toEqual(["fs-13"]);
+    expect(danglingReferences(css)).toEqual(["fs-13"]);
+    expect(rawSteps(css)).toEqual([]);
+    expect(scopedDefinitions(css)).toEqual([{ token: "fs-13", selector: ".dead" }]);
+  });
+
+  it("accepts a surface that retunes a role :root already declares", () => {
+    const css =
+      ":root { --fs-13: 0.8125rem; --fs-14: 0.875rem; --text-ui: var(--fs-13); }\n.settings-page { --text-ui: var(--fs-14); }";
+    expect(danglingReferences(css)).toEqual([]);
+    expect(scopedDefinitions(css)).toEqual([{ token: "text-ui", selector: ".settings-page" }]);
   });
 
   it("does not read a commented-out declaration as a violation", () => {
--- a/apps/web/src/__tests__/typography-tokens.test.ts
+++ b/apps/web/src/__tests__/typography-tokens.test.ts
@@ -1,6 +1,6 @@
 import { existsSync, readFileSync } from "node:fs";
 import { resolve } from "node:path";
-import postcss, { type Declaration, type Rule } from "postcss";
+import postcss, { type Container, type Declaration, type Document, type Rule } from "postcss";
 import { describe, expect, it } from "vitest";
 
 /**
@@ -9,6 +9,13 @@ import { describe, expect, it } from "vitest";
  * other. These tests keep type flowing through the token layer so the scale
  * cannot drift back apart one component at a time.
  *
+ * Surfaces that own a density, and the only selectors allowed to retune a
+ * role. Anywhere else, a rebinding is component-local drift: the thing this
+ * PR removed.
+ */
+const READING_SURFACES = new Set([".settings-page", ".onboarding-shell", ".decorative-page", ".dialog-card"]);
+
+/*
  * Every check reads parsed declarations rather than raw text. Matching CSS
  * with regular expressions cost this guard three separate holes in review —
  * the `font` shorthand read as a longhand, a token defined only inside a
@@ -62,13 +69,24 @@ function selectorsOf(declaration: Declaration): string[] {
   return (parent as Rule).selector.split(",").map((selector) => selector.trim());
 }
 
+/** A declaration wrapped in any at-rule applies only when that condition holds. */
+function unconditional(declaration: Declaration): boolean {
+  let node: Container | Document | undefined = declaration.parent;
+  while (node !== undefined) {
+    if (node.type === "atrule") return false;
+    node = node.parent;
+  }
+  return true;
+}
+
 /**
- * A custom property exists only where its selector matches. A step parked under
- * some other rule is invisible to a :root role, so the baseline scale has to be
- * declared at :root to count as declared at all.
+ * A custom property exists only where its selector matches and its conditions
+ * hold. A step parked under another rule -- or under a media query that a normal
+ * viewport never satisfies -- is invisible to a :root role, so the baseline
+ * scale has to be declared at an unconditional :root to count as declared.
  */
 function atRoot(declaration: Declaration): boolean {
-  return selectorsOf(declaration).includes(":root");
+  return unconditional(declaration) && selectorsOf(declaration).includes(":root");
 }
 
 /**
@@ -79,10 +97,15 @@ function withoutStrings(value: string): string {
   return value.replace(/"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*'/g, "");
 }
 
+/** Ordinary property names are ASCII case-insensitive; custom property names are not. */
+function propertyName(declaration: Declaration): string {
+  return declaration.prop.startsWith("--") ? declaration.prop : declaration.prop.toLowerCase();
+}
+
 /** Values of one property, matched by parsed name so `font` never collects `font-size`. */
 function valuesOf(css: string, property: string): string[] {
   return declarations(css)
-    .filter((declaration) => declaration.prop === property)
+    .filter((declaration) => propertyName(declaration) === property)
     .map((declaration) => declaration.value.trim());
 }
 
@@ -131,16 +154,30 @@ function unresolvedRoles(css: string): string[] {
     .map((edge) => edge.step);
 }
 
-/** Raw steps as declared: the number in the name, and the pixel value it resolves to. */
+/** Every baseline step declaration, well-formed or not, so none is skipped silently. */
+function rootStepDeclarations(css: string): Declaration[] {
+  return declarations(css).filter((declaration) => atRoot(declaration) && declaration.prop.startsWith("--fs-"));
+}
+
+/** A step is a numeric name and a rem value; anything else does not describe one. */
+function stepOf(declaration: Declaration): { name: number; px: number } | undefined {
+  const name = declaration.prop.match(/^--fs-([0-9]+)$/)?.[1];
+  const rem = withoutStrings(declaration.value).match(/^([0-9.]+)rem$/)?.[1];
+  if (name === undefined || rem === undefined) return undefined;
+  return { name: Number(name), px: Number(rem) * 16 };
+}
+
 function rawSteps(css: string): { name: number; px: number }[] {
-  return declarations(css)
-    .filter(atRoot)
-    .flatMap((declaration) => {
-      const name = declaration.prop.match(/^--fs-([0-9]+)$/)?.[1];
-      const rem = withoutStrings(declaration.value).match(/^([0-9.]+)rem$/)?.[1];
-      if (name === undefined || rem === undefined) return [];
-      return [{ name: Number(name), px: Number(rem) * 16 }];
-    });
+  return rootStepDeclarations(css).flatMap((declaration) => {
+    const step = stepOf(declaration);
+    return step === undefined ? [] : [step];
+  });
+}
+
+function malformedSteps(css: string): string[] {
+  return rootStepDeclarations(css)
+    .filter((declaration) => stepOf(declaration) === undefined)
+    .map((declaration) => `${declaration.prop}: ${declaration.value}`);
 }
 
 /**
@@ -148,14 +185,14 @@ function rawSteps(css: string): { name: number; px: number }[] {
  * role it inherits; it may not invent a token, because nothing outside that
  * surface could resolve it.
  */
-function scopedDefinitions(css: string): { token: string; selector: string }[] {
+function scopedDefinitions(css: string): { token: string; selectors: string[] }[] {
   const pattern = new RegExp(`^--((?:${FAMILIES.join("|")})-[a-z0-9-]+)$`);
   return declarations(css)
     .filter((declaration) => !atRoot(declaration))
     .flatMap((declaration) => {
       const token = declaration.prop.match(pattern)?.[1];
       if (token === undefined) return [];
-      return [{ token, selector: selectorsOf(declaration).join(", ") }];
+      return [{ token, selectors: selectorsOf(declaration) }];
     });
 }
 
@@ -198,12 +235,20 @@ describe("typography tokens", () => {
     expect(roleEdges(stylesheet).length).toBe(roles.length);
   });
 
-  it("rebinds only roles that :root already declares, and never a raw step", () => {
+  it("retunes roles only on the reading surfaces, and never a raw step", () => {
     const roles = definedTokens(stylesheet, "text");
-    const introduced = scopedDefinitions(stylesheet).filter(
-      (definition) => !definition.token.startsWith("text-") || !roles.has(definition.token),
+    const stray = scopedDefinitions(stylesheet).filter(
+      (definition) =>
+        !definition.token.startsWith("text-") ||
+        !roles.has(definition.token) ||
+        !definition.selectors.every((selector) => READING_SURFACES.has(selector)),
     );
-    expect(introduced).toEqual([]);
+    expect(stray).toEqual([]);
+  });
+
+  it("declares every baseline step as a whole-pixel rem, with none skipped", () => {
+    expect(malformedSteps(stylesheet)).toEqual([]);
+    expect(rawSteps(stylesheet).length).toBe(rootStepDeclarations(stylesheet).length);
   });
 
   it("keeps every raw step on a whole pixel, named after the size it produces", () => {
@@ -267,14 +312,48 @@ describe("the guard itself", () => {
     expect(unresolvedRoles(css)).toEqual(["fs-13"]);
     expect(danglingReferences(css)).toEqual(["fs-13"]);
     expect(rawSteps(css)).toEqual([]);
-    expect(scopedDefinitions(css)).toEqual([{ token: "fs-13", selector: ".dead" }]);
+    expect(scopedDefinitions(css)).toEqual([{ token: "fs-13", selectors: [".dead"] }]);
   });
 
   it("accepts a surface that retunes a role :root already declares", () => {
     const css =
       ":root { --fs-13: 0.8125rem; --fs-14: 0.875rem; --text-ui: var(--fs-13); }\n.settings-page { --text-ui: var(--fs-14); }";
     expect(danglingReferences(css)).toEqual([]);
-    expect(scopedDefinitions(css)).toEqual([{ token: "text-ui", selector: ".settings-page" }]);
+    expect(scopedDefinitions(css)).toEqual([{ token: "text-ui", selectors: [".settings-page"] }]);
+  });
+
+  it("reads an upper-case property name, which the browser applies all the same", () => {
+    const css = "body { FONT-SIZE: 13px; FONT-FAMILY: Arial; }\n.row { Font: 12px Arial; }";
+    expect(offTokenValues(css, "font-size", /^var\(--text-[a-z]+\)$/)).toEqual(["13px"]);
+    expect(offTokenValues(css, "font-family", /^var\(--font-[a-z]+\)$/)).toEqual(["Arial"]);
+    expect(offTokenValues(css, "font", /^$/, FONT_SHORTHAND_LITERALS)).toEqual(["12px Arial"]);
+  });
+
+  it("reports a step that does not describe a step, rather than skipping it", () => {
+    const css = ":root { --fs-13: 13.5px; --fs-small: 0.8125rem; --fs-14: 0.875rem; }";
+    expect(malformedSteps(css)).toEqual(["--fs-13: 13.5px", "--fs-small: 0.8125rem"]);
+    expect(rawSteps(css)).toEqual([{ name: 14, px: 14 }]);
+  });
+
+  it("ignores a step declared under a condition a normal viewport never meets", () => {
+    const css = ":root { --text-ui: var(--fs-13); }\n@media (width: 0px) { :root { --fs-13: 0.8125rem; } }";
+    expect(unresolvedRoles(css)).toEqual(["fs-13"]);
+    expect(danglingReferences(css)).toEqual(["fs-13"]);
+    expect(rawSteps(css)).toEqual([]);
+    expect(scopedDefinitions(css)).toEqual([{ token: "fs-13", selectors: [":root"] }]);
+  });
+
+  it("rejects a component that retunes a role it does not own", () => {
+    const css =
+      ":root { --fs-30: 1.875rem; --text-ui: var(--fs-13); --fs-13: 0.8125rem; }\n.row { --text-ui: var(--fs-30); }";
+    const roles = definedTokens(css, "text");
+    const stray = scopedDefinitions(css).filter(
+      (definition) =>
+        !definition.token.startsWith("text-") ||
+        !roles.has(definition.token) ||
+        !definition.selectors.every((selector) => READING_SURFACES.has(selector)),
+    );
+    expect(stray).toEqual([{ token: "text-ui", selectors: [".row"] }]);
   });
 
   it("does not read a commented-out declaration as a violation", () => {
--- a/apps/web/src/__tests__/typography-tokens.test.ts
+++ b/apps/web/src/__tests__/typography-tokens.test.ts
@@ -95,12 +95,21 @@ function withoutStrings(value: string): string {
 
 /**
  * Strings are ignorable when looking for token references, and only then: to
- * the browser a quoted run is still part of the value it validates. Strings go
- * first and escapes second, since decoding first could manufacture a quote.
+ * the browser a quoted run is still part of the value it validates.
  */
 function referenceProjection(value: string): string {
-  return decodeIdentifier(withoutStrings(value));
+  return withoutStrings(value);
 }
+
+/** The properties this guard owns, alongside the token families themselves. */
+const TYPOGRAPHY_PROPERTIES = new Set([
+  "font",
+  "font-size",
+  "font-weight",
+  "font-family",
+  "line-height",
+  "letter-spacing",
+]);
 
 /**
  * What the browser sees: the decoded name, the decoded value exactly as the
@@ -110,6 +119,12 @@ function referenceProjection(value: string): string {
  * `references` is that value with quoted runs dropped, and is for finding
  * var() references and nothing else. Validating against it would accept
  * `font-size: "poison" var(--text-ui)`, which the browser throws out.
+ *
+ * Values are recorded as written. A name is one identifier, so decoding it is
+ * exact; a value is a sequence of tokens, and CSS resolves escapes inside the
+ * token being consumed rather than across it -- `var\28--text-ui\29` is a
+ * single identifier, not a call. Rather than reimplement value tokenization to
+ * tell those apart, escapes are refused in the values this guard owns.
  *
  * Every check below reads these records. PostCSS nodes stop here on purpose --
  * twice, a check reached past the decoder to the name as written and let an
@@ -129,7 +144,7 @@ function declarations(css: string): Declared[] {
   postcss.parse(css).walkDecls((declaration) => {
     found.push({
       name: decodeIdentifier(declaration.prop),
-      value: decodeIdentifier(declaration.value).trim(),
+      value: declaration.value.trim(),
       references: referenceProjection(declaration.value),
       selectors: selectorsOf(declaration),
       unconditional: unconditional(declaration),
@@ -176,6 +191,24 @@ function definedTypographyTokens(css: string): Set<string> {
 
 function offTokenValues(css: string, property: string, allowed: RegExp, literals?: Set<string>): string[] {
   return valuesOf(css, property).filter((value) => !allowed.test(value) && !(literals?.has(value) ?? false));
+}
+
+/** A typography declaration is one this guard reads: an owned property, or a token definition. */
+function ownedByGuard(declaration: Declared): boolean {
+  return (
+    TYPOGRAPHY_PROPERTIES.has(propertyName(declaration)) ||
+    FAMILIES.some((family) => declaration.name.startsWith(`--${family}-`))
+  );
+}
+
+/**
+ * An escape inside a value changes which tokens the browser consumes, so the
+ * text is no longer what it looks like. The stylesheet has never needed one.
+ */
+function escapedValues(css: string): string[] {
+  return declarations(css)
+    .filter((declaration) => ownedByGuard(declaration) && declaration.value.includes("\\"))
+    .map((declaration) => `${declaration.name}: ${declaration.value}`);
 }
 
 function danglingReferences(css: string): string[] {
@@ -292,6 +325,10 @@ describe("typography tokens", () => {
 
   it("resolves each role to a step that is actually declared", () => {
     expect(unresolvedRoles(stylesheet)).toEqual([]);
+  });
+
+  it("spells every value it owns without escapes", () => {
+    expect(escapedValues(stylesheet)).toEqual([]);
   });
 
   it("resolves every role to a raw step rather than a bare length", () => {
@@ -455,6 +492,16 @@ describe("the guard itself", () => {
 
     const step = ':root { --fs-13: "poison" 0.8125rem; }';
     expect(malformedSteps(step)).toEqual(['--fs-13: "poison" 0.8125rem']);
+  });
+
+  it("refuses an escape in a value, which does not mean what its text says", () => {
+    const size = String.raw`body { font-size: var\28--text-ui\29; }`;
+    expect(escapedValues(size)).toEqual([String.raw`font-size: var\28--text-ui\29`]);
+    expect(danglingReferences(size)).toEqual([]);
+
+    const role = String.raw`:root { --text-ui: var\28--fs-13\29; }`;
+    expect(escapedValues(role)).toEqual([String.raw`--text-ui: var\28--fs-13\29`]);
+    expect(malformedRoles(role)).toEqual([String.raw`--text-ui: var\28--fs-13\29`]);
   });
 
   it("does not read a commented-out declaration as a violation", () => {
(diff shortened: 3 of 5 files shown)


baixiaohang (changes requested):
- Rationale: The token layer is coherent, but its regression guard does not validate the complete role-to-raw-step chain and can pass a broken font-size token.

**Risk level**: C
- Path baseline: changes are limited to `apps/web/**` -> C
- Semantic lift: none; this does not touch subscriptions, WebSocket parsing, or the data layer

**PR summary**
- Problem: Operators and Team admins currently see an inconsistent type hierarchy because component-local literals have produced dozens of nearly duplicate sizes and weights, while shadowed declarations make some edits ineffective.
- Approach: Replace point-of-use typography literals with raw size steps and semantic roles, rebind reading surfaces to a roomier density, normalize CJK fallbacks and numeric rendering, and add a stylesheet guard.
- Impacted modules: `apps/web/src/styles.css`, the Web typography test, and the document color-scheme declaration

**Review findings**
❌ 1. The guard does not validate the role-to-step reference edge. `references only tokens that exist` scans only property declarations and its `defined` set excludes `--fs-*`; the role test checks only that a value looks like `var(--fs-N)`. As a result, changing `--text-ui` to `var(--fs-15)` (with no `--fs-15` declaration) passes all seven tests while the browser drops every affected `font-size`. Please validate every raw-step reference used by a role and ensure every `--fs-*` declaration is included in the checks, so the test enforces the invariant it advertises. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:51`]

❓ 2. Please confirm whether the invitation dialog is intentionally a dense surface. The larger reading roles are scoped to `.settings-page`, but `InvitationDialog` is rendered as a sibling of the routed page under `AppShell`, so opening it while working in Team Settings still uses the root 13px/14px roles. If the invitation flow is part of the configuration surface, give `.invitation-dialog` its intended bindings explicitly. [R5 / `apps/web/src/styles.css:89`]

✅ 3. The final stylesheet consistently routes font sizes, weights, tracking, and font stacks through semantic tokens, and the CJK and tabular-numeral changes stay localized to presentation.

**Action taken**
- Submitted a request-changes review on the live replacement PR #88; #87 carries the same head but is being superseded because its branch prefix cannot pass CI.

Gandy2025:
Both findings addressed in `03f7163` and `b60d4cf`.

**R1 — the guard did not enforce its own invariant.** Confirmed and reproduced: setting `--text-ui: var(--fs-15)` with no `--fs-15` declared passed all seven assertions. Two changes:

- Reference checking now collects *every* typography token reference in the stylesheet, wherever it appears, rather than only property declarations, and `--fs-*` is in the defined set. A role pointing at an undeclared step is now a dangling reference like any other.
- A separate assertion checks the role-to-step edge directly, so the failure names the step rather than only the referencing property.

The same injection now fails both assertions with `[ 'fs-15' ]`.

**R2 — the invitation dialog was unintentionally dense.** You are right that it is a sibling of the routed page under `AppShell` and inherits nothing from `.settings-page`. It should not be dense: a modal is filled in, not scanned, whatever surface it was opened from. `.dialog-card` now binds to the reading roles directly, which covers the invitation dialog and the New Agent dialog together, and the comment above the rule records why the binding is explicit rather than inherited.

Separately, CI caught something your review did not cover: the guard located the stylesheet relative to the working directory, which is `apps/web` for the package test script but the repository root under the coverage config, so `pnpm test:coverage` failed to load the suite at all. Fixed in the same commit by probing both roots. `pnpm check`, `pnpm build`, `pnpm typecheck` and `pnpm test:coverage` pass locally now, with the web suite at 190 tests.


---

baixiaohang (changes requested):
- Rationale: The previous findings are fixed, but the expanded dangling-reference guard still omits the line-height token family and can pass a broken `--lh-*` reference.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and Team admins need a consistent, maintainable type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density by surface, normalize font fallbacks and numeric rendering, and enforce the token graph with stylesheet assertions.
- Impacted modules: Web stylesheet, typography guard, and document color-scheme metadata

**Review findings**
❌ 1. The new global reference check says it covers every typography token, but both `TOKEN_REFERENCE` and `definedTypographyTokens()` omit the `lh` family. Changing an existing declaration to `line-height: var(--lh-prsoe)` therefore still passes every dangling-reference assertion while the browser drops the declaration. Since this PR explicitly centralizes line heights behind `--lh-*`, please include `lh` in the defined/reference families (and keep any intentional literal line-height exceptions explicit) so the generalized guard enforces the complete typography graph. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:23`]

✅ 2. The original role-to-raw-step blocker is fixed: `--fs-*` is now part of the defined set and role edges are checked directly.
✅ 3. The dialog-density question is resolved coherently by binding `.dialog-card` to the reading roles, covering both invitation and New Agent dialogs regardless of their mount point.
✅ 4. Stylesheet lookup now supports both package-local and repository-root test working directories.

**Action taken**
- Submitted a follow-up request-changes review on head `b60d4cfff68dbf9f35b5aa4c3be45e4dc556e964`.

Gandy2025:
Fixed in `d633263`. You are right — `--lh-*` was missing from both the reference pattern and the defined set, so `var(--lh-prsoe)` was invisible to every assertion.

The root cause was that the family list existed twice, in a regex and in an array, and only one of them was ever updated. Both now derive from a single `FAMILIES` constant, so adding a family extends every check at once.

On keeping the literal exceptions explicit: there are exactly two, and they are box geometry rather than text rhythm — `line-height: 1` centres a single glyph in its own line box (chevron, overflow dots, dialog close), and `line-height: 38px` matches the height of the `OT` badge it labels. Both are now a named allowlist, and any other literal line height fails.

Verified by injection on the current head: `var(--lh-prsoe)` fails the dangling-reference assertion with `[ 'lh-prsoe' ]`, and changing a declaration to `line-height: 1.42` fails the new assertion with `[ '1.42' ]`.

`pnpm check`, `pnpm typecheck` and `pnpm test:coverage` pass locally; the web suite is at 191 tests.


---

yuezengwu (changes requested):
- Rationale: The earlier token-reference, dialog-density, coverage-path, and line-height findings are fixed, but the guard still permits CSS shorthand or a literal family to bypass the token layer entirely.

**Review finding**

❌ **The guard does not reject `font` shorthand or literal `font-family`.** Its property checks inspect only the `font-size`, `font-weight`, `line-height`, and `letter-spacing` longhands, while `font-family` is only included in dangling-token reference checks. On exact head `d633263eabc0ee056b737d0098429ce00215dd6d`, adding `font: 13px/1 Arial, sans-serif` to `body` still leaves all 9 typography assertions green even though that one declaration resets the size, line height, weight, and family outside every token. A separate `font-family: Arial, sans-serif` injection also leaves all 9 green. Please reject non-allowed `font` shorthand (keeping the existing `font: inherit` exception explicit) and require every `font-family` declaration to resolve through `--font-*`, with regression cases for both injections. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:42`]

✅ The prior `--lh-*` dangling-reference hole is fixed, the two geometric line-height literals are explicitly allowlisted, `.dialog-card` now receives reading density, and stylesheet lookup works from both supported working directories.

**Validation**
- `pnpm check`, `pnpm build`, `pnpm typecheck`: pass.
- `pnpm --filter @opentag/web test`: 190/190 pass on the prior head; the new exact-head typography suite is 9/9 pass.
- `pnpm test:coverage`: reaches 983/987 locally; the only four failures are unrelated `/var` versus `/private/var` temp-path canonicalization assertions outside `apps/web`.

**Action taken**
- Requested changes on exact head `d633263eabc0ee056b737d0098429ce00215dd6d`.

Gandy2025:
@yuezengwu confirmed and fixed in `e1adc47`. Reproduced both injections on `d633263`: `font: 13px/1 Arial, sans-serif` on `body` and a literal `font-family: Arial, sans-serif` each left all 9 assertions green.

- `font-family` must now resolve through `--font-*`. It was previously only visible to the dangling-reference check, which sees `var()` references and therefore nothing at all in a literal stack.
- The `font` shorthand is allowed solely as `inherit` — the declaration that lets form controls adopt the surrounding font — and the allowlist is named, like the two line-height geometry exceptions.

On the regression cases: the checks now take the CSS text as an argument rather than closing over the stylesheet, so the suite tests the guard as well as the stylesheet. `describe("the guard itself")` covers both of your injections, a literal in each of the four families, a dangling reference in two families at once, and one case pinning the boundary this fix depends on — that `font-size:` and `font-synthesis:` are not read as the `font:` shorthand.

The suite is 17 tests, and `pnpm --filter @opentag/web test` is at 199. `pnpm check` and `pnpm typecheck` pass.

Three review rounds have now found three separate holes in this guard, each a family or property I had not enumerated. The rewrite addresses that directly: the family list is a single constant that drives every check, and the guard's own behaviour is now asserted rather than assumed.

Still open on my side, unrelated to your review: `Node 26 Compatibility` is red on this head while Node 22 is green, failing in `@opentag/web#test`. The same job is red on `main` at `e1d2cd1` and was green on this branch two heads ago, so I am treating it as pre-existing flake for now — I will confirm against this push and report rather than leave it unexplained.


---

baixiaohang (approved):

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All review blockers are closed; the guard now rejects shorthand and literal-family bypasses and proves those failure paths with focused regression cases.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and enforce the complete token graph with stylesheet assertions.
- Impacted modules: Web stylesheet, typography guard, and document color-scheme metadata

**Review findings**
✅ 1. The role-to-step edge, every typography token family (including `--lh-*`), and dangling references now derive from shared guard infrastructure.
✅ 2. Literal `font-family` declarations and non-`inherit` `font` shorthand are rejected, with regression cases for both bypasses and for the longhand/shorthand parsing boundary.
✅ 3. Dialog density is explicit, and stylesheet lookup works from both supported test working directories.

**Action taken**
- Approved exact head `e1adc476b0440b4916ebf0a09e13f30835ffbece`.

yuezengwu (changes requested):
- Rationale: The previously reported `font` shorthand and literal `font-family` bypasses are fixed, but the raw-text guard treats typography token definitions inside CSS comments as live declarations.

**Review finding**

❌ **A commented token definition satisfies the guard even though the browser has no such token.** On exact head `e1adc476b0440b4916ebf0a09e13f30835ffbece`, I changed the live role to `--text-ui: var(--fs-15)` and added only `/* --fs-15: 0.9375rem; */`. All 17 typography tests still passed. `definedTokens()` scans the unparsed stylesheet with `--(${prefix}-...):`, so it counts the commented text as a definition; in the browser that custom property does not exist, leaving the live role invalid and causing affected `font-size` declarations to be dropped. Please strip comments before all definition/reference/declaration analysis or use a CSS parser, and add a regression case where a live role references a step that exists only in a comment. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:58`]

✅ I independently injected `font: 13px/1 Arial, sans-serif` and `font-family: Arial, sans-serif` into the actual stylesheet on this head; the current guard rejects both. The shared family list covers all six intended families, and the new focused guard tests exercise the real helper functions.

**Validation**
- `pnpm check`, `pnpm build`, `pnpm typecheck`: pass.
- `pnpm --filter @opentag/web test`: 199/199 pass.
- `pnpm test:coverage`: 992/996 pass locally; the four failures are unrelated macOS `/var` versus `/private/var` temp-path canonicalization assertions outside `apps/web`.
- All six GitHub checks are green.

**Action taken**
- Requested changes on exact head `e1adc476b0440b4916ebf0a09e13f30835ffbece`.

baixiaohang (changes requested):
- Rationale: A valid follow-up finding shows that the raw-text guard can still accept a typography token that exists only inside a CSS comment, so my earlier approval is no longer supported.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and enforce the token graph with stylesheet assertions.
- Impacted modules: Web stylesheet, typography guard, and document color-scheme metadata

**Review findings**
❌ 1. Confirmed by static inspection: `definedTokens()`, reference collection, and declaration checks analyze unparsed CSS text, so `/* --fs-15: 0.9375rem; */` is counted as a live definition. A live `--text-ui: var(--fs-15)` can therefore satisfy the guard while remaining invalid in the browser. Strip comments before every analysis pass or parse the CSS, and add the reported live-reference/comment-only-definition regression case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:58`]

✅ 2. The earlier role-edge, line-height, dialog-density, shorthand, and literal-family findings remain fixed.

**Action taken**
- Superseded my approval with a request-changes review on exact head `e1adc476b0440b4916ebf0a09e13f30835ffbece`.

Gandy2025:
@yuezengwu @baixiaohang confirmed and fixed in `b49206e`. Reproduced your exact case on `e1adc47`: `--text-ui: var(--fs-15)` live with only `/* --fs-15: 0.9375rem; */` alongside it, and all 17 tests stayed green. After the fix that injection fails two assertions — the dangling reference and the role edge.

The fix strips comments in `captures`, which is the single funnel every definition, reference and declaration check runs through, so one call site covers all of them rather than each check remembering to do it. That also closes the matching false positive: a rule parked inside a comment can no longer be reported as a violation. Both directions are asserted — comment-only definition, and commented-out declaration.

Suite is 19 tests, `pnpm --filter @opentag/web test` at 201, `pnpm check` and `pnpm typecheck` pass.

Worth saying plainly: this is the fourth hole found in this guard, and the pattern across all four is the same — I wrote a regex over raw text and assumed it saw what the browser sees. A parser would not have had any of these four bugs. I went with regex because the alternative is a new dependency for one test file, and I would rather that call be yours than mine: if you would prefer this parse the CSS properly, say so and I will switch it to a real parser. Otherwise the comment strip closes the class of bug you found, and the guard's own behaviour is now covered by 8 regression cases.

Separately, `Checks` failed on `e1adc47` at step 11, `Test PostgreSQL integration` — lint, build, typecheck and the full test step all passed before it. That step touches nothing this PR changes, it passed on the two earlier heads of this branch, and I re-ran the failed jobs once to confirm. This push supersedes that run; I will report the outcome either way rather than let it pass unexplained.


---

baixiaohang (changes requested):
- Rationale: Comment stripping fixes the reported case, but token-like text inside a CSS string still satisfies the raw-text guard while defining nothing in the browser.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and enforce the token graph with stylesheet assertions.
- Impacted modules: Web stylesheet, typography guard, and document color-scheme metadata

**Review findings**
❌ 1. The comment-only definition case is fixed, but the same raw-text problem remains for CSS strings. With a live `--text-ui: var(--fs-15)` and `content: "--fs-15: 0.9375rem;"`, `definedTokens()` treats the string as the missing definition; the raw-step value/name scans also accept it as a valid 15px step. The browser defines no custom property there, so affected font sizes are still dropped while every current assertion remains satisfied. Given the repeated parser-boundary failures and the author's explicit question, please switch this guard to a real CSS parser (or an equivalently complete tokenizer), then add this live-reference/string-only-definition case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:55`]

✅ 2. Centralizing comment removal in `captures()` correctly fixes both directions of the immediately reported comment case.

**Action taken**
- Requested changes on exact head `b49206ec4da020e077b2f65f65b12ebc38f40644`.

baixiaohang:
The current CI failures are not the previously mentioned PostgreSQL flake. The pull-request merge ref includes `c62a2eb` from current `main` (#90), which added:

```css
.connect-command-meta {
  font-size: 0.84rem;
}
```

The new typography guard correctly rejects that literal in the main Test job and in both Node compatibility jobs. Please refresh this branch from current `main` and route `.connect-command-meta` through the appropriate semantic text role as part of this PR. This is exactly the integration drift the invariant is meant to catch; the CLI and Container jobs are the only green jobs on the current head.

Gandy2025:
@baixiaohang you are right, and thank you for the pointer — I had called the earlier red runs a PostgreSQL flake, and that was wrong. Corrected in `8f26e7f` (merge of current `main`) and `59436c0`.

The diagnosis holds exactly as you describe: checks run against the merge ref, `main` moved under the branch, and `.connect-command-meta { font-size: 0.84rem }` from #90 is what the guard was rejecting. My local runs stayed green because they were testing the branch alone — the web suite was 201 tests locally against 217 in CI, which was the tell I should have followed several rounds ago instead of reaching for "flake".

`.connect-command-meta` now resolves through `--text-meta`, matching `.field-hint`, the closest analogue — both are muted supporting text. The rule renders inside `.onboarding-shell`, where that role is 13px, so the rendered size lands within half a pixel of the 13.44px it replaces.

On the merged head: `pnpm check`, `pnpm build`, `pnpm typecheck` pass, and `pnpm --filter @opentag/web test` is 217/217, matching the count CI sees.

Two of the four earlier red runs I attributed to flake were this same drift, and the guard was correct every time. That is a better argument for the invariant than anything in the PR description.

On your parser request: agreed, and I would rather not hand you a fifth boundary bug. I will switch the guard to a real CSS tokenizer in the next push rather than patch the string case — `postcss` is already in the dependency tree via Vite, so parsing costs no new direct dependency. That will subsume both the comment case and the `content: "--fs-15: …"` case you just raised, and I will add your string-only-definition case alongside the existing regression tests.


---

Gandy2025:
The guard now parses the stylesheet with postcss instead of matching it — `b5af01c`. @baixiaohang @yuezengwu

Three of the holes you found were one mistake repeated: reading CSS with regular expressions. `font:` matched as a longhand prefix, `/* --fs-15: … */` counted as a definition, and `content: "--fs-15: …"` counted too. Each patch closed one case and left the next, which is why the parser was the right call rather than a fourth patch.

What the parse changes:

- Comments are their own nodes and never reach a declaration check, in either direction.
- Property names are compared as parsed names, so `font` cannot collect `font-size` and no lookbehind is load-bearing.
- Quoted runs drop out of a value before it is read for token references or step definitions.

postcss is already in the tree as a Vite dependency, so declaring it in `apps/web` adds three lines to `pnpm-lock.yaml` and no new packages.

Verified against the real stylesheet, not just fixtures — each of the five injections raised across this review, applied to `styles.css` and reverted:

| injection | round | result |
| --- | --- | --- |
| `content: "--fs-15: 0.9375rem;"` with a live `--text-ui: var(--fs-15)` | this one | 2 failed |
| `/* --fs-15: 0.9375rem; */` with the same live role | previous | 2 failed |
| `font: 13px/1 Arial, sans-serif` | earlier | 1 failed |
| `font-family: Arial, sans-serif` | earlier | 1 failed |
| `font-size: 0.83rem` | first | 1 failed |

The suite is 20 tests: 10 asserting the stylesheet, 10 asserting the guard rejects what it exists to catch — now including the string-only definition, a token reference quoted inside a value (which must *not* be flagged), and a check that every `--text-*` role resolves to a raw step rather than merely looking like one.

`pnpm check`, `pnpm build`, `pnpm typecheck` pass; `pnpm --filter @opentag/web test` is 218/218. CI was green on the previous head `59436c0` across all six jobs.


---

baixiaohang (changes requested):
- Rationale: Parsing closes the syntax-boundary holes, but the guard still treats a token declared under an unrelated selector as available to root-level roles.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and parse the stylesheet to enforce the token graph.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. `definedTokens()` walks every declaration without considering its parent selector. Moving root `--fs-13: 0.8125rem` into `.dead { --fs-13: 0.8125rem; }` therefore still satisfies `definedTypographyTokens()`, `unresolvedRoles()`, and `rawSteps()`, while root `--text-ui: var(--fs-13)` cannot resolve in the browser and affected font sizes are dropped. Please require the baseline typography families/raw steps to be defined in `:root` (while validating intentional scoped `--text-*` rebindings separately), and add this moved-definition regression case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:73`]

✅ 2. PostCSS correctly eliminates the comment, string-as-declaration, and shorthand/longhand boundary failures, and it is now an explicit Web dev dependency.
✅ 3. The current-main `.connect-command-meta` drift is integrated and mapped coherently to `--text-meta` / `--lh-ui`.

**Action taken**
- Requested changes on exact head `b5af01c5d26c1d40ccf0a02fcebd166da18d0afa`.

Gandy2025:
@baixiaohang confirmed and fixed in `687513e`. A custom property exists only where its selector matches, and the guard was counting a definition wherever it appeared — parsing fixed how declarations are *read*, not where they *apply*.

The invariant is now scope-aware in both directions:

- References, raw steps and role edges resolve against `:root` alone.
- Anything declared elsewhere must be a rebinding of a role `:root` already declares.

That keeps the two intentional surface blocks legal — `.settings-page, .onboarding-shell, .decorative-page, .dialog-card` retuning `--text-*` for reading density — while rejecting both a scoped raw step and a newly invented token, since nothing outside that surface could resolve either. Your `.dead { --fs-13: … }` case now fails three assertions: the dangling reference, the unresolved role edge, and the new scoped-definition check.

Verified against the real stylesheet, all six injections raised in this review, applied to `styles.css` and reverted:

| injection | result |
| --- | --- |
| `--fs-13` moved from `:root` into `.dead` | 3 failed |
| `content: "--fs-15: 0.9375rem;"` with a live `--text-ui: var(--fs-15)` | 2 failed |
| `/* --fs-15: 0.9375rem; */` with the same live role | 2 failed |
| `font: 13px/1 Arial, sans-serif` | 1 failed |
| `font-family: Arial, sans-serif` | 1 failed |
| `font-size: 0.83rem` | 1 failed |

The suite is 23 tests. `pnpm check` and `pnpm typecheck` pass; `pnpm --filter @opentag/web test` is 221/221.

One note on my own process, since this is the sixth finding: I wrote the fixture for the new "surface retunes a role" case with a reference to `--fs-13` while declaring only `--fs-14`, and the guard failed it correctly. I fixed the fixture rather than the guard. Worth stating explicitly because on a guard this heavily revised, the tempting move is to assume a red test means the guard is wrong.


---

baixiaohang (changes requested):
- Rationale: The moved-definition case is fixed, but four remaining CSS/cascade paths can still bypass the invariant without failing the guard.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and parse the stylesheet to enforce the token graph.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. Ordinary CSS property names are ASCII case-insensitive, but `valuesOf()` compares `declaration.prop === property`. `FONT-SIZE: 13px` (and mixed-case `Font`, `FONT-FAMILY`, etc.) is applied by the browser but invisible to every longhand/shorthand check. Normalize non-custom property names before comparison and add an uppercase injection case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:83`]

❌ 2. `rawSteps()` silently drops malformed root steps instead of reporting them. Replacing `--fs-13: 0.8125rem` with `--fs-13: 13.5px` leaves the token defined, keeps role/dangling checks green, and is omitted from `rawSteps()` while the other steps satisfy the non-empty assertion; the browser then renders the off-scale 13.5px value. The same skip permits a referenced `--fs-small`. Require a one-to-one match between every root `--fs-*` declaration and a valid numeric-name/rem-value step, then validate all of them. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:135`]

❌ 3. `atRoot()` checks only the immediate rule selector, not conditional ancestors. A `:root { --fs-13: ... }` nested under `@media (width: 0px)` is counted as a global baseline, so an outside root role resolves in the guard but not for normal browser viewports. Require baseline definitions to be unconditional at `:root` (or explicitly reject conditional at-rule ancestors) and cover this case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:70`]

❌ 4. Scoped role rebindings are not restricted to the intended reading surfaces. `.row { --text-ui: var(--fs-30); }` passes because the role exists at root, allowing component-local typography drift even though this PR's density model is surface-owned. Keep a named selector/block allowlist for `.settings-page`, `.onboarding-shell`, `.decorative-page`, and `.dialog-card` (or an equivalent explicit policy), and reject arbitrary component rebindings. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:201`]

✅ 5. Parsing and root-only definition lookup correctly close the previously reported comment, string, shorthand/longhand, and unrelated-selector cases; current-main typography drift remains integrated.

**Action taken**
- Requested changes on exact head `687513e08e40419cfbff20d9fac9cd196b271351`.

Gandy2025:
@baixiaohang all four confirmed and fixed in `f2c3f55`.

1. **Case.** Ordinary property names are ASCII case-insensitive, custom property names are not. Non-custom names are now compared in lower case; `--fs-13` stays case-sensitive, as CSS defines it.
2. **Malformed steps.** `stepOf()` now runs per declaration, and `malformedSteps()` reports anything under a root `--fs-*` that does not describe a step. The suite also asserts a one-to-one match between root `--fs-*` declarations and parsed steps, so nothing can be dropped and covered by the others satisfying a non-empty check. Your `--fs-13: 13.5px` and a referenced `--fs-small` both surface.
3. **Conditional ancestors.** Baseline definitions must be unconditional — any at-rule ancestor disqualifies them, not just `@media`. Worth noting the stylesheet has fourteen typography declarations inside media queries; those are *usage*, which every property check still covers regardless of ancestry. Only *definitions* are required to be unconditional.
4. **Rebinding allowlist.** `READING_SURFACES` names the four selectors that own a density, and every selector in a rebinding rule's list must be one of them.

Verified against the real stylesheet — all ten injections raised across this review, applied to `styles.css` and reverted:

| injection | round | result |
| --- | --- | --- |
| `FONT-SIZE: 13px` | this one | 1 failed |
| `--fs-13: 13.5px` | this one | 1 failed |
| `--fs-13` moved under `@media (width: 0px)` | this one | 3 failed |
| `.row { --text-ui: var(--fs-30) }` | this one | 1 failed |
| `--fs-13` moved into `.dead` | previous | 3 failed |
| `content: "--fs-15: 0.9375rem;"` | earlier | 2 failed |
| `/* --fs-15: 0.9375rem; */` | earlier | 2 failed |
| `font: 13px/1 Arial, sans-serif` | earlier | 1 failed |
| `font-family: Arial, sans-serif` | earlier | 1 failed |
| `font-size: 0.83rem` | earlier | 1 failed |

28 tests. `pnpm check` and `pnpm typecheck` pass; `pnpm --filter @opentag/web test` is 226/226.

A note on where this has got to. Seven rounds, and the pattern is consistent: the stylesheet change itself has drawn no findings since round one, while the guard has drawn eleven. That is the honest cost of asserting an invariant over a language with a cascade — every round has been a real hole, and I would rather have them than not.

But the guard is now a large share of the diff, and the last two rounds moved from "the guard misreads CSS" to "the guard permits a policy violation" — case 4 is a design decision about who may own density, not a parsing bug. If you would rather land the typography change on its current, unchallenged stylesheet and take the guard as a follow-up PR where it can be reviewed on its own terms, say so and I will split it. If you would rather keep going here, I will keep going — your call, since it is your review time either way.


---

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All reported syntax, reference, cascade-scope, and ownership bypasses are closed; the guard now enforces the same token-layer boundary the stylesheet implements.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and parse the stylesheet to enforce the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
✅ 1. Property checks now follow browser case semantics, reject shorthand/literal bypasses, and operate on parsed declarations rather than comments or quoted content.
✅ 2. Root raw steps are exhaustively enumerated, validated for shape/value/name, and resolved only from an unconditional `:root`; dangling and role-edge checks share that baseline.
✅ 3. Scoped role rebindings are limited to the four explicit reading surfaces, preserving the surface-owned density model while rejecting component-local drift.
✅ 4. Current-main changes are integrated, including `.connect-command-meta` resolving coherently through `--text-meta` / `--lh-ui`.

**Action taken**
- Approved exact head `f2c3f557c5c0a7384555594b5a7e18789047e871`.
- The guard should remain in this PR: it is the atomic enforcement for the migration and has already caught real integration drift from `main`.

yuezengwu (changes requested):
- Rationale: The reported comment/string, scope, case, malformed-step, conditional-definition, and direct component-rebinding holes are fixed, but two browser-applied CSS paths still bypass the guard.

**Keep / Split**
- **Keep (recommended).** The parser guard is the atomic enforcement for this migration, has already caught real current-main drift, and PostCSS adds no new package to the dependency graph. Splitting would defer the invariant without reducing the review work; close the two concrete gaps below in this PR.

**Review findings**

❌ **1. Escaped CSS identifiers remain invisible to the parsed-name checks.** PostCSS preserves identifier escapes in `Declaration.prop`, while the browser decodes them before property/custom-property matching. On exact head `f2c3f557c5c0a7384555594b5a7e18789047e871`, adding `f\6f nt-size: 23px` to the actual stylesheet left all 28 typography tests green; headless Chrome applied it as `font-size` and computed `23px`. Likewise, `.row { --text\2d ui: var(--fs-30); }` left all 28 tests green, while Chrome resolves that escaped custom-property identifier as `--text-ui`, allowing a component-local role retune outside the allowlist. Please decode CSS identifier escapes before comparison, then apply ASCII case folding only to ordinary property names while keeping decoded custom-property names case-sensitive; add regressions for both escaped ordinary and custom-property identifiers. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:100`]

❌ **2. A `:root` selector list can hide a component rebinding from the allowlist.** `atRoot()` classifies the whole declaration as root-owned when any selector equals `:root`, and `scopedDefinitions()` then drops every other selector in that rule. Appending `:root, .row { --text-ui: var(--fs-13); }` to the actual stylesheet left all 28 tests green. In a `.settings-page`, the explicit `.row` definition overrides the inherited reading-surface `--text-ui: var(--fs-14)`; headless Chrome computed the row at `13px`. Please classify each selector independently or reject mixed `:root`/non-root definition lists, and add this exact regression so an arbitrary selector cannot piggyback on the baseline. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:88`]

✅ The previously reported injections are now represented by parsed-helper regression cases and the 28-case guard suite passes on the exact stylesheet. Root raw steps are exhaustively validated, at-rule baseline definitions are excluded, ordinary property case is normalized, and current-main `.connect-command-meta` drift is integrated.

**Validation**
- `pnpm check`, `pnpm build`, `pnpm typecheck`: pass.
- `pnpm --filter @opentag/web test`: 226/226 pass.
- `pnpm test:coverage`: all Web/typography coverage tests pass; the command exits on the same four unrelated macOS `/var` versus `/private/var` temp-path assertions in CLI/client tests.
- All six GitHub checks are green.

**Action taken**
- Requested changes on exact head `f2c3f557c5c0a7384555594b5a7e18789047e871`.

baixiaohang (changes requested):
- Rationale: Two browser-applied CSS paths remain invisible to the parsed guard, so my approval on this head is no longer supported.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and parse the stylesheet to enforce the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. Confirmed by static inspection: `propertyName()` compares PostCSS's escaped spelling directly. CSS identifier escapes are decoded by the browser, so `f\6f nt-size` applies as `font-size` while remaining invisible to the longhand check, and `--text\2d ui` applies as the case-sensitive custom property `--text-ui` while evading the scoped-rebinding policy. Decode identifier escapes before comparison, then case-fold only ordinary properties, and add regression cases for both ordinary and custom-property identifiers. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:100`]

❌ 2. `atRoot()` returns true when any selector in a rule is `:root`, causing `scopedDefinitions()` to discard the declaration for every sibling selector. Thus `:root, .row { --text-ui: var(--fs-13); }` is accepted as baseline even though `.row` can override the intended reading-surface binding. Classify selectors independently or reject mixed root/non-root definition lists, and cover this exact case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:88`]

✅ 3. The earlier parser-boundary, raw-step, conditional-scope, and direct component-rebinding fixes remain valid. The guard should stay in this PR as the migration's atomic enforcement.

**Action taken**
- Superseded my approval with request changes on exact head `f2c3f557c5c0a7384555594b5a7e18789047e871`.

Gandy2025:
@yuezengwu @baixiaohang both confirmed and fixed in `9c7df38`.

**Escaped identifiers.** Names are decoded before any comparison, then case-folded only when they are ordinary properties — decoded custom property names stay case-sensitive. Values are decoded too, after strings are stripped rather than before, since decoding first can manufacture a quote and break string detection. Every name-matching site now reads the decoded name: definitions, root steps, role edges, scoped definitions, and the property checks.

**Mixed selector lists.** A definition rule counts as baseline only when every selector is `:root`. A mixed list falls through to the surface policy, which rejects it — `:root, .settings-page` is refused as well, since `:root` is not a reading surface. Both directions are covered.

Regression cases added for the escaped ordinary property, the escaped custom property, decoded-but-case-sensitive custom names, and the mixed list.

I re-ran every injection from all eight rounds against the real stylesheet, not just the new two — escaped property, escaped custom property, mixed list, uppercase property, malformed step, step hidden in `@media`, component rebinding, step in a string, step in a comment, `font` shorthand. All ten fail the guard; none regressed. `pnpm check`, `pnpm typecheck` pass and the web suite is 230.

Eight rounds, thirteen holes, every one of them real. The pattern is consistent and worth naming: each hole was a place where my reading of the CSS diverged from the browser's. Moving to PostCSS fixed the lexical ones but not this class, because the parser hands over what was written while the browser matches what it resolves — escapes, case, selector lists, at-rule conditions. Thank you both for the specificity; the headless-Chrome confirmations in particular made these unambiguous rather than arguable.


---

baixiaohang (changes requested):
- Rationale: The two reported bypasses are fixed, but one raw-name role count still lets an escaped root role override evade the role-to-step invariant.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and parse the stylesheet to enforce the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. The production assertion for "every role resolves to a raw step" still counts roles with `/^--text-[a-z]+$/.test(declaration.prop)` instead of `declaredName(declaration)`. Appending a later root declaration such as `:root { --text\2d ui: 13.5px; }` therefore overrides `--text-ui` in the browser, while the escaped declaration is omitted from the role count; `roleEdges()` decodes it but drops it because the value is not a step, leaving the two counts equal and the guard green. Count and validate roles from the same decoded declaration set (preferably as a one-to-one malformed-role check), and add this escaped root-override regression case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:262`]

✅ 2. Identifier decoding now reaches the reusable name-matching helpers, preserves custom-property case sensitivity, and correctly classifies mixed `:root`/non-root selector lists as scoped. The earlier fixes remain intact.

**Action taken**
- Requested changes on exact head `9c7df386f3ea7e551f23b9565e35ebf170b75545`.

Gandy2025:
@baixiaohang confirmed and fixed in `152046c`. You were precise about it: the role count tested the name as written, so an escaped root role overrode `--text-ui` in the browser, went uncounted as a role, and was dropped by `roleEdges()` for naming no step — both counts equal, guard green.

Two things changed, because this was the second check to reach past the decoder and a third call-site patch would not have been an answer.

**The spelling is now unreachable.** The parse layer yields records carrying the decoded name, the readable value, the selectors, and whether the declaration is unconditional. PostCSS nodes stop there. `Declaration` appears in exactly two functions, both of which build those records, so no check can read a written name again — the class is closed rather than patched.

**Roles now follow the shape steps already use**, which is what you suggested when you raised the same weakness for `rawSteps()`: one helper enumerates every role declaration, and a role naming no step is reported by name instead of inferred from a count that happens to match. I had fixed that shape for steps and left it for roles.

Worth reporting plainly: refactoring the parse layer broke string removal, and the guard's own quoted-reference case caught it before I pushed. The suite testing the guard rather than only the stylesheet paid for itself here.

Twelve injections re-run against the real stylesheet, all rejected: the new escaped root-role override, escaped property, escaped custom property, mixed `:root` list, uppercase property, malformed step, step hidden in `@media`, component rebinding, step in a string, step in a comment, `font` shorthand, literal `font-family`. `pnpm check` and `pnpm typecheck` pass; the web suite is 231.


---

baixiaohang (changes requested):
- Rationale: The escaped-role hole is fixed, but the shared declaration record now makes property, role, and step validation lossy by removing CSS strings from the value the browser actually evaluates.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and parse the stylesheet to enforce the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. `Declared.value` is populated with `readableValue()`, which removes every quoted string before all downstream checks. Strings are ignorable when *scanning for token references*, but they remain semantic tokens in the value the browser validates. For example, `body { font-size: "poison" var(--text-ui); }` is recorded as `var(--text-ui)` and passes the font-size check; after substitution the browser rejects the original token sequence at computed-value time, so the declaration falls back instead of producing the token size. The same loss lets `--text-ui: "poison" var(--fs-13)` pass `malformedRoles()` and `--fs-13: "poison" 0.8125rem` pass `stepOf()`. Preserve a lossless decoded value for property/role/step validation, derive a separate string-insensitive projection only for reference collection, and add at least the direct longhand plus one token-definition regression case. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:116`]

✅ 2. PostCSS spelling is now isolated behind decoded declaration records, malformed roles are enumerated directly, and the escaped root-role override is covered. The previous identifier and mixed-selector fixes remain valid.

**Action taken**
- Requested changes on exact head `152046c64014de5d3fa3b1900cff4fcd3f013296`.

Gandy2025:
@baixiaohang confirmed and fixed in `68bf60d`. This one was mine, introduced with the record refactor one commit earlier, and your framing of it is the correction I needed: dropping strings is right when looking for `var()` references and wrong everywhere else, because to the browser a quoted run is part of the value it validates.

A record now carries two things: the decoded value as written, which every property, role and step check reads, and a separate projection with quoted runs dropped that only reference collection reads. The projection still strips strings before decoding escapes, since decoding first can manufacture a quote.

Regression cases for all three losses you named — the direct longhand, the role definition, and the step definition.

I also checked the reverse, because a fix that only tightens things would have been a different bug: `content: "var(--fs-99)"` in the real stylesheet still raises nothing, so the two projections are genuinely separated rather than collapsed into the strict one.

Twelve injections re-run against the real stylesheet, all rejected, plus that one negative case passing. `pnpm check` and `pnpm typecheck` pass; the web suite is 232.

Ten rounds, sixteen findings. Every one has been real, and this round and the last were regressions I introduced while fixing the round before — which is its own signal. If you would rather stop reviewing this file and have me revert the guard to leave only the stylesheet migration, say so and I will; my read is still that it is worth keeping, since it has already caught real drift from `main` and caught my own regression before I pushed it, but you have spent more review on the guard than on the change it protects.


---

baixiaohang (changes requested):
- Rationale: The lossless/string-specific projections are now separated correctly, but textually decoding a CSS value still changes token boundaries and can turn an invalid identifier into an apparently valid `var()` call.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and parse the stylesheet to enforce the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. `value: decodeIdentifier(declaration.value)` decodes the value as plain text, but CSS resolves escapes *inside the token being consumed*; escaped punctuation does not become new syntax. `body { font-size: var\28--text-ui\29; }` is one identifier token to the browser, not a `var()` function, so the declaration is invalid. The guard textually turns it into `var(--text-ui)` and accepts it. The same construction in `--text-ui: var\28--fs-13\29` passes the role-edge check but cannot substitute a step in the browser. Preserve and inspect value token structure with a CSS value tokenizer/parser, or adopt the simpler canonical-source policy of rejecting escapes in typography declaration values, and add both direct-property and role regressions. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:130`]

✅ 2. The lossless value and reference-only projection are now separated correctly, including the three reported poisoned-string cases and the quoted-reference negative case. Name decoding and selector ownership remain fixed.

**Keep / Split**
- Keep the guard in this PR, but stop emulating browser value tokenization with decoded strings. A canonical no-escape policy is sufficient for this stylesheet if adding a value parser is disproportionate.

**Action taken**
- Requested changes on exact head `68bf60dbca3fbd5c732000a53bb80bb4ddbcd83b`.

Gandy2025:
@baixiaohang confirmed and fixed in `bb01745`, taking the canonical-source policy you recommended rather than the value parser.

The distinction you drew is the one I had been missing: a name is one identifier, so decoding it is exact; a value is a sequence of tokens, and CSS resolves escapes inside the token being consumed rather than across it. `var\28--text-ui\29` is a single identifier, not a call.

Values are now recorded as written, and an escape in a value this guard owns — one of the six properties, or a token definition — is refused outright. The stylesheet has never contained a backslash in any value, so the policy costs nothing. Names keep their decoding, which is exact and still needed. Regressions added for the direct property and the role.

I verified the policy did not overshoot: name-side escapes are still caught, and `content: "var(--fs-99)"` still raises nothing. `pnpm check` and `pnpm typecheck` pass; the web suite is 234.

Taking your advice to stop emulating browser tokenization is the actual fix here. Seven of the last eight findings came from me modelling CSS semantics by hand and getting a boundary wrong; a policy the stylesheet already satisfies removes that whole category instead of adding another case to it. If you spot anywhere else this guard is still emulating rather than restricting, I would rather narrow it the same way than keep matching the browser.


---

baixiaohang (changes requested):
- Rationale: The canonical no-escape policy closes owned typography values, but its scope is narrower than the global reference scan and still permits an escaped missing typography reference elsewhere in the stylesheet.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and enforce a canonical source policy over the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. `escapedValues()` rejects escapes only in the six owned properties or typography token definitions, while `danglingReferences()` deliberately collects typography-token references from every declaration. The scopes therefore disagree: `padding: v\61 r(--fs-99)` is a real missing `var(--fs-99)` reference in the browser, but `padding` is not `ownedByGuard`, the escape policy ignores it, and the raw `TOKEN_REFERENCE` pattern does not see it. Either preserve the advertised stylesheet-wide reference invariant by rejecting escapes in the non-string portion of every declaration value, or explicitly restrict reference collection to owned declarations and name the narrower invariant. Add the unowned-property escaped-reference regression for the chosen boundary. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:204`]

✅ 2. For owned properties and token definitions, keeping values canonical and escape-free correctly closes the token-boundary hole without emulating browser value parsing. Name-side escapes, string poisoning, and quoted non-references remain handled correctly.

**Keep / Split**
- Keep the guard in this PR. This is a boundary-alignment fix within the new canonical policy, not a reason to defer the invariant.

**Action taken**
- Requested changes on exact head `bb01745e9b96b6a58d098232cd5cc77aeb4f5a2f`.

Gandy2025:
@baixiaohang confirmed and fixed in `f4e6868`. I took the first of your two options — keeping the stylesheet-wide reference invariant — because the stylesheet contains no backslash anywhere, so the broader policy costs nothing and the narrower one would quietly weaken what the guard claims.

The policy now matches the scan exactly: references are read from every declaration, so every declaration's value must be escape-free outside strings — which is precisely the text the scan reads. Inside a string a backslash is ordinary content and no reference is read, so those are left alone; there is a regression case each way, `padding: v\61 r(--fs-99)` refused and `content: "\201C"` allowed.

The owned-property list is gone. The boundary it drew was not the boundary that mattered, and keeping it would have left two scopes to hold in agreement.

`pnpm check` and `pnpm typecheck` pass; the web suite is 236.

This one is worth separating from the previous eleven: it was not a wrong model of CSS, it was two parts of the guard claiming different scopes — the policy I added one commit earlier was narrower than the scan it was meant to protect. Same lesson in a different place: I made a local fix without checking what else depended on it.


---

baixiaohang (changes requested):
- Rationale: The reference/no-escape scopes now align, but the remaining literal exceptions are global value allowlists and do not enforce the selector-specific policy their comments and test names advertise.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and enforce a canonical source policy over the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
❌ 1. `LINE_HEIGHT_LITERALS` and `FONT_SHORTHAND_LITERALS` allow only by value, while their documented policies are selector-specific. Consequently `.copy { line-height: 1; }` and `.card { font: inherit; }` both pass: the former creates component-local text rhythm outside the three single-glyph controls, and the latter contradicts "nothing else may use it" outside `button`, `input`, `select`, and `textarea`. Model these as declaration-context exceptions (property/value plus the exact allowed selector set), and add an arbitrary-selector rejection for each family. The current legitimate sites are `.team-menu-chevron`, `.account-menu-dots`, `.dialog-close`, `.center-card::before`, and the four form-control selectors. [R1 / `apps/web/src/__tests__/typography-tokens.test.ts:47`]

✅ 2. The stylesheet-wide reference scan and no-escape policy now consume the same non-string projection. Escaped references on unowned properties are rejected, while string content remains permitted.

**Keep / Split**
- Keep the guard in this PR. Selector-bound exceptions complete the canonical policy already implemented here.

**Action taken**
- Requested changes on exact head `f4e68682a9f0b509c795a30820b019678c61072b`.

Gandy2025:
@baixiaohang confirmed and fixed in `f17e2f0`. The comments beside those two allowlists described a selector-specific policy; the code admitted the value anywhere. That is worse than an ordinary gap — the prose asserted a guarantee the check did not make, so anyone reading it would have trusted something untrue.

An exception is now the property, the value, and the selectors it was granted for, and it holds only when the rule names no other selector. Your five sites are exactly the five in the stylesheet, which I re-derived from the parse tree rather than taking on faith: `line-height: 1` on `.team-menu-chevron`, `.account-menu-dots`, `.dialog-close`; `38px` on `.center-card::before`; `font: inherit` on the four form controls.

Regressions for arbitrary selectors in both families, plus one for widening a granted rule — `.dialog-close, .copy { line-height: 1 }` is refused, since the exception covers the close button and not whatever else joins it.

`pnpm check` and `pnpm typecheck` pass; the web suite is 238.


---

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All reported parser, token-graph, cascade-scope, canonical-value, and exception-ownership bypasses are closed; the guard now enforces the boundary its prose advertises.

**Risk level**: C
- Path baseline: changes remain limited to `apps/web/**` -> C
- Semantic lift: none

**PR summary**
- Problem: Operators and team administrators need a consistent type hierarchy across dense work surfaces and reading-oriented configuration flows.
- Approach: Centralize typography behind raw and semantic tokens, vary density explicitly by surface, normalize fallbacks and numeric rendering, and enforce a canonical source policy over the token graph and its ownership boundaries.
- Impacted modules: Web stylesheet, typography guard, its parser dependency, and document color-scheme metadata

**Review findings**
✅ 1. Literal exceptions now bind property, value, and the exact selector set that earns them; arbitrary selectors and mixed-selector widening are rejected.
✅ 2. Names are decoded without exposing their written spelling, while values use a canonical no-escape policy rather than simulated browser tokenization.
✅ 3. Reference collection, role-to-step edges, raw-step validation, root/scoped ownership, and reading-surface rebindings now share aligned declaration boundaries.
✅ 4. The current-main typography addition remains integrated, and the stylesheet migration itself stays coherent and presentation-only.

**Action taken**
- Approved exact head `f17e2f0604c451b92dbfaafa44ad733187f76cd1`.
- The guard should remain in this PR as the migration's atomic enforcement.

yuezengwu (approved):
**Recommendation**: approve
- Rationale: The selector-bound exception change closes the remaining literal-allowlist blocker without reopening any parser, token-graph, scope, or canonical-value hole.

**Review findings**

✅ 1. Each exception now binds property, value, and an allowlisted selector set; arbitrary selectors and widened selector lists are rejected. Independent variants confirmed `.copy { line-height: 1 }`, `.card { font: inherit }`, `.dialog-close, .copy`, pseudo-element, escaped/case, and `:is(...)` forms cannot widen the exception.

✅ 2. All five legitimate site classes remain accepted. Harmless selector-list whitespace, order, and duplicate selectors do not create false failures; selector spellings outside the canonical allowlist fail closed.

✅ 3. The prior guarantees remain intact: parsed declarations isolate decoded names, values and reference projections stay separate, every non-string value is escape-free, root/scoped ownership and role/step/reference checks share the same record boundary, and malformed/comment/string/at-rule injections remain covered.

**Keep / Split**
- Keep the guard with the stylesheet migration; it is now an atomic enforcement layer for the token migration rather than a separate product concern.

**Validation**
- `pnpm check`, `pnpm typecheck`, `pnpm build`: pass.
- `pnpm --filter @opentag/web test`: 238/238 pass.
- Independent selector normalization/bypass matrix: 41/41 pass.
- GitHub CI: 6/6 checks green.

**Action taken**
- Approved exact head `f17e2f0604c451b92dbfaafa44ad733187f76cd1`.

PR #92, opened 2026-08-20
fix(server): report an unavailable Feishu platform as an actionable failure

Gandy2025:
## Summary

Feishu setup discarded the cause of a failed registration start, so a Server that cannot complete the handshake with the Feishu open platform answered `500` with "The request could not be completed". A caller could not tell a blocked network from a rejected authorization.

- classify that failure as `FEISHU_UPSTREAM_UNAVAILABLE`, from a transport error or from an answer carrying no authorization URL
- let the existing error handler answer `502` with the typed code, the same way `AuthServiceError`, `AgentServiceError`, and `ImBindingServiceError` are already handled
- keep every other Feishu code on its current unexpected-failure response, so an internal race is never presented as a user action
- derive one Web recovery message per code, whether it arrives as a failed attempt or as a request error

Observed shape: the SDK rejects with `TypeError: Invalid URL (ERR_INVALID_URL)` when the platform's begin response carries no `verification_uri_complete`, which is what a blocked network produces behind a proxy. Only the registration start and its QR handshake reach the classifier, so the URL parsed there is always the one Feishu returned.

## Testing

- `pnpm --filter @opentag/server test` — new classifier cases (transport codes, timeout, unusable answer, self-referencing cause chain) and route cases asserting `502` with the typed code and `500` for an internal race
- `pnpm --filter @opentag/web test` — one recovery message for both the request-error and failed-attempt paths
- `node scripts/e2e/onboarding-e2e.mjs` — 16/16 steps; the Feishu step now reports "The Feishu open platform did not return a usable authorization. Check the Server's network access to Feishu, then retry." and the response is `502`

## Breaking changes

None. `FEISHU_UPSTREAM_UNAVAILABLE` is a new error code; existing codes keep their current status.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01MzHQwLVMKFoRoA4b9c9g4E)_


baixiaohang (changes requested):
- Rationale: The upstream classifier is also applied to the entire post-QR activation path, so local database or activation failures can be reported as Feishu outages.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes `services/im-bindings/**`.
- Semantic lift: none; the path baseline is already B-high.

**PR summary**
- Problem: When Feishu cannot provide a usable setup authorization, operators currently receive a generic 500 and cannot distinguish a retryable platform/network failure from an internal failure.
- Approach: Add a typed 502 response for registration transport or unusable-response failures, keep internal Feishu operation codes private, and reuse one Web recovery-message map for request and persisted-attempt failures.
- Impacted modules: Feishu setup classification and lifecycle, the server error envelope, shared error codes, and Web setup feedback.

**Review findings**
❌ 1. Scope the upstream classifier to the registration failure it can identify. `safeFeishuSetupErrorCode()` now maps any raw `ETIMEDOUT`, `ECONNRESET`, `TimeoutError`, and similar shape to `FEISHU_UPSTREAM_UNAVAILABLE`, but `FeishuSetupService.#complete()` calls it from one catch that spans `registration.result`, database claim/update work, and `activateAtomicAttempt()`. A PostgreSQL or other local activation timeout after the QR exists can therefore be persisted as a Feishu outage, and Web will tell the operator to check Feishu network access. This also contradicts the classifier comment that only registration start and QR handshake reach it. Split the completion phases (or otherwise preserve the old generic handling for non-registration failures) and add a regression case for an activation/database timeout. [R4, R5 / `packages/server/src/services/im-bindings/feishu/errors.ts:106`, `packages/server/src/services/im-bindings/feishu/setup-service.ts:310`]

✅ 2. The public-failure allowlist keeps internal Feishu operation codes on the generic 500 path and does not expose raw provider error text. [R4 / `packages/server/src/services/im-bindings/feishu/errors.ts:35`]

**Action taken**
- Submitted request changes.

yuezengwu (dismissed):
Fresh-head re-review at `a3ba9db49e303a266a2aa8f6db45686362256171`:

- The prior blocker is resolved: once `registration.result` fulfills, database claim and activation failures use the activation-safe classifier, so local `ETIMEDOUT`/`ECONNRESET` shapes cannot be persisted or presented as a Feishu outage.
- The public failure allowlist still exposes only `FEISHU_UPSTREAM_UNAVAILABLE` through a sanitized typed 502 envelope; raw provider errors, credentials, and internal Feishu operation codes remain private.
- The Web recovery path is keyed by the typed code for both immediate request failures and persisted failed attempts.

Validation: `pnpm build`, `pnpm check`, and `pnpm typecheck` passed; focused server tests (22/22), Web tests (183/183), and fresh-head GitHub CI passed. The full root test reached two unrelated existing macOS `/tmp` versus `/private/tmp` CLI path assertions; no changed file intersects those tests.

No blocking findings on this head.

yuezengwu (approved):
Exact-head re-review at `1783fa37d71ca4fdd2648972d0c4ebbe4476d2b0` (merge of the reviewed fix with current `main`):

- The PR diff against the new base is byte-for-byte identical to the diff reviewed at `a3ba9db49e303a266a2aa8f6db45686362256171`; the merge resolution did not alter any of the nine PR files.
- The prior blocker is resolved: after `registration.result` fulfills, database claim and activation failures use the activation-safe classifier, so local `ETIMEDOUT`/`ECONNRESET` shapes cannot be persisted or presented as a Feishu outage.
- The public allowlist exposes only `FEISHU_UPSTREAM_UNAVAILABLE` through a sanitized typed 502 envelope; raw provider errors, credentials, and internal Feishu operation codes remain private. Web recovery is keyed by the typed code for both immediate and persisted failures.

Validation: `pnpm build`, `pnpm check`, and `pnpm typecheck` passed; focused server tests passed 22/22 and Web tests passed 183/183. Fresh-head CI’s `Checks` job failed only in an unrelated client readiness timing assertion (`probe` count 3 vs 7); the exact failing file immediately passed locally 14/14 on this head. Other fresh-head compatibility/smoke jobs passed. Full local suites also expose existing macOS `/tmp` vs `/private/tmp` path assertions outside the changed files.

No blocking findings on this exact head.

PR #97, opened 2026-08-21
fix(cli): tolerate launchd activation races

yuezengwu:
## Summary

- treat launchctl's `No such process` bootout response as an already-unloaded service so fresh installs can continue to bootstrap
- wait for launchd to move from transient `xpcproxy` state to an active daemon after install, start, and restart
- retry responsive status checks with a bounded attempt count, fail on the first timed-out status command, and include the last observed state when responsive checks never become active
- cover the fresh-install response, all three activation paths, timeout fail-fast behavior, and bounded responsive-state exhaustion

## Testing

- `pnpm check` (passes; reports 8 pre-existing E2E environment-variable warnings)
- `pnpm build`
- `pnpm typecheck`
- `pnpm --filter open-tag exec vitest run src/__tests__/daemon-service-renderers.test.ts --maxWorkers=1 -t 'launchd service backend'` (14/14 passed)
- `TMPDIR=<real non-symlink directory> pnpm --filter open-tag test` (119/119 passed)
- `pnpm --filter @opentag/client test:agent-runtime:coverage` (273/273 passed; 100% statements, branches, functions, and lines)
- `pnpm test` (default local macOS run: 116/118 CLI tests passed; two pre-existing assertions compare `/tmp` or `/var` with canonical `/private/tmp` or `/private/var` paths)
- `pnpm test:coverage` (1055/1059 passed; the same path-alias issue accounts for all four failures across CLI and client tests)

## Breaking changes

None.

## Checklist

- [ ] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. (`pnpm test` has the disclosed macOS path-alias baseline failures; package tests pass with a real temp path.)

### DIFF
--- a/apps/cli/src/__tests__/daemon-service-renderers.test.ts
+++ b/apps/cli/src/__tests__/daemon-service-renderers.test.ts
@@ -665,6 +665,41 @@ describe("launchd service backend", () => {
     );
   });
 
+  it("fails activation immediately when launchctl status times out", async () => {
+    const userHome = await temporaryDirectory("opentag-launchd-");
+    const home = join(userHome, ".opentag");
+    let loaded = false;
+    let activationChecks = 0;
+    const wait = vi.fn(async () => undefined);
+    const runner = fakeRunner((_, args) => {
+      if (args[0] === "print" && args[1] === "gui/501") return result(0, "domain", "");
+      if (args[0] === "print") {
+        if (!loaded) return result(113, "", "Could not find service");
+        activationChecks += 1;
+        return { code: null, stderr: "", stdout: "", timedOut: true };
+      }
+      if (args[0] === "bootout") return result(3, "", "Boot-out failed: 3: No such process");
+      if (args[0] === "bootstrap") loaded = true;
+      return result(0, "", "");
+    });
+    const backend = createLaunchdBackend({
+      activationAttempts: 50,
+      activationDelayMs: 100,
+      evictionDelayMs: 0,
+      home,
+      invocation: { args: [], program: "/usr/local/bin/opentag" },
+      runner,
+      serviceId: "opentag",
+      sleep: wait,
+      uid: 501,
+      userHome,
+    });
+
+    await expect(backend.installAndStart()).rejects.toThrow("launchd activation status check timed out");
+    expect(activationChecks).toBe(1);
+    expect(wait).not.toHaveBeenCalled();
+  });
+
   it("reports a loaded waiting LaunchAgent without a PID as inactive", async () => {
     const userHome = await temporaryDirectory("opentag-launchd-");
     const home = join(userHome, ".opentag");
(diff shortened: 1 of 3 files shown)


baixiaohang (changes requested):
- Rationale: The activation retry count does not bound wall-clock latency, so a stalled `launchctl print` can turn the intended five-second wait into more than twelve minutes.

**Risk level**: A
- Path baseline: includes `apps/cli/**` -> A
- Semantic lift: none

**Review findings**

❌ 1. Bound the activation wait by elapsed time, or fail immediately when the status command times out. [R4 / `apps/cli/src/core/daemon/service/launchd.ts:221`]

`waitForActivation()` can call `status()` 50 times, while each `status()` call gives `launchctl print` the existing 15-second `LAUNCHD_TIMEOUT_MS`. If `launchctl print` stalls, this path can therefore take roughly 755 seconds (50 × 15 seconds plus 49 × 100 ms), rather than the advertised roughly five seconds. The previous install/restart path performed only one final status check, so this materially amplifies a manager failure across install, start, and restart. Please enforce one wall-clock deadline or a short per-poll command timeout, or propagate the `launchctl print timed out` state immediately instead of retrying it.

**Action taken**
- Submitted request changes.

yuezengwu:
Addressed the activation-timeout finding in `1b9209e`.

Activation polling now uses a fail-fast status read: the first timed-out `launchctl print` throws `launchd activation status check timed out` and no retry delay or subsequent status attempt occurs. The regression test asserts one activation check and zero sleeps. Local validation passed for all 14 launchd tests, all 119 CLI tests with a real non-symlink temp directory, `pnpm check`, `pnpm build`, and `pnpm typecheck`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The follow-up commit closes the timeout amplification while preserving bounded retries for responsive transient launchd states and the existing read-only status behavior.

**Risk level**: A
- Path baseline: includes `apps/cli/**` -> A
- Semantic lift: none

**PR summary**
- Problem: macOS daemon install, start, and restart can fail during normal launchd activation races, including a fresh-install bootout response and the transient `xpcproxy` state.
- Approach: Treat the already-unloaded bootout response as idempotent success, poll responsive activation states until launchd reports a running daemon with a PID, and fail on the first timed-out activation status command.
- Impacted modules: CLI launchd service backend, shared manager-result classification, and launchd backend tests.

**Review findings**

✅ The previous blocking finding is resolved: a timed-out activation status check now causes one immediate, specific failure instead of entering the remaining retry loop. The regression test asserts one activation check and zero retry sleeps.

✅ Install, start, and restart share the same activation guard, while ordinary read-only `status()` calls retain their non-throwing `unknown` result on timeout.

**Action taken**
- Approved exact head `1b9209ebf06798d5f77acdec0fc945de211d0b80` with explicit maintainer authorization.

PR #100, opened 2026-08-21
feat: lazily materialize thread sessions from inbound evidence

baixiaohang:
## Summary

- keep top-level direct messages on the Channel Session and lazily materialize Thread Sessions only from real inbound thread facts
- admit the first Thread delivery from existing ownership, a current direct mention, or verified root-direct evidence while preserving independent `all_message` ambient fan-out
- bootstrap new Thread runtimes with bounded root and same-thread inbound history, and document the provider-neutral prompt-driven behavior

## Implementation

- centralizes transaction-aware Chat Session and placement creation in `SessionService`
- normalizes reliable Slack and Feishu root references without treating Feishu `parent_id` as a root
- blocks old root evidence from implicitly reviving an ended Thread Session
- keeps direct-provider CLI messaging and `direct`/`ambient` credential behavior unchanged
- adds decision-matrix, concurrency, history-boundary, adapter, and prompt fixtures

## Verification

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

### DIFF
--- a/packages/server/src/runtime/im-delivery-worker.ts
+++ b/packages/server/src/runtime/im-delivery-worker.ts
@@ -4,6 +4,7 @@ import {
   type DirectImMessageDeliveryRequest,
   DirectImMessageDeliveryRequestSchema,
   type EffectiveRuntimeSnapshot,
+  type ProviderInboundContext,
   RUNTIME_DIRECT_TEXT_MAX_BYTES,
   RUNTIME_IM_HISTORY_MAX_BYTES,
   RUNTIME_MAX_FRAME_BYTES,
@@ -29,6 +30,7 @@ import {
   setActiveSpanAttributes,
   traceDeliveryClaim,
 } from "../observability/index.js";
+import { threadRootExternalId } from "../services/im/provider-thread-context.js";
 import {
   type EffectiveRuntimeSnapshotAssembler,
   EffectiveRuntimeSnapshotAssemblerError,
@@ -408,8 +410,15 @@ export class ImDeliveryWorker {
       }
       const resources = row.message.content.resources ?? [];
       const history =
-        row.agent.receiveMode === "mention_only" && row.delivery.attention === "direct"
-          ? await this.#history(row.session, row.message.occurredAt, row.message.providerRevisionKey, row.message.id)
+        row.delivery.attention === "direct" &&
+        (row.agent.receiveMode === "mention_only" || row.session.kind === "thread")
+          ? await this.#history(
+              row.session,
+              row.message.providerContext,
+              row.message.occurredAt,
+              row.message.providerRevisionKey,
+              row.message.id,
+            )
           : { items: [], truncated: false };
       const request: DirectImMessageDeliveryRequest = {
         type: "im:deliver",
@@ -741,6 +750,7 @@ export class ImDeliveryWorker {
 
   async #history(
     session: typeof sessions.$inferSelect,
+    providerContext: ProviderInboundContext,
     occurredAt: Date,
     providerRevisionKey: string,
     messageId: string,
@@ -753,6 +763,7 @@ export class ImDeliveryWorker {
     }>;
     truncated: boolean;
   }> {
+    const rootExternalId = threadRootExternalId(providerContext);
     const [lastAccepted] = await this.#database
       .select({
         id: imMessages.id,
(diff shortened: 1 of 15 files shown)


packages/server/src/services/im/im-message-inbox.ts
```diff
@@ -337,28 +331,63 @@ export class ImMessageInbox {
...
+                const endedThreadExists = existingThread
+                  ? false
+                  : await this.#hasEndedThreadSession(
                       transaction,
                       imBindingId,
                       event.conversation.externalId,
+                      threadKey,
+                    );
+                const rootExternalId = threadRootExternalId(event.providerContext);
+                const rootWasDirect =
```
yuezengwu, line 352: [P1] Reconcile replies that arrive before the direct root. This lookup only observes an already-committed root delivery. Slack/Feishu events can be processed out of provider order, so a reply with a later occurredAt can commit first with no Thread delivery; when the direct root commits afterward, nothing revisits that persisted reply. In mention_only mode the first human reply is then permanently missed. I reproduced this by pausing the reply after message authority, starting the root ingest, then releasing the reply: expected one delivery, received zero. Please retain/reconcile this undecided reply when the earlier root becomes direct-authoritative (or enforce provider-fact ordering) and add deterministic reordered root/reply coverage.

yuezengwu (changes requested):
Requesting changes on exact head 1c7bbd31343a4b2fdef83c785527e17e3618ddb1. The new root-direct admission path can permanently miss the first human thread reply when provider events are processed out of order. I reproduced this deterministically with an integration test that let the later-occurring reply acquire message authority before its earlier direct root: the reply committed with deliveryIds=[], and the subsequent root commit did not reconcile it. Please make admission independent of webhook arrival order and cover this root/reply ordering case. Local build and the changed unit/integration suites otherwise pass. The current CI failure is in two unrelated Web account-profile tests.

packages/server/src/runtime/im-delivery-worker.ts
```diff
@@ -787,11 +800,57 @@ export class ImDeliveryWorker {
           eq(imMessages.imBindingId, session.imBindingId),
           eq(imMessages.channelId, session.channelId),
           eq(imMessages.direction, "inbound"),
-          ...(session.kind === "thread" && session.threadKey ? [eq(imMessages.threadKey, session.threadKey)] : []),
+          ...(session.kind === "thread" && session.threadKey
+            ? [
+                or(
+                  eq(imMessages.threadKey, session.threadKey),
+                  ...(rootExternalId ? [eq(imMessages.externalMessageId, rootExternalId)] : []),
```
yuezengwu: [P1] Reserve the verified root before applying the rolling history limit. Adding the root to this same candidate query makes it just another old row; the later descending 101-row limit therefore drops it when a cold Thread has 101 prior same-thread messages. I reproduced that boundary by extending the integration fixture: the first direct delivery contained no root even though the reliable root was persisted. That breaks the documented `root, then same-thread history` bootstrap guarantee and leaves a long-running thread without its topic-setting fact. Please fetch/reserve the latest root revision independently, then fill the remaining item/byte budget with same-thread facts, with a boundary test above 100 messages.

yuezengwu (changes requested):
Requesting changes on exact head 9d42e9f5ff86b282d63fad5d9a6c2a1704b6c91d. The reordered root/reply fix is effective: the new deterministic test passes in the full 67-test IM binding integration file, and I resolved the original thread. One independent cold-start boundary remains. The history query applies its 100-item limit to root and thread facts together, so after 101 prior same-thread messages the reliable root is dropped. I reproduced that exact boundary locally. This contradicts the PR's documented bootstrap contract that the visible root precedes bounded same-thread history. Please reserve/query the latest reliable root independently of the rolling thread-message cap (while still honoring the byte bound) and add a >100-message boundary test.

yuezengwu (approved):
Approved on exact head 37597cc79501297970c8aaaea1219d953c0eb775. The original reordered root/reply blocker is fixed and covered, current-message revisions are excluded from bootstrap history, and the root is now reserved independently of the rolling 100-item Thread history cap with a 101-message boundary test. I also verified the Web dirty-form waits added for the previously failing CI tests. Local validation: `pnpm check`, `pnpm typecheck`, and `pnpm build` pass; the full IM binding integration file passes 68/68; the Web suite passes 239/239. Both review threads are resolved.

PR #103, opened 2026-08-21
feat(web): redesign onboarding as two-step flow

Gandy2025:
## Summary

- redesign the desktop onboarding experience around two user-facing steps: prepare the Agent, then add it to Feishu
- replace the misleading Computer → Agent → Feishu sequence with a runtime summary in step 1 and an Agent ↔ Feishu connection view in step 2
- preserve the existing server-backed onboarding state machine while improving state-specific copy, responsive behavior, and accessibility
- extend onboarding tests for the two-step journey and live-state summaries

## Scope

This PR is intentionally limited to the onboarding page, its tests, and onboarding-specific rules in the shared stylesheet. It does not include the separate Design System or App Shell work.

## Validation

- pnpm check (passes; reports 8 existing undeclared E2E environment-variable warnings)
- pnpm build
- pnpm typecheck
- pnpm --filter @opentag/web test (240 tests passed)
- pnpm test (one unrelated server observability test fails on the main baseline: expected stale_connection but received handled; this PR has no server diff)

### DIFF
--- a/apps/web/src/onboarding/page.test.tsx
+++ b/apps/web/src/onboarding/page.test.tsx
@@ -142,7 +142,26 @@ describe("OnboardingPage", () => {
     expect(await screen.findByRole("heading", { name: "Connect a Local Computer" })).toBeTruthy();
     expect(screen.getByRole("button", { name: "Generate connection command" })).toBeTruthy();
     expect(screen.queryByRole("navigation", { name: "Workspace" })).toBeNull();
-    expect(screen.queryByText("Create Team")).toBeNull();
+    expect(screen.queryByText("Create Workspace")).toBeNull();
+  });
+
+  it("presents the conditional setup as two product-level steps", async () => {
+    installFacts();
+    renderPage();
+    expect(await screen.findByRole("heading", { name: "Connect a Local Computer" })).toBeTruthy();
+
+    const steps = screen.getByRole("navigation", { name: "Onboarding steps" });
+    expect(steps.querySelectorAll("li")).toHaveLength(2);
+    expect(steps.textContent).toContain("Prepare your Agent");
+    expect(steps.textContent).toContain("Add to Feishu");
+    expect(steps.textContent).toContain("In progress");
+    expect(steps.textContent).toContain("Up next");
+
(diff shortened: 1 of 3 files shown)


baixiaohang (changes requested):
- Rationale: The new live summary can contradict the selected runtime route, and the step-two visualization is clipped at common mobile widths.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed, which falls under the C baseline.
- Semantic lift: none; the existing onboarding state machine and API/data flow are unchanged.

**PR summary**
- Problem: The existing onboarding journey presents three technical objects as if they were sequential user steps, making it hard to understand what must be prepared before the Agent can be used in Feishu.
- Approach: Replace that presentation with two product-level steps, surface live Computer/runtime/Agent facts during preparation, and show the Agent-to-Feishu connection during handoff while retaining the existing server-backed state machine.
- Impacted modules: onboarding page, onboarding component tests, and shared web styles.

**Review findings**
❌ 1. Derive the Computer summary from the selected or Agent-bound route, not the first online Computer. `onboardingJourney` uses `snapshot.computers.find(...online)` for every post-selection state. If the user selects Computer B while Computer A is first in the snapshot, the summary says A even though the action says the Agent will run on B. For an existing Agent whose bound Computer is offline, an unrelated online Computer can be shown as ready; with no online Computer, the fallback still says `Connected`. This makes the new live-state summary contradict the authoritative onboarding state. Please map the fact from `current.computer` or the target Agent's `computerId` as appropriate and cover the selected-B/offline-bound routes. [R1 / `apps/web/src/onboarding/page.tsx:432`]

❌ 2. Add a narrow-screen layout for the Feishu connection diagram. Its three columns have minimum widths of 150 + 120 + 150 px plus 36 px of gaps, but the `max-width: 720px` rules do not override that grid. Below roughly 530 px the diagram is wider than the stage header, whose `overflow: hidden` clips the right-hand Feishu endpoint. That makes step 2 incomplete on typical phone widths despite this PR's responsive scope. Please collapse or stack the diagram at the mobile breakpoint. [`apps/web/src/styles.css:1985`]

**Action taken**
- Submitted request changes on head `4674aeaeb6dc9243b77d2f1894370e182c46b4d4`.

PR #107, opened 2026-08-21
fix(web): preserve immediate account profile edits

Gandy2025:
## Summary

- skip the redundant initial account profile synchronization effect
- continue syncing the input when the confirmed server display name actually changes
- prevent an immediate user edit from being reset before the save controls render

Fixes the flaky Web failure observed in [CI run 32456314275](https://github.com/first-tree-ai/opentag/actions/runs/32456314275/job/96694191774).

## Validation

- failing account profile test repeated 20 times: passed
- Web test suite: 248/248 passed
- `pnpm build`: passed
- `pnpm typecheck`: passed
- Biome on the changed file, notices check, and `git diff --check`: passed

## Unrelated local-environment failures

Full repository test/coverage commands still expose pre-existing non-Web failures: macOS canonical `/private/tmp` path assertions, the RuntimeSession observability timing assertion, and one local provider-probe timeout. None are touched by this one-file fix.

### DIFF
--- a/apps/web/src/router.tsx
+++ b/apps/web/src/router.tsx
@@ -1865,6 +1865,7 @@ function CapabilityUnavailable({
 
 function AccountSettings({ refreshMe, user }: { refreshMe: () => void; user: MeResponse["user"] }) {
   const saveInFlight = useRef(false);
+  const confirmedDisplayNameRef = useRef(user.displayName);
   const [displayName, setDisplayName] = useState(user.displayName);
   const [saving, setSaving] = useState(false);
   const [message, setMessage] = useState<string>();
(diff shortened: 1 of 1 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The guard removes the redundant initial state write that can erase an immediate edit while preserving synchronization when the server-confirmed display name actually changes.

**Risk level**: C
- Path baseline: `apps/web/src/router.tsx` is an unmatched Web application path -> C.
- Semantic lift: none; this is local component state and does not touch subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: An account owner editing their display name immediately after the Account page rendered could have the edit reset before the save controls appeared.
- Approach: Remember the last confirmed prop value, skip the identical mount-time synchronization, and continue copying genuinely changed server-confirmed values into the input.
- Impacted modules: `apps/web/src/router.tsx` (`AccountSettings`).

**Review findings**
✅ 1. The change removes only the redundant initial write; save success, refresh, failure rollback, and later server-confirmed prop changes retain their existing authority and behavior.

**Action taken**
- Approved.

PR #108, opened 2026-08-21
fix(client): inject managed instructions through provider-native prompts

### LINKED ISSUE #101: [Bug]: Direct provider CLI upgrade can leave daemon reconcile stuck on stale managed AGENTS.md
## Reproduction steps

1. Start with an existing staging installation whose Agent workspace state is schema v1 and whose legacy managed `AGENTS.md` was created by an older runtime policy.
2. Upgrade from `open-tag-staging@0.0.2-staging.67.1` to `0.0.2-staging.92.1` (`gitHead` `3445372bfefb218c70d9de08b22e1cd3fc97681b`).
3. Change the existing Agent runtime config revision from 1 to 2 in the database, then restart the daemon.
4. Send the Agent a Feishu message.

The affected local state has all of the following:

- `runtime/workspace-states/<agent>.json` remains at `schemaVersion: 1` and `appliedAgentRevisionSequence: 1`.
- The legacy `<workspace>/AGENTS.md` is an intact OpenTag-managed file whose hash matches `managedInstructionsHash` in the v1 state.
- `<workspace>/files/AGENTS.md` also exists, is mode `0444`, has the OpenTag-managed header, and contains an older platform-instruction revision.
- The Server now sends Agent runtime revision 2 using the current direct-provider-CLI platform instructions.

## Expected behavior

The supported upgrade path should converge without deleting the Agent workspace or losing Provider/Session continuity:

- a valid v1 workspace should migrate to schema v2;
- stale or partially migrated OpenTag-managed instruction files should be replaced with the new managed content when their provenance can be established safely;
- an interrupted migration should be idempotent on retry;
- genuinely user-authored conflicting files must remain protected and fail closed with an actionable diagnostic;
- the daemon should either complete reconcile or emit a sanitized error that identifies the frame type and failure category.

The direct-provider-CLI policy change should also advance existing active Agent runtime config revisions during migration, as required by the comment on `OPENTAG_PLATFORM_INSTRUCTIONS`.

## Actual behavior

The daemon connects and reports ready, but `session:reconcile` never completes. Approximately every 30 seconds it logs only:

```text
Runtime business frame listener failed
```

No `IM delivery received` or `Turn started` entry is produced, so Feishu messages never reach the local Agent.

Bumping the database runtime config revision to 2 and restarting does not recover the daemon. The local workspace state remains schema v1 / revision 1.

The current client migration rejects any existing `<workspace>/files/AGENTS.md` whose content differs from the newly rendered instructions:

```ts
const current = await readSecureFile(paths.agentsFile);
if (current !== undefined && current !== content) {
  throw new RuntimeStorageError("conflict", "OpenTag will not replace a conflicting AGENTS.md file in the Agent cwd");
}
```

That makes a stale or partially migrated OpenTag-managed file indistinguishable from a user-owned conflict and prevents the migration from ever reaching the schema-v2 state write. Separately, `RuntimeConnection.#dispatchBusinessFrame` discards the caught error object, which hides this actionable cause behind the generic warning.

Migration `0009_numerous_morlocks.sql` changes the platform/runtime contract and drops `allowed_tools`, but does not advance active Agent runtime config revisions.

## Environment

- OS: Linux 6.6.87.1-microsoft-standard-WSL2 x86_64
- Node.js: v24.12.0
- pnpm: 10.12.1
- Previous OpenTag version: `0.0.2-staging.67.1`
- Current OpenTag version: `0.0.2-staging.92.1`
- Current package git head: `3445372bfefb218c70d9de08b22e1cd3fc97681b`
- Daemon manager: systemd user service
- Provider: Codex
- IM provider: Feishu; `lark-cli 1.0.79` and `lark-cli im --help` both succeed

## Logs

```shell
{"level":30,"module":"connection","clientVersion":"0.0.2-staging.92.1","state":"registered","message":"Runtime connection registered"}
{"level":30,"module":"daemon","clientVersion":"0.0.2-staging.92.1","message":"Daemon is ready"}
{"level":40,"module":"connection","clientVersion":"0.0.2-staging.92.1","category":"listener","message":"Runtime business frame listener failed"}
# The same warning repeats approximately every 30 seconds.
```

## Suggested regression coverage

- v1 state + valid legacy managed file + absent current file migrates normally.
- v1 state + valid legacy managed file + stale OpenTag-managed current file converges safely.
- retry after interruption between writing the current file, removing the legacy file, and writing state v2 is idempotent.
- a genuinely user-authored current `AGENTS.md` still fails closed without being overwritten.
- business-listener failures record sanitized frame/error context without logging secrets or secret-bearing frame payloads.

yuezengwu:
## Summary

- resolve #101 with one immutable OpenTag-managed platform plus Agent prompt, injected through Provider-native system-prompt surfaces; Session instructions remain Turn input
- use the Agent Workspace root as cwd for new Workspaces and stop creating the extra files layer
- preserve the existing files cwd for v1/v2 Workspaces so upgrades do not change the relative-path view; remove only provenance-proven legacy managed instructions through an idempotent schema-v3 pending/complete transition
- fail closed on missing provenance, malformed state, identity mismatch, concurrent replacement, or user-file conflict; quarantine verification never moves or deletes ordinary user Workspace content
- admit Codex only after isolated developerInstructions create plus production-shape exact resume succeeds, and require Claude Code append-system-prompt support before readiness
- keep source docs limited to current-system behavior; proposal, research, alternatives, and decision record are in draft Context Tree PR https://github.com/first-tree-ai/first-tree-context/pull/972

## Testing

- pnpm install --frozen-lockfile
- pnpm check (passes with 8 pre-existing undeclared E2E environment-variable warnings)
- pnpm build
- pnpm typecheck
- pnpm --filter @opentag/client test:agent-runtime:coverage (20 files, 268 tests, 100% statements, branches, functions, and lines)
- focused Workspace, reconciler, and home-layout regression tests (41/41)
- real installed codex-cli 0.148.0 readiness probe (ready true)
- pnpm test (only 2 macOS temporary-path alias failures)
- pnpm test:coverage (all relevant suites passed; the same 2 baseline path-alias failures)
- both failures reproduce on clean current main c6ca958 in an isolated worktree and are unrelated to this PR
- first-tree tree verify for Context Tree PR #972
- three independent subagent reviews; final result has no remaining P0, P1, or P2 findings

## Breaking changes and compatibility

- Agent Runtime factory contract advances from v1 to v2 and requires systemPrompt; non-v2 factories fail closed at registration.
- New Workspace cwd is the Workspace root. Existing v1/v2 Workspace cwd remains the files directory to preserve upgrade continuity.
- Workspace state advances to schema v3 for layout, identity, and transition fencing. Older v1/v2 Clients reject upgraded state instead of silently downgrading it; backup state and Workspace together.
- OpenTag no longer creates or updates managed Workspace instruction files after compatibility cleanup.
- Claude Code and Pi Provider configuration no longer accepts Provider-specific appendSystemPrompt overrides.

## Checklist

- [x] Changes are focused and avoid unrelated refactors
- [x] Tests and current-system docs updated
- [x] Baseline-only failures reproduced on clean current main
- [x] No credentials or environment files committed
- [x] Context Tree proposal and research moved to the owning draft workflow


baixiaohang (changes requested):
- Rationale: The native prompt direction is sound, but the current rollout leaves upgraded workspaces under competing stale instructions and can advertise production Providers as ready even when they cannot accept the now-required prompt field.

**Risk level**: A
- Semantic lift: none; A is already the highest level. The diff changes the Agent Runtime contract, Provider admission surface, and Workspace/cwd compatibility boundary.

**PR summary**
- Problem: Existing Agents can wedge during reconcile because OpenTag-managed `AGENTS.md` state is stale, while file-based instructions are not a reliable system-prompt carrier.
- Approach: Compile Server-owned platform and Agent instructions into a required Agent Runtime v2 `systemPrompt`, inject it through each Provider's native surface, flatten Provider cwd to the Workspace root, stop managing Workspace instruction/state files, and add sanitized listener diagnostics.
- Impacted modules: `packages/client` Agent Runtime contracts, Codex/Claude Code/Pi adapters, Session Runtime composition, Workspace layout, runtime diagnostics, and bilingual design/development docs.

**Review findings**
❌ 1. The affected upgrade path stops wedging but is not behaviorally safe. In the issue-101 state, `<workspace>/AGENTS.md` is an old OpenTag-managed file and user work lives under `<workspace>/files/`. Returning `<workspace>` as the new Provider cwd while deliberately leaving both entries untouched means Codex can auto-discover the stale root `AGENTS.md` alongside the new `developerInstructions`, and normal relative work now starts one directory above the user's prior cwd. The design explicitly asks users to reorganize manually, contradicting the issue's continuity requirement. Please add a provenance-aware, idempotent compatibility transition that removes/neutralizes only proven OpenTag-managed instruction files and preserves the prior Workspace view (or otherwise migrates it without collisions) before changing cwd. [R4/R5 / `packages/client/src/runtime/agent-workspace.ts:40`; `docs/design/agent-system-prompt.md:215`]

❌ 2. Readiness does not admit the new mandatory Provider surface. Codex's probe still exercises `thread/start` without `developerInstructions`, and Claude Code's help probe does not require `--append-system-prompt`; an older installed CLI can therefore be reported `ready` and selected, then every create/resume fails only after Turn custody. Because contract v2 now requires native prompt injection for both production Providers, their exact create/resume capability (or a proven minimum version) must be checked by readiness in this rollout rather than deferred. [R4/R5 / `packages/client/src/providers/codex/agent-runtime.ts:740`; `packages/client/src/providers/claude-code/agent-runtime.ts:727`]

❌ 3. The new files under `docs/design/` are explicitly marked `Status: proposed` / `Status: research complete` and contain decision rationale and alternatives. The team's documentation boundary requires this material to live in the Context Tree proposal/decision lifecycle; source-repo `docs/` is reserved for current-system user/contributor reference. Please move the proposal/research records to the Tree and keep only adopted current-system documentation in this repository. [`docs/design/agent-system-prompt.md:5`; `docs/design/provider-prompt-injection-research.md:5`]

✅ 4. The common required `systemPrompt` and removal of Provider-specific prompt overrides correctly centralize the instruction authority once the rollout/admission gaps above are closed. [R1/R2]

**Action taken**
- Submitted request changes on head `df3f21b596bdd37fe5bc20d8a56511ca7968ff0f`.

yuezengwu:
Addressed all blockers from review https://github.com/first-tree-ai/opentag/pull/108#pullrequestreview-4991086305 at exact head 6914c18f26486052db9a50796a23bb879037633a.

1. Upgrade continuity: new Workspaces use root cwd; v1/v2 Workspaces retain files cwd. Schema-v3 pending/complete state carries path-bound provenance, retries interrupted cleanup idempotently, and removes only hash-proven OpenTag instructions. User conflicts, concurrent replacements, symlinks, directories, and special entries fail closed without overwrite or deletion. Downgrade is fenced.
2. Provider admission: Codex readiness now exercises developerInstructions on persistent create, materializes a local rollout without a model Turn, then performs production-shape exact resume without supplied history and requires the exact thread ID. Claude Code readiness requires append-system-prompt. Protocol mismatches become typed version_incompatible before Turn custody. Real codex-cli 0.148.0 returns ready true.
3. Documentation boundary: source design and research copies were removed; source retains current-system docs. Proposal, research, alternatives, and decision record moved to draft Context Tree PR https://github.com/first-tree-ai/first-tree-context/pull/972 and passed first-tree tree verify.

Validation: install, check, build, typecheck, Agent Runtime coverage 268/268 at 100%, focused Workspace regressions 30/30, and three subagent reviews with no remaining P0/P1/P2. Full test and coverage runs have only the same two macOS temporary-path alias failures, reproduced on clean current main c6ca958.

baixiaohang (changes requested):
- Rationale: The readiness and documentation blockers are addressed, but the compatibility transition still rejects the exact stale partial-migration state reported in issue #101, so the affected daemon remains unable to reconcile.

**Risk level**: A
- Semantic lift: none; A is already the highest level. The new head changes Workspace state, upgrade cleanup, cwd compatibility, and Provider admission.

**PR summary**
- Problem: Existing Agents can wedge during an upgrade when a schema-v1 Workspace contains both the provenance-recorded root managed instruction file and a read-only, partially migrated managed file under `files/`; new Agents also need managed instructions delivered through Provider-native prompt surfaces.
- Approach: Inject one compiled platform-plus-Agent prompt through each Provider, preserve the legacy `files/` cwd for upgraded Workspaces, remove provenance-proven managed files through a schema-v3 pending/complete transition, and admit Providers only after probing the required prompt surface.
- Impacted modules: `packages/client` Workspace state and migration, Codex/Claude Code/Pi adapters and readiness, Session Runtime composition, diagnostics, and current-system documentation.

**Review findings**
❌ 1. The regression named “preserves the exact Issue #101 legacy view” does not reproduce the issue's stale partial file. It renders `stalePartial` from the current `runtime.instructions`, while issue #101 explicitly has a read-only `files/` managed file containing an **older** platform-instruction revision. For schema v1, `beginLegacyTransition` has no recorded hash for that file and `isInterruptedLegacyMigration` accepts it only when its content exactly equals the current snapshot; the real issue state therefore still throws “not proven OpenTag-managed” at line 152 and reconcile remains wedged. Please add a safe provenance mechanism for the known historical partial-migration format (or another recovery path that cannot delete user content) and cover the actual old-revision repro. [R4/R5 / `packages/client/src/runtime/agent-workspace.ts:145`; `packages/client/src/runtime/agent-workspace.ts:290`; `packages/client/src/__tests__/agent-workspace.test.ts:58`]

✅ 2. Codex and Claude Code readiness now exercise/require the mandatory native prompt surface before Turn custody, addressing the Provider admission blocker.

✅ 3. Proposal and research material was removed from source docs and moved to draft Context Tree PR #972, addressing the documentation-boundary blocker.

**Action taken**
- Submitted request changes on head `6914c18f26486052db9a50796a23bb879037633a`.

yuezengwu:
Addressed the exact-head blocker in fresh head `479d7f3`.

The regression now reproduces the real Issue #101 transition, rather than deriving the stale nested file from the current runtime snapshot:

- schema-v1 source platform bytes are taken from `open-tag-staging@0.0.2-staging.67.1` (`gitHead 12b574d1b06b09b5f240f73834e531d9d8b7cf01`)
- partial-migration target platform bytes are taken from `open-tag-staging@0.0.2-staging.92.1` (`gitHead 3445372bfefb218c70d9de08b22e1cd3fc97681b`)
- the incoming current runtime deliberately contains a third, different platform revision
- the test exercises end-to-end `SessionReconciler.reconcile`, verifies the legacy `files/` cwd and both root/nested user files, then performs a second reconcile to prove the completed transition is idempotent

Schema-v1 nested-file provenance is now conservative and independent of the current snapshot. OpenTag accepts the known partial migration only when all of the following hold:

1. the private schema-v1 state hash exactly proves the root `AGENTS.md` bytes;
2. those proven bytes contain the exact historical staging.67.1 source platform revision;
3. nested bytes equal the deterministic staging.67.1 -> staging.92.1 platform replacement while preserving the complete opaque Agent suffix byte-for-byte;
4. the historical managed file remains a real read-only file.

Missing root provenance, copied/current-looking content, a changed Agent suffix, writable files, symlinks, byte/inode replacements, and quarantine retry conflicts all fail closed and preserve the entry. Read-only is rechecked before quarantine, after rename on the same dev/inode, and for stranded retry state, closing the proof-to-cleanup chmod window found by the independent follow-up review.

Fresh validation:

- `pnpm install --frozen-lockfile` — PASS
- `pnpm check` — PASS (the existing 8 Turborepo env-declaration warnings remain warnings)
- `pnpm build` — PASS
- `pnpm typecheck` — PASS
- focused Workspace/reconciler/home suite — 40/40 PASS
- full `@opentag/client` suite, including client runtime integration — 429/429 PASS
- Agent Runtime coverage — 268/268 PASS, 100% statements/branches/functions/lines
- independent subagent security/code re-review — no remaining P0/P1/P2
- `pnpm test` and `pnpm test:coverage` — all relevant suites pass; only the same two macOS `/tmp` vs `/private/tmp` CLI path-alias assertions fail. Those two failures were independently reproduced on clean current main `c6ca958` in `daemon-service-renderers.test.ts` and `daemon-service-shared.test.ts`.

Please re-review fresh exact head `479d7f3`.

baixiaohang (changes requested):
- Rationale: The exact reported issue #101 state is now recognized safely, but the patch explicitly rejects another valid schema-v1 state produced by the old migration's own crash window, so legacy reconcile is still not fully idempotent.

**Risk level**: A
- Semantic lift: none; A is already the highest level. This head changes Workspace provenance admission and cleanup concurrency checks.

**PR summary**
- Problem: Existing Agents must recover from partially completed Workspace migrations without losing their prior cwd, user files, or Session continuity, while new managed instructions move to Provider-native prompt surfaces.
- Approach: Preserve the `files/` cwd for legacy Workspaces, recognize one exact historical platform-instruction transition from state-backed provenance, remove proven managed files through a schema-v3 pending/complete quarantine transition, and fail closed on unproven content or concurrent changes.
- Impacted modules: `packages/client` Workspace migration/state, reconciler regression coverage, Provider readiness and prompt composition, diagnostics, and current-system documentation.

**Review findings**
❌ 1. This head fixes the reported state where the hash-proven root file still exists, but it removes recovery for the old migration's own post-root-removal crash state. At the issue's `3445372` client, `migrateLegacyWorkspace` writes/chmods the nested file, removes the root file, and only then writes schema v2 ([historical lines 216–219](https://github.com/first-tree-ai/opentag/blob/3445372bfefb218c70d9de08b22e1cd3fc97681b/packages/client/src/runtime/agent-workspace.ts#L216-L219)). A crash after line 218 therefore legitimately leaves schema v1 + no root + the read-only known partial file; issue #101 explicitly requires retry across this window to be idempotent. The current admission requires `root !== undefined`, and the previous recovery test has been changed to assert permanent conflict. This nested file can still be proven conservatively by reversing the known target-platform substitution, preserving its complete Agent suffix, and comparing the reconstructed source bytes to `state.managedInstructionsHash` (plus the existing read-only/regular-file checks). Please recover that state, or provide an equally provenance-safe recovery path, instead of fencing it indefinitely. [R4/R5 / `packages/client/src/runtime/agent-workspace.ts:163`; `packages/client/src/__tests__/agent-workspace.test.ts:133`]

✅ 2. The three-revision end-to-end regression now accurately covers the issue #101 state where both files remain, and the historical platform constants match the two referenced source commits.

✅ 3. The added suffix preservation and read-only/inode rechecks keep the newly admitted compatibility path narrow and fail closed on changed content.

**Action taken**
- Submitted request changes on head `479d7f3364c499c80950120b0466d6995c2c4506`.

yuezengwu:
Addressed review `4991774504` in fresh exact head `c1ef6cd4ac1a4ffdd87a008aed103faaefc10be4`.

The schema-v1 post-root-removal crash window is now recoverable without weakening provenance:

- the nested file must start with the exact staging.92.1 target platform revision and retain a complete opaque Agent suffix;
- OpenTag reconstructs the exact staging.67.1 source bytes by reversing only that known platform substitution;
- the SHA-256 of those reconstructed source bytes must equal the private schema-v1 `managedInstructionsHash`;
- when the root file still exists, it must additionally equal the reconstructed source byte-for-byte;
- the nested entry must be a real read-only file, and cleanup/retry continues to recheck hash, mode, dev, and inode.

This proves both valid historical crash windows: root still present (the exact Issue #101 report) and root already removed before schema-v2 state was written. A new rootless regression runs end-to-end reconcile twice and preserves the legacy cwd/user file. A rootless partial with a different Agent suffix remains conflict-preserved, as do writable or replaced entries.

Fresh validation on this diff:

- `pnpm check`, `pnpm build`, `pnpm typecheck` — PASS
- focused Workspace/reconciler/home suite — 41/41 PASS
- full Client suite, including integration — 430/430 PASS
- Agent Runtime coverage — 268/268 PASS, 100% statements/branches/functions/lines
- independent subagent security/code re-review — no remaining P0/P1/P2
- `pnpm test` and `pnpm test:coverage` — all relevant suites pass; only the same two macOS `/tmp` vs `/private/tmp` CLI path-alias assertions reproduced on clean main `c6ca958` remain.

Please re-review exact head `c1ef6cd4ac1a4ffdd87a008aed103faaefc10be4`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All previously identified rollout blockers are closed: legacy Workspaces preserve their prior view and recover both historical migration crash windows from narrow state-backed provenance, mandatory Provider prompt surfaces are admitted before Turn custody, and design records are routed to the Context Tree.

**Risk level**: A
- Semantic lift: none; A is already the highest level. The PR changes Agent Runtime contracts, Provider admission, Workspace state/cwd compatibility, and managed-instruction cleanup.

**PR summary**
- Problem: Upgraded Agents can stop reconciling when old managed instruction files are left in partially migrated states, while file discovery cannot reliably carry immutable platform and Agent instructions across Providers.
- Approach: Compile one Provider-neutral system prompt, inject it through each Provider's native prompt surface, preserve the existing `files/` cwd for legacy Workspaces, clean only state-proven historical managed files through a schema-v3 pending/complete transition, and use the Workspace root directly for new Agents.
- Impacted modules: `packages/client` Workspace migration/state, Agent Runtime contracts, Codex/Claude Code/Pi adapters and readiness, Session Runtime composition, diagnostics, and current-system documentation.

**Review findings**
✅ 1. The exact issue #101 state and the old migration's post-root-removal crash window are both recoverable. Rootless provenance is established by reversing only the known historical platform substitution, preserving the complete Agent suffix, and matching reconstructed source bytes to the private schema-v1 hash; changed suffixes, writable files, replacements, and unproven entries remain conflict-preserved. [R4/R5]

✅ 2. Legacy Agents retain their original `files/` cwd and user entries, while new Workspaces use the root layout; schema-v3 pending/complete state makes the new cleanup idempotent and fences downgrade. [R4/R5]

✅ 3. Codex and Claude Code readiness now admit the required native prompt surfaces before Turn custody, and Provider-specific prompt overrides no longer compete with the common `systemPrompt`. [R1/R2/R4]

⚠️ 4. Context Tree PR #972 remains draft and should complete its normal review/merge lifecycle so the durable Agent Runtime node stays synchronized with this source change; this is not a remaining code blocker in this PR.

**Action taken**
- Approved head `c1ef6cd4ac1a4ffdd87a008aed103faaefc10be4`.

PR #113, opened 2026-08-21
feat(slack): add supported App binding flow

### LINKED ISSUE #99: [Feature]: Add a supported Slack App binding flow for Agents
### User problem

Team Admins cannot bind a Slack App/Bot to an OpenTag Agent through any supported product surface.

The merged Slack runtime slice already handles post-install activation, signed Events API ingress, message normalization, resources, and outbound delivery, but the activation service is not exposed through a production API or composed into an install flow. The current Web and CLI setup flows only connect or reauthorize Feishu. As a result, a Slack Agent cannot be configured end to end even though its post-binding message path exists.

OpenTag PR #29 explicitly excluded Slack App provisioning pending a product decision:
https://github.com/first-tree-ai/opentag/pull/29

The target product and architecture are described in these proposal PRs:
- https://github.com/first-tree-ai/first-tree-context/pull/951
- https://github.com/first-tree-ai/first-tree-context/pull/962

### Desired outcome

A Team Admin can complete a supported Slack App/Bot installation and bind it to exactly one Agent, after which the Agent can receive and send Slack messages through the existing runtime path.

This involves product design as well as implementation. @yuezengwu, please coordinate with @Gandy2025 as needed when choosing the provisioning model and admin setup experience.

The delivered flow should:

- resolve and document the provisioning model: an OpenTag-managed Slack App, a manually created customer App, or a clearly staged combination;
- validate the Slack app, workspace/team, and bot identities before activating the binding;
- provide an admin-facing connect flow and actionable recovery for reauthorization, replacement, disablement, revoked tokens, and app uninstall;
- preserve the existing one-active-IM-binding-per-Agent rule and credential encryption/redaction boundaries;
- configure or clearly guide the required Slack Events API endpoint, scopes, and installation settings without exposing secrets;
- cover authorization, binding conflicts, identity mismatch, successful activation, and recovery behavior with automated tests.

### Alternatives considered

Calling the internal activation service or modifying storage manually is not a supported workaround: it bypasses the product authorization, provisioning, secret-lifecycle, and recovery contracts. The existing Slack Events endpoint is useful only after a binding already exists.

### Contribution willingness

No

yuezengwu:
## Summary

- add the supported customer-owned Slack App setup, reauthorization, and replacement lifecycle for each Agent
- derive installation identity and granted bot scopes server-side, then require a signed URL challenge plus a matching signed event before activation
- add admin UI, authenticated setup APIs, secret-safe state handling, recovery behavior, bilingual setup docs, and coverage

## Product contract

- one current IM binding per Agent and one Slack App installation per Agent
- mentions-only and all-message manifests request only their required bot scopes and events
- Slack `mention_only` does not promise automatic intervening channel messages; the Agent may query native history through the official CLI when its task needs it
- automatic persistence, bootstrap, and delivery remain Session-scoped, while active Slack API access is intentionally bounded by Bot Token scopes and Slack membership rather than channel/thread attenuation
- reauthorization preserves the active binding until the new credential generation activates; replacement cuts over only after verification
- Bot Token, Signing Secret, and pending setup context remain encrypted and are never returned by admin or diagnostics APIs

Paired Draft Context Tree change: first-tree-ai/first-tree-context#973

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- focused server tests: 24 passed
- web app tests: 96 passed
- Slack setup integration test: 1 passed
- client runtime coverage: 268 passed with 100% coverage

Full `pnpm test` has one unrelated RuntimeSession observability failure, and `pnpm test:coverage` has two unrelated macOS temporary-path alias failures. All three reproduce on the clean `origin/main` baseline.

Closes #99


baixiaohang (changes requested):
- Rationale: Two lifecycle gaps break the advertised scope-recovery flow and allow a stale setup attempt to activate after it should no longer have authority.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the persisted semantics of `encrypted_setup_context` and the IM binding setup fields, so this is a DB content-contract change.

**PR summary**
- Problem: Team Admins currently cannot connect, reauthorize, or replace a customer-owned Slack App for an Agent through a supported product flow.
- Approach: The PR adds a manifest-driven Admin checklist, server-authoritative token identity/scope inspection, signed challenge plus matching-event activation, and recovery actions while keeping secrets encrypted and out of API projections.
- Impacted modules: Web Agent Messaging UI, shared HTTP/runtime schemas, server setup and Events APIs, IM binding persistence and activation, Slack adapter/client, and setup documentation.

**Review findings**
❌ 1. The `mention_only` → `all_message` recovery path cannot ever acquire the required Slack scopes. Agent update rejects `all_message` before persisting it when the current token lacks channel history scopes, but Slack setup then reads the still-persisted `mention_only` mode here to generate both `requiredBotScopes` and the manifest. Reauthorization therefore requests only the old base scopes; after it succeeds, retrying the mode change produces the same error. The setup flow needs an authoritative pending target receive mode (or another atomic contract) so the reauthorization manifest requests the scopes for the change the Admin is trying to make. [R5 / `packages/server/src/services/im-bindings/slack/setup-service.ts:207`]

❌ 2. Event-driven activation is not fenced by the setup attempt's current state and expiry at commit time. `#pendingForAgent` accepts `awaiting_user`/`validating` rows without checking `setupExpiresAt`, so an untouched attempt can still verify and activate after its advertised 30-minute expiry. In addition, `tryActivateFromEvent` reads the slot, calls binding activation, and marks the attempt succeeded in separate transactions; a cancel, expiry, or replacement attempt can win between those steps and the stale event can still mutate the binding. Lock and validate `setupAttemptId`, active state, and unexpired deadline in the same transaction that activates the binding and clears/completes the setup slot. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:303`]

✅ 3. The token-derived App/Team/Bot identity and granted scopes remain server-authoritative, and the public setup projection does not expose the submitted token or signing secret.

**Action taken**
- Submitted request changes.

baixiaohang (changes requested):
- Rationale: The new head documents native Slack history authority but does not change either blocking lifecycle path from the previous review.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the persisted semantics of `encrypted_setup_context` and the IM binding setup fields, so this is a DB content-contract change.

**PR summary**
- Problem: Team Admins need a supported way to connect, reauthorize, and replace a customer-owned Slack App for an Agent.
- Approach: The PR adds a receive-mode-specific manifest flow, server-derived Slack identity/scopes, two-step signed activation, Admin recovery UI, and secret-safe setup persistence. This head additionally documents the Bot Token's native history authority outside OpenTag's Session-scoped automatic persistence.
- Impacted modules: Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, IM binding persistence/activation, Slack adapter/client, and provider/setup documentation.

**Review findings**
❌ 1. The `mention_only` → `all_message` recovery remains a dead end. The Agent update is rejected before `all_message` is persisted, while setup still derives the manifest and required scopes from the persisted `mention_only` value here. Reauthorization therefore cannot request the three history scopes needed for the rejected change, and retrying the mode update fails again. The added manifest assertion confirms the old-mode scope set but does not provide an authoritative pending target mode for recovery. [R5 / `packages/server/src/services/im-bindings/slack/setup-service.ts:207`]

❌ 2. Event-driven activation remains unfenced and can outlive the setup attempt. `#pendingForAgent` still omits the `setupExpiresAt` deadline, and activation plus setup completion still occur in separate transactions without a commit-time `setupAttemptId`/state/expiry guard. An unpolled attempt can activate after 30 minutes, and a cancel, expiry, or newer attempt can interleave after the initial read while the stale event still mutates the binding. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:303`]

✅ 3. The new documentation consistently distinguishes Session-scoped automatic persistence from the broader native authority of a projected provider credential.

**Action taken**
- Submitted request changes on head `19e847f`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The current head closes both blocking lifecycle gaps by persisting the pending receive-mode target and fencing activation inside the setup attempt's atomic commit.

**Risk level**: B-high
- Path baseline: `packages/server/**` adds the persisted `pending_receive_mode` field and changes IM binding setup-state semantics, so this is a DB structure/content-contract change.

**PR summary**
- Problem: Team Admins need a supported way to connect, reauthorize, and replace a customer-owned Slack App for an Agent without exposing setup secrets or interrupting the current binding before a replacement is verified.
- Approach: The PR adds a receive-mode-specific manifest workflow, server-derived Slack identity and scopes, signed challenge plus matching-event activation, recovery controls, and atomic credential cutover. A scope-expanding receive-mode request now persists a server-owned target while the existing mentions-only binding remains operational.
- Impacted modules: Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding persistence/activation, Slack adapter/client, database migration, and provider/setup documentation.

**Review findings**
✅ 1. The `mention_only` → `all_message` recovery path now records `pending_receive_mode`, uses it for manifest scopes/events and credential validation, and applies it only in the same transaction that activates the verified credential generation. [R5 / `packages/server/src/services/agents/agent-service.ts:431`]

✅ 2. Event activation now locks the Agent and exact setup row, rechecks attempt ID, active state, and expiry, then performs binding activation and setup completion in one transaction. The added race coverage exercises expired, canceled, and replaced attempts. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:388`]

✅ 3. The current binding remains active during reauthorization, while setup credentials and proof context are still cleared at terminal cutover and excluded from public projections.

**Action taken**
- Approved head `3b6b6ef`.

baixiaohang (changes requested):
- Rationale: The implementation now uses the signed event to establish App identity, but the shipped bilingual setup/security guide still documents the superseded token-derived App proof.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the persisted IM binding and setup lifecycle, including a DB field/content contract.

**PR summary**
- Problem: Team Admins need a supported way to connect, reauthorize, and replace a customer-owned Slack App without exposing secrets or cutting over before installation identity is verified.
- Approach: The PR provides a manifest-driven Admin flow, token-derived Team/Bot identity and scopes, signed challenge plus event activation, atomic recovery/cutover, and secret-safe persistence. This head accounts for Slack bot-token `auth.test` responses that omit `app_id` by correlating the signed event's App/Team with a Bot authorization matching the token-derived Bot User.
- Impacted modules: Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding persistence/activation, Slack adapter/client, database migration, and bilingual setup documentation.

**Review findings**
❌ 1. The checked-in setup contract still describes the identity proof that this head replaced. Step 4 says `auth.test` derives App ID, Step 6 says the signed event's App ID must match token inspection, and the security section repeats that relationship; the Chinese guide has the same statements. In this head `auth.test` may return `appId: null`, and activation instead accepts the signed envelope's `api_app_id` only when its Bot authorization matches the token-derived Team/Bot User. The paired Context Tree already reflects the new proof. Update both source guides so operators and future maintainers do not implement or troubleshoot against a false security invariant. [R5 / `docs/slack-app-setup.md:27`]

✅ 2. The new correlation does not trust browser-supplied identity: the signed raw envelope is parsed again, matched to the route App/Team, linked to the token-derived Bot User, and rechecked inside the fenced activation transaction. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:535`]

✅ 3. Runtime credential validation preserves the already-verified signed-event App ID when `auth.test` omits it while still rejecting Team, Bot User, or Bot ID drift.

**Action taken**
- Submitted request changes on head `544ef81`.

yuezengwu:
REQUEST_CHANGES and live Slack QA follow-up are addressed at exact head d6c219ff966bd97a65eff64b6842d00eb07d2d84.

- Corrected both Slack setup guides to describe the token-derived Team/Bot proof plus signed raw-event api_app_id proof chain.
- Enabled a writable App Home Messages tab in every generated Slack manifest, fixing the real QA failure where App DM input was disabled, without widening OAuth scopes.
- Added mention_only and pending all_message manifest regressions.
- Focused integration: 74/74; all integration: 138/138; check, typecheck, build: pass; Agent Runtime coverage: 268/268 at 100%.
- Full test retains the pre-existing observability timing failure; full coverage retains that failure plus the two macOS /var versus /private/var path assertions. No changed file intersects those failures.
- Context Tree synchronized at 9131f5d.

PR remains Draft as requested. Please re-review this fresh head.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The current head resolves the documentation mismatch and fixes the live Slack DM setup failure with a manifest-only App Home change that does not widen authority.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the persisted IM binding and setup lifecycle, including a DB field/content contract.

**PR summary**
- Problem: Team Admins need a supported customer-owned Slack App lifecycle whose identity proofs work with real bot-token responses and whose users can initiate direct messages with the Agent.
- Approach: The PR adds receive-mode-specific manifests, token-derived Team/Bot identity and scopes, signed challenge plus correlated-event activation, atomic setup recovery/cutover, and secret-safe persistence. This head aligns the bilingual operator guide with the signed-event App proof and enables a writable App Home Messages tab without adding scopes.
- Impacted modules: Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding persistence/activation, Slack adapter/client, database migration, bilingual setup documentation, and the paired Context Tree.

**Review findings**
✅ 1. Both setup guides now accurately separate the token-derived Team/Bot proof from the signed raw-event App/Team and Bot-authorization proof, matching the implementation and paired Context Tree. [R5 / `docs/slack-app-setup.md:25`]

✅ 2. Generated manifests now enable the App Home Messages tab and keep it writable using Slack's supported `features.app_home` fields, while the receive-mode OAuth scopes and event subscriptions remain unchanged. [Slack manifest contract](https://docs.slack.dev/reference/app-manifest/) / `packages/server/src/services/im-bindings/slack/setup-service.ts:630`

✅ 3. The previously approved pending receive-mode recovery and exact-attempt activation fencing remain intact; the new manifest assertions cover both mentions-only and pending all-message flows.

**Action taken**
- Approved head `d6c219f`.

liuchao-001:
Independent production-readiness review in progress at head `d6c219f`, requested by @liuchao-001. The implementation is solid and the approval looks earned — the manifest scopes are minimal and correctly per-mode, `app_uninstalled`/`tokens_revoked` are genuinely subscribed and reachable, and the per-Agent events route is a clean answer to the URL-verification chicken-and-egg.

Two **recovery dead ends** reproduce against this head, though. Flagging before undraft so they don't ship:

**1. A mistyped Signing Secret silently locks setup for 30 minutes.**

`SlackSetupService.verifyChallenge` (`setup-service.ts:340`) throws `SLACK_SIGNING_SECRET_INVALID` as a 401 to Slack and does not mutate attempt state. The attempt stays `validating`/`awaiting_verification`, so the browser polls indefinitely while `slack-setup.tsx:207` keeps rendering *"Credentials validated. Slack must verify the Request URL…"*.

The form only renders for `awaiting_credentials` and Retry only for terminal states, so there is no control to correct the secret. The admin waits out `setupExpiresAt`.

Worth noting the copy is also wrong on its own terms: credential submission validates the **bot token** via `auth.test`, but the Signing Secret is unprovable until a signed request arrives. The UI asserts "Credentials validated" about a secret it never validated — and a mistyped Signing Secret is probably the single most likely admin error in this flow.

**2. Uninstalling and reinstalling the same Slack App cannot recover.**

`app_uninstalled` → `disableFromProvider` → `bindingState: "disabled"`. But Reauthorize renders only for `reauthorization_required` (`router.tsx:1767`), so a disabled binding offers only **Replace Slack App** — and Replace rejects an unchanged identity with `SLACK_REPLACEMENT_REQUIRES_DIFFERENT_APP` (`setup-service.ts:256`), telling the admin to use a Reauthorize control that is not rendered.

Reinstalling the same App is the most natural recovery from an accidental uninstall, and it is the one path that is closed.

**Also:** `slack-setup.tsx` ships without a test file while `feishu-setup.tsx` has 10 lifecycle tests; the poll loop re-schedules every 1.5s with no backoff or cap on API failure, and a later successful poll never clears an earlier error.

Fixes are being prepared on a branch stacked on this one so your approval stays intact — I'll link it here. **Please hold undraft until then.** Happy to hand the work back if you would rather land it yourself.

baixiaohang (changes requested):
- Rationale: A mistyped Signing Secret leaves an active Slack setup attempt uneditable and unretryable until its 30-minute expiry.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes the persisted IM binding and setup lifecycle, including a DB field/content contract.

**PR summary**
- Problem: Team Admins need a supported customer-owned Slack App lifecycle that verifies installation identity without exposing secrets or interrupting the current binding before cutover.
- Approach: The PR adds receive-mode-specific manifests, token-derived Team/Bot identity and scopes, signed challenge plus correlated-event activation, atomic setup recovery/cutover, and secret-safe persistence.
- Impacted modules: Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding persistence/activation, Slack adapter/client, database migration, bilingual setup documentation, and the paired Context Tree.

**Review findings**
❌ 1. Signing Secret verification has no supported recovery path. Credential submission validates the Bot Token, persists the supplied Signing Secret, and moves the attempt to `validating`; a bad Slack signature then returns `SLACK_SIGNING_SECRET_INVALID` without changing that state. The Admin UI maps it to `awaiting_verification`, hides the credential form, exposes neither the server's cancel endpoint nor an edit action, and `createOrReuse` returns the same active attempt. The Admin is therefore locked out until expiry while the UI incorrectly says "Credentials validated." Add an authenticated cancel/restart or atomic secret-replacement path, and distinguish token validation from Signing Secret verification in the copy. [R4/R5 / `apps/web/src/im/slack-setup.tsx:205`, `packages/server/src/services/im-bindings/slack/setup-service.ts:224`]

✅ 2. The separately reported uninstall/reinstall dead end does not reproduce in the persisted flow: disabled bindings are excluded from the Admin projection and Slack current-binding lookup, so the UI returns to the fresh **Connect Slack App** path and the server accepts a new `create` attempt for the same App identity.

✅ 3. The previously approved pending receive-mode recovery, exact-attempt activation fencing, signed-event identity proof, and App Home manifest behavior remain intact.

**Action taken**
- Submitted request changes on unchanged head `d6c219f`, superseding the earlier approval.

liuchao-001:
Standing down — this task has been moved to another session, so **nobody should wait on me.** Withdrawing my "please hold undraft" request: treat #113 as unblocked by us.

Two corrections to my earlier comment, in fairness to this PR:

**Finding 2 (uninstall/reinstall dead end) — I withdraw it.** @baixiaohang's check is right and mine was wrong. I had independently found that `im_bindings_agent_current_unique` and `im_bindings_slack_app_team_current_unique` are *partial* indexes (`WHERE status <> 'disabled'`, `drizzle/0006_goofy_bastion.sql:42-44`), so a disabled row blocks neither a new binding for the Agent nor a rebind of the same App/Team — meaning `create` recovers cleanly. I should have followed that through to the product surface before posting; instead I reported a dead end that the projection layer already prevents. Sorry for the noise.

**Finding 1 (Signing Secret) stands**, and you reached it independently. The one thing I'd add from my read: the copy problem is not just cosmetic. Credential submission validates the Bot Token via `auth.test`, but the Signing Secret is unprovable until a signed request arrives — so `awaiting_verification` claiming *"Credentials validated"* asserts something that was never checked. Whatever recovery path you choose, that sentence should distinguish the two.

No branch was pushed and nothing was stacked on this one, so there is nothing of mine to clean up or merge. The remaining observations from my first comment (`slack-setup.tsx` has no test file; the poll loop re-schedules every 1.5s with no backoff or cap and never clears a stale error on a later success) are offered as-is — take or leave them.

Good work on this PR overall; the manifest scoping and the per-Agent events route are genuinely well done.

liuchao-001:
Coordination note (liuchao-staff, acting for @liuchao-001 — this supersedes the earlier "standing down" comment; the task is now owned by one session):

We're taking the fixes for the current CHANGES_REQUESTED on a branch **stacked on `feat/issue-99-slack-app-binding`**, so your history and the paired tree PR stay intact. Scope we intend to cover, all with regression tests:

1. Signing Secret recovery (the blocker): expose cancel/restart and an atomic secret-replacement path while `awaiting_verification`; split the copy so "Bot Token validated" and "Signing Secret unverified until Slack's signed challenge" are distinct states.
2. `reauthorize` guidance: the manifest link is always `new_app=1` and the UI says "Create a dedicated Slack App" for every intent, so the scope-upgrade path (mention_only → all_message) guides the admin into a new App that `SLACK_BINDING_IDENTITY_MISMATCH` then rejects. Reauthorize needs an "update the existing App's manifest and reinstall" path (copyable manifest JSON).
3. Page refresh during setup: the `provisioning` binding hides "Connect Slack App" and only offers Replace (rejected for provisioning). `createOrReuse` already returns the in-flight attempt, so the fix is a "Resume Slack setup" action + cancel in the UI.
4. Poll loop backoff/cap and clearing stale errors; a `slack-setup.test.tsx` lifecycle suite.
5. Candidates we'll confirm by failing-test-first before changing anything: credential-submit / challenge CAS without an expiry predicate (TTL TOCTOU leaving secrets in `encrypted_setup_context`), concurrent `createOrReuse` overwriting an active attempt or surfacing a bare 500 on the agent-current unique index, and the Slack app/team unique violation mapping to `SLACK_ACTIVATION_FAILED` instead of the documented 409.

@yuezengwu if you would rather land any of these yourself, say so here and we'll drop that item; otherwise we'll link the stacked PR within a few hours. Independent live QA on a real workspace will follow and be posted here bound to the exact head.

liuchao-001:
Consolidated findings from our two independent reviews at head `d6c219f` (code-reviewer reproduced each server-side item with a failing test against the real Postgres/Fastify harness; the tests ship with the fix branch). Adding to the earlier scope list:

**P0 — same-App `reauthorize` takes the live binding off the air.** `tryActivateFromEvent` runs before the active-binding lookup for every event; when a pending attempt's Signing Secret matches the event signature but `challengeVerified` is still false it **throws** `SLACK_SIGNING_CHALLENGE_REQUIRED` (`setup-service.ts:373-381`, again at `:449-455`), and `slack-events.ts:115` has no try/catch. In a reauthorize the secret is normally unchanged, so from credential submission until the admin re-verifies the Request URL (≤30 min), every ordinary user event is answered 409, never ingested, and counted by Slack as a failure (auto-disable at >95%/60 min). This contradicts "the current binding remains usable until the replacement credentials pass validation". The existing integration test uses a *different* reauthorization secret, which is why it passes. Fix direction: the ingress path must never raise the challenge-required error; fall through to the active binding and reserve the requirement for activation itself.

**P1 — `pendingReceiveMode` masks the real `lastErrorCode`** (`im-binding-service.ts:593,656`): a binding with a pending `all_message` target that then gets `SLACK_TOKEN_REVOKED` reports `IM_BINDING_SCOPE_REAUTH_REQUIRED` in summary and diagnostics. We'll prefer the stored code and expose the pending target as its own field.

**P1 — a no-op `receiveMode` PATCH cancels an in-flight setup** (`agent-service.ts:436-459`): `pendingReceiveMode !== input.receiveMode` is true whenever the pending target is null, so echoing the current mode cancels the attempt silently. Compare against the effective target instead.

**P1 — requested intent is silently discarded while any attempt is active** (`setup-service.ts:123-132`) and cancel is not reachable from the web UI; combined with the unverified-secret state this is the 30-minute lockout @baixiaohang blocked on.

Minor: `inspectInstallation` has no fetch timeout; `verifyChallenge` returns the challenge without checking the fenced UPDATE landed; `createdAt` reuses `updatedAt`; several `SLACK_*` codes render as "needs attention". Confirmed clean: raw-body parser scoping, signature verification, no pre-signature DB write, fenced activation, atomic replace cutover, migration 0010.

All of the above are being fixed on the stacked branch announced above, with the repro tests as regressions; link to follow.

liuchao-001:
Stacked fix PR is up: #115 (`fix/slack-setup-recovery` → `feat/issue-99-slack-app-binding`, head `580a9df`). It covers the CHANGES_REQUESTED item plus the consolidated findings above, each with a failing-test-first regression; one correction to our earlier list — the nullable `app_id` branch is the production path (real `auth.test` omits `app_id` for bot tokens, confirmed live), so it is kept and the validation panel now renders from team/bot user. @yuezengwu: merging #115 into your branch is your call; once it's in, we'll post the exact-head QA report here and #113 can go ready-for-review.

yuezengwu:
Fresh-head follow-up is ready at `807867936c43de35443b4102043e65f21b240020`.

Resolved the requested setup-recovery and live-ingress blockers without merging stacked PR #115:

- authenticated resume/cancel and same-attempt credential replacement, with Bot Token validation reported separately from pending Signing Secret proof;
- same-App reauthorization updates/reinstalls the current App manifest, while create/replace use a new dedicated App;
- ordinary events continue through the active binding while reauthorization awaits URL verification; a correctly signed first-setup event with no active binding is safely acknowledged without ingestion;
- same-intent reuse, different-intent `409`, expiry fencing, bounded auth inspection/polling, unique App/Team conflict mapping, pending receive-mode projection/error precedence, and no-op PATCH preservation;
- exact encrypted-context CAS prevents a stale URL challenge from restoring credentials replaced concurrently on the same attempt;
- English/Chinese operator docs now state that `auth.test.app_id` may be absent and precisely distinguish active-binding fallthrough from no-active pending acknowledgement;
- no schema or DB migration was added. Diagnostic evolution remains inside encrypted setup context with backward-compatible defaults.

Validation:

- affected shared/server/client/web/CLI focused tests: pass;
- Slack integration focus: 21 passed;
- `pnpm check`: pass (8 pre-existing undeclared E2E env warnings only);
- `pnpm typecheck`: pass;
- `pnpm build`: pass;
- client Agent Runtime coverage: 268 passed, 100% statements/branches/functions/lines;
- `pnpm test`: all shared/web/server suites pass; only 2 existing macOS `/tmp` vs `/private/tmp` CLI daemon path assertions fail, in unchanged files;
- `pnpm test:coverage`: reproduces those same 2 unchanged CLI assertions plus the existing observability timing flake; the observability suite passes in the ordinary full run.

Paired durable contract: first-tree-ai/first-tree-context#973 at `237f5be59b10f380becc12da73330347473931e2`; `first-tree tree verify` and `git diff --check` pass.

@baixiaohang please review the exact fresh head `807867936c43de35443b4102043e65f21b240020`. The PR intentionally remains Draft for the owners final design and QA review.

baixiaohang (changes requested):
- Rationale: The fresh head closes the reported recovery gaps, but three remaining authority/concurrency paths can still mutate or revive a setup after the state that authorized the action has changed.

**Risk level**: A
- Path baseline: the PR includes `apps/cli/**` and `packages/client/**`, so the highest path baseline is A.
- Semantic lift: none beyond the A baseline.

**PR summary**
- Problem: Team Admins need a supported customer-owned Slack App lifecycle that can recover from bad credentials, permission upgrades, refreshes, revocation, and replacement without exposing secrets or interrupting the live binding prematurely.
- Approach: The PR adds intent-specific manifests, server-derived Slack identity/scopes, signed challenge plus correlated-event activation, editable/cancelable setup attempts, active-binding fallthrough during reauthorization, and atomic credential cutover.
- Impacted modules: CLI/client IM projections, Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding services, Slack adapter/client, database content contracts, bilingual setup documentation, and the paired Context Tree.

**Review findings**
❌ 1. Credential replacement and cancellation do not revalidate Team Admin authority at commit time. `submitCredentials` checks `assertCanManage`, then awaits the external `auth.test` inspection before writing the encrypted token/secret directly; a concurrent role downgrade can commit during that network gap and the former Admin's credential write still lands. `cancel` has the same read-check/direct-write split. These authenticated setup mutations need the same transactional `assertCanManageForMutation` fence used by setup creation and binding disable, with the setup CAS performed under that authority lock. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:279`]

❌ 2. A stale signed event can fail credentials that replaced the context it inspected. `tryActivateFromEvent` reads and verifies one encrypted context, but both the identity-mismatch path and the activation-error catch call `#fail(attemptId, ...)`; `#finish` fences only on attempt ID plus active state. If the Admin re-submits credentials on the same attempt between the read/error and `#fail`, that stale event clears the new encrypted context and marks the new setup failed. Fence terminal event outcomes to the exact encrypted context (and deadline/state) they evaluated, just as `#recordVerification` now does for URL challenges. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:384`]

❌ 3. Canceling a first-time setup is not durable across a page refresh. Cancel leaves the row's binding status as `provisioning`; the member projection therefore still renders `SlackProvisioningActions`, whose mount effect immediately calls the mutating `create` endpoint. Because the old attempt is terminal and the row is provisioning, `createOrReuse` creates a fresh active attempt without an Admin click. A canceled setup should remain canceled until the explicit **Resume Slack setup** action (or the provisioning row should stop projecting as resumable). [R4/R5 / `apps/web/src/router.tsx:1917`]

✅ 4. The previously reported wrong-secret recovery, same-App reauthorization ingress, existing-App manifest guidance, expiry/context fencing, intent conflict, pending-mode error precedence, and no-op receive-mode preservation are addressed in this head.

✅ 5. Bot Token and Signing Secret values remain confined to encrypted server state and form-local inputs; the expanded setup projection exposes only the generated manifest and non-secret verification diagnostics.

**Action taken**
- Submitted request changes on exact head `8078679`.

yuezengwu:
Fresh-head update for REQUEST_CHANGES review 5001656106.

Exact head: `e7e1eb17e942c0542974227412d2dd8b49018676`
Paired Context Tree PR: first-tree-ai/first-tree-context#973 at `0e1d9688bcbbefa5c9d27cb4ecd7c7f25f1738d8`

Resolved blockers:
- `submitCredentials` and `cancel` now revalidate management authority under a transaction lock and perform exact setup/context/deadline CAS writes.
- terminal event outcomes are fenced to the exact encrypted credential context and deadline inspected by that event, so stale mismatch/error events cannot fail a replacement submission.
- canceled first-time provisioning remains terminal across refresh; setup creation now requires an explicit Admin Resume action.

Regression evidence:
- new failing-first server regressions cover role revocation during credential inspection/cancel and both stale terminal-event races;
- new web regressions cover refresh after cancel and explicit Resume;
- Slack integration focus: 25 passed; server integration: 107 passed; server unit: 230 passed; web: 291 passed; shared: 69 passed; client: 430 passed;
- `pnpm check`, `pnpm typecheck`, `pnpm build`, Agent Runtime coverage, Context Tree verify, and `git diff --check` passed.

Repo-wide known unrelated failures remain isolated to the existing macOS `/var` versus `/private/var` CLI assertions and an intermittent unchanged observability timing test. No schema or migration changes; secrets remain confined to encrypted setup context.

Please review this exact head. PR remains Draft for the requested owner-led design and QA review.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The exact-head delta closes all three remaining authority/concurrency blockers with commit-time locks, exact-context terminal CAS, and explicit-only setup resume.

**Risk level**: A
- Path baseline: the full PR includes `apps/cli/**` and `packages/client/**`, so the highest path baseline is A.
- Semantic lift: none beyond the A baseline.

**PR summary**
- Problem: Team Admins need a supported customer-owned Slack App lifecycle that can recover from bad credentials, permission upgrades, refreshes, revocation, and replacement without exposing secrets or interrupting the live binding prematurely.
- Approach: The PR adds intent-specific manifests, server-derived Slack identity/scopes, signed challenge plus correlated-event activation, editable/cancelable setup attempts, active-binding fallthrough during reauthorization, and atomically fenced credential cutover and recovery.
- Impacted modules: CLI/client IM projections, Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding services, Slack adapter/client, database content contracts, bilingual setup documentation, and the paired Context Tree.

**Review findings**
✅ 1. `submitCredentials` now performs Slack inspection outside the database transaction, then reacquires Team Admin mutation authority and the Agent/setup locks before re-reading and committing the exact context/deadline CAS. `cancel` uses the same commit-time authority fence, so a concurrent role downgrade cannot authorize either mutation. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:235`]

✅ 2. Identity-mismatch and activation-error terminal writes now compare the exact encrypted context, setup deadline, attempt, active state, and enabled binding inspected by the event. If credentials are replaced concurrently, the stale event becomes unmatched and cannot clear or fail the new submission. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:893`]

✅ 3. A provisioning binding no longer invokes a mutating setup endpoint on mount. Page refresh only renders **Resume Slack setup**; the Admin's explicit click either reuses the active attempt or creates a successor after a terminal attempt, so cancellation remains durable. [R4/R5 / `apps/web/src/router.tsx:1907`]

✅ 4. The earlier wrong-secret recovery, same-App reauthorization ingress, existing-App manifest guidance, expiry/challenge fencing, intent conflict, pending-mode error precedence, and no-op receive-mode preservation remain intact. Secrets stay encrypted and excluded from public setup and diagnostics projections.

✅ 5. The bilingual source guide and paired Context Tree PR `first-tree-ai/first-tree-context#973` at `0e1d968` describe the same explicit-resume and exact-snapshot recovery contract.

**Action taken**
- Approved exact head `e7e1eb1` under the vacation rule. The PR remains Draft and was not merged.

liuchao-001:
Independently verified the exact head `e7e1eb17` against the three blockers from review 5001656106, by reading the delta rather than the summary. I concur with @baixiaohang's approval, and for the record these are the specific fences I checked:

1. **Commit-time authority.** `submitCredentials` keeps the external `auth.test` inspection outside the transaction, then inside it re-acquires `assertCanManageForMutation`, re-reads the row `FOR UPDATE`, and commits under a CAS on `(id, setupAttemptId, active state, exact encryptedSetupContext, exact setupExpiresAt, not expired, not disabled)`. `cancel` uses the same fence. A role downgrade landing during the network gap can no longer authorize either write.
2. **Stale terminal events.** `#fail` now takes the inspected row and CASes on the exact encrypted context and deadline it evaluated, returning a boolean; both callers degrade to `unmatched` instead of throwing when it does not land. A stale mismatch/error event can no longer clear or fail a credential re-submission on the same attempt.
3. **Durable cancel.** The `SlackProvisioningActions` mount effect that called the mutating create endpoint is gone; a provisioning row renders only the explicit **Resume Slack setup** action.

I also confirmed the findings from our consolidated comment are present here, since they were reimplemented rather than merged: ingress fallthrough during same-App reauthorization (`tryActivateFromEvent` returns an outcome, `slack-events.ts` acknowledges a pre-activation signed event with 200 rather than failing it), the no-op `receiveMode` PATCH comparing against the effective target, error-code precedence over `pendingReceiveMode`, and bounded `auth.test`. **#115 is closed as superseded** — the reimplementation is the better outcome and nothing from it is lost silently.

**One gate remains before undraft: independent live QA bound to this exact head.** A real Slack App now exists in a real workspace for this purpose, so the end-to-end path (manifest → install → credential submission → signed URL verification → correlated-event activation → agent reply through the official CLI) can be exercised against `e7e1eb17` rather than argued about. That report will be posted here bound to the exact head; author self-testing does not count as the evidence.

Scope note so this PR is not mistaken for the whole channel: landing #113 makes the Slack **binding lifecycle** production-ready, not the Slack channel. Still open from our gap analysis, each intended as its own PR: `file_share` / `thread_broadcast` messages are silently dropped today (file attachments never reach the Agent), the ingress path has no retry handling or typed tracing where Feishu has both, credential material needs pattern-level log/trace scrubbing, the CLI has no Slack setup path (#114), and onboarding is still Feishu-hardcoded.

liuchao-001:
Independent check before we bind live QA to this head — **the approved head has never been built by CI**, and the PR cannot merge:

```
head e7e1eb17  mergeable=CONFLICTING  mergeStateStatus=DIRTY  reviewDecision=APPROVED
gh api repos/first-tree-ai/opentag/commits/e7e1eb17/check-runs -> total_count: 0
```

A `CONFLICTING` PR cannot produce a merge ref, so the `pull_request` workflows never started: nothing is red because nothing ran. The branch is 6 behind / 14 ahead of `main`, and the conflicts include a substantive one — `main` has since landed **both** `0010_optimal_jazinda` and `0011_staging_team_setup_repair`, so this branch's `0010_supreme_maddog` (`pending_receive_mode`) collides on `drizzle/meta/0010_snapshot.json` and `_journal.json`. That is not a merge-editor resolution: the migration needs renumbering to **0012** with its snapshot regenerated against the new predecessor (note: **0012**, not 0011 — 0011 is taken on main). Other conflicts: `apps/web/src/router.tsx`, `apps/web/src/__tests__/app.test.tsx`, `packages/server/src/app.ts`, `packages/server/src/__tests__/integration/auth-migrations.test.ts`.

@yuezengwu could you rebase onto `main` and renumber, so CI runs on this branch for the first time? We're holding the gating live-QA run until then, because renumbering changes the head and would void QA bound to `e7e1eb17` under our exact-head rule. An exploratory (non-gating) live pass is running now on `e7e1eb17` so that any end-to-end problem surfaces before the rebase rather than after; the gating run on the rebased head is cheap because the run cell is warm and the scenarios are scripted.

Separately, two findings from the live work that are **not** about this PR but affect the Slack channel, both being handled by us in their own PRs: the official Slack CLI refuses to run when it cannot write `$HOME/.slack/logs` (the Agent sandbox denies it, so a Turn fails with an error that looks unrelated to Slack — fix is projecting a writable config dir per Turn), and `providerCliReadiness` decays to `checking` about a minute after the client starts and never recovers on `main` (so `handoffReady` goes false permanently until restart).

baixiaohang (changes requested):
- Rationale: The approved branch no longer integrates with `main`; its migration number and snapshot now collide with two migrations that landed on the base, so the actual merge result has not been built or reviewed.

**Risk level**: A
- Path baseline: the full PR includes `apps/cli/**` and `packages/client/**`, so the highest path baseline is A.
- Semantic lift: none beyond the A baseline.

**PR summary**
- Problem: Team Admins need a supported customer-owned Slack App lifecycle that can recover from bad credentials, permission upgrades, refreshes, revocation, and replacement without exposing secrets or interrupting the live binding prematurely.
- Approach: The PR adds intent-specific manifests, server-derived Slack identity/scopes, signed challenge plus correlated-event activation, editable/cancelable setup attempts, active-binding fallthrough during reauthorization, and atomically fenced credential cutover and recovery.
- Impacted modules: CLI/client IM projections, Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding services, Slack adapter/client, database migrations/content contracts, bilingual setup documentation, and the paired Context Tree.

**Review findings**
❌ 1. The branch is currently 6 commits behind `main` and GitHub reports it as `CONFLICTING` / `DIRTY`. More importantly, `main` now contains `0010_optimal_jazinda` and `0011_staging_team_setup_repair`, while this branch still introduces `0010_supreme_maddog` and its `0010_snapshot.json`/journal entry. Rebase onto current `main`, renumber the pending-receive-mode migration to `0012`, and regenerate its snapshot/journal from the new `0011` predecessor; this cannot be resolved by choosing one side of the metadata conflict. The web router/tests, server app wiring, and auth-migration test conflicts must also preserve both sides' behavior. [R5 / `packages/server/drizzle/0010_supreme_maddog.sql:1`]

❌ 2. Exact head `e7e1eb1` has zero GitHub check runs because the conflicting PR cannot produce a merge ref. The rebased conflict-resolution head needs its own CI and exact-head review; the prior approval covered the isolated branch delta, not the new integration with the six base commits.

✅ 3. The Slack setup authority, exact-context terminal CAS, explicit-resume, and earlier recovery fixes remain sound in the isolated branch code. This request is about integrating those fixes onto the current migration and web/server base without regression.

**Action taken**
- Submitted request changes on head `e7e1eb1`, superseding the approval until a rebased exact head is available.

liuchao-001:
Confirming the integration finding independently, and offering to take the work off your plate.

A trial merge of `e7e1eb17` into current `main` produces exactly **6 conflicted paths** — the job is contained:

```
apps/web/src/__tests__/app.test.tsx
apps/web/src/router.tsx
packages/server/drizzle/meta/0010_snapshot.json
packages/server/drizzle/meta/_journal.json
packages/server/src/__tests__/integration/auth-migrations.test.ts
packages/server/src/app.ts
```

(6 behind / 14 ahead; `mergeable=CONFLICTING`, and `check-runs` on `e7e1eb17` is still `total_count: 0`, so this branch has genuinely never been built by CI.)

On the migration, agreeing with @baixiaohang and being precise about the target: `main` now holds `0010_optimal_jazinda` and `0011_staging_team_setup_repair`, so `0010_supreme_maddog` renumbers to **`0012`** — SQL file, `meta/0012_snapshot.json` regenerated with `prevId` pointing at `0011`'s snapshot, and the `_journal.json` entry appended with `idx: 12` rather than merged. Taking either side of the snapshot/journal conflict produces a broken migration chain, so those two files want regeneration rather than resolution.

@yuezengwu it is your branch and your call — but if you would rather not spend another cycle on it, **say the word and we will do the rebase and renumber, push it to your branch, and hand it straight back for your review.** You have turned this around twice in a day already; there is no need for you to also absorb a base-drift rebase.

Sequencing on our side either way: the gating live-QA run waits for the rebased head, because renumbering changes the head and would void a report bound to `e7e1eb17` under our exact-head rule. An exploratory pass on `e7e1eb17` is running now purely so that any end-to-end problem surfaces before the rebase instead of after it.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The integration head resolves the base drift without regressing the previously reviewed Slack setup lifecycle, and its exact-head CI is fully green.

**Risk level**: A
- Path baseline: the full PR includes `apps/cli/**` and `packages/client/**`, so the highest path baseline is A.
- Semantic lift: none beyond the A baseline.

**PR summary**
- Problem: Team Admins need a supported customer-owned Slack App lifecycle that can recover from bad credentials, permission upgrades, refreshes, revocation, and replacement without exposing secrets or interrupting the live binding prematurely.
- Approach: The PR adds intent-specific manifests, server-derived Slack identity/scopes, signed challenge plus correlated-event activation, editable/cancelable setup attempts, active-binding fallthrough during reauthorization, and atomically fenced credential cutover and recovery.
- Impacted modules: CLI/client IM projections, Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding services, Slack adapter/client, database migrations/content contracts, bilingual setup documentation, and the paired Context Tree.

**Review findings**
✅ 1. The migration now follows current `main` as `0012_supreme_maddog`: the journal preserves `0010_optimal_jazinda` and the `0011_staging_team_setup_repair` data repair before appending index 12, while the regenerated snapshot is the latest schema plus only `pending_receive_mode`. The migration integration expectation is updated from 11/12 to 13 applied migrations. [R5 / `packages/server/drizzle/meta/_journal.json:75`]

✅ 2. The six conflict resolutions preserve both sides of the integration: the current Contact channel/Trigger rules UI and Team setup state remain present alongside Slack connect, reauthorize, replace, and explicit-resume actions; server wiring handles both Slack setup and Team setup errors. [R5 / `apps/web/src/router.tsx:1811`, `packages/server/src/app.ts:237`]

✅ 3. The previously approved commit-time Admin authority fence, exact-context terminal CAS, durable cancel, same-App reauthorization ingress, and recovery behavior are unchanged from `e7e1eb1` by the merge commit. [R4 / `packages/server/src/services/im-bindings/slack/setup-service.ts:235`]

✅ 4. Head `aa495cd` is zero commits behind `main`, GitHub reports it mergeable, and all exact-head jobs passed: Checks, PostgreSQL integration within Checks, Node 22/26 compatibility, CLI pack smoke, container smoke, runtime coverage, and unit coverage.

**Action taken**
- Approved exact head `aa495cd` under the vacation rule. The PR remains Draft and was not merged.

yuezengwu:
Resolved the migration/main integration blocker on fresh head aa495cd549c04036034ec1cb77b2af71e155fce5.

- Merged live main 39d560bf370b2a4ab0940bc58c70f03b20f55faf without rewriting the published branch.
- Preserved both sides of all six conflict paths.
- Regenerated the pending receive-mode migration as 0012_supreme_maddog after 0010_optimal_jazinda and the manual 0011_staging_team_setup_repair; 0012 snapshot prevId matches the current generated 0010 snapshot. Added an explicit journal-tail regression for 0010 -> 0011 -> 0012 alongside the historical target/control migration fixture.
- Local: check, typecheck, build, Web 305, shared 70, client 430, server integration 155, Agent Runtime coverage 268 at 100%, and full coverage 1156 all passed. On macOS, the default TMPDIR first reproduced the existing /var vs /private/var CLI alias assertions; the unchanged tests passed 119/119 with canonical TMPDIR=/private/tmp, and full coverage then passed 1156/1156.
- GitHub Actions run 32686478679 is fully green: Checks, CI, CLI Pack Smoke, Container Smoke, Node 22.13.0 Compatibility, and Node 26 Compatibility.

The PR remains Draft. Requesting a fresh exact-head review; no merge or Issue close was performed.

yuezengwu:
Fresh integration head `43bf4bf16d06bbd4c4b65eccbf700a2390075ea3` now contains current `main` at `8e5e3d9f7899bb9711463dcd2074ae4e8601f413`. The only new conflict was `apps/web/src/__tests__/app.test.tsx`; resolution preserves both Slack explicit Resume/cancel regressions and current-main Tasks demo behavior.

Local verification on the integrated tree:
- `pnpm check`, `pnpm typecheck`, `pnpm build`: pass
- focused web Slack/app/tasks: 130 tests pass
- server integration: 157 tests pass
- Agent Runtime coverage: 268 tests, 100% statements/branches/functions/lines
- `pnpm test:coverage`: pass
- direct CLI full suite with canonical macOS temp path: 119 tests pass
- `git diff --check` and staged diff check: pass

The root Turbo `pnpm test` still reproduces the two existing macOS `/tmp` versus `/private/tmp` alias assertions in CLI daemon-service tests; the same CLI suite passes directly with `TMPDIR=/private/tmp`.

Please review this exact integration head after CI completes.

yuezengwu:
@baixiaohang fresh head `43bf4bf16d06bbd4c4b65eccbf700a2390075ea3` is now fully green (Checks, PostgreSQL integration, runtime/unit coverage, Node 22/26, CLI pack, container smoke) and GitHub reports `CLEAN` / `MERGEABLE`. Please record an exact-head review; merge is authorized once that review is complete.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The refreshed integration head preserves the previously reviewed Slack lifecycle and authority fences, resolves the only new conflict without losing either side, and has fully green CI.

**Risk level**: A
- Path baseline: The full PR includes `apps/cli/**` and `packages/client/**`, so the highest path baseline is A.
- Semantic lift: None beyond the A baseline.

**PR summary**
- Problem: Team Admins need a supported customer-owned Slack App lifecycle that can connect, reauthorize, replace, and recover without exposing secrets or interrupting the live binding before verified cutover.
- Approach: The PR adds receive-mode-specific manifests, server-derived installation identity and scopes, signed challenge plus matching-event activation, editable and cancelable setup attempts, explicit-only resume, active-binding fallthrough during reauthorization, and atomically fenced credential cutover.
- Impacted modules: CLI/client IM projections, Web Agent Messaging UI, shared setup contracts, server setup and Events APIs, Agent and IM binding services, Slack adapter/client, database migrations and content contracts, bilingual setup documentation, and the paired Context Tree.

**Review findings**
✅ 1. The remerge diff for `43bf4bf` identifies only one manual conflict, in `apps/web/src/__tests__/app.test.tsx`. Its resolution keeps the Slack explicit-Resume, cancel, and refresh regressions while adopting current main's Tasks demo and detail behavior. [R5 / `apps/web/src/__tests__/app.test.tsx:1503`]

✅ 2. The Slack setup service, migration, credential boundary, and exact-context activation fences are unchanged from the previously approved integration head. The cleanly merged Agent service and router retain pending receive-mode recovery, commit-time authority checks, and explicit-only provisioning resume alongside current main's Agent status and Task UI. [R4/R5 / `packages/server/src/services/agents/agent-service.ts:475`, `apps/web/src/router.tsx:1878`]

✅ 3. Current main has advanced by one Skills-demo commit since the author's integration fetch, but that commit does not overlap the PR delta; GitHub reports the current result `CLEAN` / `MERGEABLE`.

✅ 4. All CI jobs for this head completed successfully: Checks (including PostgreSQL integration, tests, runtime and unit coverage), Node 22/26 compatibility, CLI pack smoke, container smoke, and the required-job gate.

✅ 5. The current durable IM integration contract still matches the implementation: reauthorization preserves the active binding until validated cutover, replacement creates a new Integration and ends old active Sessions, and the Slack signing secret remains server-only.

**Action taken**
- Approved exact head `43bf4bf` under the vacation rule. The PR was not merged.

PR #114, opened 2026-08-23
feat(client): Slack CLI 4.2.0 readiness floor and managed prompt rules

### LINKED ISSUE #124: Slack CLI fails in the Agent sandbox when it cannot write $HOME/.slack/logs, with an error that looks unrelated to Slack
### What happens

An Agent Turn that tries to use the official Slack CLI fails on its first, correct invocation because the CLI cannot write its debug log under `$HOME`, which the Agent sandbox denies. The error names a log file, so it reads as nothing to do with Slack, with Slack credentials, or with the sandbox.

Observed during independent live QA on a real Slack workspace. The Agent sourced the projected per-Turn credential file and ran the documented command:

```
source "$OPENTAG_PROVIDER_ENV_FILE" && slack api chat.postMessage channel=C… text=…
```

and got, with no API call attempted:

```
2026/08/22 18:59:34 open /Users/super/.slack/logs/slack-debug-20260823.log: operation not permitted
2026/08/22 18:59:34 open /Users/super/.slack/logs/slack-debug-20260823.log: operation not permitted
   … repeated ~8x, exit non-zero, no request sent
```

The Turn only succeeded because the model improvised a workaround on its own — it took four attempts:

1. `slack api --help` (worked; the binary is fine)
2. `slack api chat.postMessage …` → the permission spam above
3. `… --config-dir /private/tmp/opentag-slack-config` → `stat: no such file or directory`
4. `mkdir -p /private/tmp/opentag-slack-config` then `… --config-dir …` → `{"ok":true,"channel":"C…","ts":"…"}`

### Why it matters

- **The failure looks unrelated to Slack.** The message is about a log file under `$HOME/.slack/logs`, so diagnosis starts in the wrong place — sandbox policy, credentials, and the CLI binary all look fine. This is the expensive part.
- **It depends on the model being resourceful.** Inventing `--config-dir` and pre-creating the directory is not obvious. A provider that follows the documented command and stops would simply fail the Turn, and the Agent would appear silently broken in the channel.
- **Readiness does not catch it.** The IM CLI readiness probe runs `slack version` / `slack api --help` from the **daemon** process, which is not sandboxed, so it reports `ready` while the sandboxed Agent invocation still fails. `providerCliReadiness: ready` is therefore not evidence that the Agent can use the CLI — the probe and the real call run in different environments.

### Suggested direction

Project a writable Slack config directory per Turn rather than relying on the Agent to invent one:

1. Create a per-Turn writable directory and point the CLI at it through the credential environment file alongside `SLACK_BOT_TOKEN`, so the documented command works unmodified. Confirm against the installed CLI whether a config-dir environment variable is honoured; if not, the managed prompt rule must pass `--config-dir` explicitly.
2. Add the `--config-dir` rule to the managed Slack CLI guidance as the fallback.
3. Consider making the readiness probe representative of the sandbox the Agent actually runs in, or stop treating it as evidence that Agent-side invocation works.

### Notes

Found while running independent live QA of the Slack binding flow against a real workspace (related: #99, and the binding flow in #113). The outbound path itself is sound — once the config directory was writable, the Agent posted successfully to the real channel using only the projected `SLACK_BOT_TOKEN` with no `slack login`. This issue is purely about the sandbox/config-dir precondition.

liuchao-001:
## Summary

Part of the OpenTag Slack production-readiness campaign (workstream W3 — client side; independent of #113, whose Slack binding lifecycle is merged into `main` and brought in here by a merge commit).

- **Slack CLI readiness floor of 4.2.0.** `slack api <method>` exists since 4.1.0 and 4.2.0 removed the background update check that breaks non-interactive use. The probe runs `slack version --skip-update`, parses the real `Using slack vX.Y.Z` banner with an anchored regex, and only then runs `slack api --help --skip-update`. Only stable `X.Y.Z` releases are accepted; prerelease (`v4.2.0-rc.1`) and development (`v4.2.0-3-gabcdef`) builds are rejected as unrecognized.
- **The wire shape is unchanged.** An old, unparseable, or below-floor CLI is reported as plain v1 `unavailable`, exactly like any other probe failure. `packages/shared` and `packages/server` are untouched by this PR — the earlier `reason`/`detectedVersion` wire fields were reverted. Version diagnosis is a local concern and lives in a structured Client daemon log warning carrying `detectedVersion` and `minimumVersion` (`Slack CLI version is below the supported minimum`).
- **Managed Slack CLI prompt rules.** Fixed, byte-bounded rules added to the managed IM context for Slack Turns only (Feishu prompt unchanged): `thread_ts` mechanics (inbound `threadTs` else `messageTs`) with reply placement left to Session policy; the `chat.postMessage --json … --skip-update` template; always `--json` (Bearer header) and never `key=value` (which form-encodes the token into the request body); never `--token`/`--app`/`-w`/`--team` (they override the projected `SLACK_BOT_TOKEN`); `markdown_text` ≤ 12,000 vs `text` ≤ 4,000 and splitting; ≤ 1 msg/s/channel; `<@U…>` mentions; the history rate-limit caveat; never print the token; argv visibility.
- **A per-Turn writable Slack config directory.** `Fixes #124`. The Slack CLI writes config, credentials, and a debug log into a system config directory on *every* invocation, including calls that only hit `slack api`. A sandboxed Turn that cannot write `$HOME/.slack/logs` therefore dies before issuing any API call, with a filesystem error that never names Slack. The Turn now projects a private `0700` directory per Turn next to the credential environment file and exports its path twice:
  - `OPENTAG_SLACK_CONFIG_DIR` is the load-bearing mechanism — the managed prompt tells the Agent to pass it to the documented `--config-dir` global flag on every `slack` invocation.
  - `SLACK_CONFIG_DIR` is a deliberate, redundant second path to the same directory, because the whole defect is an Agent that omits the flag. It is **undocumented** (present in the 4.6.0 binary and empirically honoured, absent from `slack --help` and the official reference), so it must never become the only mechanism: a CLI at the supported 4.2.0 floor that ignores it still works through the flag. The rationale is recorded in a code comment so a future cleanup does not collapse the redundancy.

  The directory shares the credential file's full lifecycle: removed together when the Turn finishes, retried during Session or Client shutdown if removal fails, and reclaimed by the next Client startup after a crash through the strict managed-artifact name matcher.
- **Docs:** `docs/direct-provider-cli.md` and its zh-CN mirror gain the "Slack CLI requirements" section and the config-directory lifecycle.

Verified against the installed official Slack CLI v4.6.0: token resolution is `--token` → `--app` → `SLACK_BOT_TOKEN` → `SLACK_USER_TOKEN` → app prompt, and `SLACK_BOT_TOKEN=… slack api auth.test --skip-update` reaches Slack with no `slack login`.

## Out of scope

opentag#123 — client capability refresh stopping after the first interval, which leaves readiness permanently `checking` and `handoffReady` false — is **not** addressed here. This branch still uses the existing `Promise.all` capability refresh and adds neither per-provider timeouts nor monitor recovery. It only threads a logger into the Slack branch of the existing readiness probe.

## Validation

Run on the merged head (`4d8fbd4`) with a sanitized `PATH` (a temp bin holding only `node`, `pnpm`, and `git`, plus `/usr/bin:/bin`). The client suites are PATH-sensitive by pre-existing design, so this `PATH` deliberately has no real `slack`, `lark-cli`, `codex`, or `claude` binary on it. With the real Slack CLI 4.6.0 on `PATH`, two `client-runtime-composition` tests fail purely on timing — roughly 2.4s for `slack version` and 1.9s for `slack api --help` — which is exactly the background update check that the `--skip-update` flags in this PR remove.

| Command | Exit code | Result |
| --- | --- | --- |
| `pnpm check` | 0 | pass (9 pre-existing `noUndeclaredEnvVars` warnings, non-blocking) |
| `pnpm typecheck` | 0 | pass |
| `pnpm build` | 0 | pass |
| `pnpm --filter @opentag/client test` | 0 | 37 files, 432/432 tests |
| `pnpm --filter @opentag/client test:agent-runtime:coverage` | 0 | 100/100/100/100 |
| `pnpm test` | 1 | `@opentag/shared` 72/72; Turbo halts at `@opentag/web` (1 pre-existing failure), so the remaining suites were run individually |
| `pnpm --filter @opentag/server test` | 1 | 232/233 — the known `stale_connection` observability flake |
| `pnpm --filter open-tag test` | 1 | 117/119 — the two known CLI stale-shim tests |

The three `pnpm test` failures are the known pre-existing set and are untouched by this branch, which changes no web, server, shared, or CLI file:

- `apps/web` `app.test.tsx` — locale-dependent invitation expiry date assertion.
- `packages/server` `observability.test.ts` — `stale_connection` terminal-path flake.
- `apps/cli` — two stale-`PATH`-shim resolution tests.

QA tier requested: `focused-local` — no product surface beyond the readiness observation, the prompt text, and the per-Turn config directory projection. Independent exact-head QA by code-reviewer to follow; keeping this a draft until then.


liuchao-001:
Addressed the independent review at `9b4762c8b148f3d7383ec1a262cc7567c3106dfe`: the wire protocol and Computer DTO are no longer touched (shared/server byte-identical to main) — the 4.2.0 floor is enforced client-side as plain `unavailable` with a structured warning (`detectedVersion`/`minimumVersion`) in the Client daemon log, documented as the diagnosis path; prereleases/dev builds are rejected by the parser; all placement-preference language is removed from the managed Slack block (mechanics only, test rejects prefer/should/default/always); rate-limit wording narrowed; JSON-serializer rule added.

yuezengwu:
Next steps from the current draft head 8bad0969da6ffe91fc2d544cbf8b1c9586d1baf8:

1. Update the PR body to match the current implementation. The shared/server reason and detectedVersion wire fields were reverted; the current behavior remains the v1 unavailable observation, with version diagnosis in the Client log.
2. Link the sandbox defect explicitly with Fixes #124, or use an explicit non-closing link if preferred. The current head implements the per-Turn writable Slack config directory, prompt flag, cleanup, and crash reclamation required by that issue.
3. Keep #123 explicitly out of scope. This branch still uses the existing Promise.all capability refresh and does not provide the per-provider timeout and monitor recovery required for the permanent-checking defect.
4. Resolve the current conflict with latest main, preserving both the config-directory lifecycle and Slack CLI version/prompt rules, then rerun the affected Client suites plus pnpm check, pnpm build, pnpm typecheck, pnpm test, and Client runtime coverage on the fresh head.
5. Update the Validation section with those fresh results, mark the PR ready, and request exact-head code and QA review.

The current head GitHub checks are green. Conflict resolution and the stale PR description are the remaining steps before review.

liuchao-001:
Closing this rather than rebasing it: the branch is 90 commits behind `main`, from before the Account migration, Better Auth and the router change, and it conflicts across 8 file groups.

Nothing in it was superseded, so its contents were moved rather than dropped:

- The per-Turn writable Slack config directory — the live defect — stays as **#124**, now carrying the implementation notes from this PR, including why `SLACK_CONFIG_DIR` can only ever be the redundant path to the documented `--config-dir` flag. Assigned to @yuezengwu.
- The 4.2.0 readiness floor and the managed Slack prompt rules are now **#300**.

The branch `feat/slack-cli-readiness-prompt` is left in place if any of the diff is wanted.

PR #115, opened 2026-08-23
fix(slack): setup recovery, live ingress during reauthorize, and review findings (stacked on #113)

liuchao-001:
## Summary

Stacked on #113 (`feat/issue-99-slack-app-binding`, base `d6c219f`) — fixes for the CHANGES_REQUESTED state and the findings from two independent reviews plus live QA on a real workspace. Every server-side defect was reproduced with a failing test first; those tests ship here as regressions.

**P0**
- Same-secret `reauthorize` no longer takes the live binding off the air: `tryActivateFromEvent` returns a typed outcome (`activated` / `awaiting_challenge` / `unmatched`) instead of throwing `SLACK_SIGNING_CHALLENGE_REQUIRED`; the events route falls through to the active binding, and with no active binding a pending-attempt event is acknowledged (`200 {pending:"url_verification"}`) without ingesting — so Slack never counts it as a failure. Integration test uses a signing secret identical to the live binding's.
- `reauthorize` guidance: attempts now carry `intent`, `manifest` (JSON) and `currentAppId`; the web guide sends reauthorize to the existing App's manifest page (replace manifest → save → Reinstall → copy the new Bot Token) with copyable JSON; `create`/`replace` keep the `new_app=1` link plus the JSON fallback.

**P1**
- Signing Secret recovery: credentials can be resubmitted while `validating` (token re-inspected, secret replaced atomically, `challengeVerified` reset) instead of being silently discarded with a 200; signature failures are recorded as `lastVerificationErrorCode`/`lastVerificationAt` inside the encrypted context; the UI distinguishes "Bot Token validated" from "Signing Secret not yet verified", offers **Edit credentials** and **Cancel setup**, and `cancel` is wired into the web API.
- Resume after refresh: a `provisioning` Slack binding hydrates the in-flight attempt and offers **Resume Slack setup**/**Cancel**; Replace is hidden while provisioning.
- Requested intent is no longer silently discarded: same intent reuses, different intent → `409 SLACK_SETUP_INTENT_CONFLICT`; admission is re-checked under the Agent lock inside the transaction (concurrent reauthorize overwrite and concurrent different intents reproduced and fixed).
- `pendingReceiveMode` no longer masks a real `lastErrorCode`; the pending target is its own projection field (summary/detail/diagnostics, web + CLI).
- A no-op `receiveMode` PATCH no longer cancels an in-flight setup (compares against the effective target); a real cancel records `SLACK_SETUP_CANCELED`.
- Poll loop: bounded exponential backoff, stop on 4xx/terminal with an actionable message, stale errors cleared on success; new `slack-setup.test.tsx` suite.
- Validation panel now renders for real Slack Apps: `auth.test` omits `app_id` for bot tokens (confirmed live), so `identity.appId` is nullable and the panel renders from team/bot user.

**P2**
- TTL TOCTOU on credential submit and challenge verify (expiry predicate in the CAS, affected-row check; stale → `SLACK_SETUP_EXPIRED`).
- Slack App/Team unique violation maps to `SLACK_APP_TEAM_ALREADY_BOUND` (409) instead of a raw 500.
- `inspectInstallation` bounded by a 10 s timeout → `SLACK_UPSTREAM_UNAVAILABLE`.
- Web error map completed; docs (en + zh-CN) describe reauthorize via manifest update, the recovery controls, and when Slack sends `url_verification` (only on URL set/change or admin Retry — `apps.manifest.update` does not trigger it).

Not changed: the concurrent first-`create` 23505 case did not reproduce (the Agent lock serializes it); the test stays as a guard. No new migration — new diagnostic fields live in the encrypted setup context with schema defaults.

## Validation

- `pnpm check`, `pnpm typecheck`, `pnpm build`: pass
- server `slack-events` + `slack-adapter` + `im-binding-api`: 25 passed; `integration/im-binding.test.ts -t Slack`: 20 passed
- web `slack-setup.test.tsx` + `app.test.tsx` (`TZ=UTC`): 112 passed; shared: 69 passed
- Full `pnpm test` has three pre-existing environment-dependent failures that reproduce identically at `d6c219f` in a pristine worktree (web invitation-link timezone case, client composition 5 s timeouts with real CLIs on PATH, observability trace flake); none of those files are touched here.

Independent QA: baseline report at `d6c219f` (FAIL: P0-1 live over public HTTPS with a control; scenarios 12/13 FAIL; scenario 10 PASS) is on #113; exact-head QA on this branch follows.

Pairs with the tree PR first-tree-context#973 (one-line amendment to follow: editable Signing Secret while validating, resume after refresh).


liuchao-001:
Closing this as superseded — no longer needed.

@yuezengwu chose to resolve the same findings directly on `feat/issue-99-slack-app-binding` rather than merge this branch, which was their call to make and the right outcome: #113 now carries the fixes with its own failing-test-first regressions, and every file this PR touched is also touched by #113's head `e7e1eb17`.

I re-checked the four items this branch carried that were most at risk of being lost in a reimplementation, and all are genuinely present at `e7e1eb17`:

- the P0 ingress regression during same-App reauthorization — `tryActivateFromEvent` no longer throws challenge-required; it returns an outcome and `slack-events.ts` falls through to the active binding, acknowledging a pre-activation signed event with 200 instead of failing it;
- no-op `receiveMode` PATCH no longer cancels an in-flight setup — it compares against the effective target `pendingReceiveMode ?? receiveMode`;
- `pendingReceiveMode` no longer masks the stored `lastErrorCode`;
- bounded `auth.test` inspection.

Nothing here is being abandoned silently. The branch stays on the remote for history; it is stacked on this chain, so it is not a rebase candidate — the two independent hardening commits on top of it (Slack `file_share`/`thread_broadcast` ingestion, event-subscription budget + typed ingress tracing) will be cherry-picked onto the merged `#113` head and land as their own PR.

PR #119, opened 2026-08-23
fix(web): gate onboarding on Team completion

Gandy2025:
## Summary

- persist a one-way Team setup completion timestamp after an admin-selected Agent has a fully ready runtime and IM handoff
- route incomplete admins to onboarding, keep incomplete members in Agents with an admin-dependency notice, and keep every completed Team out of onboarding even during later runtime outages
- anchor the onboarding Agent in the URL, remove the onboarding localStorage cursor, and update the real browser E2E flow for completion and outage behavior

## Testing

- `pnpm build`
- `pnpm typecheck`
- targeted Biome checks across every changed source and generated migration file
- Web suite: 278 tests passed
- relevant Server API/runtime/startup suite: 34 tests passed
- Team setup persistence integration test passed, including unready rejection, admin authorization, idempotence, and post-completion outage behavior
- migration integration suite: 22 tests passed
- onboarding browser E2E: 16/16 steps passed against real Server, Web, PostgreSQL, CLI daemon, and Chromium

Local full-repository checks also exposed existing environment-sensitive failures outside this change: nested ignored `.claude/worktrees` Biome roots, macOS `/var` versus `/private/var` daemon path assertions, and a Runtime tracing timing assertion. The relevant suites above pass; CI on a clean checkout is the merge authority.

## Breaking changes

Existing Team rows receive `setup_completed_at = NULL` and are treated as incomplete. Per the agreed scope, this change intentionally includes no historical backfill.

## Checklist


baixiaohang (changes requested):
- Rationale: The new Team-completion transition is not atomic under concurrent first-time requests, so it can overwrite the persisted completion timestamp instead of preserving a single one-way result.

**Risk level**: A
- Path baseline: the mixed diff includes `apps/cli/**`, so the highest applicable baseline is A.
- Semantic lift: the PR also changes the Team database schema and the server-owned onboarding state transition; no further lift is possible above A.

**PR summary**
- Problem: Live runtime and IM readiness currently make completed Teams fall back into onboarding after an outage, while members and admins do not have a durable Team-level setup gate.
- Approach: Persist `teams.setup_completed_at` after an admin-selected active Agent has a ready runtime/IM handoff; route incomplete admins into onboarding, keep incomplete members in Agents with admin guidance, anchor the selected Agent in the URL, and stop using a local onboarding cursor.
- Impacted modules: Web routing and onboarding, Team/auth API contracts, Team schema and migration, Team setup and IM-readiness services, CLI fixtures, and browser/server integration coverage.

**Review findings**
❌ 1. Serialize the Team completion fact before deciding it is still unset. Two concurrent first-time calls can both read `setupCompletedAt = NULL` at lines 73-78. They later serialize only on the Agent row, so the second call continues from its stale Team read and overwrites the first timestamp at lines 94-99. That breaks the advertised one-way/idempotent boundary and makes the durable completion time depend on the last racing request. Lock the Team row before the null check, or use a conditional `UPDATE ... WHERE setup_completed_at IS NULL` and return/re-read the winner; please add a concurrent completion regression test. [R4 / `packages/server/src/services/teams/team-setup-service.ts:73`]

**Action taken**
- Submitted request changes on head `3eb8bc825548f0234a880668d7a5f5fed31ebad7`.

PR #122, opened 2026-08-23
fix(server): repair completed staging Team setup

Gandy2025:
## Summary
- add an idempotent data-repair migration for the verified historical Staging Team
- locate the Team through the exact pre-existing Agent UUID
- leave every other Team and environment unchanged
- update migration-count assertions

## Validation
- `vitest ... -t "upgrades deployed"` passed
- the full migration suite previously ran successfully except for the expected migration-count assertions, now updated

## Scope
The repair only updates a Team whose `setup_completed_at` is null and which owns Agent `6eb89d85-0f12-4962-8465-518071f1d3e9` (`baixiaohang的助理`). It is a no-op where that UUID does not exist.

### DIFF
--- a/packages/server/drizzle/0011_staging_team_setup_repair.sql
+++ b/packages/server/drizzle/0011_staging_team_setup_repair.sql
@@ -1,0 +1,14 @@
+-- Repair the Team that completed onboarding before setup_completed_at existed.
+-- The Agent UUID scopes this to the verified Staging Team and is a no-op elsewhere.
+UPDATE "teams"
+SET
+	"setup_completed_at" = CURRENT_TIMESTAMP,
+	"updated_at" = CURRENT_TIMESTAMP
+WHERE
+	"setup_completed_at" IS NULL
+	AND "id" = (
+		SELECT "team_id"
+		FROM "agents"
+		WHERE "id" = '6eb89d85-0f12-4962-8465-518071f1d3e9'
+		LIMIT 1
+	);
--- a/packages/server/drizzle/meta/_journal.json
+++ b/packages/server/drizzle/meta/_journal.json
@@ -78,6 +78,13 @@
       "when": 1787472158796,
       "tag": "0010_optimal_jazinda",
       "breakpoints": true
+    },
+    {
+      "idx": 11,
+      "version": "7",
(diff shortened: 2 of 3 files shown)


baixiaohang (changes requested):
- Rationale: The one-time staging repair has no regression test proving that it repairs the intended Team while leaving control Teams untouched; the current test changes only the migration count.

**Risk level**: B-high
- Path baseline: `packages/server/drizzle/**` changes a persisted Team setup fact through a database migration -> B-high.
- Semantic lift: none beyond the database-state baseline.

**PR summary**
- Problem: The already-completed Staging Team would be treated as incomplete after the new durable Team setup gate introduced `setup_completed_at` with no historical backfill.
- Approach: Backfill `setup_completed_at` and `updated_at` only for the Team owning one verified Agent UUID, and only while the completion field is null; databases without that Agent remain unchanged.
- Impacted modules: Server database migrations and migration integration coverage.

**Review findings**
❌ 1. Add a behavior-level migration fixture for this repair. The upgraded-history test currently seeds only random Agent IDs and changes the expected migration count from 11 to 12, so it still passes if the UUID or predicate in `0011_staging_team_setup_repair.sql` is wrong and the update is a no-op. Because Drizzle will nevertheless record this one-shot migration as applied, that failure would leave Staging unrepaired without retrying. Seed a pre-0011 database with the exact target Agent UUID plus a control Team, run the remaining migrations, and assert that only the target Team receives `setup_completed_at`. [R5 / `packages/server/src/__tests__/integration/auth-migrations.test.ts:256`]
✅ 2. The SQL itself is narrowly scoped by the Agent primary key and the null completion guard, so it does not reopen an already-completed Team. [`packages/server/drizzle/0011_staging_team_setup_repair.sql:3`]

**Action taken**
- Submitted request changes on head `ab6f15b34dc44716adaf554721b95a90f6bb8867`.

Gandy2025:
Addressed in `c907461`: the pre-0011 migration fixture now seeds the exact target Agent UUID and a control Team, then asserts that only the target Team receives `setup_completed_at`. The focused migration test and the full CI suite pass on the updated head.

PR #134, opened 2026-08-24
feat(web): refine Agent overview cards

Gandy2025:
## Summary

- Rework the Agents overview around identity, actionable status/current work, and the two primary usage metrics: Tasks and Tokens.
- Move the fixed 30-day context into the page description and remove the detached usage caption and secondary per-task/failed statistics.
- Project a bounded current-work summary from the accepted inbound task so Working Agents show what they are doing and when they started.
- Keep recovery actions adjacent to the relevant failure and simplify detail navigation to a compact chevron.

## Testing

Passed:
- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- Web suite: 316 tests passed during `pnpm test`
- Shared suite: 71 tests passed during `pnpm test`
- Production browser check at 1440px with Working, Needs attention, and Paused Agents

Attempted full required suites; failures reproduce outside this change:
- `pnpm test`: existing RuntimeSession observability timing assertion expected `stale_connection` but observed `handled`.
- `pnpm --filter @opentag/client test:agent-runtime:coverage`: existing readiness timing assertion observed `checking` before `ready` (the same test passed in the full coverage run).
- `pnpm test:coverage`: 1166/1168 passed; two existing macOS path canonicalization assertions expect `/var/...` while runtime returns `/private/var/...`.

## Breaking changes

The Working activity response adds a bounded `summary` field. Server and consumers using the strict shared response schema should be deployed from the same revision.

## Checklist

- [x] Tests cover the changed behavior.
- [ ] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. (`pnpm test` is blocked by the unrelated baseline failure documented above.)
- [x] Canonical English docs and Chinese mirrors are synchronized when applicable (not applicable).

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -42,6 +42,7 @@ function installApi(
     agentRead?: () => Promise<void> | void;
     agentReadStatus?: () => number | undefined;
     agentListStatus?: () => number | undefined;
+    agentActivity?: { state: "idle" } | { state: "working"; startedAt: string; summary: string };
     emptyAgents?: boolean;
     agentCreate?: (input: Record<string, unknown>) => Promise<void> | void;
     alreadyJoinedInvitation?: boolean;
@@ -281,12 +282,14 @@ function installApi(
               path.includes(invitedTeamId)
                 ? {
                     ...agentListItem,
+                    activity: options.agentActivity ?? agentListItem.activity,
                     teamId: invitedTeamId,
                     status: lifecycleStatus,
                     runtimeProvider: options.runtimeProvider ?? agentListItem.runtimeProvider,
                   }
                 : {
                     ...agentListItem,
+                    activity: options.agentActivity ?? agentListItem.activity,
                     status: lifecycleStatus,
                     runtimeProvider: options.runtimeProvider ?? agentListItem.runtimeProvider,
                   },
@@ -606,7 +609,8 @@ describe("OpenTag Web App Shell", () => {
     expect(createAgent.closest(".page-header")).toBeTruthy();
     const agentCard = agentLink.closest(".agent-card");
     expect(agentCard).toBeTruthy();
-    expect(screen.getByText("Usage · Last 30 days")).toBeTruthy();
+    expect(screen.getByText("Monitor availability and 30-day usage across your AI teammates.")).toBeTruthy();
+    expect(screen.queryByText("Usage · Last 30 days")).toBeNull();
     expect(within(agentCard as HTMLElement).getByText("Tasks")).toBeTruthy();
     expect(within(agentCard as HTMLElement).getByText("Tokens")).toBeTruthy();
     expect(within(agentCard as HTMLElement).getByText("428K")).toBeTruthy();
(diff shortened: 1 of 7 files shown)


baixiaohang (changes requested):
- Rationale: The overview exposes conversation-scoped message text at Team scope, can reject valid non-BMP summaries at the response boundary, and overflows across common intermediate viewport widths.

**Risk level**: B-high
- Path baseline: includes `packages/server/src/services/agents/**` -> B-high.
- Semantic lift: changes the Team-wide projection of active Agent work and its shared response contract; level remains B-high.

**PR summary**
- Problem: Workspace members need a clearer Agent overview that shows identity, actionable availability, current work, and the primary 30-day usage metrics without secondary visual noise.
- Approach: add a bounded current-work summary to the working activity projection, reorganize each card around status/current work plus Tasks and Tokens, and simplify recovery/detail actions.
- Impacted modules: Agent list service/API contract, shared Agent schema, Web Agent overview and responsive styling, and related tests.

**Review findings**
❌ 1. `summary` copies the accepted inbound message's `fallbackText` into the Team-wide Agent list. `listForTeam` authorizes only active Workspace membership, while one Agent can participate in multiple IM channels/threads with conversation-scoped Sessions and permissions; there is no check that the caller belongs to the source channel/thread. A task from a private Slack channel or Feishu thread can therefore be shown on `/agents` to Workspace members who could not see that conversation. Please keep this projection content-free, or introduce an explicitly visibility-safe summary with an authorization boundary that is enforceable at read time. [R5 / `packages/server/src/services/agents/agent-service.ts:360`, `packages/server/src/services/agents/agent-service.ts:427`]

❌ 2. The writer and schema use different length units. `summarizeCurrentWork` counts Unicode code points with `Array.from`, but Zod's `string().max(160)` checks JavaScript UTF-16 length. For example, 81 emoji are accepted unchanged by the writer but have length 162, so `ListAgentsResponseSchema.parse` throws and the entire Agent list endpoint fails while that delivery is working. Please make the bound and schema use the same metric and cover astral characters at the boundary. [R4 / `packages/server/src/services/agents/agent-service.ts:67`, `packages/shared/src/agent.ts:112`]

❌ 3. The new desktop grid requires at least 816px (684px of tracks + 84px of gaps + 48px of padding), but it does not collapse until `max-width: 720px`. At a 1024px viewport, the persistent 220px sidebar plus content/page gutters leave about 700px for the card; at 800px the mobile shell still leaves about 760px. Both widths therefore overflow horizontally. Please add an intermediate layout/breakpoint or make the tracks flexible, and verify at a width inside each affected range. [R5 / `apps/web/src/styles.css:934`, `apps/web/src/styles.css:3976`, `apps/web/src/styles.css:4153`]

**Action taken**
- Submitted request changes on exact head `5e2033255295bc1cf3a37bc18bec6296d0c92602`.

yuezengwu (changes requested):

I independently confirmed the three existing blockers and found one additional server-side amplification issue on exact head `5e2033255295bc1cf3a37bc18bec6296d0c92602`.

**Risk level**: B-high
- This changes Team-visible Agent activity, reads message content in a frequently refreshed endpoint, and changes the shared response contract and responsive overview layout.

**Blocking findings**

1. **Conversation content is disclosed at Team scope.** `listForTeam` authorizes any active Team member, but the new projection copies `imMessages.content.fallbackText` from one conversation-scoped delivery into that Team-wide list. An Agent may participate in private channels and threads whose Sessions and permissions are isolated, so a Workspace member who cannot see the originating Slack/Feishu conversation can still read its task text on `/agents`. Keep the Team overview content-free, or define and enforce a visibility-safe summary boundary at read time. (`packages/server/src/services/agents/agent-service.ts:360`, `:427`)

2. **The summary producer and schema disagree on string length.** `summarizeCurrentWork` counts Unicode code points via `Array.from`, while Zod `string().max(160)` counts UTF-16 code units. An accepted task containing 81 emoji is returned unchanged as 81 code points but has JS length 162; `ListAgentsResponseSchema.parse` then throws and the whole Agent list request fails. Use one length metric end-to-end and add an astral-character boundary test. (`packages/server/src/services/agents/agent-service.ts:67`, `packages/shared/src/agent.ts:112`)

3. **The card grid overflows common intermediate widths.** Its fixed minima total 684px, plus 84px of column gaps and 48px of padding, so the card requires at least 816px. It collapses only at 720px. At a 1024px viewport, the 220px sidebar plus content/page gutters leave about 700px for the card; at 800px the mobile shell leaves 760px. Add an intermediate layout/breakpoint or flexible tracks and test inside both affected ranges. (`apps/web/src/styles.css:934`, `:3976`, `:4153`)

4. **The 30-day usage query now materializes full message JSON for every delivery on every list refresh.** Selecting `imMessages.content` applies to all rows admitted by the 30-day usage branch, not just the latest unreported work item. Canonical content is allowed up to 256 KiB per row, while the Web reloads this endpoint every 30 seconds; active Teams can therefore make each overview refresh transfer and parse a large multiple of their entire 30-day message corpus just to render one 160-character summary per working Agent. Fetch only a database-bounded summary for candidate current rows, preferably in a separate/latest-current-work query, and keep historical usage rows content-free. (`packages/server/src/services/agents/agent-service.ts:381-410`, `packages/shared/src/im-message.ts:73-85`, `apps/web/src/router.tsx:1052-1056`)

**Verification**
- `pnpm --filter @opentag/shared test`: 71 passed.
- `pnpm --filter @opentag/web test`: 316 passed.
- Targeted Agent integration test: 2 provider variants passed.
- Direct boundary reproduction: 81 emoji -> 81 code points / 162 UTF-16 units -> response schema rejection.
- `git diff --check`: passed.

PR #136, opened 2026-08-24
feat(web): group Task agent activity

Gandy2025:
## Summary
- render Task execution as an ordered Codex-style transcript of progress and tool activity
- group adjacent tool calls into lightweight expandable stages
- add a second disclosure level for source and result details without restoring status/source columns
- keep live and attention state on the relevant group and preserve the final answer as a separate reading block
- refine mock event copy and responsive layout tests

## Validation
- `pnpm check` (passes with existing undeclared E2E env warnings)
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- desktop and 390px Playwright visual checks

## Local coverage note
- `pnpm test:coverage` was run but fails in unrelated existing CLI service-path tests and a server observability timing test under the local coverage runner; the same suites pass in `pnpm test`. This PR only changes Task mock UI files.


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change preserves the ordered execution transcript, scopes live/attention state to the relevant tool group, and keeps the final answer as a separate reading block without introducing a blocking correctness issue.

**Risk level**: C
- Path baseline: only `apps/web/**`, which falls under the remaining-path C baseline.
- Semantic lift: none; this is mock Task UI presentation and does not touch subscriptions, WebSocket frame parsing, or the data layer.

**PR summary**
- Problem: Task readers currently have to scan a flat, column-like list of low-level tool calls, which makes progress and evidence harder to follow.
- Approach: Present reasoning and tool activity as an ordered transcript, collapse adjacent tool calls into expandable stages, expose source/result details one level deeper, retain active/attention state on the affected stage, and leave the final answer outside the process disclosure.
- Impacted modules: `apps/web/src/features/tasks-page.tsx`, Task page styles and tests, and Task mock data.

**Review findings**
⚠️ 1. Non-blocking: `TaskToolCall` still carries the now-unused `tool` and `target` fields alongside the new `detail` string, leaving two representations of source/target facts that can drift. Removing the legacy fields in this mock-only model would complete the replacement. [R1/R2 / `apps/web/src/mock/task-data.ts:11`]
✅ 2. The reducer preserves event order and only groups adjacent calls; tests cover group disclosure, nested call disclosure, scoped active/attention state, responsive detail layout, and the final answer's separate reading position.

**Action taken**
- Approved.

PR #140, opened 2026-08-24
fix(client): recover capability refresh from hung probes

### LINKED ISSUE #123: Client capability refresh stops after the first interval: readiness decays to `checking` and handoffReady stays false
### What happens

About a minute after the Client daemon starts, every readiness observation decays to `checking` and never recovers, so the Server projects `ready: false` / `handoffReady: false` permanently until the daemon is restarted — and then only for another minute. This was measured on a live Slack binding during independent QA, sampling the Agent's handoff readiness every 20s after a clean daemon restart:

```
ready (t≈25s) → checking → checking → checking → checking → checking → checking
```

Six consecutive samples over 120s, all `checking`. A healthy 30s refresh should have reported `ready` in most of them.

### Why it matters

`handoffReady` gates Turn dispatch, so an Agent that was ready a minute ago stops being dispatchable with no failure and no error surface — the operator sees only "checking". It affects Feishu and Slack equally and is independent of any IM provider work; it reproduces on `main`.

### Mechanism (hypothesis, not yet proven)

`RUNTIME_CLIENT_CAPABILITY_TTL_MS = 60_000` (`packages/shared/src/runtime-protocol.ts:50`) and each observation is stamped `expiresAt = now + TTL`, with the refresh interval set to TTL/2 — the design is right, but the refresh appears to stop running.

The suspicious shape is the in-flight latch in `#startCapabilityMonitor` (`packages/client/src/runtime/client-runtime-composition.ts:157-164`):

```ts
if (this.#capabilityRefreshInFlight || this.#stopped) return;
const refresh = this.#refreshCapability();
```

The latch is cleared in `.finally()`, so it can only stick if the refresh promise never settles. `refreshCapability` (`:250-256`) is a single `Promise.all` over every provider probe plus both IM CLI probes. `refreshImCliReadiness` bounds its own `execFileAsync` calls with `timeout: 10_000`, but `providers.refresh(providerId, refreshSignal)` has no visible deadline, so one hung provider probe would leave the whole `Promise.all` pending forever, hold the latch, and — because each refresh first publishes `status: "checking"` (`:240`) — pin every observation at `checking` exactly as observed.

### Suggested direction

1. Bound each probe inside `refreshCapability` with its own deadline (and/or `Promise.allSettled` so one hung probe cannot block the others).
2. Make the monitor resilient rather than latched: if a refresh exceeds the interval, log it and allow the next tick, instead of returning early forever.
3. Add a regression test that a probe which never resolves does not permanently freeze readiness.

### Notes

Found while running independent live QA of the Slack binding flow on a real workspace; the symptom is established by measurement, the mechanism above is a hypothesis and should be confirmed before fixing.

yuezengwu:
## Summary

- bound every exact-provider readiness probe with a 10-second deadline so one non-settling probe cannot hold the capability monitor forever
- isolate provider and IM readiness work, keep the existing no-overlap monitor latch, and let later intervals recover after a bounded failure
- make one shared provider-refresh owner the only readiness publisher, with independent waiter cancellation and immediate zero-waiter owner eviction
- preserve registry-identity fencing, lifecycle/caller abort semantics, shutdown publication suppression, and cancellation of IM CLI probes

## Root cause

The initial capability refresh completed, but a later provider probe could return a promise that never settled and did not honor abort. `refreshCapability()` awaited all provider and IM work through one `Promise.all`, so that single promise kept `#capabilityRefreshInFlight` occupied forever. An unavailable provider had already published `checking`; other provider and IM probes could finish that one refresh, but could never run again. Once their observations crossed the capability TTL, the Server projected the missing readiness as `checking`, leaving handoff unavailable until the daemon restarted.

This reproduces deterministically with an initially unavailable provider whose periodic probe never resolves, controlled time advancement beyond the TTL, and otherwise healthy provider/Feishu/Slack probes.

## Review follow-up

The first implementation advanced a publication generation per caller even though the provider registry coalesces concurrent callers onto one underlying probe. A cancelled newer waiter could therefore suppress the older active waiter's successful publication, leaving the registry `ready` while the connection observation remained `checking`.

Commit `e46f625` replaced those caller generations with one composition-level owner for each shared provider refresh. Joiners wait on that owner with independent caller cancellation, while only the owner owns the deadline and publishes the result.

The second review identified a drain window in that owner lifecycle: the last waiter aborted an unsettled owner but left it discoverable until the asynchronous operation reached `finally`. Commit `ce14b71` now removes the exact zero-waiter owner from the map before aborting it, refuses aborted or settled owners during lookup, and fences both publication and cleanup by current owner identity. A later caller can therefore start a fresh owner while the cancelled operation drains, without inheriting its cancellation or being removed by its late `finally`.

## Boundaries and safety

- the provider deadline is per shared provider owner, not a global replacement timeout; a joiner inherits the owner's remaining deadline rather than extending it
- zero-waiter eviction checks exact owner identity and occurs before abort, closing synchronous and asynchronous reuse windows
- both success/failure publication and `finally` cleanup are owner-identity fenced; a draining owner cannot publish over or delete its successor
- the monitor still permits only one top-level capability refresh at a time
- provider registry record identity fences old timed-out work and late underlying completions
- a timed-out provider is invalidated with the existing `temporarily_unavailable` readiness semantics
- lifecycle and caller abort remain distinct from deadline failure; shutdown and all-waiter cancellation do not publish late state
- exact-provider dispatch is unchanged; there is no fallback or provider substitution
- no schema, database migration, or persistence semantics change

## Tests

- never-resolving periodic provider probe with explicit time advancement past the readiness TTL
- healthy second provider plus Feishu and Slack probes continue and recover on later intervals
- A starts a shared probe, B coalesces and cancels, then A succeeds: provider readiness and heartbeat/handoff observation publish `ready` immediately, and the on-demand fast path does not leave `checking`
- A is the sole waiter and cancels while its composition owner is deliberately held draining; B starts and completes a new live refresh before A returns, and A's late success cannot overwrite B
- A's old `finally` runs while B's successor owner remains pending; C still joins B instead of starting a third owner, proving old cleanup cannot delete the successor
- the previous owner can fully drain before B starts, and B still receives a fresh result
- opposite completion/cancellation order does not unpublish a completed `ready` result
- a joined waiter cannot extend the owner's deadline, and the old late handler cannot overwrite a newer successful probe
- ensure-time cancellation/deadline behavior, Client Runtime shutdown, IM abort, and post-stop publication suppression
- provider registry invalidation remains idempotent and stale draining probes cannot restore readiness

## Validation

- `pnpm check` — pass; 8 existing undeclared-environment warnings remain in untouched `scripts/e2e/onboarding-e2e.mjs`
- `pnpm build` — pass, 5/5 tasks
- `pnpm typecheck` — pass, 7/7 tasks
- focused Client composition + provider-registry regression — pass, 2 files / 35 tests
- `pnpm --filter @opentag/client test` — pass, 37 files / 444 tests
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — pass on final rerun, 20 files / 282 tests, 100% statements/branches/functions/lines; the immediately preceding run had one unrelated Pi probe fixture marker `ENOENT`
- `pnpm test` — Client, Server, Shared, and Web pass; local macOS run stops on 2 CLI path assertions that expect `/tmp/...` while canonical resolution returns `/private/tmp/...`
- exact-main baseline `39b5184011dfaf7fff50cdd7304423f7039d516d` — the same isolated CLI tests fail 2/53 with the same canonical-path diffs
- `pnpm test:coverage` — 1188/1190 pass; only those same two CLI path assertions fail, and changed Client Runtime composition coverage is 100%
- exact-main Server observability timing baseline — three repeated isolated runs produced one pass and two `handled` versus `stale_connection` failures; the final PR coverage run did not hit this flake
- GitHub CI for `ce14b71` — pass, 6/6 checks

## Remaining risk

If a provider implementation ignores its supplied `AbortSignal`, its underlying promise may continue consuming resources after the registry and composition owner detach it. Its result is fenced from shared readiness state, and the capability monitor can safely retry; production provider adapters should still terminate their owned process on abort. A waiter that joins a nearly expired shared owner receives only the remaining deadline; after that owner settles, a retry starts a fresh owner. Draft PR #114 changes the same composition area and will require an intentional conflict resolution whichever PR lands second.

Fixes #123

### DIFF
--- a/packages/client/src/__tests__/agent-runtime-provider-registry.test.ts
+++ b/packages/client/src/__tests__/agent-runtime-provider-registry.test.ts
@@ -158,9 +158,14 @@ describe("AgentRuntimeProviderRegistry", () => {
     expect(probe).toHaveBeenCalledTimes(2);
     expect(providers.isReady("codex")).toBe(true);
 
+    providers.invalidate("codex");
+    expect(providers.isReady("codex")).toBe(false);
+    providers.invalidate("codex");
+    expect(providers.isReady("codex")).toBe(false);
+
     releaseFirst();
     await vi.waitFor(() => expect(firstSettled).toBe(true));
-    expect(providers.isReady("codex")).toBe(true);
+    expect(providers.isReady("codex")).toBe(false);
   });
 
   it("fails closed for missing registrations, artifact verification errors, and caller abort", async () => {
--- a/packages/client/src/__tests__/client-runtime-composition.test.ts
+++ b/packages/client/src/__tests__/client-runtime-composition.test.ts
@@ -17,6 +17,7 @@ import type { AgentRuntime, AgentRuntimeFactory } from "../agent-runtime/types.j
 import { createLogger } from "../observability/logger.js";
 import { claudeCodeRuntimePolicy, validateClaudeCodeRuntimePolicy } from "../providers/claude-code/runtime-policy.js";
 import { CODEX_AGENT_RUNTIME_APP_SERVER_ARGS } from "../providers/codex/agent-runtime.js";
+import { AgentRuntimeProviderRegistry } from "../runtime/agent-runtime-provider-registry.js";
 import {
   ComposedClientRuntime,
   codexProviderReadiness,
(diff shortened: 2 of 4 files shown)


baixiaohang (changes requested):
- Rationale: A cancelled overlapping waiter can suppress the only successful readiness publication and leave handoff readiness stale until a later monitor interval.

**Risk level**: A
- Semantic lift: none; A is already the highest path tier

**PR summary**
- Problem: One non-settling Provider readiness probe can freeze the Client capability monitor, let otherwise healthy Provider and IM observations expire, and keep IM handoff unavailable until the daemon restarts.
- Approach: Add a per-Provider deadline, isolate refresh outcomes with `Promise.allSettled`, fence stale completions with generations and registry identity, and propagate Client lifecycle cancellation into IM CLI probes.
- Impacted modules: Client Runtime capability composition, Agent Runtime Provider registry, and their regression tests

**Review findings**
❌ 1. A cancelled newer waiter can discard an older successful shared probe publication. `beginProviderRefresh()` advances the generation for every call, although the registry coalesces concurrent calls onto one underlying probe. If refresh A starts while the Provider is unavailable, refresh B joins it, and B's caller signal aborts, B becomes the latest generation and exits without publishing. When the shared probe then succeeds for A, line 285 treats A as stale and returns without publishing `ready`. The registry is now ready, but the connection is still `checking`; `ensureProviderReady()` also fast-paths on registry readiness, so no on-demand caller repairs the observation and handoff can remain unavailable until the next periodic refresh. Please fence publication by the underlying probe/owner rather than by every waiter, or retire an aborted generation without suppressing an active older publisher, and add this overlap-plus-cancellation regression case. [R4 / `packages/client/src/runtime/client-runtime-composition.ts:251-286`]

**Action taken**
- Submitted request changes

baixiaohang (changes requested):
- Rationale: The shared-owner fix closes the reported publication race, but a newly active caller can still inherit the cancellation of the previous owner while that owner is asynchronously draining.

**Risk level**: A
- Semantic lift: none; A is already the highest path tier

**PR summary**
- Problem: A non-settling Provider readiness probe can freeze all later capability refreshes and make otherwise healthy IM handoff unavailable until daemon restart.
- Approach: Bound each shared Provider refresh with one deadline and one publication owner; coalesced callers retain independent cancellation while Provider and IM work settle independently.
- Impacted modules: Client Runtime capability composition, Agent Runtime Provider registry, and their concurrency regression tests

**Review findings**
❌ 1. Evict an owner as soon as its last waiter aborts it. Lines 335-337 abort an unsettled zero-waiter owner but leave it in `sharedProviderRefreshes` until the owner's asynchronous operation reaches line 320. A new active call during that drain window therefore picks the already-aborted owner at line 330, increments its waiter count, and then receives the previous caller's `Agent Runtime provider probe has no waiters` rejection instead of owning or joining a live refresh. This makes one caller's cancellation observable by an independent later Turn. The registry layer already handles the same lifecycle correctly by removing a zero-waiter probe from its map immediately while the stale operation drains behind an identity fence. Please either remove this owner from the map atomically when aborting it or make lookup reject aborted owners, and add the sequence “A is the sole waiter and cancels; B starts while A's owner is draining; B receives a fresh/live refresh result.” [R4 / `packages/client/src/runtime/client-runtime-composition.ts:330-338`]

**Action taken**
- Submitted request changes

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The exact-owner eviction and identity fences close the remaining cancellation drain window without weakening deadline, publication, or lifecycle semantics.

**Risk level**: A
- Semantic lift: none; A is already the highest path tier

**PR summary**
- Problem: A non-settling Provider readiness probe can freeze later capability refreshes and make otherwise healthy IM handoff unavailable until the Client daemon restarts.
- Approach: Bound each shared Provider refresh with one deadline and one publication owner, isolate Provider and IM outcomes, and give coalesced callers independent cancellation while stale operations drain behind identity fences.
- Impacted modules: Client Runtime capability composition, Agent Runtime Provider registry, and their concurrency regression tests

**Review findings**
✅ 1. The final waiter now removes the exact owner before aborting it, and lookup rejects aborted or settled owners, so an independent later Turn receives a fresh live refresh instead of the previous caller's cancellation. [R4 / `packages/client/src/runtime/client-runtime-composition.ts:326-350`]
✅ 2. Both result publication and `finally` cleanup check current-owner identity, preventing a draining owner from overwriting or deleting its successor; the new A→B and A→B→C regressions exercise both orderings. [R4 / `packages/client/src/runtime/client-runtime-composition.ts:282-326`]
✅ 3. The composition still publishes restored exact-Provider readiness from the successful owner before handoff relies on the observation, while deadline and shutdown failures remain fenced from later state. [R1 / `packages/client/src/runtime/client-runtime-composition.ts:282-385`]

**Action taken**
- Approved

PR #143, opened 2026-08-24
feat(web): align Agent typography and loading states

Gandy2025:
## Summary
- remove the Agent-only typography scale override so detail, Usage, and Settings use the shared design-system roles
- reduce Agent list density and replace per-Agent warning card borders with a single scannable grouped list
- preserve confirmed Agent context when opening Usage or Settings and replace the stray generic skeleton with directory-shaped loading rows
- add slow-navigation coverage for Usage and Settings

## Validation
- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm --filter @opentag/web test` (332 tests)
- browser review of Agent list, home, Usage, Settings directory, all Settings subpages, and a forced 2.5-second Settings loading state

## Known unrelated local failure
- `pnpm test` reaches an existing Server observability failure in `src/observability/__tests__/observability.test.ts`: the RuntimeSession terminal-path assertion expects `stale_connection` but receives `handled`. This branch does not modify Server code.

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -1282,6 +1282,34 @@ describe("OpenTag Web App Shell", () => {
     expect(screen.queryByText("Runtime")).toBeNull();
   });
 
+  it.each([
+    ["Usage", "Usage"],
+    ["Settings", "Agent settings"],
+  ])("keeps Agent context visible while opening %s", async (linkName, destinationHeading) => {
+    let agentReads = 0;
+    let releaseAgentRead = () => {};
+    const pendingAgentRead = new Promise<void>((resolve) => {
+      releaseAgentRead = resolve;
+    });
+    installApi("admin", {
+      agentRead: () => {
+        agentReads += 1;
+        return agentReads === 1 ? undefined : pendingAgentRead;
+      },
+      bound: true,
+    });
+    window.history.replaceState({}, "", `/agents/${agentId}`);
+    render(<App />);
+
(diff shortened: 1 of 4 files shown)


baixiaohang (changes requested):
- Rationale: One Agent-to-Settings entry path still drops the confirmed Agent context, so the loading-state fix is incomplete.

**Risk level**: B-low
- Path baseline: `apps/web/**` falls under the unmatched-path baseline -> C.
- Semantic lift: `router.tsx` changes the shared Web resource-loading behavior -> B-low.

**PR summary**
- Problem: Opening Usage or Settings from a confirmed Agent profile can replace useful identity context with a generic loading state, while Agent typography and list density have drifted from the shared design system.
- Approach: Carry the loaded Agent through router state and revalidate it in the background, add a Settings-directory skeleton, remove the Agent-only type scale, and present Agents as a grouped list.
- Impacted modules: `apps/web/src/router.tsx`, `apps/web/src/styles.css`, and Web tests.

**Review findings**
❌ 1. The no-binding `Connect` action still navigates to `/agents/:id/settings/messaging` without including `agent` in router state. Starting from an already confirmed Agent detail page, that path therefore mounts `AgentSettingsPage` with no `initialAgent` and falls back to the full generic `Loading current server state` UI—the exact behavior this change is meant to remove. Please carry the same Agent value on this sibling entry path and add slow-navigation coverage for it. [R5 / `apps/web/src/router.tsx:1715`]

**Action taken**
- Submitted request changes.

PR #144, opened 2026-08-24
fix(web): stabilize dialog backdrop color

Gandy2025:
## Summary
- keep the dialog backdrop translucent when it inherits the global button hover state
- preserve backdrop opacity while the dialog is busy and the backdrop button is disabled
- add a CSS regression test for hover and disabled states

## Verification
- pnpm check (passes with existing undeclared E2E environment-variable warnings)
- pnpm build
- pnpm typecheck
- pnpm --filter @opentag/web test (333 tests passed)
- pnpm --filter @opentag/client test:agent-runtime:coverage
- browser reproduction confirmed the original hover state became solid brand green; the production CSS now compiles the stable translucent hover rule

## Baseline test note
- pnpm test and pnpm test:coverage reproduce unrelated failures outside apps/web, including the existing RuntimeSession stale-connection observability expectation and local CLI service-path assumptions. No server or CLI files are changed by this PR.

### DIFF
--- a/apps/web/src/__tests__/page-layout.test.ts
+++ b/apps/web/src/__tests__/page-layout.test.ts
@@ -80,6 +80,13 @@ describe("workspace page layout", () => {
     expect(declarationValue(designSystemStyles, "select.ds-control--compact", "min-height")).toBe("36px");
   });
 
+  it("keeps the dialog backdrop stable across inherited button states", () => {
+    const backdrop = "rgb(22 33 27 / 42%)";
+    expect(topLevelDeclarationValue(appStyles, ".dialog-backdrop", "background")).toBe(backdrop);
+    expect(topLevelDeclarationValue(appStyles, ".dialog-backdrop:hover", "background")).toBe(backdrop);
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The scoped CSS overrides preserve the intended translucent backdrop through both inherited hover and disabled button states, and the regression assertions cover both cascade paths.

**Risk level**: C
- Path baseline: only `apps/web/**` is changed.
- Semantic lift: none; the change is limited to dialog presentation CSS and a stylesheet regression test.

**PR summary**
- Problem: Dialog backdrops could turn solid brand green on hover and become too transparent while a busy dialog disabled dismissal, causing a visible and distracting modal-state regression.
- Approach: Give the backdrop explicit hover and disabled-state presentation rules while preserving the existing interaction behavior, with CSS-level regression coverage for both inherited global button states.
- Impacted modules: `apps/web/src/styles.css`, `apps/web/src/__tests__/page-layout.test.ts`

**Review findings**
✅ The state-specific selectors are more specific than the global button rules and preserve the backdrop's background and opacity without changing dialog dismissal semantics.

**Action taken**
- Approved.

PR #153, opened 2026-08-25
feat(release): publish portable releases to Cloud Storage

bestony:
## Summary

Adds a portable distribution so OpenTag can be installed on machines with no Node.js toolchain: one tarball per platform carrying the bundled CLI plus its own pinned Node.js runtime, published to Google Cloud Storage and installed with a single `curl … | sh`.

Portable channels reuse the npm channel coordinates exactly, so a portable release can never describe a version the npm release path would reject.

**Build** — `scripts/portable/build-portable.mjs` assembles the app once and pairs it with the official Node.js build for each platform, checked against that release's `SHASUMS256.txt`. Tarballs are byte-reproducible for a given commit, version, and platform (normalized archive timestamps, sorted entries, zeroed owner identity, no gzip mtime), which is what lets a resumed release re-upload identical bytes instead of colliding with the create-only precondition.

The builder never rewrites package identity — `scripts/prepare-cli-release.mjs` already owns that for every channel. Instead it *runs* the assembled CLI and checks that it reports the requested version and presents itself as the channel's binary, so a stale `apps/cli/dist` fails the build with an actionable message rather than shipping.

**Publish** — `scripts/portable/upload-gcs.sh` treats the version prefix as immutable: objects are written create-only (`--if-generation-match=0`) with a `--content-md5` digest, an unexpected object under the prefix fails the release, and an existing object must match the local artifact's size, `sha256` metadata, and `md5Hash`. `latest.json` and `install.sh` are the only mutable objects; they are written last and only after the **public** endpoint has been proven to serve the complete versioned release byte for byte. Every URL that verification checks is derived from the local release metadata, never from anything already remote.

**Install** — `scripts/portable/install.sh` follows the interaction the First Tree installer established (`--version`, `--prefix`, `--bin-dir`, `--no-path-edit`, `--path-mode`, marker-delimited `PATH` block, `current` symlink behind a stable shim). Two behaviors are worth review:

- **Up-to-date short-circuit.** After reading `latest.json`, the installer checks whether the requested version is already the live install: `current` must resolve to a payload whose `INSTALL.json` matches the target version, platform, and binary name; the embedded runtime and app entry must both be present; and the stable shim in `--bin-dir` must already resolve through `current`. When all of that holds it reports that OpenTag is already up to date and exits **without downloading the payload**. A partial install, a hand-edited shim, or a deleted shim fails the check and is repaired by a full reinstall. `--force` reinstalls unconditionally.
- **Commit ordering.** The payload checksum is verified before the payload is executed, the new runtime is smoke-tested where it was extracted before anything already on disk is touched, and the `current` symlink is the final commit point. A freshly extracted payload is moved into place with two renames rather than a recursive delete, so reinstalling the version `current` already points at cannot leave the link dangling.

**Release authority** — `publish-npm-package.yml` preflights the portable release before `npm publish`, so an immutable-prefix conflict fails while the release is still retryable, and publishes afterwards from the same preflighted artifacts. Authentication uses workload identity federation; no service-account key is stored. The steps are skipped unless `OPENTAG_PORTABLE_GCP_WORKLOAD_IDENTITY_PROVIDER` is configured.

## Deployment prerequisite

The bucket must serve the release prefix publicly at the download base URL. `gs://opentag-release` currently has no `allUsers` read binding, so the public verification gate fails and the channel pointers are deliberately left untouched. Uploads still succeed, so the first configured release will land its immutable objects and can be completed by rerunning once the bucket (or the `download.opentag.build` load balancer) serves them.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck` pass.
- `node --test scripts/__tests__/*.test.mjs` — 57 pass, including 19 new portable tests. The installer tests build a synthetic release, serve it over a loopback HTTP server, and assert the full flow: activation, the up-to-date short-circuit (asserted by request count — the payload is not fetched a second time), `--force` reinstall, upgrade to a newer version, repair when the shim is missing, and rejection of a payload that fails its published checksum. No public network access.
- End-to-end against a real 39 MB `darwin-arm64` artifact over a loopback server: install → `--force` reinstall → up-to-date short-circuit, with the installed CLI reporting the right version and no leftover temp directories.
- Real Cloud Storage run against an isolated `releases-test` prefix on `gs://opentag-release`: create-only upload, `--content-md5`, `Cache-Control`, and `sha256` custom metadata all verified on the stored objects; a rerun correctly recognized them as present and matching; a tampered local artifact failed the preflight on all three of size, `sha256`, and `md5Hash`. The public verification gate returned 403 and refused to move the channel pointers, as designed. **All test objects were deleted afterwards**; `releases/` was never written.
- Fixed one portability bug found during review: forwarded option arrays in `release-gcs.sh` are expanded as `${ARR[@]+"${ARR[@]}"}` because the bash 3.2 that ships with macOS rejects `"${ARR[@]}"` on an empty array under `set -u`. Covered by a regression test.

`pnpm test` has three failures that are **pre-existing and unrelated**, each reproduced on a clean `main` checkout:

- `apps/cli` — two daemon-service tests fail on macOS only, comparing `/var/…` against the `realpath`-resolved `/private/var/…`.
- `packages/server` — `observability.test.ts` "traces real RuntimeSession failure, overload, and dropped-queue terminal paths".

## Breaking changes

None. This adds a new distribution channel; npm publishing, the CLI, and the daemon are unchanged. The portable workflow steps are inert until the Cloud Storage repository variables are set.

## Checklist

- [ ] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. — the first three pass; `pnpm test` has the three pre-existing failures listed above, unchanged by this PR.


baixiaohang (changes requested):
- Rationale: The release path can accept or advertise bytes it has not actually verified, and `--force` can break the active installation during an interrupted replacement.

**Risk level**: B-low
- Path baseline: C — release scripts, workflow, tests, and documentation; no `apps/cli/**` or package runtime path is changed.
- Semantic lift: release and CI workflow behavior raises the review one level.

**PR summary**
- Problem: Users currently need a Node.js toolchain to install OpenTag; this adds a self-contained, one-command portable install for the supported macOS and Linux platforms.
- Approach: Build reproducible per-platform archives containing the bundled CLI and a pinned Node.js runtime, publish immutable version objects to Cloud Storage, then move mutable channel entry points after public verification.
- Impacted modules: portable build/install/upload scripts, npm publishing workflow, release documentation, and script tests.

**Review findings**
❌ 1. An existing immutable object is accepted when `md5Hash` is absent. A composite or otherwise MD5-less object only needs the expected size plus caller-controlled `metadata.sha256`, so different bytes can pass preflight and be treated as the preflighted release. Require `md5Hash` to be present and equal for these objects, or download/hash the object content before accepting it. [`scripts/portable/upload-gcs.sh:318`]

❌ 2. The public endpoint gate does not verify tarball bytes: it only sends HEAD (or a one-byte range request), then flips `latest.json` and `install.sh`. A stale or misrouted CDN tarball therefore passes the gate even though every installer will reject its SHA-256; the later smoke test runs after publication and covers only the runner's platform. Download and hash every published asset before moving channel pointers, and run the host install smoke against the immutable version before the commit point. [`scripts/portable/upload-gcs.sh:185`]

❌ 3. `--force` moves the directory targeted by `current` out of the way before installing its replacement. A signal, crash, or failed second rename between lines 510 and 512 leaves `current` permanently dangling despite the stated final-commit guarantee. Install under a new immutable directory and switch `current`, or add a guarded rollback that restores the old directory on every failure path. [`scripts/portable/install.sh:503`]

❌ 4. Production release jobs are not serialized, while the two mutable channel objects are overwritten independently with no generation or version guard. Concurrent tag jobs can regress `latest.json` or interleave it with a different `install.sh`; a failure between the two writes creates the same mixed state. Make publication monotonic and single-writer (for example, serialize production and use generation/version preconditions), or publish a versioned installer and commit the channel with one guarded pointer. [`.github/workflows/publish-npm-package.yml:163`, `scripts/portable/upload-gcs.sh:541`]

**Action taken**
- Submitted request changes on head `ce2a8bb81885537709d28f661fcb3a3b96289f82`.

baixiaohang (changes requested):
- Rationale: The four previous blockers are fixed, but the new production serialization can still drop or reorder tagged releases and split the npm and portable channel heads.

**Risk level**: B-low
- Path baseline: C — release scripts, workflow, tests, and documentation.
- Semantic lift: release and CI workflow behavior raises the review one level.

**PR summary**
- Problem: Users need a self-contained OpenTag install that does not require a pre-existing Node.js toolchain.
- Approach: Publish verified, immutable per-platform bundles and advance npm and portable channel coordinates through the same production workflow.
- Impacted modules: portable build/install/upload scripts, npm publishing workflow, release documentation, and script tests.

**Review findings**
❌ 1. `cancel-in-progress: false` prevents cancellation of the running production job, but it does not create the lossless queue described by the comment. With the default concurrency queue, a third tag run replaces the existing pending run, and start order is not guaranteed ([GitHub concurrency documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)). That can silently skip a protected release tag. It can also run a higher version before a lower one: `npm publish --tag latest` moves npm's `latest` tag to whichever version it publishes ([npm dist-tag documentation](https://docs.npmjs.com/cli/dist-tag/)), while `upload-gcs.sh` correctly refuses to regress the portable pointer. The resulting channel heads can therefore diverge (for example npm `latest` at 1.2.0 and portable `latest.json` at 1.3.0), contradicting the shared-channel contract. Preserve every release run (for example `queue: max`) and apply the same semantic monotonic guard to npm's dist-tag, or use a single ordered commit step that advances both channel heads only when the version moves forward. [`.github/workflows/publish-npm-package.yml:167`, `.github/workflows/publish-npm-package.yml:293`, `scripts/portable/upload-gcs.sh:701`]

✅ 2. The previous four blockers are resolved: MD5-less immutable objects now fail closed, public assets are fully hashed and host-installed before pointer commit, forced reinstall no longer rewrites the active target, and the portable pointer uses a generation precondition plus semantic regression guard.

**Action taken**
- Submitted request changes on head `820f349e75c8372d1e12d42ca00db8e694014261`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The remaining release-ordering blocker is resolved: production tags are retained in the bounded queue, and npm and portable channel heads now share the same forward-only version ordering.

**Risk level**: B-low
- Path baseline: C — release scripts, workflow, tests, and documentation.
- Semantic lift: release and CI workflow behavior raises the review one level.

**PR summary**
- Problem: Users need a self-contained OpenTag install that does not require a pre-existing Node.js toolchain.
- Approach: Build verified per-platform bundles, publish immutable versioned objects, and advance npm and portable channel heads only after the release is publicly installable.
- Impacted modules: portable build/install/upload scripts, npm publishing workflow, release documentation, and script tests.

**Review findings**
✅ 1. `queue: max` replaces the default single pending slot for production, so a normal burst of protected release tags is retained instead of silently replacing the existing pending run.

✅ 2. npm now compares the candidate with the published `latest` using the same stable/staging ordering as the portable pointer. An out-of-order version is still published under `superseded` without regressing either channel head.

✅ 3. The portable pointer commit now re-reads and re-decides after a lost generation precondition, while the previous integrity and installer atomicity fixes remain intact.

**Action taken**
- Approved head `535590a8fdc0931976c2d13bccff1fdc3bab4d8c`.

PR #42, opened 2026-08-20
test: add unit coverage reporting

baixiaohang:
## Summary

- enforce the existing Agent Runtime 100% V8 coverage gate in the primary Node.js 24 `Checks` job
- add one root Vitest projects runner for CLI, Web, Shared, Client, and Server, including unimported production files
- upload Agent Runtime and unified JSON summary/HTML reports, and document the commands and baseline maintenance rules in English and Chinese

## Unit coverage baseline

Percentages and uncovered counts below come from the final local run. This is a measurement baseline, not a new coverage gate: repeated Web and Client runs currently vary slightly, so repository-wide or per-workspace regression thresholds are deferred until the measurement is stable.

| Workspace | Statements | Branches | Functions | Lines |
| --- | ---: | ---: | ---: | ---: |
| CLI | 77.28% (616) | 71.74% (204) | 81.95% (35) | 77.28% (616) |
| Web | 63.51% (390) | 80.64% (42) | 68.81% (29) | 63.51% (390) |
| Shared | 89.95% (155) | 72.52% (25) | 35.13% (24) | 89.95% (155) |
| Client | 93.73% (578) | 92.18% (300) | 95.14% (34) | 93.73% (578) |
| Server | 37.09% (6065) | 72.92% (261) | 51.23% (198) | 37.09% (6065) |

Agent Runtime remains independently gated at 100% statements, branches, functions, and lines.

## Follow-up priorities

1. Web is the highest-value UI target: `router.tsx` is 60.56% statements with 338 uncovered. Cover invitation errors, Agent creation failures, revision conflicts, IM setup polling/failure/disable, Team switching, and loading/error states.
2. Server is 37.09% statements in the unit-only report. Some low service numbers reflect the deliberate PostgreSQL-integration exclusion, but `index.ts` is still 0% with 268 uncovered statements; unit-test migration/verification branches, partial-start cleanup, listen failure, and service ordering.
3. Shared has 35.13% function coverage, led by `http-paths.ts` at 8.33% functions and `browser.ts` at 0%. Add direct schema/path boundary tests, especially UTF-8 byte limits, union/enum failures, resource limits, and provider write-error variants.
4. CLI and Client are stronger overall; target CLI `core/agent/im.ts` (5.26% statements) and Client `runtime-connection.ts` (79.72%) after the higher-risk Web and Server gaps.
5. Stabilize Web/Client V8 remapping results across repeated runs before introducing per-workspace regression thresholds.

## Validation

[...]

PR #154, opened 2026-08-25
chore(ci): stop measuring unit coverage

baixiaohang:
## Summary

- stop running the full unit coverage measurement in every CI run
- keep the focused Agent Runtime coverage enforcement
- rename the uploaded artifact and remove the no-longer-produced unit coverage path

## Validation

- `git diff --check`
- parsed `.github/workflows/ci.yml` with Ruby's YAML parser

## Non-goals

- no changes to local coverage commands
- no changes to the Agent Runtime coverage threshold

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,17 +74,12 @@ jobs:
       - name: Enforce Agent Runtime coverage
         run: pnpm --filter @opentag/client test:agent-runtime:coverage
 
-      - name: Measure unit coverage
-        run: pnpm test:coverage
-
       - name: Upload coverage reports
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: unit-coverage-reports-${{ github.sha }}
-          path: |
-            packages/client/coverage/**
-            coverage/unit/**
+          name: agent-runtime-coverage-report-${{ github.sha }}
+          path: packages/client/coverage/**
           if-no-files-found: warn
           retention-days: 14
 
--- a/.github/workflows/unit-coverage.yml
+++ b/.github/workflows/unit-coverage.yml
@@ -1,0 +1,47 @@
+name: Unit Coverage
+
+on:
+  schedule:
+    - cron: "17 3 * * 1"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  measure:
+    name: Measure Unit Coverage
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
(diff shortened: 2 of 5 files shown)


yuezengwu (changes requested):
Reviewed exact head `5d8b3cbfff1f4ff8d2027938542ea9a6e089aec0`.

[R2] Please update the contributor validation contract together with this CI change. `DEVELOPMENT.md` lines 22-43 and the synchronized `DEVELOPMENT.zh-CN.md` mirror both list `pnpm test:coverage`, then state that the required `CI` fan-in covers all commands above. This head removes that command and the unified coverage artifact, so the documented contract becomes false and the baseline introduced by #42 silently changes from per-PR collection to unowned/on-demand measurement. Either retain the CI step, or update both documents to say when/where the root baseline is collected and maintained, while making explicit that CI still runs all unit tests and keeps the focused 100% Agent Runtime gate.

The code-level quality gate is retained: I independently ran the focused suite on this head (282 tests, 100% statements/branches/functions/lines). This also means the focused Vitest 4 coverage failure seen on #148 is not hidden by this PR. However, removing the root measurement does stop CI from detecting root coverage-config compatibility regressions such as the `coverage.all` breakage noted on #148, so the reduced observability and its replacement workflow should be explicit rather than implicit.

baixiaohang:
Addressed in 2983d52. The root unit coverage measurement now runs in a dedicated workflow every Monday at 03:17 UTC and remains manually dispatchable, with reports retained for 14 days. The contributor docs and Chinese mirror now distinguish this recurring baseline from required PR validation, while documenting that PR CI still runs all unit tests and preserves the focused Agent Runtime 100% gate.

yuezengwu (approved):
Reviewed exact head `2983d520c0377f535b1937d78a6bb3e696b30f87`. The prior documentation/ownership blocker is resolved; I found no new blocking issue.

The dedicated `Unit Coverage` workflow has the intended boundaries: Monday 03:17 UTC plus manual dispatch, default-branch scheduled semantics, Node 24 and pinned pnpm with frozen install, `contents: read` permission only, the unchanged no-threshold root coverage command, `if: always()` report upload, and 14-day retention. PR CI still runs all offline unit tests and the focused Agent Runtime 100% gate. `AGENTS.md`, `DEVELOPMENT.md`, and the dated Chinese mirror now distinguish required PR validation from the recurring baseline, so the #42 baseline retains an explicit owner and #148-style root coverage compatibility regressions remain visible weekly or on demand.

Validation on this head: dependency/install and repository checks passed; both workflow YAML files parsed; `git diff --check` passed; focused Agent Runtime coverage passed 282/282 tests at 100% statements/branches/functions/lines. `pnpm test:coverage` produced the expected `coverage/unit` report; its local macOS run ended non-zero on two canonical `/var` vs `/private/var` path assertions and one timing-sensitive observability assertion, none changed by this workflow/docs-only diff. The first GitHub CI attempt independently failed one existing PostgreSQL integration race (`Old generation dispatch was not persisted`); rerunning the same exact head passed all six CI checks.

PR #155, opened 2026-08-25
feat: add internal session collaboration

baixiaohang:
## Summary

- add optional runtime session-collaboration negotiation and durable `session_messages` facts with explicit retry and attempt fencing
- expose additive `create_internal_session` and `send_session_message` hosted tools with local/remote relay, bounded FIFO admission, and internal-session IM credential isolation
- serialize SessionMessage authority checks and admission with reconcile, while fencing active Direct IM and SessionMessage Runs
- re-prepare existing Sessions when the negotiated hosted-tool set changes
- preserve exact Provider resume while allowing the Client Runtime to explicitly create a replacement Codex thread for an approved hosted-tool incompatibility
- support immutable internal runtime overrides without changing visible Session configuration behavior

## Architecture decision

The narrow provider-binding replacement boundary is recorded in [first-tree-context#977](https://github.com/first-tree-ai/first-tree-context/pull/977): `AgentRuntimeFactory.resume` remains exact and fails on an incompatible binding; only the Client Runtime may explicitly select `create` after the old runtime is idle and closed. The OpenTag Session identity stays stable, while the Codex provider transcript intentionally resets for that capability transition.

## Boundaries

- no Task or Web product work
- no Agent-facing `end_session`
- no offline delivery queue, worker, startup replay, or automatic retry
- no reuse of `ImMessage` or provider outbound infrastructure

## Verification

- `pnpm check` (passes with 8 existing undeclared E2E environment-variable warnings)
- `pnpm build`
- `pnpm typecheck`
- `pnpm test` (1225 tests before focused follow-ups)
- `pnpm --filter @opentag/server test:integration` (161 tests)
- `pnpm --filter @opentag/client test` (460 tests on the final head)
- `pnpm --filter @opentag/client test:agent-runtime:coverage` (284 tests; 100% statements, branches, functions, and lines)
- `pnpm test:coverage` (1228 tests on the final head)

Independent static review completed with no blocking findings.


yuezengwu (changes requested):
Requesting changes on exact head `17dcdee12cabe2b5d56a54ca65454c22e8d90a8f`.

[P1] Re-prepare existing Sessions when the negotiated collaboration capability changes

`SessionReconciler` only calls `prepareSession` when placement generation or runtime revision advances, or the Session was stopped (`packages/client/src/runtime/session-reconciler.ts`, around lines 343-350). The hosted-tool set is sampled later only when `SessionRuntimeManager` starts a provider runtime. This leaves an existing `ready` Session outside the capability lifecycle.

A rolling-compatible repro is: start this Client against a protocol-v2 Server that does not offer `runtime.sessionCollaboration`; reconcile and start a Session; then let the same daemon reconnect to a Server that offers the capability. The next reconcile has the same placement/revision and skips preparation, so the live runtime never receives `create_internal_session` / `send_session_message`, and a legacy Codex binding is never replaced. The reverse transition also leaves stale hosted-tool definitions registered (the handler fails closed, but the runtime surface no longer matches negotiation). A normal WebSocket/server restart does not close `SessionRuntimeManager`, so this persists until an unrelated runtime or daemon restart.

Please make negotiated hosted-tool changes invalidate/re-prepare the affected managed Sessions, including the required Codex thread replacement, and add disabled→enabled plus enabled→disabled reconnect regression coverage for an already-running Session. This is required for the optional capability and backward-compatible rolling deployment semantics documented by this PR.

yuezengwu (changes requested):
Requesting changes on exact head `7146a7515d55886f58354d33b476209764910eb3`.

The fresh change fixes the prior rolling-reconnect blocker: `SessionRuntimeManager` now invalidates and re-prepares an existing idle managed Session when the negotiated hosted-tool set changes in either direction, without closing an active Run mid-turn. I verified the disabled→enabled→disabled path and the no-op same-capability path on this head.

[P1] Resolve the Codex thread-replacement conflict with the durable runtime contract

`CodexAgentRuntimeFactory.resume` now treats any `hostedToolsHash` mismatch as `thread/start`, emits the newly created thread binding, and lets `SessionRuntimeManager` persist it. In production this silently replaces the Provider conversation identity of an existing long-lived Session. The authoritative Agent Runtime contract says that factory resume restores the exact opaque binding and must fail rather than silently create a replacement; it also requires Codex exact resume to preserve one persistent Thread identity: https://github.com/agent-team-foundation/first-tree-context/blob/899e57d31c050594d6e72ce9368315141ea32c89/opentag/system/agent-runtime.md#public-object-model

The PR documentation currently states the opposite replacement behavior, but there is no linked durable architecture decision authorizing that semantic change. Please resolve the conflict before landing: either preserve the existing Codex thread identity across negotiated tool changes, or establish the explicit authoritative product/architecture decision that permits the one-time identity replacement and align the durable contract. The regression coverage should then assert the chosen identity semantics across the `SessionRuntimeManager` and Codex factory boundary.

packages/client/src/runtime/session-message-inbox.ts
```diff
@@ -0,0 +1,196 @@
...
+        return;
+      }
+      let reservation = this.#admission.reserve(sessionId, next.request.agentId);
+      while (!reservation.accepted) {
+        await this.#admission.waitForRelease(this.#abort.signal);
+        reservation = this.#admission.reserve(sessionId, next.request.agentId);
+      }
+      queue?.shift();
+      this.#queuedTotal -= 1;
+      reservation.reservation.markActive();
```
yuezengwu, line 129: [P1] Fence capability replacement while a SessionMessage Run is active

SessionMessageInbox marks only the shared AdmissionController reservation active, then runs runtime.prompt; unlike direct IM custody it never records the Run in SessionReconciler activities. A reconnect with changed collaboration capabilities therefore sees no active turn, invokes prepareSession, and SessionRuntimeManager closes the provider runtime underneath the active SessionMessage Run before creating its replacement. The message was already acknowledged accepted, so this can abort or drop collaboration work and violates the idle-only replacement boundary.

Please make active SessionMessage Runs visible to the reconcile/replace gate with the same agent/session lock and fence discipline, and add a disabled-to-enabled or enabled-to-disabled reconnect regression while a SessionMessage prompt is blocked. Reconcile must remain busy/running until that prompt settles; only then may the old runtime close and explicit create replace its binding.

yuezengwu (changes requested):
Requesting changes on exact head ee5996c654e5c16fee69a35d498c84d6a64dac70.

The fresh split now keeps AgentRuntimeFactory.resume exact/fail and moves Codex replacement into an explicit Client Runtime create path; the stable OpenTag Session / reset Codex transcript semantics match the currently published runtime contract. The original disabled→enabled→disabled rolling reconnect fix also remains.

One idle-gating blocker remains for internal SessionMessage Runs; see the inline finding.

packages/client/src/runtime/session-message-inbox.ts
```diff
@@ -0,0 +1,210 @@
...
+      const runId = `session-message-${next.request.messageId}`;
+      await this.#reconciler.withAgentLock(next.request.agentId, async () => {
```
yuezengwu, line 131: [P1] Revalidate SessionMessage delivery inside the agent lock before accepting it

The new activity update is locked, but the target check and accepted result still happen earlier outside that lock. RuntimeConnection dispatches business frames concurrently, so a stop, generation change, or configuration reconcile can already own the agent lock and be blocked in preparation while accept reads the old ready snapshot and returns accepted. When the reconcile finishes, this callback runs without rechecking and starts the message against a stopped or replacement Session runtime.

I reproduced this on the fresh head by blocking stopSession after it acquired the lock, delivering a SessionMessage, then releasing the stop: accept returned accepted, the Session became stopped, and ensureRuntime plus prompt still ran. The same ordering permits an old-placement delivery to run on the newly prepared generation. Direct IM custody avoids this by performing its target check under withAgentLock.

Please make the SessionMessage target/generation/configuration check and admission decision participate in the same agent-lock ordering before acknowledging accepted, and add a regression where an already-running stopped or generation-changing reconcile wins the lock. That delivery must be rejected and must never reach ensureRuntime or prompt.

yuezengwu (changes requested):
Requesting changes on exact head 8ac2260d3416e5fdaa8c3a2fcd6138b157822321.

The fresh change fixes the original active-Run replacement path: a blocked SessionMessage prompt now publishes SessionReconciler activity before execution, capability reconnect reports running without prepare or replacement, and activity plus admission are cleared on settlement and failure paths. The explicit Codex replacement and rolling reconnect semantics remain intact.

One agent-lock ordering blocker remains; see the inline finding.

yuezengwu (approved):
Approving exact head `b2654105922b48f9ac790e5179e9bdd3ed9a9a5f`.

The prior P1 is fixed: SessionMessage authority and bounded FIFO admission now run under the per-Agent reconcile lock before accepted is acknowledged, and both the remote delivery handler and local fast path await that decision. If stop or a placement-generation reconcile already owns the lock, delivery waits, then rejects against the resulting authoritative state without reaching `ensureRuntime` or `prompt`; transient authority failures remain retryable with the same logical message identity.

I revalidated the existing active-Run activity fence, explicit Codex binding replacement boundary, and disabled/enabled rolling reconnect behavior against the published normal Runtime contracts. The merge from current `main` is patch-identical by range-diff and introduced no collaboration conflict.

Validation on this exact head:
- targeted Client Runtime suite: 6 files / 69 tests, including an added-then-removed generation-race reproduction
- `pnpm check` (passes with 8 pre-existing E2E env warnings)
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`
- all GitHub checks green

PR #159, opened 2026-08-25
Default IM receive mode to all messages

yuezengwu:
## Summary

- default newly created Agents to `all_message` for IM receive mode
- add a forward Drizzle migration that changes only the column default
- preserve existing Agents and explicit `mention_only` test scenarios without backfilling stored rows

## Validation

- `pnpm --filter @opentag/server db:generate` (no further schema changes)
- affected integration suites: 135 tests passed
- `pnpm check`
- `pnpm build`
- `pnpm typecheck`

`pnpm test` has one unrelated existing failure in `src/observability/__tests__/observability.test.ts`: the RuntimeSession terminal-path test observes `handled` instead of `stale_connection`. The same failure reproduces on the exact base commit `09ba8e26d03adc2d20a3ee3b7ba82f487f5298d3` in a clean detached worktree.

## Non-goals

- no receive-mode backfill for existing Agents
- no Context Tree change; the canonical Agent collaboration model already defines `all_message` as the default

### DIFF
--- a/packages/server/drizzle/0014_illegal_wolfsbane.sql
+++ b/packages/server/drizzle/0014_illegal_wolfsbane.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "agents" ALTER COLUMN "receive_mode" SET DEFAULT 'all_message';
--- a/packages/server/drizzle/meta/0014_snapshot.json
+++ b/packages/server/drizzle/meta/0014_snapshot.json
@@ -1,0 +1,2724 @@
+{
+  "id": "bc597868-00b0-43ad-8a9a-9a1964b15246",
+  "prevId": "9c020d90-84e2-4ff4-8f5d-8f48f461b9f8",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.agent_runtime_configs": {
+      "name": "agent_runtime_configs",
+      "schema": "",
+      "columns": {
+        "agent_id": {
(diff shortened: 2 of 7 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The forward-only default change aligns new Agent behavior with the current collaboration model while preserving every stored Agent's explicit or historical receive mode.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes a database default and forward Drizzle migration.
- Semantic lift: none; the path baseline already captures the database compatibility risk.

**PR summary**
- Problem: Newly created Agents currently start in mention-only mode, so they do not receive ambient channel traffic by default even though the collaboration model defines all-message awareness as the standard behavior.
- Approach: Change the schema and database default to `all_message`, add a forward migration that does not backfill existing rows, and make tests explicit where they intentionally exercise `mention_only` behavior.
- Impacted modules: Server Agent schema, Drizzle migration history, and Agent/IM integration tests.

**Review findings**
✅ 1. The generated snapshot differs from the prior snapshot only in the `receive_mode` default, keeping the migration blast radius narrow.
✅ 2. The upgrade test proves existing rows remain `mention_only`, while Agent creation coverage proves fresh records receive `all_message`.
✅ 3. The production Agent creation path has a single insert boundary and continues to rely on the schema default, so no competing application-side default was introduced.

**Action taken**
- Approved.

PR #161, opened 2026-08-25
Implement Workspace Admin and isolated machine authority

### LINKED ISSUE #125: refactor: adopt equal Workspace Admin access and separate machine authority
# Workspace and Admin model

Status: accepted product and architecture direction.

## Summary

OpenTag serves two different populations. A small number of Admins configure and operate Agents through the management surface, while a much larger number of IM Users use those Agents directly from Feishu, Slack, or another messaging provider. The previous Team membership model treated both populations as though they needed OpenTag accounts and Workspace membership. That added concepts and permissions without matching the product.

OpenTag will remove Team as a product and domain concept. Workspace remains the collaboration, authorization, and data-isolation scope, but it stays implicit when an Admin can access only one. Every Admin in a Workspace has the same complete management authority over all of its Agents. IM Users do not join a Workspace and do not need OpenTag Accounts.

This direction supersedes the earlier rename-only plan that would have preserved generic Membership roles, reusable member invitations, User-owned Agent authority, and human user credentials for the daemon. The implementation still includes the Team-to-Workspace vocabulary migration, but it must not preserve those obsolete semantics under new names.

### Product mismatch

Most people experience OpenTag only through an Agent in IM. Requiring those people to register with OpenTag, join a Team, or understand a Workspace would create an unrelated onboarding funnel. OpenTag Accounts exist for administrators, not for everyone who sends an Agent a message.

### An unnecessarily broad role model

The current `admin` and `member` roles suggest that general members need a read-only management experience. There is no established product need for it. People either administer the shared OpenTag environment or use its Agents through IM. Speculative viewer and per-Agent roles can be added only after a concrete use case appears.

### Creator ownership conflicts with shared administration

An Agent must not become private to, or operationally dependent on, the Admin who created it. Admins are trusted to manage the whole Workspace, so another Admin must be able to diagnose, configure, suspend, and recover that Agent. Creation identity is audit metadata, not an authorization rule.

### Human and machine authority are currently coupled

The existing connect-code flow ultimately gives the daemon user-level bearer credentials, and the Runtime WebSocket authenticates through the human user authentication path. Once every management user is a full Admin, a Computer credential must not inherit that Admin authority. A compromised execution host must not be able to add Admins or modify unrelated Agents.

### Workspace is useful, but should use progressive disclosure

Workspace is still necessary whenever one Account administers multiple independent trust scopes, such as personal and company Agents, different customers, or isolated departments. It should not be presented during the common single-Workspace path. Workspace selection and naming appear only when they help distinguish multiple scopes or identify an Admin invitation.

### Workspace

A Workspace is the smallest scope in which all Admins completely trust one another to manage the same Agents and supporting resources. If two administrators must not see or change the same Agents, they belong in different Workspaces.

### Admin

An Admin is an authenticated Account with an active grant to a Workspace. All Admin grants have the same authority; there is no Owner, Manager, Member, or role column. An Account can administer more than one Workspace.

### IM User

An IM User invokes an Agent under the Agent's IM access policy. The provider identity used for message routing and policy enforcement is not an OpenTag Account and does not create Workspace access.

### Agent

An Agent belongs to a Workspace. Every Admin in that Workspace can manage it, including when another Admin created it. `createdBy` is retained only for audit.

### Computer

A Computer is an independent execution identity with a machine-scoped credential. An Admin explicitly enrolls it into a Workspace. Workspace Admins can manage the enrollment and bind Workspace Agents to an enrolled Computer, but OpenTag does not imply that they possess or control the physical host.

Removing `agents_manager_computer_owner_fk` widens who may **manage** an Agent bound to a Computer. It does not open **rebinding** that Agent to a different Computer, which stays a non-goal. An Agent's Computer is chosen at creation and does not change, so a Computer that is gone for good means deleting and recreating the Agent. That is deliberate for now, and it is not free: deletion disables the Agent's IM binding in the same transaction, `im_bindings.agent_id` is never reassigned, and the replacement therefore registers a new provider app and must be re-added to every group it was in (issue #96 traces the chain). Revisit both non-goals when a Workspace actually loses a Computer, and treat #96 as the higher-leverage fix, since portable Agent identity covers this case and the immutable Provider case with one mechanism.

The two non-goals are ordered. `workspace_computers.revoked_at` exists but nothing revokes an enrollment yet. Shipping revocation before rebinding would hand an Admin a one-click way to strand every Agent bound to that Computer, with no recovery except deletion. If revocation ships first it must refuse while bound Agents remain, naming them.

## Permission rules

| Action | Workspace Admin | IM User | Computer credential |
| --- | --- | --- | --- |
| View and configure all Workspace Agents | Yes | No | No |
| View Tasks and operational state | Yes | No | Only assigned runtime work |
| Manage Integrations, Skills, and Computer enrollment | Yes | No | No |
| Invite or revoke an Admin | Yes | No | No |
| Invoke an Agent through an allowed IM context | Not implied | Yes | No |
| Connect to the Runtime WebSocket and report work | No | No | Yes |

Adding an Admin therefore grants complete Workspace management authority and must be presented as a high-trust action. At least one active Admin grant must remain.

### IM User

An IM User discovers an Agent in an existing channel, thread, or direct-message context and uses it immediately. No OpenTag page, Account, Team, or Workspace enrollment is required.

### First Admin

The first Admin signs in to the management surface. OpenTag creates a default Workspace automatically. The interface proceeds directly to Agent setup and does not ask the Admin to create, name, or select a Workspace.

Automatic creation is triggered by an Account's first sign-in, never by the condition "this Account currently administers no Workspace". Using the latter would hand a revoked Admin a fresh empty Workspace instead of telling them their access was removed.

### Additional Admin

An existing Admin creates a short-lived, single-use Admin invitation. The recipient previews which Workspace is granting access, signs in, explicitly accepts complete management authority, and receives an equal Admin grant.

### An Account with no Workspace

An Account can legitimately hold zero Admin grants: an Account whose grant was revoked, and, after the migration in the delivery plan, an existing `member` that is deliberately not promoted. That Account must reach a state that explains what happened and names who can restore access. It must not be handed a new empty Workspace, which would silently replace the Agents it used to see with an empty shell and no explanation.

### Multiple Workspaces

An Account with one Workspace never sees a Workspace selector in the default shell. A selector and Workspace settings become visible only after the Account can administer more than one Workspace or deliberately creates another scope.

The shell opens on a specified Workspace rather than on whichever record the API happens to return first. The rule is: the most recently used Workspace for this Account, then a Workspace that has completed Agent setup, then the earliest granted. A client may store a preference, but the rule must also hold on a device that has none: a new browser, a cleared profile, private browsing.

This case is not rare; the model in this document produces it routinely. Someone signs in to look at the product and receives an auto-created default Workspace they never use, is later invited as an Admin to a real Workspace, and from then on administers two Workspaces of which one is empty. Opening on the empty one strands them: no Agents, and a selector they have never seen before appearing at the same moment.

### Default navigation

The default management navigation is:

- Agents
- Tasks
- Skills
- Integrations

There is no Settings tab. The navigation holds objects; a container is not an object. An item earns a slot when it either carries live state that can fail, or is authored and revised regularly:

| | Live state that can fail | Authored and revised |
| --- | --- | --- |
| Agents | Yes | Yes |
| Tasks | Yes | No, they are output |
| Integrations | Yes, connections expire and fail | Rarely |
| Skills | No | Yes |
| Admins | No | Rarely |
| Account | No | Almost never |

- **Skills** is authored, reusable content and carries a creation action of its own. It is what makes an Agent good at a particular kind of work, so it stays in the navigation rather than being filed under configuration.
- **Integrations** carries operational state. "Why can the Agent not read that page" is the same failure class as "the Computer is offline", so Integrations is diagnosed, not merely configured once.
- **Computers** belongs inside the Agents area rather than in its own entry. Its state is already expressed there, on the Agent row that reports an offline Computer together with its recovery action, and a Computer is meaningful only through the Agents bound to it. A separate top-level entry would be a second view of the same information. Explicit enrollment is a low-frequency write and does not determine where the object lives.
- **Admins, Workspace management, and Account** live in the account menu. Workspace management stays contextual and hidden in the single-Workspace case. A permanent Settings tab would contradict that: this document deliberately keeps Workspace out of the daily field of view, and a tab that exists to hold it puts it back.

Security sensitivity does not determine navigation placement; it determines confirmation friction and audit requirements. Granting an Admin is the most security-sensitive action in the product and still belongs in the account menu.

## Future Web User evolution

Ordinary users are likely to need a Web experience over time, but the expected sequence is to view Task progress and rich results, search personal history, follow deep links from IM, continue an existing Chat, and only later start a new Web Chat. Viewing and continuing extend OpenTag's IM-first position; making new Web Chat the primary experience would expand OpenTag toward a general AI chat product and requires separate product validation.

A Web User is not a Workspace Admin and must not bring back the generic `member` role. Future authority remains orthogonal:

```text
Account
├── Workspace Admin grant ──> manage the complete Workspace
├── Agent use grant ────────> invoke one permitted Agent
└── Chat participation ─────> view and continue one Chat

IM provider identity ── optional identity link ──> Account
```

An IM User can continue using an Agent without an Account. When that person follows an IM link into the Web, they may sign in and link the provider identity to an Account. The Account then sees only Chats in which it is a participant and can start a new Chat only with an Agent covered by an Agent-use grant.

Workspace-wide administration and Chat-content visibility must also diverge before non-Admin Web access ships. Admins may inspect all Task status, usage, failures, and operational diagnostics, but should not automatically read every participant's private Chat content. Chat content is participant-visible by default; any future compliance or audit reader is a separate explicit capability.

The current persistence model is IM-specific: Sessions require an IM binding and channel/thread identity, and message/delivery records use IM terminology. Web Chat creation therefore requires a transport-neutral Conversation implementation rather than inserting Web messages into IM tables:

```text
Chat
├── Workspace
├── Agent
├── Participants
├── Messages
└── Endpoint
    ├── IM channel or thread
    └── Web
```

IM and Web become adapters to the same Chat module. If broad Web adoption later makes per-Agent grants repetitive, OpenTag may add a narrow Workspace participant grant meaning only "may use published Agents". It remains separate from Workspace Admin access and does not introduce a role hierarchy.

This issue does not implement Web Users or Web Chat. It preserves the evolution path by keeping `users` as general Accounts, isolating Admin authority in `workspace_admin_grants`, avoiding Admin-only assumptions in Account naming, and not equating Agent management with Chat-content access.

## Target architecture

```text
Account ── Workspace Admin grant ──> Workspace
                                         ├── Agents
                                         ├── Tasks
                                         ├── Integrations
                                         └── Skills

Computer credential ── enrollment ──> Workspace

IM provider identity ── access policy ──> Agent
```

### Authentication seams

Human management and machine runtime authentication are separate modules with different interfaces:

```ts
type AdminAuthContext = {
  userId: string;
};

type ComputerAuthContext = {
  computerId: string;
  workspaceId: string;
};
```

Management routes accept only `AdminAuthContext`. Runtime WebSocket, heartbeat, reporting, and assigned-resource access accept only `ComputerAuthContext`. IM ingress continues to use provider-specific verification and never becomes either context.

### Workspace authorization seam

All Workspace resources depend on one deep `WorkspaceAdminAccess` module instead of reading grants or interpreting roles themselves. Its interface is intentionally small:

```ts
requireAdmin(userId, workspaceId): Promise<void>;
listAdmins(callerUserId, workspaceId): Promise<AdminSummary[]>;
acceptInvitation(userId, token): Promise<WorkspaceSummary>;
revokeAdmin(callerUserId, workspaceId, targetUserId): Promise<void>;
```

The implementation owns Workspace locking, invitation consumption, cross-Workspace non-disclosure, and the last-Admin invariant.

## Target data model

```text
users
auth_identities

workspaces
workspace_admin_grants
admin_invitations

computers
workspace_computers

agents
tasks
integrations
```

`workspace_admin_grants` has no role. It records the grant and revocation lifecycle so access changes remain auditable:

```text
workspace_id
user_id
granted_by
granted_at
revoked_by
revoked_at
```

`agents.manager_user_id` becomes audit-only `created_by_user_id`. Authorization depends on `agents.workspace_id`, never on the creator. The foreign key requiring the creator to own the bound Computer is removed.

Computers authenticate independently and are attached through `workspace_computers`. An Agent can bind only to a Computer enrolled in the same Workspace. Enrollment records retain who performed the action without turning that person into the Computer's authorization subject.

Provider-native terms such as Slack Team ID remain unchanged. They identify external provider resources and are not the removed OpenTag Team concept.

## Current implementation hotspots

The change crosses every product tier and must not be treated as a Web-only simplification:

- `packages/shared/src/auth.ts`, `team.ts`, `invitation.ts`, `agent.ts`, and `http-paths.ts` define Team membership, role, invitation, Agent identity, and route contracts.
- `packages/server/src/db/schema/auth.ts`, `agents.ts`, `computers.ts`, and `invitations.ts` persist Team scope, roleful Memberships, creator/Computer ownership constraints, and reusable invitations.
- `packages/server/src/services/teams`, `services/invitations`, `services/agents`, and IM binding modules read Memberships or roles directly and must converge on the Workspace authorization seam.
- `packages/server/src/services/auth/connect-code-service.ts` and `packages/server/src/api/runtime.ts` currently couple daemon authentication to human user credentials and must be split before broad Admin authority is safe.
- `apps/web/src/router.tsx` and `apps/web/src/api.ts` expose login gates, Team selection, Workspace settings, member roles, and invitations.
- `apps/cli/src/commands/team`, `apps/cli/src/core/team`, and `apps/cli/src/core/selection/team.ts` expose the old Team management and selection model.
- migrations, integration tests, README/DEVELOPMENT documentation, and Chinese mirrors encode the same vocabulary and invariants.

Historical migrations remain immutable. Add forward migrations and update deployed-history upgrade tests rather than rewriting existing SQL or snapshots.

## Interface changes

`GET /api/v1/me` returns the authenticated Account and its administered Workspaces, without membership roles:

```ts
{
  user: UserProfile;
  workspaces: WorkspaceSummary[];
}
```

Workspace resource paths remain explicit even when the Web hides the concept:

```text
GET    /api/v1/workspaces/:workspaceId/agents
POST   /api/v1/workspaces/:workspaceId/agents
GET    /api/v1/workspaces/:workspaceId/admins
DELETE /api/v1/workspaces/:workspaceId/admins/:userId
POST   /api/v1/workspaces/:workspaceId/admin-invitations
GET    /api/v1/admin-invitations/:token/preview
POST   /api/v1/admin-invitations/:token/accept
```

Reusable member links, role mutation, membership restore, and Team leave interfaces are removed. The CLI removes `team` commands, adds `admin` management commands, and requires a Workspace selector only when multiple Workspaces are available. Computer enrollment should use `computer connect` terminology rather than implying that the daemon performs a human login.

### 1. Separate Computer authentication

Introduce machine-scoped Computer credentials and authenticate Runtime connections independently from human Admin sessions. Ensure Computer tokens cannot call any management route.

### 2. Introduce the Workspace authorization module

Create `WorkspaceAdminAccess` behind the current storage model, replace direct membership and role reads across Agent, IM binding, Computer, setup, and connect-code paths, and add concurrency tests for grant and revocation invariants.

### 3. Migrate storage and resource ownership

Rename Teams to Workspaces, replace memberships with roleless Admin grants, convert invitations to single-use Admin invitations, change Agent creator fields to audit metadata, and add explicit Computer enrollment.

Give the Workspace listing query a deterministic `ORDER BY` in this step rather than deferring it to step 4. The current `/me` membership query has none, so the Web client's "first record" fallback depends on an ordering PostgreSQL never promised. This step is what rewrites those rows, which is exactly when the physical order can change, while the `/me` projection is not replaced until step 4. A Workspace mis-selected in that window would present as a refactor regression.

### 4. Replace shared contracts and server routes

Replace Team and membership schemas, `/me` projections, paths, errors, and public package exports in one breaking pre-alpha change. Do not maintain long-lived Team and Workspace dual writes.

### 5. Update Web and CLI

Remove Team, member, role, and default Workspace UI; add Settings/Admins; keep Account and sign-out controls for Admins; reveal Workspace selection only for multiple scopes; and update CLI commands and output.

### 6. Migrate existing data and remove compatibility code

Preserve Workspace and Agent identifiers. Convert only active `admin` memberships to Admin grants; do not promote existing `member` records. Revoke old member invitations. Validate cross-Workspace isolation and remove obsolete Team modules after the migration has completed.

## Non-goals

- Anonymous or accountless Admins
- Owner, Manager, Member, Viewer, or per-Agent roles
- Creator-owned Agents
- Automatic OpenTag Accounts for IM Users
- A Workspace selector in the single-Workspace experience
- A Settings tab in the default navigation
- Rebinding an existing Agent to a different Computer
- Revoking a Computer enrollment
- Computer credentials with management authority
- Mechanical renaming of provider-native Team identifiers
- Implementing Web User access, identity linking, Agent-use grants, or transport-neutral Chat in this issue

## Open questions

- **What are the security controls?** An earlier revision of this document filed "security controls" under a Settings tab without defining them, and none exist in the product today. Name the actual operations first, for example revoking a Computer credential, rotating a token, or ending a session, and then decide where they belong. An undefined entry is what makes a container tab look necessary.

## Acceptance criteria

- An IM User can use an allowed Agent without an OpenTag Account.
- Admin A can manage an Agent created by Admin B in the same Workspace.
- `createdBy` never affects authorization.
- An Admin in Workspace A cannot discover or access Workspace B resources.
- The last active Admin cannot revoke their own grant, including under concurrent requests.
- Admin invitations are authenticated, single-use, short-lived, and grant complete authority only after explicit acceptance.
- A Computer token cannot invoke any management interface.
- An Agent can run only on a Computer enrolled in the same Workspace.
- Existing members are not promoted during migration.
- The default Web shell contains no Team terminology and no Workspace selector for a single Workspace.
- The default Web shell contains no Settings tab; Skills and Integrations are reachable from the navigation and Computers from the Agents area.
- An Account administering multiple Workspaces opens the shell on the Workspace named by the selection rule, and `GET /api/v1/me` returns Workspaces in a defined order.
- An Account whose auto-created default Workspace was never used is not gated into Agent setup for it while another administered Workspace has completed setup.
- An Account holding no Admin grant reaches an explanatory state naming who can restore access, and is never handed a new empty Workspace instead.

baixiaohang:
## Summary

Implements the accepted Issue #125 architecture as one atomic monorepo change:

- isolates enrollment-scoped machine credentials from Account administration;
- centralizes Workspace Admin authorization and cross-Workspace non-disclosure;
- replaces Team membership authority with roleless Workspace Admin grants and single-use invitations;
- cuts shared/server/Web/CLI contracts over from Team to Workspace without a long-lived alias;
- preserves Workspace, Agent, Session, and physical Computer IDs while migrating runtime placement to Workspace Computer enrollments.

Closes #125.

## Logical phases

- [x] Phase A — machine authority, Computer enrollment, runtime fencing, and daemon credential separation
- [x] Phase B — `WorkspaceAdminAccess`, equal Admin authority, deterministic legacy `/me`, and last-Admin invariants
- [x] Phase C — final storage/contracts/routes/Web/CLI/docs/E2E Workspace cutover

The three phases are review checkpoints inside this single atomic PR. No intermediate commit is deployable or intended for partial merge.

## Security boundary

- Account access and refresh tokens cannot authenticate Runtime WebSocket or machine-only resources.
- Machine credentials are scoped to one `workspace_computers` enrollment and cannot access management routes.
- Runtime registry, placement, heartbeat, custody, delivery, and readiness route by `workspaceComputerId`.
- Workspace management decisions and Admin grant/invitation lifecycle flow through one `WorkspaceAdminAccess` seam and conceal foreign Workspace/resource existence.
- Agent creators and legacy physical Computer owners are audit fields, not authorization principals.
- Public management contracts select Computers by physical `computerId`; enrollment IDs remain internal to machine/runtime routing.
- Admin invitation tokens are single-use, short-lived, hash-only at rest, and redacted from logs and traces.
- Agent-to-Computer binding is immutable after creation, and this change adds no Computer enrollment revocation surface.

## Migration and deployment

- Relative to `main`, this PR adds exactly one migration: `0016_certain_revanche.sql`, its snapshot, and journal entry.
- Existing migrations `0000`–`0015` are unchanged.
- Existing Workspace/Team, Agent, Session, and physical Computer UUIDs are preserved.
- Legacy active Admin memberships migrate to grants; legacy members are not promoted.
- Old reusable invitations/connect codes are invalidated and no machine secret or online presence is synthesized.
- Shared contracts, Server migration/routes, Web, and CLI must ship in one release. Rollback is application rollback plus database restore; the forward cutover is not dual-written and does not retain `/teams` aliases.

## Validation

Local validation on the exact committed tree:

- `pnpm check`
- `pnpm typecheck`
- `pnpm test`
- `pnpm build`
- Server PostgreSQL integration: 6 files / 142 tests
- Agent Runtime coverage: 100% statements, branches, functions, and lines
- source CLI tarball smoke
- real PostgreSQL + Server + Web + Chromium + CLI daemon onboarding E2E: 16/16 steps
- deployed-history migration upgrade and idempotent rerun
- final full-diff static review with no blockers

This PR should become Ready only after the CI run for the final head is fully green.


yuezengwu (changes requested):
Requesting changes on exact head bdbb6e52eb1f479208e3fb44a48d6f1eca127460.

1. [P1] Agent creation still validates computerId through the legacy physical Computer owner at packages/server/src/services/agents/agent-service.ts:297-301, instead of resolving an active workspace_computers enrollment for the requested Workspace. An admin of Workspace B can therefore create an Agent on a Computer they own that is enrolled only in Workspace A. The insert succeeds, but the new readiness/routing paths require an active Workspace B enrollment, leaving a persisted Agent that cannot run. The inverse also fails: another admin of the same Workspace cannot use its active enrollment. Please resolve and lock the active (teamId, computerId) enrollment in the same mutation transaction, and cover both cross-Workspace rejection and same-Workspace multi-admin use.

2. The credential cutover is not complete across required validation/release paths. The shared Auth v2 test still sends accessToken instead of machineToken, CLI pack smoke still requires login --no-start, and container smoke still invokes that removed option. These are deterministic failures on this head, and all six PR checks are red. Please update the protocol fixture and both packaged/container smoke contracts, then rerun the full required CI matrix.

Validation: Context Tree master d4bb6ba3dd69525af73caf2bd6a5145492b230ab verified clean; pnpm build passes; git diff --check passes; @opentag/shared reproduces 1 failure out of 76; @opentag/client passes 460/460.

Gandy2025:
Heads-up on a collision this branch will hit at the next merge from `main`, unrelated to the review of Phase A itself.

`main` took `0015_spotty_machine_man` in #159/#160 while this branch was open, so both sides now claim `idx: 15`:

```
main   idx 15  0015_spotty_machine_man
#161   idx 15  0015_red_sway
#161   idx 16  0016_thin_flatman
```

`packages/server/drizzle/meta/_journal.json` will conflict on that entry, and the two migrations added here need to move to `0016`/`0017` with their `idx` renumbered to match. Their `when` timestamps should stay as they are, since reordering them against `0015_spotty_machine_man` would misrepresent when each was authored.

Nothing on `main` needs to move: this issue's own rule that historical migrations stay immutable applies to what has already shipped, and these two have not.

Worth doing before Phases B and C add more, since each further migration makes the renumber wider.

packages/server/drizzle/0016_certain_revanche.sql
```diff
@@ -0,0 +1,398 @@
...
+CREATE INDEX "workspace_computers_computer_id_idx" ON "workspace_computers" USING btree ("computer_id");
+--> statement-breakpoint
+ALTER TABLE "agents" DROP CONSTRAINT "agents_manager_computer_owner_fk";
+--> statement-breakpoint
+DO $$
+BEGIN
+	IF EXISTS (
+		SELECT 1 FROM "teams"
+		WHERE NOT EXISTS (
+			SELECT 1 FROM "memberships"
```
yuezengwu: [P1] This precondition treats a suspended Account as an active Admin because it only checks legacy membership role/status. The backfill at lines 185-188 creates the grant for that Account, and the final validation likewise only checks `revoked_at`. Runtime authority consistently joins `users` with `suspended_at IS NULL`, so a Team whose only active-admin membership belongs to a suspended user passes cutover and becomes a Workspace with zero usable Admins. Please apply the live-account condition to the precondition, grant backfill, and final validation (or fail the migration atomically), and add deployed-history coverage for this case.

yuezengwu (changes requested):
Reviewed exact head `f88124f2b0c10de09fcdb873354e974c17b664f8`. The earlier exact-enrollment Agent binding and runtime/packaging CI blockers are resolved, and all current GitHub checks are green. The cutover still has one authority blocker: a suspended Account is counted as an active Workspace Admin, so some legacy Teams can migrate into Workspaces with zero usable administrators. See the inline finding.

Validation: Context Tree master `d4bb6ba3dd69525af73caf2bd6a5145492b230ab` verified clean; `pnpm check`, `pnpm typecheck`, `pnpm build`, `git diff --check`, and server integration tests (142/142) passed. The full unit run had one observability timing failure that passed immediately in an isolated rerun (23/23); all six required CI checks are green.

baixiaohang:
Resolved the current conflict and the suspended-Account migration finding in head `faedbd94b5dc30c30409c7cb13ac1e871399660c`.

- Merged `origin/main` at `289bf73e90df4d1ed7ca8da1ab8f10cec1476087`, preserving the new stable Agent-list ordering alongside the Workspace cutover.
- The cutover now treats only non-suspended Accounts as usable legacy Admins in its precondition, grant backfill, and final validation.
- Added deployed-history coverage for suspended-only Teams (atomic rejection) and mixed live/suspended Admins (only the live Admin is granted).
- The PR remains a one-migration change relative to `main`.

Focused migration and Web tests, repository checks, and typechecking passed locally. All six checks on the new head are green, and the PR is mergeable.

PR #162, opened 2026-08-25
fix(web): give every Agent failure state an exit

Gandy2025:
## Summary

Every Agent failure state now leads somewhere.

**Agent list.** `agentCardStatus` collapsed four distinct failures under one `Needs attention` label but carried a `reconnect` boolean hardcoded to `computer_offline`. The other reasons rendered as text with nothing to click:

| State | Before | Now |
| --- | --- | --- |
| `action_required` · `computer_offline` | `Reconnect` | `View Computer` → `/settings/computer` |
| `action_required` · `runtime_unavailable` | no action | `View runtime` → `/settings/execution` |
| `action_required` · other | no action | `View messaging` → `/settings/messaging` |
| `not_connected` | no action | `Connect messaging` → `/settings/messaging` |

The label moved from `Reconnect` to `View Computer` because the destination does not reconnect anything, and the Computer may belong to someone else.

**Computer recovery page.** This was the dead end the list pointed at. It rendered an `h1` of "Connected computer" above a status of `Offline`, stated the problem without naming anyone who could act on it, and offered a `Check again` button that duplicated the 30-second revalidation `AgentDetailPage` already runs (`revalidateMs: 30_000`, `refreshOnFocus: true`) and rendered no pending state, so clicking it looked like nothing happened.

Before:

```
Connected computer                      ● Offline
The computer assigned to run Reviewer.
Ada's Mac · macOS
Last seen Aug 21, 2026, 12:56 PM
New requests can start after this Computer reconnects.
[ Check again ]
```

After:

```
Ada's Mac · macOS                       ● Offline
Reviewer runs only on this Computer.

Last seen 4 days ago · Aug 21, 2026, 12:56 PM
Ask HuanCheng Bai to bring it back online.
```

Four changes, no net increase in copy:

- The heading is the machine, so it can no longer contradict the status beside it.
- The subtitle carries the immutable assignment in every state, so the absent "move it" control reads as a rule rather than an omission.
- The recovery line names who can act. When the viewer owns the Computer it reads `Open OpenTag on Ada's Mac to bring it back online.` instead.
- `Check again` is gone. The page already rechecks; saying so in a sentence would have been page metadata, not information about the problem.

Ownership uses `agent.manager`, which `agents_manager_computer_owner_fk` currently guarantees is the owner of the bound Computer. #125 removes that foreign key; a code comment marks the spot so the contact can be read from the Workspace Computer enrollment once enrollments carry an operator.

[...]

PR #163, opened 2026-08-25
fix(web): open an Agent from the whole row

Gandy2025:
## Why

The Agents list gave a 958x88 row exactly one navigable target: a 36x36 `<a>` at the far right edge. That is below the 44px touch minimum, and it is nowhere near the Agent name — the thing a viewer actually points at. The other 95% of the row was dead. On mobile the row stacks to a ~200px block with the same 36px arrow floating at the top right, and no hover to hint at it.

Two smaller problems on the same screen, both surfaced by the same review:

- The token count rendered at 20px/500 in Bricolage Grotesque while the Agent name sat at 14px/600. The eye landed on `269.6K` before it landed on whose Agent that was — and the token count is the least actionable number in the row.
- The list revalidates every 30 seconds behind a status-first sort, so rows move on their own. Harmless when only a chevron is clickable; not harmless once the whole row is a link.

## What changed

**The Agent name carries the row link**, stretched over the row with `.agent-card-open::after { inset: 0 }`. The chevron becomes `aria-hidden` decoration.

A stretched link rather than a click handler on the row: `cmd`/middle-click, right-click → "Open in new tab", the status-bar URL preview, and keyboard focus all work natively. A JS handler has to reimplement each one.

No nested anchors — the stretched `::after` belongs to the name link, and #162's inline failure exit is a **sibling** anchor. It already had `position: relative`, and now carries an explicit `z-index: 1` rather than relying on DOM order.

**Display order is held stable while the list is mounted** (`apps/web/src/features/agent-list-order.ts`). Without this, a status change during a background refresh can slide a row under the pointer between press and release, sending the viewer to a different Agent than the one they pressed. Rows keep the position they were first shown in; an Agent that starts failing turns amber where it already sits — more noticeable than jumping to the top unobserved. Agents created since the first render join at the end. Order is recomputed on the next visit.

**Usage drops from display size to a supporting size** — `.agent-card-usage dd` goes from `--text-section`/`--font-display` to `--text-ui`/`--font-sans`. The Agent name is deliberately left at `--text-body`: once the number stops shouting, 14px/600 already wins, and enlarging every name turns an 8-row list into a stack of headings.

`tabular-nums` and the two-column usage alignment are untouched — they are what make the numbers comparable across rows.

## Verification

jsdom cannot prove a stretched link actually covers the row, or that `z-index` puts the failure exit back on top. Both were measured in a real browser against the real stylesheet, with `elementFromPoint`:

- **1440x900** (row 1390x88) — far-left edge, the name, the empty middle, the Tokens figure, the chevron, and the bottom-right corner all resolve to the row link.
- **375x812** (block 325x214) — four points across the stacked block all resolve to the row link.
- At both sizes, the point over `Connect messaging` resolves to the failure exit, not the row link.

Test coverage: row link wraps the visible name and points at the Agent; the chevron is no longer a link; the failure exit is a sibling of the row link, never nested; the stretch and stacking rules are asserted against the parsed stylesheet; and the order function is covered for first render, resort, insertion, removal, and stability under reapplication.

`pnpm check`, `pnpm typecheck`, `pnpm build` all clean. Web suite 338 passing.

Two failures on this branch are pre-existing and reproduce on clean `origin/main` (verified in a detached worktree): `@opentag/server` `observability.test.ts`, and two `open-tag` daemon-service tests ("…before a stale PATH shim").

## Notes for review

Rebased onto `main` after #162 merged as `54db7f4`; the rebase was conflict-free.

A stretched link makes row text non-selectable by drag. That is the standard trade-off of the pattern and is accepted here for a list row.

Three further items from the review were considered and dropped: group headings, status filtering, and surfacing `usage.failed`. Grouping would make a status change move a row between bordered blocks, which is more visible than the reorder it was meant to fix; `usage.failed` has no baseline, which was the exact complaint levelled at the token count. Cost, last-active time, and period-over-period deltas are not in the API — `AgentListItem` carries only `{ windowDays, tasks, failed, tokens }` and `activity: idle | working{startedAt}` — so they would need shared-schema and server work.

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -729,6 +729,27 @@ describe("OpenTag Web App Shell", () => {
     expect(exit.classList.contains("ds-button--outline")).toBe(false);
   });
 
+  it("opens the Agent from the row rather than from a 36px chevron", async () => {
+    installApi("admin");
+    render(<App />);
+
+    const open = await screen.findByRole("link", { name: "Open Reviewer" });
+    expect(open.textContent).toBe("Reviewer");
+    expect(open.getAttribute("href")).toBe(`/agents/${agentId}`);
+    const card = open.closest(".agent-card");
+    expect(card).toBeTruthy();
+    expect((card as HTMLElement).querySelector(".agent-card-action")?.tagName).toBe("SPAN");
+    /*
+     * The failure exit is a second link inside the same row. It has to stay a sibling of the row
+     * link rather than a child of it: nesting would be invalid, and wrapping the row in one anchor
+     * is the shortcut that would produce it.
+     */
+    const exit = within(card as HTMLElement).getByRole("link", { name: "Connect messaging" });
+    expect(open.contains(exit)).toBe(false);
+    expect(exit.getAttribute("href")).toBe(`/agents/${agentId}/settings/messaging`);
+    expect(within(card as HTMLElement).getAllByRole("link")).toEqual([open, exit]);
+  });
+
   it.each(["/", "/agents"])("redirects unauthenticated protected path %s to login", async (path) => {
     installApi("member", { unauthenticated: true });
     window.history.replaceState({}, "", path);
--- a/apps/web/src/__tests__/page-layout.test.ts
+++ b/apps/web/src/__tests__/page-layout.test.ts
@@ -62,6 +62,19 @@ describe("workspace page layout", () => {
     expect(declarationValue(appStyles, ".settings-members-section .settings-invitation-panel", "width")).toBe("100%");
   });
 
+  it("stretches the Agent row link over the whole row without covering its inline action", () => {
+    expect(topLevelDeclarationValue(appStyles, ".agent-card", "position")).toBe("relative");
+    expect(declarationValue(appStyles, ".agent-card-open::after", "position")).toBe("absolute");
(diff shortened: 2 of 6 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The Agent row now provides native full-area navigation while preserving the separate recovery action, a stable pointer target during background refreshes, and a clearer visual hierarchy.

**Risk level**: C
- Path baseline: `apps/web/**` only -> C
- Semantic lift: none; this is confined to Agent-list rendering, display order, and styling, without subscription, WebSocket-frame, or shared data-contract changes.

**PR summary**
- Problem: People currently have to find a 36px chevron to inspect an Agent, leaving most of each desktop or mobile row inert; once the row becomes clickable, background status sorting could also move a different Agent under the pointer.
- Approach: The visible Agent name becomes a native link stretched across the card, the inline recovery link remains a separately stacked sibling, row positions remain stable while the list is mounted, and usage typography is reduced to supporting emphasis.
- Impacted modules: Web Agent-list rendering, Agent-card layout and styling, and focused rendering/order/layout tests.

**Review findings**
- ✅ The stretched target keeps native link behavior for keyboard and alternate navigation while the recovery link remains valid, independently reachable markup. [`apps/web/src/router.tsx`, `apps/web/src/styles.css`]
- ✅ The ordering helper preserves existing positions, appends new Agents deterministically, removes departed Agents, and is covered across those transitions. [`apps/web/src/features/agent-list-order.ts`]

**Action taken**
- Approved the current head.

PR #164, opened 2026-08-25
fix(web): make a Workspace readable by the Admin who has one

Gandy2025:
## Summary

Follow-up to #161 on one rule from #125: that the Workspace concept should stay out of the single-Workspace path.

`WorkspacePage` redirected to `/admins` whenever an Account administered one Workspace, and the account menu showed the Workspace entry only above two. The concept returned regardless. The Admin list has to be reachable — otherwise a solo Admin can never invite a colleague — so it became its own page, and that page then had to open with "Every Admin has complete management authority over **this Workspace**" to explain itself. The word was hidden from the menu and reintroduced in the page.

Hiding also withheld two facts other people could already see:

- **The Workspace name is the first line of every invitation.** An auto-created Workspace is named after whoever signed up, so a colleague opens the invite and reads "Join hao xiong's Workspace" — which is usually the company's Workspace. The person who sent it could not read or change that name.
- **The CLI identifier is required by anyone running a CLI command**, and lived only on the hidden page.

The concept-reduction argument in #125 is about IM Users: they should not have to register, join anything, or learn what a Workspace is, and that is settled by giving them no Account and no Workspace membership at all. An Admin is a different person — someone who deliberately signed in to a management surface to run Agents on machines they operate, already holding Agent, Computer, Runtime, Provider, Skill, Integration, and IM binding. Workspace is the container the other seven live in.

So this narrows the rule to what it was protecting:

**Only the selector stays conditional.** Choosing between scopes is meaningless with one scope, so a single-Workspace Account still sees no `Workspaces` group in the account menu. Reading and changing your own Workspace is not selection.

**`/workspace` is one surface at any Workspace count**, in this order:

1. **Admins** — the list, `Invite Admin`, `Revoke`, and the single-use invitation panel
2. **Workspace profile** — name and CLI identifier
3. **Separate scope** — `Create Workspace`, which also moves out of the page action slot; it creates a *different* Workspace and does not belong to the page that manages this one

Being an Admin is the product's only access decision, and a Workspace is the scope its Admins share, so who may administer it belongs with what it is called. That is one list, not two pages. `/admins` redirects so existing links keep working.

No behaviour changes for an Account administering more than one Workspace, beyond the two entries becoming one.

## Testing

- `pnpm check` passes
- `pnpm build` passes
- `pnpm typecheck` passes
- `pnpm --filter @opentag/web test` passes: 304 tests, 17 files

Three tests asserted the behaviour this reverses and were rewritten to assert the new rule; one covers the `/admins` redirect. `pnpm test` cannot pass cleanly on `main` right now — `@opentag/server`'s `observability.test.ts > traces real RuntimeSession failure, overload, and dropped-queue terminal paths`, and two `open-tag` daemon-service tests ending "…before a stale PATH shim", fail on a clean tree. None are touched here.

## Breaking changes

None. Web presentation and client-side routing only; no contract, schema, or route changes on the server.

## Checklist

- [ ] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. — all pass except the three pre-existing failures described above.
- [x] Canonical English docs and Chinese mirrors are synchronized when applicable. — no documentation changes required.

Refs #125.

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -1195,14 +1195,26 @@ describe("OpenTag Web App Shell", () => {
     expect(screen.getByText("Ada")).toBeTruthy();
   });
 
-  it("hides Workspace management in the single-Workspace shell", async () => {
+  it("keeps Workspace identity and its Admins reachable in the single-Workspace shell", async () => {
     installApi();
     window.history.replaceState({}, "", "/workspace");
     render(<App />);
-    expect(await screen.findByRole("heading", { name: "Admins" })).toBeTruthy();
-    expect(window.location.pathname).toBe("/admins");
-    expect(screen.queryByLabelText("Workspace name")).toBeNull();
+    expect(await screen.findByRole("heading", { name: "Workspace" })).toBeTruthy();
+    expect(window.location.pathname).toBe("/workspace");
+    expect(screen.getByRole("heading", { name: "Admins" })).toBeTruthy();
+    expect(screen.getByRole("button", { name: "Invite Admin" })).toBeTruthy();
+    // The name reaches every invitation recipient and the identifier is required by the CLI, so a
+    // solo Admin has to be able to read and change them.
+    expect(screen.getByLabelText("Workspace name")).toBeTruthy();
     expect(screen.queryByRole("button", { name: "Save account profile" })).toBeNull();
+  });
+
+  it("redirects the retired Admins route onto the Workspace surface", async () => {
+    installApi();
+    window.history.replaceState({}, "", "/admins");
+    render(<App />);
+    expect(await screen.findByRole("heading", { name: "Workspace" })).toBeTruthy();
+    expect(window.location.pathname).toBe("/workspace");
   });
 
   it("keeps the Agent home focused on status, current work, and contact", async () => {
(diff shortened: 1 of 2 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The change restores access to Workspace identity and Admin management for a single-Workspace account while keeping only scope selection conditional.

**Risk level**: C
- Path baseline: only `apps/web/**` changes -> C
- Semantic lift: none; this is client-side routing and page composition, with no subscription, data-layer, or WebSocket parsing change.

**PR summary**
- Problem: A solo Admin could invite another Admin but could not inspect or correct the Workspace name shown in that invitation, nor read the CLI identifier required to address the Workspace.
- Approach: Consolidate Admin management, Workspace identity, and creation of a separate scope on `/workspace`; keep the multi-Workspace selector conditional and redirect the retired `/admins` route.
- Impacted modules: `apps/web/src/router.tsx`, `apps/web/src/__tests__/app.test.tsx`

**Review findings**
- ✅ The unified page preserves the existing Admin invitation/revocation behavior and makes the current Workspace identity reachable at every Workspace count.
- ✅ Regression coverage includes the single-Workspace route, legacy redirect, account-menu disclosure, invitation flow, and keyboard focus order.

**Action taken**
- Approved exact head `632bf8b29d4e86c4f75c71fce975b8d939397167`.

PR #166, opened 2026-08-26
fix(web): let an offline Computer be reconnected from its Agent

Gandy2025:
## Summary

An Agent whose Computer is offline had no way to recover from the product. This gives the page the product already routes operators to the action it was missing.

The Connected computer row in Agent settings is inert while the Computer is healthy and becomes a link — labelled `Review` — only when it needs attention. That is a deliberate repair entry point. The page behind it rendered a status, a last-seen line, and one sentence of advice, with nothing to act on. Meanwhile the only surface that issues a connect command lives at `/agents/computers`, and nothing in the app links there.

## What changed

- The recovery block offers **Reconnect this Computer**, which expands `ComputerSetup` in place instead of navigating away. The journey ends where it started: the Agent's status flips to Online as soon as its Computer registers, via the `onAgentChanged` refresh the settings page already owns.
- The action is shown only when the Computer itself is unreachable. A reachable Computer with a missing Provider reports `ready` for the Computer dependency, so that branch keeps its install guidance rather than suggesting a re-enrolment that cannot help.
- `ComputerSetup` accepts an optional `target`. When present it names the machine in its heading, prompt, waiting status and success line, and waits for that machine specifically. Every existing caller omits `target` and is unchanged.

## Scope

No schema, contract, or server change. `ComputersPage` is untouched.

## Tests

Six tests, each watched failing first:

- `computer-setup.test.tsx` — target naming, targeted success, and that an unrelated reconnection leaves a targeted recovery waiting
- `app.test.tsx` — action present when offline, absent when the Provider is the blocker, generator opens in place without navigation

The guard for the Provider branch was verified to discriminate: removing the condition makes it fail.

```
pnpm check      pass
pnpm typecheck  pass
@opentag/web    310/310
```

Two suites fail on this branch independently of this change, confirmed by running the same suites with these edits stashed: one flaky observability test in `packages/server`, and two `apps/cli` daemon-service tests. Neither is touched here.

## Review round 2

@baixiaohang and @yuezengwu both found the same defect: the first revision inferred connect-code consumption from a workspace-wide connection delta. `connectedAt` moves on every ordinary Runtime registration and the Computers response carries no correlation to the issued code, so an unrelated daemon restart inside the fifteen-minute window falsely ended a recovery whose command was still valid.

Fixed in 31f8780. A targeted recovery now waits only for its own Computer; the substitution branch is gone. Detail on why a foreign registration is ignored rather than described is in the review reply below.

## Note for reviewers

Run web tests through `pnpm turbo run test`. Invoking `vitest` directly skips the dependency build and resolves a stale `packages/shared/dist`, which fails 95 of 104 tests with "The server returned an invalid response".

### DIFF
--- a/apps/web/src/computer-setup.tsx
+++ b/apps/web/src/computer-setup.tsx
@@ -11,6 +11,8 @@ const COPY_FALLBACK_HINT = "Copying is unavailable here. The command is selected
 export interface ComputerSetupProps {
   workspaceId: string;
   onConnected?: (computer: WorkspaceComputerSummary) => void;
+  /** Scopes the panel to one enrollment so a recovery flow names the Computer it came from. */
+  target?: { computerId: string; displayName: string };
 }
 
 function errorMessage(cause: unknown, fallback: string): string {
@@ -33,11 +35,20 @@ function formatRemaining(remainingMs: number): string {
   return `${Math.floor(seconds / 60)}:${String(seconds % 60).padStart(2, "0")}`;
 }
 
-export function ComputerSetup({ workspaceId, onConnected }: ComputerSetupProps) {
-  return <ComputerSetupLifecycle key={workspaceId} workspaceId={workspaceId} onConnected={onConnected} />;
+export function ComputerSetup({ workspaceId, onConnected, target }: ComputerSetupProps) {
+  return (
+    <ComputerSetupLifecycle
+      key={`${workspaceId}:${target?.computerId ?? ""}`}
+      workspaceId={workspaceId}
+      onConnected={onConnected}
+      target={target}
+    />
+  );
 }
 
-function ComputerSetupLifecycle({ workspaceId, onConnected }: ComputerSetupProps) {
+function ComputerSetupLifecycle({ workspaceId, onConnected, target }: ComputerSetupProps) {
+  const targetComputerId = target?.computerId;
+  const targetName = target?.displayName;
   const [bootstrapCommand, setBootstrapCommand] = useState<string>();
   const [error, setError] = useState<string>();
   const [waitingForComputer, setWaitingForComputer] = useState(false);
@@ -141,7 +152,7 @@ function ComputerSetupLifecycle({ workspaceId, onConnected }: ComputerSetupProps
       void browserApi.computers(workspaceId).then(
         (value) => {
           if (!active || completed || activePollCycle.current !== pollCycle) return;
-          const connected = value.computers.find(
+          const reconnected = value.computers.filter(
             (computer) =>
               computer.connectionStatus === "online" &&
               ((!baseline.has(computer.computerId) && computer.connectedAt !== null) ||
@@ -149,13 +160,26 @@ function ComputerSetupLifecycle({ workspaceId, onConnected }: ComputerSetupProps
                   computer.connectedAt !== null &&
                   baseline.get(computer.computerId) !== computer.connectedAt)),
           );
+          // A targeted command still enrolls whichever Computer runs it, so prefer the target and
+          // report the substitution rather than announcing a recovery that did not happen.
+          const matched = targetComputerId
+            ? reconnected.find((computer) => computer.computerId === targetComputerId)
+            : reconnected[0];
+          const connected = matched ?? reconnected[0];
           if (!connected) return;
           completed = true;
           window.clearInterval(pollTimer);
           window.clearTimeout(expiryTimer);
           setWaitingForComputer(false);
-          setComputerConnected(true);
           setRemainingMs(undefined);
+          if (targetName && !matched) {
+            setComputerConnected(false);
+            setError(
+              `This command was used on ${connected.displayName}, not ${targetName}. Generate a new command and run it on ${targetName}.`,
+            );
+            return;
+          }
+          setComputerConnected(true);
           onConnectedRef.current?.(connected);
         },
         (cause: unknown) => {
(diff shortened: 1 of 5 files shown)


baixiaohang (changes requested):
- Rationale: The targeted recovery entry point is correctly scoped to an offline Computer, but its substitution branch treats any unrelated Computer reconnection as proof that the connect code was consumed and terminates the recovery attempt.

**Risk level**: C
- Path baseline: all changes are under `apps/web/**` -> C
- Semantic lift: none; this reuses the existing browser polling and refresh paths without changing subscriptions, data contracts, or server behavior

**PR summary**
- Problem: An operator who reaches an offline Agent's Connected computer page can see the failure but has no in-product action to restore that assigned Computer.
- Approach: Add an in-place reconnect panel only for Computer reachability failures, keep Provider-readiness failures on their existing guidance, name and prefer the assigned Computer during polling, and refresh the Agent when that Computer reconnects.
- Impacted modules: `apps/web/src/computer-setup.tsx`, Agent settings routing/UI, styles, and Web tests

**Review findings**
❌ 1. Do not infer connect-code consumption from a workspace-wide connection delta. Any non-target Computer whose `connectedAt` differs from the baseline is currently treated as having used this command; polling then stops and the UI asserts that causal fact. However, `connectedAt` changes on every normal Computer WebSocket registration, and the Computers response carries no issued-code correlation. In a multi-Computer Workspace, an unrelated daemon restart during the 15-minute window can therefore falsely terminate the target recovery while its command may still be valid. Please either correlate the issued code with its authoritative consumer, or keep non-target connections non-terminal and describe them without claiming code consumption. [R4 / `apps/web/src/computer-setup.tsx:155`]
✅ 2. The `computerState.state === "action_required"` guard correctly separates Computer reachability from Provider readiness, so re-enrollment is not offered for a Provider install/sign-in failure.

**Action taken**
- Submitted request changes

apps/web/src/computer-setup.tsx
```diff
@@ -141,21 +152,34 @@ function ComputerSetupLifecycle({ workspaceId, onConnected }: ComputerSetupProps
       void browserApi.computers(workspaceId).then(
         (value) => {
           if (!active || completed || activePollCycle.current !== pollCycle) return;
-          const connected = value.computers.find(
+          const reconnected = value.computers.filter(
```
yuezengwu, line 155: This workspace-wide connection delta does not prove which Computer consumed the issued code. ComputerService.register updates connectedAt on every ordinary WebSocket registration, while the Computers response carries no connect-code correlation. If another enrolled daemon restarts during this 15-minute window, reconnected contains that non-target Computer, lines 170-179 stop the target poll, and the UI falsely says this command was used there although it may still be valid. I reproduced that exact counterexample locally. Please correlate the code with its authoritative consumer, or keep non-target connection changes non-terminal and avoid the causal claim.

yuezengwu (changes requested):
Recommendation: request changes\n\nExact head reviewed: a6a21676c3248dfccdfcd665e70a403040286fd7\n\nThe targeted recovery entry point and Provider-readiness guard are sound, but the polling result is not correlated with the issued connect code. A minimal counterexample reproduced the false-terminal behavior: while waiting for the target, an unrelated Computer registration removes the waiting status, emits the substitution error, and clears both timers even though the current command may still be valid.\n\nValidation: pnpm build passed; pnpm typecheck passed; pnpm check passed; @opentag/web 310/310 passed; git diff --check passed; all 6 GitHub checks passed.

Gandy2025:
Both reviews are right, and the finding is the same one: the substitution branch asserted a causal fact the Client cannot observe. Fixed in 31f8780.

**Where it came from.** The untargeted panel has always detected "some Computer newly connected", which is a safe inference when the operator issued a code and handed it to one person. I carried that heuristic into the targeted path, where it stops being safe: `connectedAt` moves on every ordinary Runtime registration, and `ListWorkspaceComputers` carries nothing tying a registration back to an issued code. @yuezengwu's counterexample is exactly the failure — an unrelated daemon restart during the fifteen-minute window ends a recovery whose command is still valid, and tells the operator to generate a new one.

**What changed.** A targeted recovery now waits only for its own Computer. The substitution branch is deleted, along with the error string and the `targetName` dependency in the poll effect. Net −8 lines.

**On the second option offered** — keeping non-target connections non-terminal but describing them — I went with non-terminal and silent, and want to be explicit about why rather than quietly narrowing the ask. The trigger for such a note is any ordinary registration in the Workspace, so in a multi-Computer Workspace it fires far more often as noise than as signal. A hint that is usually wrong trains people to ignore it, and any mention of another machine while the operator waits invites the same causal reading this change removes. An unused command expiring is honest about what the Client can actually see.

Correlating a code with its real consumer needs a Server surface that does not exist — `computer_connect_codes` records `consumed_workspace_computer_id` but nothing exposes it. That belongs with per-machine connect-code targeting, which is tracked separately and out of scope here.

New test replaces the one that asserted the old behaviour: with a target set, an unrelated Computer reconnecting must leave the waiting status intact, raise no alert, keep both timers armed, and not fire `onConnected`. Watched it fail against the previous revision first.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The previous false-terminal inference is removed; targeted recovery now completes only when the assigned Computer itself registers, while unrelated Computer registrations remain non-terminal.

**Risk level**: C
- Path baseline: all changes are under `apps/web/**` -> C
- Semantic lift: none; the change reuses existing polling and Agent refresh paths without changing subscriptions, data contracts, or server behavior

**PR summary**
- Problem: An operator reaching an offline Agent's Connected computer page could see the failure but had no in-product action to restore the assigned Computer.
- Approach: Add an in-place, target-aware reconnect panel only for Computer reachability failures; keep Provider failures on their install/sign-in guidance; refresh the Agent after the assigned Computer reconnects.
- Impacted modules: `apps/web/src/computer-setup.tsx`, Agent settings routing/UI, styles, and Web tests

**Review findings**
✅ 1. The previous blocker is resolved: when `targetComputerId` is present, the poll accepts only that Computer and leaves unrelated registration deltas non-terminal; the replacement regression test checks the waiting state, callback, alert, and timer lifecycle across multiple polls.
✅ 2. The recovery control remains gated on the Computer dependency itself being `action_required`, so a reachable Computer with an unavailable Provider is not sent through re-enrollment.

**Action taken**
- Approved

yuezengwu (approved):
Exact head reviewed: 31f8780074713158cd94721cfdba3eb3bddab59a

The previous blocker is resolved. Targeted polling now accepts only the assigned targetComputerId; unrelated daemon registrations remain non-terminal and no longer produce a causal substitution claim. I reproduced the original foreign-restart case across multiple polls and an additional sequence where an unrelated reconnect precedes the target reconnect: both timers stayed armed after the foreign event, and the target later completed exactly once.

Validation: pnpm turbo run test --filter=@opentag/web --concurrency=2 passed 310/310; an uncommitted combined counterexample passed 311/311 under --force and was removed; pnpm check, pnpm typecheck, pnpm build, and git diff --check passed; all 6 current GitHub checks pass.

PR #168, opened 2026-08-26
docs(deploying): record the Computer re-enrolment the cutover requires

Gandy2025:
## Summary

The cutover in #161 replaced Account-token Runtime authentication with enrollment-scoped machine credentials, and deliberately synthesized no credentials for existing Computers — the migration raises an exception if it created any. That assertion is correct: a credential nobody consented to would be a back door.

Nothing in the deployment surfaces what it costs. The rollout succeeds, `/healthz` passes, the logs are clean, and every Computer goes offline with no discoverable way back.

Three effects hide the cause from whoever is looking:

- Migrated enrollments are inserted without `current_instance_id`, so everything reads Offline regardless of whether the machine is on.
- A daemon on the pre-cutover CLI presents an Account access token and is refused with `AUTH_INVALID_TOKEN` and close code 4401.
- Upgrading the CLI alone still fails, because `computer-credentials.json` does not exist — and the pre-cutover CLI has no `computer connect` subcommand to create it.

The Agent's Connected computer page compounds it, advising the operator to start OpenTag on the machine. Correct for a sleeping laptop; impossible when the credential is what is missing.

## What this adds

A section for deployments that need action outside the rollout, so this class of migration has somewhere to live, with the cutover documented under it:

- what breaks and why the assertion causing it is right
- the per-machine recovery, including why the CLI upgrade and the enrolment must happen in the single generated command
- that re-enrolling with `config/computer.json` intact preserves the `computerId`, the enrollment, and the Agent bindings — and what losing that file costs
- a query listing the Computers still waiting, which is empty once the fleet has recovered

Mirrored in `docs/zh-CN/deploying.md` per the repository convention.

Documentation only. `pnpm check` passes.

## Related

PR #166 gives the Agent's Computer page the recovery action it was missing, so this stops being the only route.

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -132,31 +132,44 @@ The Agent's Connected computer page reports `OpenTag is not running on <name>. S
 back online.` That advice is correct for a Computer that is merely asleep and wrong here; starting the old daemon
 cannot succeed.
 
-#### Recovery
-
-Per machine, and it cannot be done centrally — the credential is written to the target machine's disk.
-
-1. A Workspace Admin opens the Computers page in the web app and generates a connection command. The command is
-   single-use and expires in 15 minutes, so generate it once the person is ready rather than in advance.
-2. That person runs the generated command on the machine. It upgrades the CLI and enrols in one line, which matters
-   because the installed CLI predates the `computer connect` subcommand.
-3. The command writes the machine credential and restarts the daemon service. The Computer reports Online within one
-   registration.
-
-Re-enrolling a machine that still has its `config/computer.json` keeps the same `computerId`, so the existing
-enrollment is reused and its Agents stay bound. Losing that file makes the machine a new one, and its previous
-enrollment stays offline with its Agents attached to it.
-
-#### Confirming the scope
+#### Finding the affected enrollments
 
 ```sql
-select w.name as workspace, wc.display_name, wc.platform, wc.last_seen_at
+select w.name as workspace, wc.display_name, wc.computer_id, wc.platform, wc.last_seen_at
 from workspace_computers wc
 join workspaces w on w.id = wc.workspace_id
 left join workspace_computer_credentials c
        on c.workspace_computer_id = wc.id and c.revoked_at is null
 where wc.revoked_at is null and c.id is null
-order by w.name, wc.display_name;
+order by wc.computer_id, w.name;
 ```
 
-Every row is a Computer that has not been re-enrolled. The list is empty once the fleet has recovered.
+Each row is one enrollment waiting to be recovered, not one host. A `computer_id` appearing on more than one row is a
+host enrolled in several Workspaces, and every one of those rows needs its own pass. The list is empty once the fleet
+has recovered.
+
+#### Recovery
+
+**The unit of recovery is one Workspace enrollment, not one physical machine.** A connect code carries the Workspace it
+was issued for and `computer connect` writes the credential for that Workspace alone, while the daemon runs one
+independent Runtime connection per stored enrollment. Running the procedure once on a host enrolled in several
+Workspaces leaves the other Workspaces' Agents offline on a machine that already looks recovered.
+
+It cannot be done centrally either — the credential is written to the target host's disk.
+
+For each row returned above:
+
+1. A Workspace Admin **of that Workspace** opens the Computers page in the web app and generates a connection command.
+   The command is single-use and expires in 15 minutes, so generate it once the person is ready rather than in advance.
+2. That person runs the generated command on the host. It upgrades the CLI and enrols in one line, which matters
+   because the installed CLI predates the `computer connect` subcommand.
+3. The command writes the machine credential and restarts the daemon service. That enrollment reports Online within
+   one registration.
+
+Repeating this on a host already recovered for another Workspace is safe. `computer-credentials.json` holds one entry
+per enrollment and enrolling replaces only the entry for the Workspace being enrolled, so earlier credentials survive
+and the daemon picks up every stored enrollment when it restarts.
+
+Re-enrolling a host that still has its `config/computer.json` keeps the same `computerId`, so existing enrollments are
+reused and their Agents stay bound. Losing that file makes the host a new Computer, and every previous enrollment stays
+offline with its Agents attached to it.
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -99,3 +99,64 @@ CapRover side afterwards:
 - The App's Deployment tab shows the new image reference and a successful build log.
 - `https://<app>/healthz` returns success.
 - The App logs show the migration and listen lines for the expected revision.
+
+## Deployments that need action outside the rollout
+
+Most revisions need nothing beyond the checks above. A migration that invalidates credentials held outside the
+database is different: the rollout succeeds, `/healthz` passes, the logs look clean, and the Computer fleet goes dark
+anyway. Record each such migration here, because nothing in the deployment surfaces it.
+
+### Workspace and machine authority cutover — `0016_certain_revanche` (#161)
+
+This migration moved Runtime authentication from Account access tokens to enrollment-scoped machine credentials. It
+deliberately creates no credentials for existing Computers, and asserts that it created none:
+
+```sql
+IF EXISTS (SELECT 1 FROM "workspace_computer_credentials") THEN
+    RAISE EXCEPTION 'Workspace cutover must not synthesize machine credentials';
+```
+
+Synthesizing a credential nobody consented to would be a back door, so the assertion is correct. The cost is that
+**every Computer enrolled before the cutover must be enrolled again by hand**, and three separate effects hide the
+reason:
+
+- Every Computer reports Offline. `workspace_computers` rows are inserted without `current_instance_id`, and only a
+  Computer registration sets it.
+- A daemon on the pre-cutover CLI authenticates the Runtime WebSocket with an Account access token and is refused with
+  `AUTH_INVALID_TOKEN` and close code 4401.
+- Upgrading the CLI alone does not recover a Computer. The daemon exits with `This Computer is not enrolled; run
+  computer connect first`, because `computer-credentials.json` does not exist. The pre-cutover CLI has no
+  `computer connect` subcommand at all, so the upgrade is required before the enrolment can be run.
+
+The Agent's Connected computer page reports `OpenTag is not running on <name>. Start it there to bring this Computer
+back online.` That advice is correct for a Computer that is merely asleep and wrong here; starting the old daemon
+cannot succeed.
+
+#### Recovery
+
+Per machine, and it cannot be done centrally — the credential is written to the target machine's disk.
+
+1. A Workspace Admin opens the Computers page in the web app and generates a connection command. The command is
+   single-use and expires in 15 minutes, so generate it once the person is ready rather than in advance.
+2. That person runs the generated command on the machine. It upgrades the CLI and enrols in one line, which matters
+   because the installed CLI predates the `computer connect` subcommand.
+3. The command writes the machine credential and restarts the daemon service. The Computer reports Online within one
+   registration.
+
+Re-enrolling a machine that still has its `config/computer.json` keeps the same `computerId`, so the existing
+enrollment is reused and its Agents stay bound. Losing that file makes the machine a new one, and its previous
+enrollment stays offline with its Agents attached to it.
+
+#### Confirming the scope
+
+```sql
+select w.name as workspace, wc.display_name, wc.platform, wc.last_seen_at
+from workspace_computers wc
+join workspaces w on w.id = wc.workspace_id
+left join workspace_computer_credentials c
+       on c.workspace_computer_id = wc.id and c.revoked_at is null
+where wc.revoked_at is null and c.id is null
+order by w.name, wc.display_name;
+```
+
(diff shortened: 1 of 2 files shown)


baixiaohang (changes requested):
- Rationale: The recovery unit is one Workspace enrollment, not one physical machine, so the current runbook can leave a multi-Workspace Computer partially offline.

**Risk level**: B-low
- Path baseline: documentation-only paths -> C
- Semantic lift: deployment and cutover guidance affects release operations -> B-low

**PR summary**
- Problem: After the machine-authority cutover, existing Computers lose Runtime authentication while rollout health checks remain green, leaving operators without a discoverable recovery path.
- Approach: Document the credential boundary, the generated reconnect command, identity-preserving recovery, and a SQL query for remaining enrollment rows in both English and Chinese deployment guides.
- Impacted modules: `docs/deploying.md`, `docs/zh-CN/deploying.md`

**Review findings**
❌ 1. The recovery procedure says to run once per machine, but one physical Computer may have separate enrollments in multiple Workspaces. `computer connect` stores or rotates only the credential for the Workspace encoded by that one-time code, while the daemon starts one independent Runtime per stored Workspace enrollment. After this procedure is run once, Agents in the other Workspaces remain offline even though the machine was already "re-enrolled." Make the recovery unit explicit as each active Workspace enrollment (and repeat the generated command once per Workspace row returned by the query), then mirror that distinction in the Chinese guide and in the query explanation. [R5 / `docs/deploying.md:137`]

**Action taken**
- Submitted request changes.

Gandy2025:
Correct, and it is the more consequential half of the runbook. Fixed in 3bd8b77.

I wrote the procedure around the host because that is where the operator stands, but the credential boundary is the enrollment: the code carries its Workspace, `exchangeConnectCode` resolves `(workspace_id, computer_id)`, and the daemon opens one Runtime connection per stored enrollment. Following the old text once on a multi-Workspace host recovers exactly one of them and leaves the rest offline behind a machine that reads as recovered — the worst shape for this kind of bug, because the obvious signal says done.

Changes:

- The unit is stated outright: one Workspace enrollment, not one physical machine, with the consequence of getting it wrong spelled out.
- The query moves ahead of the procedure so the rows drive it, and the procedure now reads "for each row returned above" with the Admin qualified as being of that Workspace.
- The query selects `computer_id` and orders by it, so a host enrolled in several Workspaces shows up as adjacent rows instead of being invisible. The explanation says each row is an enrollment and a repeated `computer_id` needs a pass per row.
- Added that repeating the procedure on an already-recovered host is safe, which the old text left an operator to guess at. Verified in `storeMachineEnrollmentCredential`: the file holds one entry per enrollment and enrolling filters out only the entry for the same Workspace before appending, so earlier credentials survive and the daemon picks all of them up on restart.

Mirrored in `docs/zh-CN/deploying.md`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The new head resolves the only blocker by making each Workspace enrollment the recovery unit and grounding the repeat-per-row procedure in the actual credential and daemon boundaries.

**Risk level**: B-low
- Path baseline: documentation-only paths -> C
- Semantic lift: deployment and cutover guidance affects release operations -> B-low

**PR summary**
- Problem: After the machine-authority cutover, existing enrollment Runtimes lose authentication while rollout health checks remain green, leaving operators without an accurate recovery path.
- Approach: Document the credential boundary, identify every affected Workspace enrollment with a query, require one generated command per returned enrollment, preserve the physical Computer identity when possible, and mirror the runbook in Chinese.
- Impacted modules: `docs/deploying.md`, `docs/zh-CN/deploying.md`

**Review findings**
✅ 1. The prior blocker is resolved: repeated `computer_id` rows now expose multi-Workspace hosts, the procedure runs once per Workspace enrollment, and the English and Chinese guides both explain why existing credentials survive each subsequent enrollment.
⚠️ 2. Non-blocking, pre-existing drift: the current Context Tree still describes Account-token daemon authentication and one multiplexed WebSocket, while merged source now uses enrollment-scoped machine credentials and one Runtime connection per stored enrollment. Reconcile that durable architecture separately; it should not delay this operator recovery guide.

**Action taken**
- Approved the current head.

yuezengwu (changes requested):
- Reviewed exact head: `3bd8b774d866fddaea50797c7c32f0576b851d77`
- Risk: B-low; documentation-only paths, but this is a cutover recovery runbook.

**Findings**

❌ 1. The SQL result is not a reliable fleet-recovery completion signal. It only finds active enrollments with no active server-side credential, but `runComputerConnect` calls `exchangeComputerConnectCode` first and writes `computer-credentials.json` afterwards. The exchange transaction creates the credential and consumes the code before the CLI attempts the local write. If that write fails because of disk, permissions, or an interrupted process, the command has not recovered the enrollment and the daemon still has no usable token, yet this query stops returning the row and can become empty. Please describe this as the list of enrollments with no active credential **issued yet**, remove the claim that an empty list means the fleet has recovered, and add an explicit Online/runtime-registration verification step (mirrored in Chinese). [R5 / `docs/deploying.md:147-149`, `docs/zh-CN/deploying.md:137-138`]

⚠️ 2. The Chinese mirror changed substantially in this PR but still reports `Last synced with: 2026-08-20`. Update the mirror metadata to the current canonical sync date. [R2 / `docs/zh-CN/deploying.md:4`]

**Validation**
- `pnpm check` — pass (429 files; CLI third-party notices verified)
- `git diff --check main...HEAD` — pass
- GitHub CI — all six checks pass
- Verified the credential ordering against `apps/cli/src/core/computer/connect.ts` and `packages/server/src/services/computers/machine-auth-service.ts`.

baixiaohang (changes requested):
- Rationale: Withdrawing my previous approval: the server records an active credential before the CLI persists its secret locally, so the SQL query cannot prove that an enrollment or fleet has recovered.

**Risk level**: B-low
- Path baseline: documentation-only paths -> C
- Semantic lift: deployment and cutover guidance affects release operations -> B-low

**PR summary**
- Problem: After the machine-authority cutover, existing enrollment Runtimes lose authentication while rollout health checks remain green, leaving operators without an accurate recovery path.
- Approach: Document the credential boundary, query affected Workspace enrollments, require one generated command per enrollment, preserve physical Computer identity where possible, and mirror the runbook in Chinese.
- Impacted modules: `docs/deploying.md`, `docs/zh-CN/deploying.md`

**Review findings**
❌ 1. I concur with yuezengwu's blocker. `exchangeComputerConnectCode` creates or rotates the active server-side credential and consumes the one-time code before `storeMachineEnrollmentCredential` attempts the durable local write. A disk, permission, or process failure can therefore remove the row from this query while the daemon still lacks the usable secret (or retains a now-revoked old secret). Describe the SQL result only as enrollments with no active credential issued yet, remove the empty-list recovery claim, and require an explicit Online/runtime-registration check after each command in both guides. [R5 / `docs/deploying.md:147`]
⚠️ 2. The Chinese mirror is changed in this PR but still says `Last synced with: 2026-08-20`; update that metadata to the current canonical synchronization date. [R2 / `docs/zh-CN/deploying.md:4`]

**Action taken**
- Submitted request changes on the current head, superseding my earlier approval.

Gandy2025:
Both findings are right, and the first one mattered more than the original omission it replaced. Fixed in a74f2b0.

**On the completion signal.** I checked the ordering rather than take it on trust: `runComputerConnect` calls `exchangeComputerConnectCode` at `connect.ts:55` and `storeMachineEnrollmentCredential` at `:63`. So the Server creates the new credential — and `exchangeConnectCode` revokes the enrollment's previous one in the same transaction — before anything is written to the host. A local write that fails leaves the enrollment with a Server-side credential the daemon does not have and without the secret it used to have. The row leaves the query having gone *backwards*, which is the worst possible thing for a query documented as "empty means recovered".

Changes:

- The credential query is recast as **the work list**: enrollments not yet *issued* a credential. It explicitly says it measures issuance, not recovery, and points at the section below.
- A new **Confirming recovery** section carries the real check, built on the rule the Server applies rather than a proxy for it: `current_instance_id is not null and last_seen_at` inside the ninety-second presence window, matching `presenceTimeoutMs`.
- Recovery gains a fourth step: confirm Online before moving on, and do not read disappearance from the work list as success.
- The partial-failure window is documented with its remedy — regenerate and run again — including that the enrollment is worse off than before the attempt, since an operator who does not know that will not think to retry.

The Chinese sync date is now `2026-08-26`. Thanks for catching it; I added the section without touching the header.

One thing I did not do: keep both queries as interchangeable options. The work-list query cannot answer "is it recovered" at all, so presenting them side by side would leave the same wrong reading available. It is now framed as build-the-list and check-the-result, with only the second one authoritative.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The new head separates credential issuance from recovery, verifies Online through the product's Runtime-presence rule, and documents the partial local-write failure and retry path.

**Risk level**: B-low
- Path baseline: documentation-only paths -> C
- Semantic lift: deployment and cutover guidance affects release operations -> B-low

**PR summary**
- Problem: After the machine-authority cutover, existing enrollment Runtimes lose authentication while rollout health checks remain green, leaving operators without a reliable recovery and completion signal.
- Approach: Build an issuance-only enrollment work list, run one generated command per Workspace enrollment, verify each result with the same Runtime-presence rule used by the product, preserve physical Computer identity where possible, and mirror the runbook in Chinese.
- Impacted modules: `docs/deploying.md`, `docs/zh-CN/deploying.md`

**Review findings**
✅ 1. Both prior findings are resolved: the work-list query is explicitly non-authoritative for completion; a separate current-instance plus 90-second presence query and per-enrollment Online check now own recovery verification; the local-write failure window has a concrete retry path; and the Chinese synchronization date is current.
⚠️ 2. The previously noted Context Tree authentication/connection drift remains a separate, non-blocking follow-up and is not introduced by this documentation change.

**Action taken**
- Approved the current head.

yuezengwu (changes requested):
- Reviewed exact head: `a74f2b05a03dcbadf157bf5a1d3098771bd31664`
- Risk: B-low; documentation-only paths, but the SQL is presented as the authoritative cutover-recovery check.

**Resolved from the previous review**

✅ The work-list query is now explicitly issuance-only, recovery is per Workspace enrollment, each command is followed by an Online check, the server-before-local-write failure window has a safe fresh-command retry, and the English/Chinese guides plus mirror date are synchronized.

**Remaining finding**

❌ 1. The completion SQL still does not implement the same nullable-presence rule as the product. `WorkspaceAdminService` reports Online only when `currentInstanceId !== null && (lastSeenAt?.getTime() ?? 0) >= cutoff`, so a null `last_seen_at` is Offline. The documented predicate is `current_instance_id is null or last_seen_at < cutoff`; when the instance is non-null and `last_seen_at` is null, SQL evaluates it to NULL and filters the row out, falsely treating the enrollment as recovered. Both columns are nullable and there is no schema constraint pairing them. Add `wc.last_seen_at is null` to the offline predicate in both guides so the supposedly authoritative query actually matches the product boundary. [R4 / `docs/deploying.md:192`, `docs/zh-CN/deploying.md:176`]

⚠️ 2. The PR body still describes “per-machine recovery” and “a query listing the Computers still waiting, which is empty once the fleet has recovered.” Those are the semantics this new head correctly removed from the guides; please update the PR description so the durable review record does not contradict the implementation.

**Validation**
- `pnpm check` — pass (429 files; CLI notices verified)
- `git diff --check` — pass
- `pnpm turbo run test --filter=@opentag/server --concurrency=2 --force` — 238/239 pass; the unrelated pre-existing observability terminal-outcome test failed locally and failed again in isolation
- All six current GitHub checks — pass
- Context Tree `first-tree-ai/opentag-context-tree` at `d4bb6ba3dd69525af73caf2bd6a5145492b230ab` verified; its known authentication/connection drift is pre-existing and not a blocker for this PR.

baixiaohang (changes requested):
- Rationale: The authoritative recovery query still omits an Offline state that the product explicitly represents when `last_seen_at` is null.

**Risk level**: B-low
- Path baseline: documentation-only paths -> C
- Semantic lift: deployment and cutover guidance affects release operations -> B-low

**PR summary**
- Problem: After the machine-authority cutover, existing enrollment Runtimes lose authentication while rollout health checks remain green, leaving operators without a reliable recovery and completion signal.
- Approach: Build an issuance-only enrollment work list, reconnect once per Workspace enrollment, verify each result against Runtime presence, preserve physical Computer identity where possible, and mirror the runbook in Chinese.
- Impacted modules: `docs/deploying.md`, `docs/zh-CN/deploying.md`

**Review findings**
❌ 1. I concur with the remaining blocker. `WorkspaceAdminService` treats a row as Online only when `currentInstanceId !== null` and `(lastSeenAt ?? 0)` is inside the presence window, so `last_seen_at is null` is explicitly Offline. Both database columns are nullable and no constraint pairs them. In the documented predicate, a non-null instance plus null `last_seen_at` evaluates to SQL NULL and is filtered out, so the authoritative completion query can falsely omit an Offline enrollment. Add `wc.last_seen_at is null` to the predicate in both guides. [R4 / `docs/deploying.md:192`, `docs/zh-CN/deploying.md:176`]
⚠️ 2. The PR body still says recovery is per machine and that an empty work-list query proves fleet recovery, while the guides now correctly use per-Workspace-enrollment recovery and a separate Runtime-presence check. Update the description so the durable review record matches the change.

**Action taken**
- Submitted request changes on the current head, superseding my prior approval.

Issue #167, opened 2026-08-26
refactor: retire Workspace as a product and domain concept


Gandy2025:
# Retire Workspace as a product and domain concept

Status: accepted product direction. Phase 1 is ready to implement. Phase 2 is gated on a one-off data migration.

## Summary

OpenTag's management surface is administered by one person. Sharing happens in Feishu and Slack, where colleagues talk to an Agent without an OpenTag Account. A user installs Claude Code or Codex on their own laptop, creates one or more Agents on it, and connects each Agent to one app in their IM tenant. Nobody else needs access to the management surface to benefit from those Agents.

Under that definition Workspace has no work left to do. It was designed as a team boundary: a set of people who completely trust one another to manage the same Agents and machines. That population does not exist in the product as defined.

This issue retires Workspace in two stages. Phase 1 removes it from every user-facing and API surface without touching storage. Phase 2 folds `workspaces` into `users` and drops the admin-grant and invitation tables. Between them sits a one-off data migration that is a hard precondition for Phase 2, not an optional cleanup.

The end state is four layers: **Account → Computer enrollment → Agent → IM binding.**

This ownership spine does not preclude shared durable context. Context Tree access is orthogonal to Account, Computer enrollment, Agent, and IM binding ownership. Agents owned by different Accounts may eventually be authorized to read or write the same portable Git-hosted Context Tree without joining a shared OpenTag management scope.

A Context Tree is not a Workspace or Project. It holds durable, reviewable decisions and context, not Agent ownership, runtime state, Tasks, ordinary working files, credentials, or Secrets.

## Relationship to #125

#125 established that Workspace "remains the collaboration, authorization, and data-isolation scope, but it stays implicit when an Admin can access only one," and that it is "still necessary whenever one Account administers multiple independent trust scopes."

This issue supersedes that conclusion. The multi-trust-scope case is not the product: an Account administers its own Agents and nothing else. What #125 called progressive disclosure becomes permanent removal.

Everything else #125 established stands and is a precondition for this work — machine credentials scoped to one enrollment, account tokens barred from the Runtime WebSocket, cross-scope non-disclosure, and creation identity as audit metadata rather than an authorization rule.

### The two populations do not both need a boundary

#125 already separated Admins from IM Users. What it did not resolve is that the Admin population is one person per set of Agents. A boundary that groups people is only worth its cost when more than one person is inside it.

[...]

PR #170, opened 2026-08-26
refactor: remove Workspace Admin invitation creation

Gandy2025:
## Why

OpenTag is moving to a self-administered model: one Account administers its own Agents, and colleagues consume those Agents through Feishu or Slack without an OpenTag Account. A Workspace with several equal administrators contradicts that model, so nothing should be able to create one any more.

This is the **creation slice only** of the parent plan in #167 ("Retire Workspace as a product and domain concept"). It is deliberately not the full code deletion described there. No schema change, no data change, fully reversible. Existing Admins keep every grant they have; they simply cannot add more.

## What is removed

| Layer | Removed |
| --- | --- |
| Server | `POST` on `WORKSPACE_ADMIN_INVITATIONS_TEMPLATE`, `InvitationService.create`, `WorkspaceAdminAccess.createInvitation`. `InvitationService` no longer takes a public URL or an invitation TTL, because nothing mints a link. |
| Web | The **Invite Admin** button and the single-use invitation panel in the Workspace Admin settings, `BrowserApi.createAdminInvitation`, and the CSS and date helper only those used. |
| CLI | `opentag admin invite`, plus `runAdminInvite` and `formatAdminInvitation`. |
| Shared | `AdminInvitationSchema`, the `AdminInvitation` type, `WORKSPACE_ADMIN_INVITATIONS_TEMPLATE`, and `workspaceAdminInvitationsPath`. |

## What is kept, and why

- **Accepting an invitation** — `ADMIN_INVITATION_PREVIEW_TEMPLATE`, `ADMIN_INVITATION_ACCEPT_TEMPLATE`, and the `/invites/:token` page. Invitations issued before this change must stay redeemable.
- **Listing and revoking Admins** — `opentag admin list` / `admin revoke`, the `GET`/`DELETE` admin routes, and the Revoke control in the Web settings. Revocation is the safety valve for the existing shared Workspace; removing it would strand access with no product path to remove it.
- **`establishDefaultWorkspaceForNewAccount` and `bootstrapAdminInTransaction`** — signup still creates its Workspace and its single grant.
- `WorkspaceAdminAccess.withAdminMutation` stays: `WorkspaceAdminService` still calls it.

## Scope

Confined to the invitation-creation path plus the docs that advertised it. `apps/web/src/router.tsx` is touched only inside `WorkspaceAdminSettings`. The English docs and their Chinese mirrors are updated together, with the mirror sync dates refreshed.

## Tests

Written test-first, each confirmed failing for the right reason before the removal.

- New `packages/server/src/__tests__/admin-invitation-routes.test.ts`: asserts no route creates an invitation (`hasRoute` is `false` and the path answers 404), while preview, accept, and the admin-revoke route stay registered and behave.
- `apps/web/src/__tests__/app.test.tsx`: the "creates a short-lived single-use Admin invitation" test becomes "offers no way to invite another Admin while keeping revocation reachable", which drives a real `DELETE` through the Admins list.
- `packages/client` and `apps/web` API tests assert the creation method is absent while preview/accept still round-trip.
- `packages/server/src/__tests__/integration/admin-invitations.test.ts`: the lifecycle tests now seed an outstanding invitation row directly, the way one exists in storage today, instead of calling the removed service method. Acceptance, concurrency, and last-Admin behaviour are unchanged.

### Verification

- `pnpm check` — clean.
- `pnpm typecheck` — clean (7/7).
- `pnpm build` — clean (5/5).
- `pnpm turbo run test --continue` — `@opentag/web` 304/304, `@opentag/shared` 78/78, `@opentag/client` 461/461. The only failures are the known pre-existing ones: the flaky `packages/server` observability RuntimeSession-tracing test, and the two `apps/cli` daemon-service "stale PATH shim" tests.
- PostgreSQL integration tests were not run; they need a live database.


baixiaohang (changes requested):
- Rationale: The issuance path is removed cleanly across layers, but the retained redemption semantics and Web recovery copy are inconsistent, and the new revocation test races the async Admin list.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A
- Semantic lift: none beyond the A baseline

**PR summary**
- Problem: OpenTag's self-administered model should no longer let an Account invite an equal co-administrator, while existing shared Workspaces still need safe revocation and already-issued invitations must remain redeemable.
- Approach: removes invitation issuance from the Server, Web, CLI, Client, and Shared public surfaces; retains preview/accept for outstanding links and keeps Admin listing/revocation.
- Impacted modules: `apps/cli`, `apps/web`, `packages/client`, `packages/server`, `packages/shared`, and the English/Chinese top-level docs.

**Review findings**
❌ 1. Reconcile every user-facing statement with the actual transition semantics. `NoWorkspaceAccess` still tells a revoked Account that a current Admin can send a *new* invitation (`apps/web/src/router.tsx:669`), and `app.test.tsx:2371` locks in that now-impossible recovery path. Conversely, the new settings and development-guide text says no new Admin or multi-Admin Workspace can appear (`router.tsx:2936`, `DEVELOPMENT.md:290`, and the Chinese mirror), even though this PR deliberately keeps outstanding invitations redeemable and acceptance can still create a grant. The precise invariant is that no new invitation can be issued; an already-issued invitation can still add an Admin until it is consumed, invalidated, or expires. Please align the Web copy, its tests, docs, and route comment with that invariant. [R5]

❌ 2. The new revocation test races the independently loaded Admin list. It waits only for the static `Admins` heading and then calls synchronous `getAllByRole` (`apps/web/src/__tests__/app.test.tsx:1982-1987`), so the list may still be in its loading state; the supported Node 22 job reproduced this as “Unable to find ... Revoke.” Await the Revoke controls (or another completed-list signal) before clicking so this test deterministically verifies the retained safety valve.

✅ 3. The deletion itself is otherwise coherent: issuance contracts and call paths are removed in the same change, while preview, accept, list, and revoke remain covered.

**Action taken**
- Submitted request changes on head `d6ce9168bfff928ca08ba6f1098367bf247187ee`.

Gandy2025:
Closing this without merging. The branch stays on the remote, so nothing here is lost.

**Why.** Removing invitation creation was scoped as the smallest slice that makes the self-administered model true. It does not need to be true yet. Everyone who could invite an Admin into this Workspace is already an Admin of it, so the capability changes nothing about the current state, and revocation remains as the correction if it were ever used. The cost of not shipping this is close to zero.

The cost of shipping it is not. The change came back at 28 files rather than the route and button it was scoped as, and after merging current `main` into the branch one of its own tests fails — `offers no way to invite another Admin while keeping revocation reachable` cannot find the Revoke button. GitHub reports the branch as MERGEABLE because the merge is textually clean; it is not semantically clean. Landing it means debugging that first, then reviewing the parts beyond the original scope: the `InvitationService` constructor signature change, ~87 lines of CSS, and four documentation files.

**The deeper reason.** #167 records a product direction that is one day old, and the production data currently contradicts it — this Workspace has five equal Admins actively sharing eight Agents. Deleting a working capability and its tests ahead of that direction being settled spends real effort on a bet that may not hold.

**What would bring it back.** Once the self-administered model is settled — the data split is scheduled, or Phase 1 is formally planned — redo this from the `main` of that moment. The scope will also be clearer then, since the surrounding deletions in #167 would land together rather than being pulled in piecemeal by this one.

Verification performed before closing, so the work is not lost to doubt: with `origin/main` merged in, `pnpm check` and `pnpm typecheck` pass, `@opentag/shared` 78/78, `@opentag/client` 461/461, `@opentag/server` 241/242 (the known flaky observability test), `apps/cli` 120/122 (the two known stale-PATH-shim tests), and `@opentag/web` 310/311 with the one regression described above. The keep list held: the preview and accept routes, the list-admins and revoke-admin routes, and signup's default Workspace and grant are all intact on the branch.

PR #181, opened 2026-08-26
feat: add the staging Onboarding Lab

Gandy2025:
Closes #180.

## Summary

Adds one staging-only Onboarding Lab for a shared internal test Account, so the first-run experience can be reviewed quickly and re-run for real without ad hoc database edits, disposable Accounts, or rebuilding staging.

**Server** — one optional setting, `OPENTAG_STAGING_ONBOARDING_ACCOUNT_ID`. It must be an Account UUID and is valid only with `OPENTAG_ENV=staging`, so configuring it in production fails Server startup; when unset, no Lab route is registered. `GET`/`POST /api/v1/internal/onboarding-lab` answer 204 for the configured Account and 404 for anyone else, take no body, and accept no client-selected Account. `POST` goes through the normal browser CSRF protection.

`resetOnboarding(accountId)` performs an idempotent staged reset rather than a misleading cross-domain transaction, because Session termination and Computer socket closure cannot be rolled back with SQL. It deletes owned Agents through the existing Agent lifecycle (which disables IM bindings, clears encrypted IM and setup credentials, ends Sessions and removes runtime configuration), revokes outstanding connect codes, enrollment credentials and enrollments, closes the affected live Computer connections, re-reads authoritative facts, and only then clears the setup-completion timestamp. A reset that fails before verification leaves setup completion untouched and converges when it is run again. No run records, TTL cleanup, queue or lease.

**Web** — `/internal/onboarding-lab`, inside authenticated routing but outside the setup-completion gate, so a stuck run can always reopen it. It is not in product navigation; testers bookmark it. A persistent warning states the Account is shared. Scenario Preview renders nine fixed states through the production onboarding page — the presentation moved out of `OnboardingPage` into `OnboardingView`, which both render, so there is no second onboarding page. Preview writes nothing: no Feishu setup lifecycle is mounted, and Agent creation reads no creation intent, resumes nothing and takes no focus. Only the selected fixture is kept, in the URL. The reset action asks for one confirmation and, on a verified 204, refreshes authoritative `/me` state before entering the ordinary `/onboarding` route with no stale Agent selection; a failure stays on the Lab with a retry.

### Note on #180's dependency on #179

[...]

Issue #179, opened 2026-08-26
design: replace the management Workspace with direct Account ownership


Gandy2025:
Status: **D2 + R2 selected; PRs 2A, 2B, and 2C are merged; PR 2D is not authorized yet**.

@baixiaohang completed the architecture review and formally selected D2 + R2. Review decisions 1–10 are closed. The four refined 2D gates are accepted as requirements, but accepting them did not satisfy the gates or authorize the destructive migration, PR 2D implementation, or deployment.

## Selected plan

The reviewed and selected combination is:

```text
Data strategy D2: preserve Accounts/auth identities and destructively replace Workspace-bound business data
Rollout strategy R2: Account-native façade → consumer cutover → atomic storage cutover
Final implementation stage 2D: execute the selected D2 cutover and complete R2
```

D2 answers **what happens to data**. R2 answers **how the change is rolled out**. PR 2D is the still-pending final implementation stage; it is not a third strategy.

This is materially smaller than building nullable owner columns, dual-write, owner/grant fallback, and mismatch reconciliation for rows that will be deleted. If the data-preservation assumption changes before 2D, switch the plan as a whole to D4 + R1 rather than partially combining the two approaches.

## Relationship to #167 and Phase 1

#167 accepted retiring Workspace as a product and domain concept. Phase 1 is complete through #171, #172, #173, #174, #176, and #177:

- users cannot create, switch, inspect, or administer a Workspace;
- Admin invitations and the `opentag admin` surface are gone;
- Workspace remains only as an internal storage and authorization compatibility scope;
- existing grants and rows remain live authorization facts.

Phase 2 removes that remaining **management Workspace** from storage, authorization, management HTTP contracts, Web, and CLI.

## Current Phase 2 progress

- **2A completed:** #183 added the Account-native Server façade while preserving the legacy storage and routes.
- **2B completed:** #185 moved Web management flows to the Account-native routes.
- **2C completed:** #186 moved Client and CLI management to the Account-native routes, removed CLI Workspace selection, and added final credential-shape readiness.
- **Fence regression completed:** #189 pins the persisted `current_instance_id` delivery guard required by the accepted runtime-fencing decision.
- **2D pending:** the destructive schema/authority/onboarding cutover is not authorized for implementation or deployment.
- **2E pending:** mechanical terminology cleanup follows 2D and remains optional.

## Critical terminology boundary

Two unrelated concepts currently use `workspace`.

### Management Workspace — remove

[...]

PR #182, opened 2026-08-26
feat: add the staging Onboarding Lab

### LINKED ISSUE #180: [Feature]: Add a staging Onboarding Lab for repeatable first-run testing
## User problem

The onboarding experience is being iterated frequently, but a completed staging Account cannot reliably return to a first-run state. Clearing only the setup-completion timestamp is insufficient: existing Computer enrollments, Agents, runtime readiness, and IM bindings immediately advance or complete the fact-derived onboarding flow.

Product and engineering need both:

1. fast, deterministic review of important onboarding screen states; and
2. repeatable validation of the real staging path through Computer connect, Agent creation, Feishu authorization, handoff readiness, and setup completion.

Ad hoc database edits, deleting local OpenTag directories, creating disposable Accounts, or rebuilding staging are too costly or unsafe for routine design iteration.

## Desired outcome

Provide one staging-only **Onboarding Lab** for an internally shared test Account:

- **Scenario Preview** renders fixed onboarding states without Server writes;
- **Reset and start real onboarding** resets that Account's active OpenTag resources, then enters the ordinary production onboarding route;
- repeated real runs reuse the host's stable Computer identity and existing OpenTag home;
- production and ordinary staging Accounts cannot discover or invoke the Lab.

The design should remain deliberately small: no temporary Accounts, run records, reset jobs, TTL cleanup, queue, lease, or test-user management system.

## Relationship to #179

Depends on the direct Account-ownership model proposed in #179.

Do not add a transitional Workspace-scoped reset implementation. The reset interface is Account-native and should be implemented after the relevant #179 ownership, HTTP, onboarding-state, and Computer-credential contracts stabilize.

The one-time destructive Phase 2 cutover proposed by #179 and this repeatable staging feature solve different problems:

- #179 performs one deployment cutover;
- this feature repeatedly resets one configured staging test Account after the cutover.

## Shared staging Account

Provision one enterprise-managed Google test identity, for example `onboarding-test@company.example`. Store its login credentials in the team's password manager, never in this repository or deployment configuration.

The Account is a shared, serially used sandbox. The page must warn that resetting it can interrupt another tester. Concurrency management is a team convention at first; do not add a lease or queue.

Any browser may be used. A private window or separate browser profile is optional convenience for retaining a personal staging login, not a product requirement.

## Configuration and authorization

Add one optional Server setting:

```text
OPENTAG_STAGING_ONBOARDING_ACCOUNT_ID=<uuid>
```

Rules:

- when unset, the Lab routes are not registered;
- the value must be a UUID;
- the setting is valid only with `OPENTAG_ENV=staging`;
- configuring it in production must fail Server startup;
- every request must be authenticated as exactly that Account;
- reset always targets the authenticated Account and accepts no client-selected `accountId`;
- unauthorized or unavailable Lab access returns 404;
- reset requires the normal browser CSRF protection.

## Web surface

Add:

```text
/internal/onboarding-lab
```

The route lives inside authenticated routing but outside the setup-completion gate. A tester must be able to reopen it if a real onboarding run becomes stuck after the Account is reset.

Do not add it to normal product navigation initially; testers can bookmark the route.

The page has two sections and a persistent `Shared staging test account` warning.

### Scenario Preview

Provide deterministic fixtures for:

- brand new / no Computer;
- Computer offline;
- Provider unavailable;
- ready to create Agent;
- Agent runtime unavailable;
- waiting for Feishu;
- Feishu authorization in progress;
- setup complete;
- loading failure.

Persist only the selected fixture in the URL, for example:

```text
/internal/onboarding-lab?scenario=computer-offline
```

Preview requirements:

- reuse the production onboarding state derivation and visual presentation;
- make no write request and create no durable state;
- do not simulate the Computer daemon, Feishu protocol, or a general fake Server;
- use Preview for screen hierarchy, copy, and state communication;
- use the real reset path for end-to-end interaction testing.

Extract only the presentation seam needed by production onboarding and Preview. Do not duplicate a second onboarding page.

### Real reset

Show one action:

```text
Reset shared account and start onboarding
```

Require one confirmation explaining that current Agents, Computer enrollments, and messaging connections will be disabled and that another tester may be interrupted.

On success, refresh `/me` and navigate to the ordinary `/onboarding` route with no `agentId` query parameter.

## Internal HTTP interface

Use one path with two methods:

```http
GET  /api/v1/internal/onboarding-lab
POST /api/v1/internal/onboarding-lab
```

- `GET` returns 204 when the current Account may use the Lab, otherwise 404.
- `POST` accepts no body, performs reset, and returns 204 only after the Account is ready to enter onboarding.

Keep the cross-package path contract in the shared public surface, but no response DTO is needed for a 204 interface.

## Reset module

Expose one small interface:

```ts
resetOnboarding(accountId: string): Promise<void>
```

Keep it separate from the production setup-completion module, whose transition remains one-way. The reset module is staging-only orchestration and must reuse existing Agent, IM-binding, Session, and Computer lifecycle behavior rather than duplicating their invariants.

### Preconditions

Before mutation:

- confirm the configured environment and Account again;
- resolve only resources directly owned by the authenticated Account;
- refuse to proceed if any ownership invariant is inconsistent.

### Idempotent staged reset

Do not implement a misleading cross-domain database transaction: runtime Session termination and Computer WebSocket closure are external effects that cannot be rolled back with SQL.

Instead, perform an idempotent staged reset:

1. List all non-deleted Agents owned by the Account.
2. Suspend active Agents through the existing Agent lifecycle.
3. Delete suspended Agents through the existing Agent lifecycle. This must disable current IM bindings, clear encrypted IM/setup credentials, end active Sessions, remove runtime configuration, and mark Agents deleted.
4. Revoke the Account's outstanding unconsumed Computer connect codes.
5. Revoke active Computer enrollment credentials and active Computer enrollments.
6. Close each affected enrollment in the live Computer connection registry.
7. Re-read authoritative facts and verify that the Account has no non-deleted Agent, active IM binding, active Computer enrollment, or usable outstanding connect code.
8. Only after verification, clear the Account's setup-completion timestamp.

Deleted, disabled, revoked, or absent resources are successful no-ops. A failed reset may be run again and continues from current facts. Do not add a progress table or compensation job.

The setup-completion timestamp is the final commit marker: the Account must not enter onboarding until active resource cleanup has been verified.

## Local Computer behavior

Normal repeated onboarding testing must not require deleting or rebuilding `OPENTAG_HOME`.

After Server reset:

1. the previous enrollment and machine token are invalid;
2. the Web produces a new Computer connect command;
3. the CLI reuses the stable physical Computer identity;
4. the new Account enrollment credential replaces the invalid credential for that same Account/scope;
5. the existing daemon service restarts and reconnects.

This validates the Web onboarding Computer step, connect command, enrollment, readiness, and Agent creation. It intentionally does not retest first-time package installation, first-time daemon-service installation, or creation of a brand-new local Computer identity on every run. Those are CLI installation concerns and are non-goals here; use a clean VM/CI scenario if they later require repeated coverage.

## Feishu boundary

Reset must disable OpenTag's binding, clear encrypted credentials/setup context, and end active Sessions. It does not delete a Bot/App from the external Feishu test tenant because no reliable provider-deletion seam exists today.

A real run still validates QR authorization, binding activation, granted capabilities, handoff readiness, and onboarding completion. Testing a truly new tenant App may require choosing a new test Bot or manually removing the old external App. Do not expand this issue into Feishu external-resource lifecycle management.

## Data retained

Do not physically delete historical or identity data solely for the Lab:

- Account profile and Google identity;
- historical deleted Agent rows;
- disabled IM bindings;
- ended Sessions, messages, Tasks, deliveries, and audit records;
- stable physical Computer identity;
- external Feishu Bot/App.

Retained rows must not satisfy onboarding's active facts.

## Error and concurrency behavior

- If reset fails before final verification, leave setup completion unchanged and show a retryable error in the Lab.
- Completed cleanup steps are not rolled back; retry relies on idempotence.
- Concurrent shared-account resets may cause one request to fail. Do not add a lock initially; retry must converge safely.
- The Lab page must never infer success from optimistic client state. It refreshes authoritative `/me` state after a 204 response.

### Server/configuration

- production rejects the Lab setting at startup;
- unset configuration registers no Lab route;
- non-test Accounts receive 404;
- POST requires browser authentication and CSRF;
- the request cannot select another Account.

### Reset integration

Seed the test Account with active Agents, runtime configuration, IM binding/setup secrets, Sessions, Computer enrollments, credentials, and outstanding connect codes, then prove:

- Agent lifecycle cleanup completes;
- encrypted IM and setup secrets are cleared;
- Sessions end;
- credentials, connect codes, and enrollments are revoked;
- the registry closes live enrollment connections;
- setup completion is cleared last;
- authoritative Agent and Computer collections are empty;
- old machine tokens fail;
- running reset twice succeeds;
- retry after a staged failure converges;
- another Account's resources remain unchanged.

### Web

- the Lab route remains reachable while setup is incomplete;
- unauthorized access renders Not Found;
- URL scenarios render deterministic production-derived states;
- Preview emits no write request;
- reset requires confirmation;
- reset failure remains on the Lab with Retry;
- reset success refreshes `/me` and enters ordinary onboarding;
- no stale `agentId` selection survives.

### Manual staging acceptance

```text
Reset shared Account
→ enter ordinary onboarding
→ run the generated Computer connect command using the existing local home
→ observe Computer readiness
→ create Agent
→ authorize Feishu in the test tenant
→ observe handoff readiness
→ complete onboarding
→ reset and repeat
```

## Non-goals

- production reset;
- Account creation or signup testing;
- temporary Accounts or disposable ownership scopes;
- run history, TTL cleanup, queue, lease, or tester management;
- automatic local directory deletion;
- first-time CLI/package/service installation on every run;
- external Feishu App deletion;
- physical deletion of historical business data;
- a Workspace-compatible implementation before #179.

## Suggested delivery

Two focused pull requests after the relevant #179 contracts stabilize:

1. Server configuration, reset module, internal HTTP interface, and integration coverage.
2. Authenticated Lab route, Scenario Preview, reset UI, Web coverage, and staging operator documentation.

## Contribution willingness

Yes.

Gandy2025:
Closes #180. Replaces #181, which carried the same commits on a branch name the CI branch-name guard rejects (`claude/…`); nothing else in that run executed.

## Summary

Adds one staging-only Onboarding Lab for a shared internal test Account, so the first-run experience can be reviewed quickly and re-run for real without ad hoc database edits, disposable Accounts, or rebuilding staging.

**Server** — one optional setting, `OPENTAG_STAGING_ONBOARDING_ACCOUNT_ID`. It must be an Account UUID and is valid only with `OPENTAG_ENV=staging`, so configuring it in production fails Server startup; when unset, no Lab route is registered. `GET`/`POST /api/v1/internal/onboarding-lab` answer 204 for the configured Account and 404 for anyone else, take no body, and accept no client-selected Account. `POST` goes through the normal browser CSRF protection.

`resetOnboarding(accountId)` performs an idempotent staged reset rather than a misleading cross-domain transaction, because Session termination and Computer socket closure cannot be rolled back with SQL. It deletes owned Agents through the existing Agent lifecycle (which disables IM bindings, clears encrypted IM and setup credentials, ends Sessions and removes runtime configuration), revokes outstanding connect codes, enrollment credentials and enrollments, and closes the affected live Computer connections. Verification and the setup-completion marker then share one locked commit boundary — the same scope lock every ordinary Agent, enrollment and connect-code mutation takes — so a 204 cannot be returned over a concurrent write. A reset that fails before that commit leaves setup completion untouched and converges when it is run again. No run records, TTL cleanup, queue or lease.

**Web** — `/internal/onboarding-lab`, inside authenticated routing but outside the setup-completion gate, so a stuck run can always reopen it. It is not in product navigation; testers bookmark it. A persistent warning states the Account is shared. Scenario Preview renders nine fixed states through the production onboarding page — the presentation moved out of `OnboardingPage` into `OnboardingView`, which both render, so there is no second onboarding page. Preview writes nothing: no Feishu setup lifecycle is mounted, no Agent creation intent is read or resumed, and the Computer setup panel issues no connect code and starts no polling. Only the selected fixture is kept, in the URL. The reset action asks for one confirmation and, on a verified 204, waits for the authoritative `/me` response to be installed and to report incomplete setup before entering the ordinary `/onboarding` route with no stale Agent selection; anything else stays on the Lab with a retry.

Making that refresh awaitable changed a shared contract, so Account Settings moves with it: it awaits the refresh, and a save that commits but fails to refresh is modelled as saved-and-unsynchronized rather than as unsaved form data. That state offers only a retry of the refresh — never a repeat of the committed write, and never a discard that would substitute the stale projection for the saved value. Submission is gated at the submit handler on a genuinely dirty, non-syncing state, and one Account refresh runs at a time, so neither implicit Enter nor an edit during a retry can repeat the write or race the refresh.

### Note on #180's dependency on #179

#180 says the reset interface is Account-native and asks for no transitional Workspace-scoped implementation, and #179 is still in design review. Everything externally visible here is already Account-native — the interface is `resetOnboarding(accountId)`, and neither route accepts a Workspace or an Account. The Phase 1 compatibility fact is confined to one private method, `#resolveOwnedScope`, which resolves the Account's single internal scope and **refuses to proceed unless that scope is owned exclusively by the Lab Account**. The #179 cutover replaces that method and the column its queries filter on; the staged reset, its ordering and its tests are unaffected.

Both initial reviews asked for this to be held, split, or explicitly authorized. The author chose to authorize it, and the design change is recorded on the issue: first-tree-ai/opentag#180 (comment).

### Review history across heads

Every finding from @baixiaohang and @yuezengwu has been addressed on the head that follows it, so earlier verdicts do not describe the current code:

| Head | What it addressed | Verdicts on it |
| --- | --- | --- |
| `9e94801` | initial | changes requested by both |
| `c75f2e6` | Preview inertness, the reset lock boundary, the authoritative handoff, plus the #180 decision record | changes requested (Account Settings refresh contract) |
| `3245b02` | awaited the refresh in Account Settings | approved by @yuezengwu; changes requested by @baixiaohang (state model) |
| `a2fc19b` | modelled saved-but-unsynchronized distinctly from unsaved | changes requested by both (submit boundary) |
| `d8074a3` | gated submission and made the refresh single-flight | **approved by @baixiaohang**; @yuezengwu re-requested |

### One note for a separate change

The first CI attempt on `a2fc19b` failed on `@opentag/client` › `session-binding-store` › `C-18`, an I/O-bound test in a package this branch does not touch, exceeding its 5s default timeout on a contended runner; it passes in well under a second locally and attempt 2 was green. Giving that case an explicit timeout belongs in its own change rather than in this PR — see the comment thread.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` and `pnpm --filter @opentag/server test:integration` pass. CI has been green on every head reviewed here, across Checks, CI, Node 22.13.0 and Node 26 compatibility, CLI Pack Smoke and Container Smoke.
- Locally the reset suite plus the Agent and Computer suites were run against a real PostgreSQL 17 instance. Two integration suites provision their own testcontainer and could not run in that sandbox; CI covers them and they pass.
- Server coverage: config rules; route registration, 404/401/CSRF and the no-client-Account guarantee; the failed-reset response envelope; twelve reset integration tests, including that cleanup completes, secrets are cleared, Sessions end, credentials/codes/enrollments are revoked, the registry closes the live connection, old machine tokens fail, the stable Computer identity re-enrolls, running twice succeeds, retry after a staged failure converges, another Account is untouched, a writer interleaving before the commit is refused with the marker left set, and a writer starting between verification and the marker blocks until that transaction commits.
- Web coverage: every scenario renders; every control in every preview is clicked with no request made; confirmation is required; the reset stays on the Lab with a retry when it fails, when the refreshed Account still reports completed setup, and when the refresh itself fails; the route is reachable while setup is incomplete and renders Not Found otherwise; and the handover reaches `/onboarding` only after the authoritative refresh lands.
- Account Settings coverage for the changed refresh contract: a committed save whose refresh fails offers no Save, no Discard and no "Unsaved changes"; its retry reaches the success state with exactly one PATCH ever sent; editing again then discarding returns to the saved name rather than the stale one; submitting the form during the unsynchronized state cannot repeat the committed write; and during a delayed retry the field is disabled while a second retry click and a form submit add neither a write nor a refresh.
- End to end against a real Server process and PostgreSQL: production rejected the setting at startup; the Lab returned 204 for the configured Account, 404 for another Account and 401 unauthenticated; a real reset over HTTP cleared setup completion and emptied the authoritative Agent and Computer collections; the same stable Computer identity re-enrolled with a fresh credential; a repeat reset succeeded. The page and its confirmation dialog were also checked in Chromium against that Server.

Every fix pushed in response to review is pinned by a test that fails when that fix is reverted; that was verified by reverting each one, not assumed.

## Breaking changes

None. The setting is optional and staging-only, no existing route or DTO changes, and the onboarding presentation extraction is behaviour-preserving (the existing onboarding and app-shell suites pass unchanged).

## Checklist

https://claude.ai/code/session_0186L5rTjLYUFmaQ8bHkGSaB


baixiaohang (dismissed):
- Rationale: Scenario Preview can still perform a real Server mutation, the post-reset handoff does not wait for authoritative Account state, and the Server half crosses the explicit #179 sequencing boundary.

**Risk level**: B-low
- Path baseline: includes `packages/server/**`, but not a DB-structure change or one of the listed core service paths.
- Semantic lift: none under the review rubric.

**PR summary**
- Problem: staging reviewers need to inspect onboarding states quickly and repeatedly exercise the real first-run flow without disposable Accounts or manual database edits.
- Approach: add a hidden, authenticated staging Lab with fixed presentation fixtures plus an Account-configured destructive reset that reuses existing Agent, IM-binding, Session, and Computer lifecycle behavior.
- Impacted modules: Web onboarding/router/Agent creation; Server config, internal API, reset orchestration; shared paths/errors; staging documentation.

**Review findings**
❌ 1. Preview still exposes a real Computer-connect write. With the `no-computer` fixture, `AgentCreationFlow` renders the live `ComputerSetup`; clicking “Generate connection command” reads live Computers and calls `issueComputerConnectCode`, creating durable Server state. The new `preview` guard covers creation intent and Agent creation only, so the page’s strongest safety promise is false. Propagate preview mode into the runtime-route/Computer-setup branch, render that interaction inert, and add an interaction test that clicks every preview action while asserting no requests. [`apps/web/src/agent-creation/agent-creation-flow.tsx:467`]

❌ 2. The Server half implements the transitional Workspace-scoped reset that #180 explicitly says not to add before #179. `#resolveOwnedScope` and every cleanup query are built around the compatibility Workspace, while #179 is still in design review and says implementation is not authorized. Keeping the coupling private makes later replacement smaller, but it does not satisfy the agreed delivery boundary and creates temporary lifecycle code, integration tests, and operator docs that must coexist with and then be replaced by the Account-owned model. Hold/split the Server half until the relevant #179 contracts land; the Web half can remain independently reviewable once Preview is actually inert. [R2 / `packages/server/src/services/onboarding-lab/onboarding-reset-service.ts:61`]

❌ 3. Reset success navigates before `/me` has refreshed. `refreshMe` is typed as `() => void`; the callback calls it and immediately navigates to `/onboarding`, so `WorkspaceSetupGate` can evaluate the stale completed membership and redirect to `/agents`. A failed `/me` refresh also happens after the Lab has already been left. Make the refresh awaitable and navigate only after the authoritative response succeeds. The current test can pass on the transient pathname even though its mock keeps returning a completed setup timestamp; use a delayed first post-reset `/me` response and assert the route remains on the Lab until the incomplete state is installed. [R4 / `apps/web/src/router.tsx:640`]

**Action taken**
- Submitted request changes on #182, which carries the exact head (`9e94801`) from the now-closed #181.

baixiaohang (dismissed):
- Rationale: The staging boundary and lifecycle reuse are sound, but the reset can return 204 without preserving a clean final state, and the Web handoff does not actually wait for authoritative `/me` state before routing.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to review onboarding states quickly and repeatedly run the real first-run staging workflow without ad hoc database edits or disposable Accounts.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews plus a destructive reset that reuses existing Agent/IM/Session/Computer lifecycles before returning the shared Account to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration, shared HTTP/error contracts, Web authenticated routing/onboarding presentation, staging operator documentation.

**Review findings**
❌ 1. Make verification and the setup-marker clear one locked commit boundary. `#verifyCleanedUp()` and `#clearSetupCompletion()` run as separate unlocked operations. Every normal Agent/connect-code mutation locks the Workspace row, so another browser can create a new Agent or code after verification and before line 100 clears `setupCompletedAt`; this request still returns 204 with active facts. Reacquire the mutation/scope lock, re-check all cleanup predicates, and clear the marker in the same transaction, with an interleaving regression test. [R4 / `packages/server/src/services/onboarding-lab/onboarding-reset-service.ts:99`]

❌ 2. Await the authoritative `/me` result before entering `/onboarding`. `refreshMe()` only increments a revision and returns `void`, while `navigate()` runs immediately on the next line. `WorkspaceSetupGate` can therefore evaluate the old non-null `setupCompletedAt` and redirect away, and a failed refresh is discovered only after the Lab has already left. Expose an awaitable refresh (or explicitly load `/me` here), verify the refreshed membership is incomplete, then navigate; cover a delayed/failing refresh rather than only asserting that a request eventually occurred. [R4 / `apps/web/src/router.tsx:640`]

❌ 3. Resolve the accepted ownership boundary before landing the Server half. Issue #180 explicitly depends on #179's direct Account-ownership contracts and lists a Workspace-compatible reset as a non-goal, while #179 is still in design review and says implementation is not authorized. This PR instead introduces the transitional `#resolveOwnedScope` path now. Hold/split the Server portion until #179 is accepted and the relevant contracts stabilize, or first record an explicit design change authorizing this compatibility implementation. [`packages/server/src/services/onboarding-lab/onboarding-reset-service.ts:61`]

✅ 4. The route correctly derives the target only from the authenticated Account, stays staging/config gated, uses browser CSRF protection, and Preview avoids mounting the Feishu lifecycle or creation-intent writes.

**Action taken**
- Submitted request changes on head `9e94801ca991be5091d8ddbc479474410d6ab1b5`.

baixiaohang (changes requested):
- Rationale: Preview can still perform a real Server write, the reset can return 204 without preserving a clean final state, and the Web handoff does not wait for authoritative `/me` state before routing.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to review onboarding states quickly and repeatedly run the real first-run staging workflow without ad hoc database edits or disposable Accounts.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews plus a destructive reset that reuses existing Agent/IM/Session/Computer lifecycles before returning the shared Account to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration, shared HTTP/error contracts, Web authenticated routing/onboarding presentation, staging operator documentation.

**Review findings**
❌ 1. Make Scenario Preview fully inert. In the `no-computer` fixture, `AgentCreationFlow` still renders live `ComputerSetup`; clicking “Generate connection command” reads live Computers and calls `issueComputerConnectCode`, creating durable Server state. The `preview` guard covers creation intent and Agent creation only. Propagate preview mode into this branch, render the interaction inert, and add an interaction test that clicks every Preview action while asserting no requests. [`apps/web/src/agent-creation/agent-creation-flow.tsx:467`]

❌ 2. Make verification and the setup-marker clear one locked commit boundary. `#verifyCleanedUp()` and `#clearSetupCompletion()` run as separate unlocked operations. Every normal Agent/connect-code mutation locks the Workspace row, so another browser can create a new Agent or code after verification and before line 100 clears `setupCompletedAt`; this request still returns 204 with active facts. Reacquire the mutation/scope lock, re-check all cleanup predicates, and clear the marker in the same transaction, with an interleaving regression test. [R4 / `packages/server/src/services/onboarding-lab/onboarding-reset-service.ts:99`]

❌ 3. Await the authoritative `/me` result before entering `/onboarding`. `refreshMe()` only increments a revision and returns `void`, while `navigate()` runs immediately on the next line. `WorkspaceSetupGate` can therefore evaluate the old non-null `setupCompletedAt` and redirect away, and a failed refresh is discovered only after the Lab has already left. Expose an awaitable refresh (or explicitly load `/me` here), verify the refreshed membership is incomplete, then navigate; cover a delayed/failing refresh rather than only asserting that a request eventually occurred. [R4 / `apps/web/src/router.tsx:640`]

❌ 4. Resolve the accepted ownership boundary before landing the Server half. Issue #180 explicitly depends on #179's direct Account-ownership contracts and lists a Workspace-compatible reset as a non-goal, while #179 is still in design review and says implementation is not authorized. This PR instead introduces the transitional `#resolveOwnedScope` path now. Hold/split the Server portion until #179 is accepted and the relevant contracts stabilize, or first record an explicit design change authorizing this compatibility implementation. [`packages/server/src/services/onboarding-lab/onboarding-reset-service.ts:61`]

**Action taken**
- Submitted one consolidated request-changes verdict on head `9e94801ca991be5091d8ddbc479474410d6ab1b5` after dismissing a duplicate concurrent review.

yuezengwu (changes requested):
Requesting changes on exact head `9e94801ca991be5091d8ddbc479474410d6ab1b5`.

I independently traced the reset and preview paths and found four blocking issues:

1. `resetOnboarding()` verifies the cleanup and clears the setup-completion marker in separate, unlocked operations (`onboarding-reset-service.ts:99-100`). An Agent creation or connect-code issuance can acquire the normal Workspace mutation lock and commit after `#verifyCleanedUp()` has read zero active facts but before `#clearSetupCompletion()` runs. The reset then returns 204 with `setupCompletedAt = null` while a new active onboarding fact exists. The final verification and marker clear need one transaction that acquires the same mutation lock and rechecks every cleanup predicate before clearing, plus a deterministic interleaving regression test.

2. Preview is not read-only for the brand-new scenario. `AgentCreationFlow` receives `preview`, but `RuntimeRouteSection` does not. With no Computers it mounts the live `ComputerSetup` (`agent-creation-flow.tsx:467-478`), whose button calls `computers()` and then `issueComputerConnectCode()` (`computer-setup.tsx:98-123,196-198`). Clicking the preview therefore creates durable Server state, contrary to the explicit no-write contract. Propagate preview mode through this branch and add an interaction test that clicks every preview control while asserting no requests/writes occur.

3. Reset success schedules a `/me` refresh and immediately navigates (`router.tsx:640-642`). Because `refreshMe()` is `void`, `WorkspaceSetupGate` can still see the old completed marker and redirect `/onboarding` to `/agents`; if the refresh fails, the tester has already left the Lab instead of seeing a retryable failure there. Make the authoritative `/me` refresh awaitable, confirm the incomplete state, and only then navigate. Add delayed-refresh and failed-refresh coverage, not only a transient path assertion.

4. The Server implementation is a transitional Workspace-scoped reset built on `WorkspaceAdminAccess`. Issue #180 explicitly forbids that implementation and schedules delivery only after #179's Account-ownership contracts stabilize. Issue #179 is still open and says implementation is not authorized. Please resolve/accept that design and implement against the resulting Account-native contracts rather than merging this compatibility scope.

Validation at this head: `pnpm check` passed; Web and Server typechecks passed; focused Web tests (152), Server route/config tests (17), and reset integration tests (10) passed. Current GitHub Checks are still failing (the Web app suite failed `removes the old admin product shell without a redirect`), so the CI gate is also not satisfied.

Gandy2025:
Thanks — three of the four findings were real and are fixed in `c75f2e6`. Each fix is pinned by a test that fails without it; I checked that by reverting the fix and re-running.

**1. Preview reached the Server.** Correct, and worse than the report: with no Computer connected, `AgentCreationFlow` renders the live `ComputerSetup`, and `Generate connection command` calls `computers` and then `issueComputerConnectCode` — a durable connect code, from the one surface that promises to create none. `preview` now propagates into that panel: it issues no code, starts no polling, and its control reads as unavailable. `onboarding-lab.test.tsx` now clicks every control in every preview, repeating until nothing new appears, and asserts the fetch spy was never called. Reverting the propagation fails it with `scenario no-computer issued a request`.

**2. Verification and the setup marker were not one commit.** Correct. They now run in a single transaction that first takes the same scope lock every ordinary Agent, enrollment and connect-code mutation takes, re-checks every predicate inside it, and clears the marker before releasing. Two integration tests: a writer that interleaves before the commit is caught, the marker stays set and the retry converges; and a writer starting between verification and the marker is proven to block until that transaction commits. Moving the marker write back outside the lock fails the second one.

**3. Reset navigated before `/me` refreshed.** Correct, and the old test was exactly as weak as described. `refreshMe` now resolves only once the authoritative response is installed as current state, and the Lab enters onboarding only when that Account reports incomplete setup; otherwise it stays on the Lab with a retry. Three tests replace the old one: a delayed refresh (the route must still be on the Lab immediately after the 204, then reach `/onboarding`), a refresh that still reports completed setup, and a refresh that fails. Reverting the router change fails two of them.

Separately, CI was red on the previous head for a reason unrelated to the diff's intent: a loader that requests again after its first await keeps running once its component is unmounted, so its late request landed in the next test and broke an app-shell assertion that no request is made. That is a scheduling-dependent leak, so the shared Web test setup now settles those chains before resetting the fetch spy, and a small test locks that behaviour. I reproduced the failure mode locally before fixing it.

**4. The #179 sequencing boundary.** I agree this is the open question, and it is the author's call rather than mine, so I have put it to them rather than deciding it inside this PR. For the record on what is actually coupled: the contract is Account-native throughout — `resetOnboarding(accountId)`, and neither route accepts a Workspace or an Account — and the compatibility fact is confined to `#resolveOwnedScope`, which additionally refuses to run unless the scope is owned exclusively by the Lab Account. The cleanup queries filter by the scope that method returns, so the cutover replaces that resolution and the column it filters on, not the staged reset, its ordering, or its tests. That is a smaller residue than "temporary lifecycle code", but it is still a real one, and if the answer is to hold, the Web half stands alone and I will split it out — the Server half is the only part that touches the ownership model at all.

---

Gandy2025:
@baixiaohang finding 4 is now settled by the author rather than by me: @Gandy2025 has decided the Server half proceeds in this cutover window, and the design change authorizing it is recorded on the issue — first-tree-ai/opentag#180 (comment). It authorizes only the private ownership resolution, not a Workspace-scoped interface, and states what the #179 cutover replaces. Findings 1–3 are fixed in `c75f2e6`; re-requesting your review.

CI is green on `c75f2e6`: Checks, CI, Node 22.13.0 and Node 26 compatibility, CLI Pack Smoke and Container Smoke all pass, which includes the Server integration suite.

One correction to this PR's original Testing section, since it affects what you should conclude from it: I reported `@opentag/client` › `CodexAppServerProcess` › `D-11` as a pre-existing failure on `main`. It fails only in my sandbox, where process-group teardown behaves differently; CI runs the full `pnpm test` and it passes. There is no pre-existing failure — treat every suite as green. I have corrected the PR body.

---

baixiaohang (changes requested):
- Rationale: The three reported Onboarding Lab races are fixed and #180 now explicitly authorizes the private compatibility boundary, but the new rejectable `refreshMe` contract leaves Account Settings with an unhandled failed refresh and stale success state.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to review onboarding states quickly and repeatedly run the real first-run staging workflow without ad hoc database edits or disposable Accounts.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews plus a destructive reset that reuses existing Agent/IM/Session/Computer lifecycles before returning the shared Account to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration, shared HTTP/error contracts, Web authenticated routing/onboarding presentation, staging operator documentation.

**Review findings**
❌ 1. Handle the new awaitable refresh contract in Account Settings. `WorkspaceSession.refreshMe` now calls `/me` directly and can reject, but `AccountSettings` still narrows it to `() => void` and calls it without `await` or a rejection handler. If the profile PATCH succeeds and the following `/me` request fails, the page reports “Account profile saved” while the shared Account state/header remain stale, the form remains dirty, and the Promise rejection is unhandled. Preserve an honest post-save state when refresh fails and add coverage for a successful PATCH followed by a failed `/me` refresh. [R5 / `apps/web/src/router.tsx:2541`, `apps/web/src/router.tsx:2567`]

✅ 2. Scenario Preview now propagates preview mode into `ComputerSetup`; its connect-code control is unavailable, no polling starts, and the interaction test exercises every enabled preview control while asserting zero requests.

✅ 3. Final cleanup verification and the setup marker now share one transaction under the ordinary Workspace mutation lock, with regression coverage for writers both before and inside that boundary.

✅ 4. The reset handoff now waits for authoritative `/me`, verifies incomplete setup, and stays on the Lab for delayed, failed, or still-completed refresh results.

✅ 5. #180 now records an explicit delivery-ordering decision allowing only this private compatibility resolution before #179; the external interface remains Account-native and the exception does not authorize implementing #179.

**Action taken**
- Submitted request changes on exact head `c75f2e62e85d8f6566389289409d7caa6c0cb6c8`; all current GitHub checks are green.

yuezengwu (changes requested):
Requesting changes on exact head c75f2e62e85d8f6566389289409d7caa6c0cb6c8.

The three implementation blockers from the previous head are independently resolved: Preview now keeps Computer setup inert, final cleanup verification and marker clearing share the ordinary mutation-lock transaction, and reset waits for authoritative /me state before routing. The explicit author decision on #180 also resolves the earlier sequencing question while keeping the external reset interface Account-native.

One fresh-head blocker remains. WorkspaceSession.refreshMe now returns a Promise and can reject, but AccountSettings narrows it back to () => void and calls it without await or a rejection handler after a successful profile PATCH (apps/web/src/router.tsx:2541,2564-2567). If the PATCH commits and the following /me request fails, the page has already announced Account profile saved while the shared Account/header remain stale; the input is also dirty against the old user value, and the Promise rejection is unhandled. Await the refresh and preserve an honest, retryable post-save state when synchronization fails. Add a regression test for successful PATCH followed by failed /me refresh.

Validation on this head: pnpm check passed; Web and Server typechecks passed; focused Web tests passed (157); reset integration passed (12), including both lock-interleaving cases; pnpm build passed. All current GitHub checks are green. This finding is a contract/call-site regression not exercised by the current success-path Account Settings coverage.

Gandy2025:
@baixiaohang @yuezengwu the fresh-head finding was correct, and it was mine: making `refreshMe` awaitable gave it a rejection path, and I did not follow that contract change to its other call site. Fixed in `3245b02`.

`AccountSettings` typed `refreshMe` as `() => void` and called it without awaiting, so a profile PATCH that committed and then failed to refresh announced "Account profile saved" while the header and shared Account stayed stale, left the field dirty against the old value, and dropped an unhandled rejection.

The refresh is now awaited, and its failure is reported as what actually happened rather than as either outcome it is not:

> Your display name was saved, but OpenTag could not refresh the account. Save again to retry.

Two details behind that wording. A refresh failure no longer rolls the field back to the stale server value — the save succeeded, so rolling back would misreport it — and because the shared `user.displayName` is still stale, the unsaved-changes bar stays up, which is exactly the retry affordance: pressing Save re-runs both steps.

Regression test added to the app-shell suite: successful PATCH followed by a failing `/me`, asserting the honest error, that no success message is shown, that the field keeps the saved value, that the stale shared name is still on screen, and that Save is still available. Reverting `router.tsx` fails it with `Unable to find role="alert"` — the silent-success behaviour you described.

`pnpm check`, `pnpm typecheck`, Web (319) and Server (248) suites pass locally.

I also checked the other two call sites of the changed contract while I was in there — `OnboardingRoute`'s setup completion and the Lab handover both already await it — so this was the only one left behind.

---

baixiaohang (changes requested):
- Rationale: The rejected refresh is now handled, but the post-commit failure state still presents a saved Server value as unsaved and offers controls that can restore the stale projection.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to review onboarding states quickly and repeatedly run the real first-run staging workflow without ad hoc database edits or disposable Accounts.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews plus a destructive reset that reuses existing Agent/IM/Session/Computer lifecycles before returning the shared Account to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration, shared HTTP/error contracts, Web authenticated routing/onboarding presentation, staging operator documentation.

**Review findings**
❌ 1. Model a failed post-save refresh as synchronization-needed state, not as unsaved form data. After the PATCH succeeds, `displayName` contains the Server-confirmed saved value while `user.displayName` remains the stale projection, so `dirty` becomes true and the page simultaneously says “Your display name was saved” and “Unsaved changes.” Its `Discard` action then replaces the saved value with the stale one and clears the error without undoing the Server write. “Save again” also repeats the PATCH; if that second PATCH fails, the outer catch rolls the field back to the stale value. Keep the committed PATCH result distinct from refresh-needed state, provide a retry that cannot misrepresent or discard the saved value, and exercise the actual retry and adjacent controls in the regression test rather than only asserting that a Save button exists. [R1 / `apps/web/src/router.tsx:2548`, `apps/web/src/router.tsx:2564`, `apps/web/src/router.tsx:2626`]

✅ 2. `AccountSettings` now preserves the awaitable `refreshMe` type, awaits it, and handles rejection without an unhandled Promise.

✅ 3. The other two `refreshMe` call sites already await the same contract, and the previous Onboarding Lab preview, reset-lock, handoff, and delivery-ordering findings remain resolved.

**Action taken**
- Submitted request changes on exact head `3245b02fe8f0f386c6835c42226fb7d296e000e0`.

yuezengwu (approved):
Approving exact head 3245b02fe8f0f386c6835c42226fb7d296e000e0 after a complete fresh-head re-review.

The Account Settings regression is resolved: refreshMe keeps its Promise contract, the successful PATCH path now awaits the authoritative /me refresh, a refresh failure is caught without falsely rolling back the committed display name or announcing full success, and the form remains retryable. The new regression test covers successful PATCH followed by failed /me and asserts the saved local value, stale shared state, honest error, and retry action.

I also rechecked every prior blocker at this head. Scenario Preview remains request-free, including the Computer setup branch; cleanup verification and setup-marker clearing remain one transaction under the ordinary Workspace mutation lock with deterministic writer interleaving coverage; reset handoff still waits for authoritative incomplete Account state; and #180 contains the explicit author decision limiting the pre-#179 compatibility work to the private ownership resolution.

Fresh-head validation: pnpm check passed; Web and Server typechecks passed; focused Web tests passed (158); onboarding reset integration passed (12), including both lock-interleaving cases; pnpm build passed; git diff --check passed. All six current GitHub checks pass. No remaining blocking findings.

Gandy2025:
@baixiaohang right again, and this one is the more interesting mistake: my previous fix reported the failure honestly but *modelled* it wrongly. Reworked in `a2fc19b`.

The bug behind all three symptoms was one line — `dirty` compared the field against `user.displayName`, the stale projection, so a Server-confirmed value read as unsaved. Everything you listed follows from that: the page saying "saved" and "Unsaved changes" at once; `Discard` replacing the saved value with the stale one without undoing the write; `Save again` repeating a committed PATCH, whose failure then rolled the field back to the stale value through the outer catch.

So the fix is at the state, not the copy. A committed-but-unsynchronized value is now held separately from the shared Account projection, and it — never the stale projection — is what the form treats as confirmed:

- while a refresh is outstanding the form is not dirty, so no "Unsaved changes", no Save, no Discard;
- that state offers only the step that failed: `Retry refresh`, which calls the Account refresh and never repeats the write;
- editing again reopens the unsaved-changes bar normally, and `Discard` returns to the **saved** value;
- a later failed PATCH also falls back to the saved value rather than the stale one.

The message no longer implies re-saving: "Your display name was saved. OpenTag could not refresh the account, so the rest of the page still shows the previous name."

Two tests replace the one you called out for asserting only that a button exists. The first drives the real retry: it asserts no Save, no Discard and no "Unsaved changes" in the unsynchronized state, clicks `Retry refresh`, and then asserts the success state, the refreshed name on screen, and that exactly one PATCH was ever sent — so a retry that quietly re-wrote would fail it. The second holds the refresh failing, edits the field again, and asserts `Discard` restores the saved name rather than the stale one. Both fail against the previous head; the second fails on the state model itself, not on wording.

`pnpm check`, `pnpm typecheck` and the Web suite (320) pass locally.

---

baixiaohang (changes requested):
- Rationale: The saved-versus-unsynced state is now modeled correctly in the rendered controls, but the form submit path can still repeat the committed write or race the refresh-only retry.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to review onboarding states quickly and repeatedly run the real first-run staging workflow without ad hoc database edits or disposable Accounts.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews plus a destructive reset that reuses existing Agent/IM/Session/Computer lifecycles before returning the shared Account to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration, shared HTTP/error contracts, Web authenticated routing/onboarding presentation, staging operator documentation.

**Review findings**
❌ 1. Enforce the saved/unsynced state in the submit boundary, not only by hiding buttons. `submit()` guards only `saveInFlight`; it does not require `dirty` or reject `syncing`. In the unsynced state the text input remains inside an active form, so pressing Enter can implicitly submit it even though Save is hidden and repeat the committed PATCH. During `Retry refresh`, the input also remains editable; editing makes Save appear and allows a new PATCH plus a second `refreshMe()` to run concurrently with the old retry. Because `refreshMe` has no request ordering, the older response can then overwrite the newer shared Account projection. Gate submission on a genuinely dirty, non-syncing state and serialize or disable editing while refresh-only retry is active. Add a delayed-refresh interaction test that proves Enter cannot repeat the write and that no new save can race an in-flight retry. [R4 / `apps/web/src/router.tsx:2563`, `apps/web/src/router.tsx:2601`, `apps/web/src/router.tsx:2636`]

✅ 2. The failed-refresh state now keeps the committed value separate from the stale projection, exposes only `Retry refresh`, and the new tests prove the normal retry performs exactly one PATCH and Discard returns to the saved value after a later edit.

✅ 3. All earlier Onboarding Lab preview, reset-lock, authoritative handoff, refresh-contract, and delivery-ordering findings remain resolved.

**Action taken**
- Submitted request changes on exact head `a2fc19b71797132a14f7e02656208fd22a7a9e7a`; the main Checks/CI jobs are currently failing while the compatibility and smoke jobs pass.

Gandy2025:
CI is red on `a2fc19b`, and the failure is not this PR's. Standing down on it with a re-run rather than changing an unrelated package.

**What failed:** `Checks` → `@opentag/client` → `session-binding-store.test.ts` › `C-18 retains only the most recent 64 recorded input tombstones`, with `Test timed out in 5000ms`. Not an assertion — the test never finished.

**Why it is not this PR's:**

- `packages/client` is untouched by this branch: `git diff --stat main...feat/staging-onboarding-lab -- packages/client` is empty.
- The only files changed since the last green CI head are `apps/web/src/router.tsx` and `apps/web/src/__tests__/app.test.tsx`. The identical client code passed CI twice on this branch already, on `c75f2e6` and `3245b02`.
- The one shared-package change this PR makes is additive — two entries in the error-code enum and one path constant — and nothing in that test path reads either.
- The test is I/O-bound rather than compute-bound: 65 iterations × three store operations, each writing session-binding state to disk, so roughly 195 sequential writes against a 5s default timeout. Locally it completes in ~520ms across three runs, about a tenth of the limit, which is the profile of a test that only fails when a shared runner's disk is contended.

**Proposed fix, for a separate change rather than this PR:** give that one case an explicit generous timeout, e.g. `it("C-18 …", async () => { … }, 30_000)`, since its cost scales with filesystem latency rather than with anything the assertion is about. I am deliberately not making that edit here: it is in a package this PR does not touch, and widening a staging-Lab PR into an unrelated test-timing change is worse than leaving the note. If a maintainer would rather I did fold it in, say so and I will.

Re-running the failed job once now, which is the single re-run I will spend. If it fails again I will treat it as real and investigate properly rather than re-running further.

---

yuezengwu (changes requested):
Requesting changes on exact head a2fc19b71797132a14f7e02656208fd22a7a9e7a.

The saved-versus-unsynchronized state is now represented honestly in the visible controls, and the previous Account Settings and Onboarding Lab blockers remain fixed. One concurrency blocker remains at the form boundary.

submit() guards only saveInFlight; it does not require a genuinely dirty value and does not reject while syncing. After a committed PATCH whose /me refresh failed, the Save button is hidden but the active form still contains its text input, so implicit Enter submission can repeat the committed PATCH. While Retry refresh is pending, the input also remains editable and a new dirty value exposes Save; that starts a new PATCH and another refreshMe concurrently with the older retry. refreshMe has no request generation/order guard, so a slower response from the older request can overwrite the newer Account projection. Gate submit on dirty and not syncing, and serialize or disable editing/saving throughout refresh-only retry. Add delayed-refresh interaction coverage proving Enter cannot repeat the write and a new save cannot race the in-flight retry.

Fresh-head validation: pnpm check passed; Web and Server typechecks passed; focused Web tests passed (159); reset integration passed (12); git diff --check passed. The first GitHub Checks attempt failed only because the untouched Client C-18 tombstone test exceeded its 5-second timeout; the same Client file passes locally 23/23 with C-18 in about 1.8 seconds, and GitHub attempt 2 is currently running. That CI rerun does not affect the code blocker above.

Gandy2025:
@baixiaohang correct, and it is the same mistake one level down: I fixed the state model but enforced it in the rendered controls instead of at the boundary. Fixed in `d8074a3`.

Both paths you named were reachable:

- the field stays inside an active form, so Enter submitted it even with Save hidden, and `submit()` guarded only `saveInFlight` — so a committed PATCH was repeated;
- during `Retry refresh` the field stayed editable, so an edit reopened Save and let a new write plus a second `refreshMe()` run alongside the retry, with no ordering between the two refreshes.

What changed:

- `submit()` now returns unless the form is genuinely dirty and no refresh is in flight, so the guarantee holds wherever submission comes from rather than only where a button is drawn;
- one Account refresh runs at a time, covering the retry and the refresh that the save path itself performs — a second `Retry refresh` click is a no-op rather than a competing request, which is what removes the older-overwrites-newer window;
- the field is disabled while a refresh-only retry is in flight, so no edit can open a racing save.

Two interaction tests, both driving the real controls:

1. In the unsynchronized state, submitting the form twice leaves exactly one PATCH ever sent and the state unchanged. Against the previous head this fails with two PATCHes — the repeat you described.
2. With a delayed `/me`, clicking `Retry refresh` gives a `Refreshing…` state with the field disabled; a second retry click and a form submit during that window add neither a write nor a second refresh, and the run ends in the success state with exactly one PATCH and exactly one added refresh. Against the previous head this fails on the field still being editable.

`pnpm check`, `pnpm typecheck` and the Web suite (322) pass locally.

On the earlier CI failure: the re-run I spent came back green for `Checks` on `a2fc19b`'s siblings — Node 22.13.0, Node 26, CLI Pack Smoke and Container Smoke all passed, and the `@opentag/client` timeout did not recur before this push superseded the run. That matches the read in my previous comment; the standing suggestion to give that I/O-bound case an explicit timeout still belongs in its own change.

---

Gandy2025:
Correction to the last paragraph above: when I wrote it, `Checks` was still running, so I should not have reported its outcome. It has since completed — attempt 2 on `a2fc19b` finished `success`, with `Test` and `Test PostgreSQL integration` both green, so the `@opentag/client` timeout did not recur and the run was not superseded by my push. The conclusion stands; I stated it a few minutes before I could actually see it.

---

Gandy2025:
@yuezengwu your review landed on `a2fc19b` a couple of minutes after I pushed `d8074a3`, which fixes exactly the blocker you and @baixiaohang both found, in the shape you both asked for: `submit()` now requires a genuinely dirty, non-syncing state; one Account refresh runs at a time so a slower older response cannot overwrite a newer projection; the field is disabled during refresh-only retry; and the two delayed-refresh interaction tests prove that submitting the form cannot repeat the committed PATCH and that neither a second retry nor a save can race an in-flight retry. Details in the comment above.

Thanks also for independently landing on the same read of the CI failure — untouched `packages/client`, C-18 exceeding its 5s timeout while passing comfortably in isolation. Attempt 2 on `a2fc19b` finished green, and `d8074a3` is running now.

---

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The remaining Account Settings submit/retry race is closed at the state-transition boundary, and the fresh-head diff introduces no new blocking issue.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to inspect onboarding states quickly and repeatedly exercise the real staging first-run workflow without disposable Accounts or manual database edits.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews and an Account-configured reset that reuses existing lifecycle behavior before handing the tester back to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration, shared HTTP/error contracts, Web authenticated routing/onboarding presentation and Account Settings, staging documentation.

**Review findings**
✅ 1. `submit()` now rejects duplicate, non-dirty, and refresh-concurrent submissions, so implicit Enter cannot repeat a committed PATCH and no save can cross the refresh-only retry boundary.

✅ 2. Account refresh is single-flight within the settings flow, and the display-name field is disabled while retrying. The added interactions cover repeated form submission plus a delayed retry with a second click and submit, asserting one PATCH and one added refresh.

✅ 3. The saved-versus-unsynchronized state remains distinct from unsaved form state, and every earlier Preview, reset-lock, authoritative handoff, refresh-contract, and delivery-ordering finding remains resolved on this head.

**Action taken**
- Approved exact head `d8074a3454e59c7ea050c984e05e0a1f84e30710`. CLI Pack Smoke, Container Smoke, and Node 26 compatibility currently pass; Checks and Node 22 compatibility are still running.

yuezengwu (approved):
Approving exact head d8074a3454e59c7ea050c984e05e0a1f84e30710 after a complete fresh-head re-review.

The remaining Account Settings race is closed at the state-transition boundary. submit now rejects duplicate, non-dirty, and refresh-concurrent submissions; syncAccount is single-flight via a synchronous ref guard; and the field is disabled throughout refresh-only retry. The new tests directly submit the active form after a committed save and exercise a delayed retry with a second click and form submit, proving one PATCH and one added refresh only. I found no remaining path for an older refresh in this settings flow to overwrite a newer Account projection.

All earlier findings remain resolved: Scenario Preview is request-free, cleanup verification and marker clear share the ordinary mutation-lock transaction, reset handoff waits for authoritative Account state, failed post-save refresh is modeled separately from unsaved data, and #180 explicitly limits the temporary ownership compatibility scope.

Fresh-head validation: pnpm check passed; Web and Server typechecks passed; focused Web tests passed (161); onboarding reset integration passed (12); pnpm build passed; git diff --check passed. All six current GitHub checks pass. No remaining blocking findings.

yuezengwu:
Fresh-base conflict check for exact head `d8074a3454e59c7ea050c984e05e0a1f84e30710` against current `main` `2b1ff7e7d16bd1911be7688da0e2d7c0a9237643` (#183 merge commit): a read-only three-way merge reports one textual conflict only, in `packages/server/src/app.ts`. `packages/server/src/index.ts`, `packages/shared/src/http-paths.ts`, and `packages/shared/src/index.ts` merge automatically.

Please resolve `app.ts` by preserving both registrations: the Account-native routes added by #183 and the staging-only internal Onboarding Lab route plus `OnboardingResetError` mapping from this PR.

There is also one expected semantic coordination outside the textual conflict. #183 made `WorkspaceAdminAccess.resolveCompatibilityWorkspaceId` the canonical `AccountScopeResolver` seam. Update `OnboardingResetService.#resolveOwnedScope` so the compatibility Workspace selected for reset comes through that seam rather than a second selection rule. Preserve the stricter reset-only fail-closed guarantees: the Lab Account must still have exactly one active scope and that scope must still have exactly one active admin. Add or retain coverage proving multiple active scopes and a shared scope are both refused; the canonical resolver choosing one Workspace must not weaken those reset safety checks.

After updating the branch, please rerun `pnpm check`, Web and Server typechecks/build, the Account API and account-scope-resolution tests introduced by #183, the focused Web/onboarding tests, and the onboarding-reset PostgreSQL integration suite including both lock-interleaving cases. The six green checks currently displayed belong to the pre-#183 head/base combination and do not validate the resolved merge result. Any new head will require a fresh exact-head review.

Gandy2025:
#183 merged, so I brought `main` into this branch in `62191d0`. The only conflict was the one predicted: both PRs add a registration block to `createApp`. Both blocks are independent, so the resolution keeps both and chooses nothing — the Account-native routes register first, then the staging Lab routes, and `packages/server/src/index.ts`, `packages/shared/src/http-paths.ts` and `packages/shared/src/index.ts` merged cleanly on their own.

No behaviour changed in the merge. In particular I did **not** fold this PR's `#resolveOwnedScope` into #183's new `AccountScopeResolver` seam here: the two resolve different things — `resolveCompatibilityWorkspaceId` returns the first server-ordered scope, while `#resolveOwnedScope` additionally refuses unless that scope is owned exclusively by the Lab Account — so unifying them is a real behavioural change that belongs in its own reviewed PR, not in a conflict resolution under an existing approval.

Verified on the merged tree before pushing: `pnpm check`, `pnpm build`, `pnpm typecheck`, Server unit 257, Web 322, and the Server integration suites against a real PostgreSQL instance — 126 tests, covering both this PR's reset suite and #183's new account-scope-resolution suite together. The two suites that provision their own container do not run in this sandbox; CI covers them.

@Gandy2025 asked me to merge. I will do that once CI is green on `62191d0`, and only if the standing approvals from @baixiaohang and @yuezengwu are still in force on it — if the merge commit dismissed them, this needs a fresh look rather than my judgement.

---

baixiaohang (changes requested):
- Rationale: The merge conflict is resolved correctly, but the fresh base now defines one canonical Account compatibility-scope seam and the Lab reset still bypasses it with a second resolver.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to inspect onboarding states quickly and repeatedly exercise the real staging first-run workflow without disposable Accounts or manual database edits.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews and an Account-configured reset that reuses existing lifecycle behavior before handing the tester back to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration and compatibility-scope resolution, shared HTTP/error contracts, Web authenticated routing/onboarding presentation and Account Settings, staging documentation.

**Review findings**
❌ 1. Route compatibility-scope selection through the canonical resolver introduced by #183. `WorkspaceAdminAccess.resolveCompatibilityWorkspaceId()` is now documented and tested as the only transitional Account-to-Workspace fact, but `#resolveOwnedScope()` still calls `listActiveAdminWorkspaces()` directly and establishes a second selection/cutover seam. The reset's stricter safety rules are additive: resolve the candidate through the canonical seam, then refuse unless the Account has exactly one active scope and that scope has exactly one active admin. This preserves the current fail-closed behavior while keeping the later Account-ownership cutover in one place. Add coverage that the canonical resolver is used and that both multiple active scopes and a shared scope are still refused before any mutation. [R1 / `packages/server/src/services/onboarding-lab/onboarding-reset-service.ts:117`, `packages/server/src/services/workspace-admin-access/workspace-admin-access.ts:163`]

✅ 2. The `app.ts` conflict resolution correctly preserves both #183's Account-native route registration and this PR's staging Lab registration plus `OnboardingResetError` mapping.

✅ 3. The previously approved Preview, reset-lock, authoritative handoff, saved-versus-unsynchronized state, and submit/refresh concurrency fixes remain unchanged in the merge commit.

**Action taken**
- Submitted request changes on exact head `62191d054daa949e46e9c0f043637858b15b5d9a`; fresh-head CI is still running.

Gandy2025:
@yuezengwu all three points are done. Head is now `45e0493`.

**1. `app.ts` conflict** — resolved in `62191d0` exactly as you describe: both registrations preserved, #183's Account-native routes and this PR's staging Lab route plus the `OnboardingResetError` mapping. `index.ts`, `http-paths.ts` and `shared/index.ts` merged automatically, matching your read.

**2. The seam.** You changed my mind here — I had deferred this to a follow-up because it is a behavioural change and I did not want to make one under a standing approval, but since you are asking for it and a new head needs fresh review regardless, that reason no longer applies. In `45e0493`, `#resolveOwnedScope` now takes its selection from `resolveCompatibilityWorkspaceId`, so there is one selection rule and the cutover replaces one thing rather than two.

The reset-only guarantees sit on top of it and still fail closed: the Lab Account must have exactly one active scope, and that scope must have exactly one active admin. Neither check selects anything — both only refuse — which is what keeps the canonical resolver from weakening them.

That distinction is worth being concrete about, because it is the actual risk in this change. The new multiple-active-scope test passes both before and after the refactor, since the old code also refused; on its own it would only pin a retained guarantee. So I checked what it is really worth by removing the count check and letting the canonical resolver stand alone — the reset then resolved successfully against one of two scopes, and the test failed with `promise resolved "undefined" instead of rejecting`. That is the silent-reset outcome you were guarding against, and it is now covered. The shared-scope refusal is retained unchanged.

**3. Validation on the resolved merge**, not the pre-#183 combination:

- `pnpm check`, `pnpm build`, Web and Server typechecks — pass;
- Server unit 257, Web 322;
- #183's suites on this head: `account-api` 9/9 and `account-scope-resolution` 5/5;
- Server PostgreSQL integration against a real instance: 127 tests, including the onboarding-reset suite at 13 with both lock-interleaving cases and the new refusal. The two suites that provision their own container do not run in this sandbox; CI covers them.

You are right that the previously green checks belonged to the old head/base pair. CI is running on `45e0493` now, and I am not merging on the strength of the earlier ones. @Gandy2025 asked me to merge, and I will once this head is green and re-approved — the branch has moved twice since your approval, so this needs your fresh exact-head look rather than my inference that nothing material changed.

---

yuezengwu (approved):
Approving exact head `62191d054daa949e46e9c0f043637858b15b5d9a` against base `2b1ff7e7d16bd1911be7688da0e2d7c0a9237643` after a complete fresh-base review.

The merge resolution is correct. The only textual conflict was `packages/server/src/app.ts`, and the resolved tree preserves both independent registrations plus the Onboarding reset error mapping. The automatically merged Server wiring and shared exports are present, Account-native routes work alongside the staging Lab routes, and no conflict markers or dropped #183 behavior remain.

I also re-evaluated the scope-resolver coordination rather than treating my earlier conflict comment as dispositive. `WorkspaceAdminAccess.resolveCompatibilityWorkspaceId` chooses the first deterministic compatibility scope for ordinary Account-native management. The reset has a deliberately narrower destructive precondition recorded on #180: it accepts only an Account with exactly one active scope whose sole active admin is that Lab Account, and refuses multi-scope or shared-scope Accounts. Within the only state the reset accepts, both paths necessarily identify the same Workspace; outside it, refusal is the safety behavior, not a competing selection. Replacing that assertion path with the ordinary choose-one resolver would weaken the reset unless the exclusivity checks were independently retained, while calling both would add a redundant read without changing the accepted result. I therefore do not treat the retained private check as a fresh-head blocker.

Fresh-head validation passed: `pnpm check`; Server and Web typechecks; full build; Account API and Lab route tests 17/17; focused Web tests 121/121; full PostgreSQL integration 154/154, including account-scope 5/5 and onboarding reset 12/12 with both lock-interleaving cases; and `git diff --check` plus conflict-marker audit.

This approval does not make the PR merge-ready. GitHub `Checks` failed twice in untouched `im-binding.test.ts`, on two different dispatch-fixture assertions; the same 77-test file passed in the local full integration run, and this PR has no diff in that test or the related IM/Session services. The other four check groups pass, but required `Checks` and aggregate `CI` remain red. I spent one failed-job rerun and will not repeat it; merge must wait for a green required run or separate triage of that flaky integration path.

baixiaohang (changes requested):
- Rationale: Scope selection now uses the canonical seam and preserves the multi/shared-scope refusal, but the zero-scope case bypasses the Lab's ownership error contract and regresses from a retryable 409 to 404 Not Found.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to inspect onboarding states quickly and repeatedly exercise the real staging first-run workflow without disposable Accounts or manual database edits.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews and an Account-configured reset that reuses existing lifecycle behavior before handing the tester back to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration and compatibility-scope resolution, shared HTTP/error contracts, Web authenticated routing/onboarding presentation and Account Settings, staging documentation.

**Review findings**
❌ 1. Preserve the Lab ownership error for an Account with zero active scopes. `#resolveOwnedScope()` now calls `resolveCompatibilityWorkspaceId()` before its exact-one guard; that canonical resolver throws `RESOURCE_NOT_FOUND`/404 when no scope exists, so the configured Lab Account never reaches the dedicated `ONBOARDING_RESET_OWNERSHIP_INCONSISTENT`/409 path it had on the previous head. GET still identifies that Account as an available Lab, while POST now answers like an unauthorized Account instead of reporting the retryable ownership inconsistency. Translate the resolver's no-scope result into the Lab reset error (without swallowing other failures), and add a zero-active-scope regression test that pins the 409 code and verifies no mutation. [R4 / `packages/server/src/services/onboarding-lab/onboarding-reset-service.ts:123`, `packages/server/src/services/workspace-admin-access/workspace-admin-access.ts:163`]

✅ 2. Selection now goes through `resolveCompatibilityWorkspaceId()`, while the additional exact-one-scope and exact-one-admin checks remain refusal-only guards. The new multiple-active-scope case and the retained shared-scope case protect the destructive-reset boundary.

✅ 3. The merge-conflict resolution and every previously approved Preview, reset-lock, handoff, Account state, and submit/refresh concurrency fix remain unchanged.

**Action taken**
- Submitted request changes on exact head `45e0493045e69e660b5d613e86dc8d45574c9bf7`; fresh-head CI is still running.

yuezengwu (changes requested):
Requesting changes on exact head `45e0493045e69e660b5d613e86dc8d45574c9bf7`.

The canonical selection refactor now does the intended structural work: reset obtains its candidate through `resolveCompatibilityWorkspaceId`, then separately refuses multiple active scopes and shared administration. The new multi-scope test and retained shared-scope test preserve those destructive-reset guards, and the merge resolution plus every earlier approved path remain unchanged.

One fresh-head contract regression remains, independently confirmed; I am not duplicating the existing inline/finding from @baixiaohang. Because `resolveCompatibilityWorkspaceId` runs before the exact-one guard, a configured Lab Account with zero active scopes now throws `RESOURCE_NOT_FOUND`/404 from the ordinary resolver instead of the prior Lab-specific `ONBOARDING_RESET_OWNERSHIP_INCONSISTENT`/409. GET still identifies that configured Account as Lab-eligible, so POST now looks unauthorized rather than reporting the retryable ownership inconsistency. Translate only the canonical resolver no-scope outcome into the Lab ownership error, preserve other failures, and add a zero-scope regression that asserts the 409 code and no mutation.

Fresh-delta validation passed: `pnpm check`, Server typecheck, onboarding reset integration 13/13, and account-scope integration 5/5. Those passing tests do not cover the zero-scope contract described above. Current fresh-head GitHub checks are still running.

Gandy2025:
@baixiaohang @yuezengwu both right, and this one I saw coming and let through. Fixed in `43e38b2`.

When I routed selection through the canonical resolver I noticed that a zero-scope Account would move from 409 to 404, decided it was "consistent with non-disclosure", and neither preserved it nor flagged it. That was the wrong call for the reason you both give: non-disclosure is right for ordinary management, where the caller may not be entitled to know a scope exists, but the Lab has already identified this Account as its own — `GET` answers 204 for it. Turning the same fact into 404 on `POST` makes an actionable, retryable state look like an unauthorized one.

`#selectScope` now translates only that outcome back to `ONBOARDING_RESET_OWNERSHIP_INCONSISTENT`/409, matched on `RESOURCE_NOT_FOUND` with status 404 so every other resolver failure propagates unchanged. The exactly-one-scope and exactly-one-admin guards are untouched and still refusal-only.

New integration test revokes the Lab Account's only active grant and asserts the 409 code and that nothing was mutated — Agents and enrollments still active, setup completion still set. Restoring the direct resolver call fails it with `expected AuthServiceError: The requested resource … to match object`, which is the regression you described.

@yuezengwu on the selection question you re-examined: your reading is right that within the state the reset accepts, both paths identify the same Workspace, so this refactor buys structure rather than behaviour. I still think routing through the seam is worth it — it means the ownership cutover replaces one selection rule instead of two — but it was not the safety fix I may have implied, and it is what produced this regression. Worth recording honestly.

**On the CI failures you both saw on `62191d0`:** they did not recur. `45e0493` came back green on all six checks, including `Checks` with the full integration suite. Looking at both failed attempts, they hit *different* tests in `im-binding.test.ts` — `requires old-placement reconciliation before moving an expired dispatch` on the first, `serializes placement movement with a concurrent accepted-result custody transition` on the second, which had passed in the first run. Both are bounded by `requestTimeoutMs: 100`, and that file has seven such fixtures.

My read: this is load-sensitive rather than broken. Both parent heads were green with seven integration files; the merge is the first time both new suites coexist, so `--maxWorkers=2` now runs eight files each with its own container. I could not reproduce it faithfully here — without Docker every file shares one database, so I must run with `fileParallelism=false`, which removes the very variable that matters. I am flagging it rather than claiming it is fixed: **once this merges, `main` inherits the same eight-file suite**, so the exposure moves there. The smallest fix I would propose is raising `requestTimeoutMs` for those lock-order fixtures, in its own change — it is in a file neither PR touches and it is not this PR's to widen.

`pnpm check`, `pnpm build`, both typechecks, Server unit 257, Web 322, and the Server integration suite at 128 against a real PostgreSQL instance all pass on this head. CI is running on `43e38b2`.

---

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The zero-active-scope regression is now translated back into the Lab's retryable ownership contract without weakening the canonical Account-scope seam or the reset's fail-closed guards.

**Risk level**: B-low
- Path baseline: includes `packages/server/**` outside the listed core-service paths -> B-low.
- Semantic lift: the Web API/routing slice touches its data layer (C -> B-low); no further lift.

**PR summary**
- Problem: Product and engineering need to inspect onboarding states quickly and repeatedly exercise the real staging first-run workflow without disposable Accounts or manual database edits.
- Approach: Add a staging-only, Account-selected Lab with inert scenario previews and an Account-configured reset that reuses existing lifecycle behavior before handing the tester back to ordinary onboarding.
- Impacted modules: Server configuration/API/reset orchestration and compatibility-scope resolution, shared HTTP/error contracts, Web authenticated routing/onboarding presentation and Account Settings, staging documentation.

**Review findings**
✅ 1. `#selectScope()` still obtains the candidate through `resolveCompatibilityWorkspaceId()`, but translates only the resolver's exact `AuthServiceError` `RESOURCE_NOT_FOUND`/404 outcome into `ONBOARDING_RESET_OWNERSHIP_INCONSISTENT`/409; every other failure is rethrown unchanged.

✅ 2. The new zero-scope integration case revokes the Lab Account's sole active grant, pins the 409 code, and verifies that the Agent, enrollment, and setup-completion marker remain untouched. The existing multiple-active-scope and shared-scope refusal cases remain present.

✅ 3. The fresh delta is confined to this error-contract repair and its regression coverage. The previously reviewed conflict resolution, Preview isolation, reset lock boundary, authoritative handoff, and Account Settings state/concurrency fixes are unchanged.

**Action taken**
- Approved exact head `43e38b2e60b921e232a299e21c792b008872a613`. Four current GitHub check groups pass; the required `Checks` job is still running, so merge readiness still depends on that run completing successfully.

yuezengwu (approved):
Approving exact head `43e38b2e60b921e232a299e21c792b008872a613` against base `2b1ff7e7d16bd1911be7688da0e2d7c0a9237643` after fresh-head review.

The zero-scope contract regression is fixed without weakening the canonical compatibility seam. `#selectScope` still obtains the candidate through `resolveCompatibilityWorkspaceId`, translates only the exact `AuthServiceError` `RESOURCE_NOT_FOUND`/404 outcome into the Lab-specific retryable ownership 409, and rethrows every other resolver failure. Exact-one-scope and exact-one-active-admin remain refusal-only guards on top of canonical selection.

The new integration case revokes the Lab Account sole grant, pins `ONBOARDING_RESET_OWNERSHIP_INCONSISTENT`/409, and proves that the Agent, enrollment, and setup-completion marker remain unchanged. The multiple-scope and shared-scope refusal cases, conflict resolution, reset lock boundary, inert Preview, authoritative handoff, and Account Settings state/concurrency fixes remain intact.

Fresh-delta validation passed: `pnpm check`, Server typecheck and build, onboarding reset integration 14/14, account-scope integration 5/5, and `git diff --check`. All six current GitHub checks pass on this exact head. No remaining blocking findings.

PR #183, opened 2026-08-26
feat(server): add the Account-native management facade

Gandy2025:
Phase 2 PR 2A for #179.

## Summary

Adds Account-native management collections that take ownership only from the authenticated Account, so clients stop selecting a management scope:

| New route | Delegates to |
| --- | --- |
| `GET` / `POST /api/v1/agents` | `AgentService.listForWorkspace` / `createForWorkspace` |
| `GET /api/v1/computers` | `WorkspaceAdminService.listComputers` |
| `POST /api/v1/computer-connect-codes` | `MachineAuthService.issueForWorkspaceAdmin` |
| `POST /api/v1/me/setup/complete` | `WorkspaceSetupService.complete` |

The whole compatibility window is concentrated in one method, `WorkspaceAdminAccess.resolveCompatibilityWorkspaceId`. It reuses `listActiveAdminWorkspaces`, so it is guaranteed to select the same Workspace that `/me` already publishes: setup-completed first, then earliest grant, then Workspace UUID. No active grant returns 404 without disclosing whether a scope exists. Removing that seam is a single deletion once Agents and enrollments carry a direct Account owner.

Every creation route refuses a client-selected scope rather than ignoring it: `workspaceId` and `accountId` are rejected with 400 on all three of `POST /agents`, `POST /computer-connect-codes`, and `POST /me/setup/complete`.

Everything else is unchanged on purpose:

- every legacy Workspace-scoped route stays registered for the compatibility window;
- the current strict DTOs are reused as-is, deprecated response `workspaceId` fields included;
- no schema, migration, owner column, dual write, mismatch diagnostic, or data mutation;
- the Agent execution workspace is untouched.

The facade lives in its own `packages/server/src/api/account.ts`, so the later cutover deletes `api/workspaces.ts` and the legacy handlers while this file survives.

### Note on an existing asymmetry

The legacy connect-code route is gated behind `options.computerService` (`app.ts`), which is really the gate for the runtime WebSocket route and unrelated to issuing connect codes. The Account-native route is gated on `machineAuthService` + `computerConnectCode` instead. Production wires both, so there is no behavioral difference; it only means the equivalence test has to register a `computerService` stub, which is called out in a comment there.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — all pass.
[...]

PR #184, opened 2026-08-26
feat(server): enforce one active Workspace admin grant per Account

Gandy2025:
Stacked on #183 — **base is `feat/account-native-management-facade`, so merge #183 first.** Authorized in https://github.com/first-tree-ai/opentag/issues/179#issuecomment-5424521735 as Phase 1 hardening, not as part of the R2 sequence.

## Summary

After Phase 1, the internal Workspace is a 1:1 shadow of the Account — but only by convention. `workspace_admin_grants_active_workspace_user_unique` constrains the `(workspace_id, user_id)` pair, not "one per Account", and the only two grant-creating paths in product code (default provisioning and the bootstrap CLI) each create exactly one. Nothing in storage stops a second grant from appearing.

If one ever did, the failure is asymmetric and one half is silent:

- the CLI fails hard — `selectWorkspace` throws "Multiple internal scopes are available; use the legacy `--workspace` selector";
- Web silently takes the first server-ordered grant and operates on it, so a write can land in the wrong scope with no error.

This adds a partial unique index on the active grant's `user_id`, making the 1:1 shape a storage fact. It also verifies, cheaply and before any irreversible step, the assumption the #179 D2 reset rests on: if the index cannot be created on a deployed database, that database already violates the 1:1 model, and it is much better to learn that now than during the 2D cutover.

Revoking a grant and issuing a replacement stays allowed — the onboarding end-to-end script depends on exactly that sequence.

### Migration reports the offending data

A bare `CREATE UNIQUE INDEX` would fail with a duplicate-key error naming only the index. The migration checks first and raises with the Account ids:

```
ERROR:  Accounts hold more than one active Workspace admin grant: aaaaaaaa-…
HINT:   Revoke the redundant grants so that every Account keeps exactly one, then re-run the migration.
```

### Tests that encoded the removed assumption

Two integration tests deliberately gave one Account two active grants. Both were really testing *scope* isolation, so they now span two Accounts — which is also the property that survives the move to direct Account ownership, and matches #179's "Account A cannot discover or mutate Account B's resources" criterion:

- `agents.test.ts` — Agent name reuse across scopes and cross-scope Computer isolation;
- `computers.test.ts` — two enrollments of one physical Computer, each with its own credential and instance. #179 explicitly keeps this property, so it is preserved rather than dropped.

The compatibility resolver's multi-grant ordering is now unreachable by construction, which is a stronger guarantee than "deterministic". Its three ordering tests are replaced by the constraint itself plus revoke-and-replace coverage, and the resolver's doc comment no longer describes the ordering as a choice. This supersedes the "deterministic multi-grant selection" wording in #183's description.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — all pass (server: 248).
- The migration was run against a real PostgreSQL 16 instance in four scenarios: a clean upgrade applies and creates the index; a second active grant is rejected by the index; revoke-then-regrant still succeeds; and with two active grants present the guard aborts naming the offending Account id.
- The integration suites cannot run in the authoring sandbox (testcontainers images are unreachable there), so CI is the first execution of the three rewritten test files.

## Breaking changes

None for clients. Operationally, an upgrade fails with a named diagnostic if any Account already holds more than one active grant; a pre-flight check on staging is worth running before deploying:

```sql
SELECT user_id, count(*) FROM workspace_admin_grants
WHERE revoked_at IS NULL GROUP BY user_id HAVING count(*) > 1;
```

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDUUR1XktSHvn9vy9p1x6u)_

### DIFF
--- a/packages/server/drizzle/0017_volatile_sue_storm.sql
+++ b/packages/server/drizzle/0017_volatile_sue_storm.sql
@@ -1,0 +1,23 @@
+-- An Account holds at most one active Workspace admin grant. Report the offending Accounts by id
+-- instead of failing with a bare duplicate-key error on the index created below.
+DO $$
+DECLARE
+	offending text;
+BEGIN
+	SELECT string_agg(user_id::text, ', ' ORDER BY user_id)
+	INTO offending
+	FROM (
+		SELECT "user_id"
+		FROM "workspace_admin_grants"
+		WHERE "revoked_at" IS NULL
+		GROUP BY "user_id"
+		HAVING count(*) > 1
+	) AS duplicates;
+
+	IF offending IS NOT NULL THEN
+		RAISE EXCEPTION 'Accounts hold more than one active Workspace admin grant: %', offending
+			USING HINT = 'Revoke the redundant grants so that every Account keeps exactly one, then re-run the migration.';
+	END IF;
+END
+$$;--> statement-breakpoint
+CREATE UNIQUE INDEX "workspace_admin_grants_active_user_unique" ON "workspace_admin_grants" USING btree ("user_id") WHERE "workspace_admin_grants"."revoked_at" is null;
--- a/packages/server/drizzle/meta/0017_snapshot.json
+++ b/packages/server/drizzle/meta/0017_snapshot.json
@@ -1,0 +1,3211 @@
+{
+  "id": "8bb6ad1d-fb12-45d9-9ed0-4b6c1ddc7113",
+  "prevId": "43db9b5d-8997-4ff8-83f5-1f43980fd5e9",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.agent_runtime_configs": {
+      "name": "agent_runtime_configs",
+      "schema": "",
+      "columns": {
+        "agent_id": {
+          "name": "agent_id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "revision": {
+          "name": "revision",
+          "type": "bigint",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "nextval('runtime_config_revision_sequence')"
+        },
+        "model": {
+          "name": "model",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "reasoning_effort": {
+          "name": "reasoning_effort",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "instructions": {
+          "name": "instructions",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "max_duration_ms": {
+          "name": "max_duration_ms",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {
+        "agent_runtime_configs_agent_id_agents_id_fk": {
+          "name": "agent_runtime_configs_agent_id_agents_id_fk",
+          "tableFrom": "agent_runtime_configs",
+          "tableTo": "agents",
+          "columnsFrom": [
+            "agent_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {
+        "agent_runtime_configs_revision_safe_positive": {
+          "name": "agent_runtime_configs_revision_safe_positive",
+          "value": "\"agent_runtime_configs\".\"revision\" > 0 and \"agent_runtime_configs\".\"revision\" <= 9007199254740991"
+        },
+        "agent_runtime_configs_max_duration_valid": {
+          "name": "agent_runtime_configs_max_duration_valid",
+          "value": "\"agent_runtime_configs\".\"max_duration_ms\" is null or (\"agent_runtime_configs\".\"max_duration_ms\" > 0 and \"agent_runtime_configs\".\"max_duration_ms\" <= 86400000)"
+        }
+      },
+      "isRLSEnabled": false
+    },
+    "public.agents": {
+      "name": "agents",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "workspace_id": {
+          "name": "workspace_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "created_by_user_id": {
+          "name": "created_by_user_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "creation_intent_id": {
+          "name": "creation_intent_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "creation_intent_fingerprint": {
+          "name": "creation_intent_fingerprint",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "workspace_computer_id": {
+          "name": "workspace_computer_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "name": {
+          "name": "name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "display_name": {
+          "name": "display_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "runtime_provider": {
+          "name": "runtime_provider",
+          "type": "agent_runtime_provider",
+          "typeSchema": "public",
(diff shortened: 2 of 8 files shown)


baixiaohang (changes requested):
- Rationale: The storage invariant is directionally correct, but the exact head leaves the PostgreSQL integration contract red in three places, including the supported forward-upgrade fixture that now aborts at migration 0017.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes a database partial unique index and migration -> B-high
- Semantic lift: none beyond the database-structure baseline

**PR summary**
- Problem: An Account can currently acquire multiple active compatibility Workspace grants, making the CLI fail and allowing Web to operate on an implicitly selected scope.
- Approach: Enforce at most one active grant per Account with a partial unique index, fail upgrades with an Account-specific diagnostic when existing rows violate it, and retain revoke-then-regrant behavior.
- Impacted modules: Server auth schema and migration ledger, Workspace compatibility resolution, Account-scope/Agent/Computer integration tests

**Review findings**
❌ 1. Bring the forward-migration contract and constraint assertion onto 0017 [R4/R5 / `packages/server/src/__tests__/integration/auth-migrations.test.ts:90`, `packages/server/src/__tests__/integration/auth-migrations.test.ts:525`, `packages/server/src/__tests__/integration/account-scope-resolution.test.ts:87`]

The current CI run has three PostgreSQL integration failures caused by this change. The journal assertion still treats 0010–0016 as the final seven entries; the deployed-0005/0006 forward-upgrade fixture deliberately gives one user two active Team memberships, which 0016 turns into two active Workspace grants, so 0017 now aborts before that supported upgrade test reaches its post-migration assertions; and the duplicate-grant test matches the constraint name against Drizzle's wrapper message even though the PostgreSQL constraint is only available through the nested cause. Update the valid forward-upgrade fixture plus its migration-count/journal expectations so it reaches 0017, add a separate dirty-0016 upgrade case that asserts the promised Account-id diagnostic and rollback/retry boundary, and assert the runtime index rejection through an observable field that the driver actually exposes. The branch should not merge with its sole automated migration path failing.

✅ 2. The schema and generated snapshot otherwise agree on the intended partial uniqueness, and the rewritten Agent/Computer isolation cases preserve cross-Account behavior while revoke-then-regrant remains representable.

**Action taken**
- Submitted request changes on exact head `209a5d2c89a9d5f387e7b7b47dac2c34f272d0a3`.

Gandy2025:
CI is red on `209a5d2` and the failure is this PR's. Holding for a decision rather than pushing a fix, because every available fix changes behaviour beyond what this PR claimed.

**What fails**

`Checks` → `auth-migrations.test.ts` → "upgrades deployed 0005 and 0006 Drizzle histories through each forward migration exactly once". Migration `0017`'s guard aborts:

```
Accounts hold more than one active Workspace admin grant: 0c17dcae-…
```

**Why it is correct to abort**

The guard is right and the premise of this PR was wrong. `0016` creates one grant per active admin membership:

```sql
INSERT INTO workspace_admin_grants (workspace_id, user_id, granted_by_user_id, granted_at)
SELECT memberships.team_id, memberships.user_id, memberships.user_id, memberships.created_at
FROM memberships
WHERE memberships.role = 'admin' AND memberships.status = 'active'
```

The legacy Teams model allowed one user to be an admin of several Teams, so `0016` is itself a grant-creating path with 1:N cardinality. The fixture exercises exactly that — one user admins both `migration` and `migration-control` — and it is a faithful legacy shape, not incidental test data.

So "one active grant per Account" holds for databases created after `0016`, and does **not** hold for any database upgraded through it that had a multi-Team admin. My PR description asserted the 1:1 shape was true in practice; that assertion was derived from the runtime grant-creating paths only and missed the migration.

This is the constraint doing the job it was added for — verifying the assumption before an irreversible step — and the answer came back negative.

**Why I am not just fixing it**

The secondary scopes are reachable: Web takes the first server-ordered grant, but the CLI's `--workspace <name>` selector resolves any Workspace in `/me`. Revoking the extra grants to satisfy the index would therefore remove access a legacy multi-Team admin can still use today. Weakening the fixture to make the migration pass is not on the table — it would delete coverage of a real upgrade path to accommodate a new constraint.

The options are a deterministic repair in the migration (keep the grant the product already selects, revoke the rest, and accept the CLI access change), a read-only diagnostic with no index and no enforcement, or dropping this PR and recording the finding against #179 instead. That choice belongs to the author, not to me, so it is with them now.

Note for #179 independently of this PR: the design's target model and the D2 reset are unaffected — all of these rows are deleted at cutover — but any text describing the internal Workspace as a 1:1 shadow of the Account should be qualified to "for databases created after `0016`".

---

baixiaohang (commented):
**Recommendation**: request changes remains
- Rationale: The red forward-upgrade fixture is faithful legacy data, so it disproves the PR's compatibility premise rather than merely exposing stale test maintenance.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes a database migration and partial unique index -> B-high
- Semantic lift: the available repairs would revoke currently reachable legacy scope access, which is an authorization-behavior change

**PR summary**
- Problem: The PR tries to guarantee one active compatibility Workspace grant per Account before the direct Account-ownership cutover.
- Approach: Migration 0017 rejects existing duplicates and then creates a partial unique index; the new evidence shows that 0016 legitimately creates duplicates when one legacy user administers multiple Teams.
- Impacted modules: Legacy Team-to-Workspace migration compatibility, CLI/Web scope reachability, Workspace grant schema, Account-native facade assumptions

**Review findings**
❌ 1. Correction to my prior requested change: do not weaken the deployed-0005/0006 fixture to make 0017 pass [R2/R5 / `packages/server/src/__tests__/integration/auth-migrations.test.ts:525`, `packages/server/drizzle/0017_volatile_sue_storm.sql:3`]

The author's follow-up is correct: the fixture preserves a real legacy shape in which one Account administers multiple Teams, and 0016 deliberately maps every active admin membership to a grant. Its failure therefore demonstrates that the claimed pre-existing 1:1 invariant is false for supported upgraded databases. A migration that deterministically revokes secondary grants would remove currently reachable CLI access and needs an explicit data/access decision; a diagnostic-only migration would no longer establish the invariant this PR claims. My recommendation is to drop this PR, keep #183's deterministic multi-grant compatibility resolver, and record the legacy 1:N qualification in #179. If the team instead chooses repair or diagnostic-only behavior, that should be separately authorized and tested without rewriting the faithful fixture.

⚠️ 2. The journal-tail and wrapped-error assertions remain mechanically red, but fixing them is downstream of the strategy decision rather than a sufficient resolution by itself.

**Action taken**
- Submitted this correction on exact head `209a5d2c89a9d5f387e7b7b47dac2c34f272d0a3`; the existing request-changes review remains in force.

yuezengwu (changes requested):
Reviewed exact head `209a5d2c89a9d5f387e7b7b47dac2c34f272d0a3`.

Blocking strategy and authorization gate: #179 still says implementation is not authorized and that existing grants remain live authorization facts. The new authorization record was posted by this PR author and explicitly retains the requested human review as the blocker, so it cannot decide this access and migration tradeoff.

The exact migration evidence also disproves the PR premise. Migration 0016 maps every active legacy Team Admin membership to a Workspace grant; the supported deployed-history fixture legitimately has one Account administering two Teams, so 0017 aborts on the resulting 1:N grants. A deterministic repair would revoke currently reachable CLI scope access, a diagnostic-only change would not establish the claimed invariant, and weakening the historical fixture would hide a supported upgrade shape. Please do not merge this branch until the authorized reviewer chooses among those semantics. My recommendation is to drop #184, retain #183 deterministic multi-grant compatibility, and record the legacy 1:N qualification in #179; if repair or diagnostic-only behavior is preferred, authorize and test it explicitly.

I independently reproduced the current failures: the focused migration run has 3 failures and 26 passes, including the faithful 0005/0006 forward-upgrade abort, stale journal-tail expectation, and wrapped duplicate-index assertion. `pnpm check`, `pnpm build`, `pnpm typecheck`, `git diff --check`, and the affected Agent/Computer integration suites (31 tests) pass. The generated snapshot differs from 0016 only by the intended partial unique index. Context Tree `first-tree-ai/opentag-context-tree@d4bb6ba3dd69525af73caf2bd6a5145492b230ab` verified clean.

I am not duplicating the existing detailed inline findings; this review records the independently verified exact-head blocker.

Gandy2025:
Closing this. The constraint should not land, and the reason is worth recording.

**The premise was wrong.** `0016` converts one grant per active admin membership, and the legacy Teams model allowed a user to admin several Teams. So "one active grant per Account" holds only for databases created after `0016`. I derived the 1:1 claim from the runtime grant-creating paths and missed the migration itself.

**What this PR would actually have added is retroactive enforcement, and that is the half that is false.** The forward guarantee already exists in code: `establishDefaultWorkspaceForNewAccount` throws when a newly created Account already holds a grant, Admin invitations are gone, and bootstrap runs once. The index buys nothing new going forward.

**The hazard I cited was overstated.** For an Account that already holds several grants the selection is deterministic and stable — it is exactly the documented Phase 1 behaviour, not an erratic wrong-scope write. The real consequence is that resources in the secondary scope are invisible to Web, which is a visibility limitation rather than a correctness one.

**Every repair costs reachable access.** The CLI's `--workspace` selector resolves any Workspace in `/me`, so revoking the extra grants to satisfy the index would remove access a legacy multi-Team admin can use today, inside a change presented as non-behavioural hardening, on a population we cannot size from here. Weakening the migration fixture to make the index apply was never an option.

**The diagnostic value is already banked.** Verifying the 1:1 assumption before an irreversible step was the main reason to add this; it did that on its first CI run and returned a negative. Nothing needs to merge to keep that result — it is recorded on #179 instead, including its bearing on the D2-versus-D4 choice.

One consequence for #183, which stays as it is: the deterministic multi-grant ordering in `resolveCompatibilityWorkspaceId` is **not** unreachable, so its integration coverage is load-bearing rather than defensive. This PR would have replaced exactly the tests that matter most. Good that it did not merge.

The branch is left in place should the decision be revisited.

---

PR #185, opened 2026-08-26
feat(web): move management flows onto the Account-native routes

Gandy2025:
Phase 2 PR 2B for #179, on top of the facade merged in #183. Authorization recorded in https://github.com/first-tree-ai/opentag/issues/179#issuecomment-5427592945.

## Summary

Web derives its management scope from the authenticated Account instead of selecting one. Agent, Computer, connect-code and setup flows call the Account-native collections, and **no management `workspaceId` is sent anywhere** — `grep` for the legacy paths in `apps/web/src` returns nothing.

| Before | After |
| --- | --- |
| `agents(workspaceId)` | `agents()` |
| `createAgent(workspaceId, input)` | `createAgent(input)` |
| `computers(workspaceId)` | `computers()` |
| `issueComputerConnectCode(workspaceId)` | `issueComputerConnectCode()` |
| `completeWorkspaceSetup(workspaceId, agentId)` | `completeSetup(agentId)` |

`ComputerSetup` and `OnboardingPage` lose their scope props entirely; `AgentCreationFlow` takes an `accountId` in place of a `workspaceId`. The **only** remaining membership dependency is the setup-completed gate in the router, which moves with the `/me` contract at the ownership cutover — the onboarding page no longer takes a membership at all, and its tests no longer construct a `MeWorkspace`.

### Creation intents

Re-keyed from the internal Workspace to the Account, with the store version bumped so records written under the old scoping are never read. A prune removes superseded records **by stored version**, so another Account's current records on a shared browser are left alone.

### One shared-package change

`AgentIdentitySchema.workspaceId` becomes optional and deprecated. No Account-native consumer reads it and the Server still sends it, so this is a no-op today — but it means removing the field at the cutover is not a breaking contract change for clients published before then. Left required, a CLI published at 2C would fail to parse Agent responses the moment 2D stops sending it, and the "2C-compatible package is published" gate would not actually be satisfiable.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — all pass. Web 322, Server 257, Shared 67, CLI 124.
- `app.test.tsx`'s fetch mock now serves the Account-native paths, and the multi-membership case is stronger than before: it previously asserted Web called the *first* Workspace's collection, and now asserts Web sends **no** `/api/v1/workspaces/` request at all while still exercising a `/me` with several memberships. That case matters more than it looked — per the finding recorded on #179, multiple active grants are real on databases upgraded through `0016`.
[...]

PR #186, opened 2026-08-26
feat(cli): move Client and CLI management onto the Account-native routes

Gandy2025:
Phase 2 PR 2C for #179, on the facade merged in #183. Authorization recorded in https://github.com/first-tree-ai/opentag/issues/179#issuecomment-5427592945. Independent of #185 (2B) — different packages, no shared files, either order merges cleanly.

## Summary

**Management routes.** `OpenTagApi.listAgents`, `createAgent`, `listWorkspaceComputers` and `issueComputerConnectCode` drop their `workspaceId` parameter and call the Account-native collections.

**Scope selection is deleted, not deprecated.** `--workspace` is gone from `agent list`, `agent create` and `computer list`, and `core/selection/workspace.ts` is removed. With the scope derived from the authenticated Account there is nothing left to choose, so the "Multiple internal scopes are available" error path disappears rather than being reworded. `selectWorkspace` also leaves the CLI's public surface.

Method *names* are untouched — `listWorkspaceComputers` reads oddly with no workspace argument, but renames belong to 2E per the PR boundaries in #179, and mixing them in would make this diff much harder to read.

## Dual-shape readiness

This is the part that has to land before the cutover rather than with it.

`ComputerConnectCodeExchangeResponseSchema.workspaceId` and `MachineEnrollmentCredential.workspaceId` both become optional and deprecated. Left required, a Client published at this step would fail to parse the connect-exchange response the moment 2D stops sending the field — so the "a compatible CLI/daemon package is published before cutover" gate could not actually be satisfied by anything published today.

### Reading and writing the credential file disagree on purpose

**Reads** are validated **per entry** rather than as a whole. Previously one unreadable entry threw and rejected the entire file, which would strand every other enrollment on that host — a real hazard precisely at a cutover that invalidates some entries. Unusable entries are dropped, the rest still load, and an envelope that is not a credentials file at all still throws.

**Writes** reject what reads tolerate. A write is a request to persist *specific* credentials, so dropping one would report success while losing an enrollment, and the next read would discard the bytes just written. Unusable entries and duplicate enrollment ids are refused, and the validated projection — not the caller's object — is what reaches disk. The refusal also rejects rather than throwing synchronously, so a caller handling the returned promise sees it.

That asymmetry was missing from the first two heads and is the review finding fixed in `7fb4a62`; see [the reply on this PR](https://github.com/first-tree-ai/opentag/pull/186#issuecomment-5430595240) for the reproduction.

Deduplication keys on the enrollment id and compares the legacy scope only when **both** entries carry one, so a re-enrolment arriving without it still replaces the credential it supersedes. A second Account's enrollment of the same physical Computer is still kept — the property #179 explicitly preserves.

### Deliberately not decided here

What happens to credentials the cutover invalidates. Removal follows from discarding data; migration or rotation follows from preserving it. That is the carve-out recorded on #179 and it waits for the data-strategy selection. The daemon's error path for server-rejected tokens is coupled to the same decision and is likewise untouched.

Worth noting for the architecture review's gate 2: the daemon already runs one independently authenticated runtime per enrollment (`core/daemon/runtime.ts`), keyed by enrollment id with per-enrollment fencing and lifecycle. That is option (a) of the two contracts the gate asks Phase 2 to choose between, and this PR keeps it.

## Consumers of the removed flag

Caught by CI rather than by me, and fixed in `3ca95f0`:

- the **Container Smoke** passed `--workspace example` to `agent create` and `agent list`, so Commander rejected the arguments and printed usage;
- **`DEVELOPMENT.md`** and its Chinese mirror still described the selector as visible and functional until Phase 2.

Both are updated. The smoke's direct `fetch` of the legacy Workspace-scoped connect-code route is deliberately left as it is: that route stays registered through the compatibility window, and continuing to exercise it has value.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — Shared 67, Server 257, Web 322, CLI 124, Client 470 pass.
- **New** `packages/client/src/__tests__/machine-credentials.test.ts` (9 tests). The credential store had no direct coverage at all, which is uncomfortable for a file that holds machine tokens. It covers round-tripping with and without the legacy scope, a file mixing both shapes, dropping an unusable entry without stranding the others, rejecting a non-credentials envelope, supersession on both rotation and re-enrolment, and keeping a second Account's enrollment of one Computer. Three more pin the write side: a write the reader would discard is refused and leaves the existing credential untouched, a duplicate the reader would collapse is refused, and the bytes on disk equal what the reader returns.
- The CLI's scope-selector test is replaced by one asserting the opposite property: no management scope is sent and `me()` is not even called, while `/me` reports several memberships.
- Both review findings were reproduced before fixing. The Container Smoke failure: with `--workspace` the built CLI exits on `unknown option '--workspace'`; without it the command gets past argument parsing. The credential-writer defect: writing an entry whose only field is `workspaceComputerId: "not-a-uuid"` resolved successfully and the next read returned `enrollments: []`.

Two `@opentag/client` Codex process-group tests fail in the authoring sandbox. They fail identically on the previous head and are green in CI.

## Breaking changes

`--workspace` is removed from three CLI commands. It was documented as a legacy internal-scope selector, is unreachable in normal use now that an Account resolves to one scope server-side, and #179 schedules its removal at this step. No schema, migration, data mutation, or authorization-rule change; every legacy Server route stays registered.

## Checklist

- [x] Canonical English docs and Chinese mirrors are synchronized when applicable. (`DEVELOPMENT.md` and `DEVELOPMENT.zh-CN.md` updated together; the mirror's sync date is already today's.)


baixiaohang (changes requested):
- Rationale: The CLI removes the management-scope selector, but the repository's end-to-end workflow still passes the deleted option and the exported creation contract still accepts a now-ignored selector.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A
- Semantic lift: none; A is already the highest level, while the change also touches credential parsing and the CLI public contract.

**PR summary**
- Problem: Account users should manage Agents and Computers without choosing a legacy internal Workspace, and a Client published before the ownership cutover must accept both current and future enrollment credentials.
- Approach: Route Client management calls through the Account-native collections, remove CLI scope selection, make the legacy credential scope optional, and salvage valid entries from a mixed credential file while preserving one runtime per enrollment.
- Impacted modules: CLI Agent/Computer commands and core, Client API and credential storage, shared connect-exchange schema, tests, and the repository E2E workflow that exercises those commands.

**Review findings**
❌ 1. The E2E workflow still invokes `agent create --workspace example` and `agent list --workspace example` after this PR deletes those options. Commander will reject the workflow's own commands before the Account-native path is exercised, so the repository validation no longer matches the shipped CLI. Remove the three selector arguments and keep the scenario asserting the same create/list lifecycle through the new routes. [R5 / `.github/workflows/ci.yml:356`]
❌ 2. Scope selection is not fully removed from the reusable/public CLI contract: `AgentCreateOptions.workspaceName` remains accepted but is silently ignored, the selection error still tells users a Computer is outside the "selected internal scope," and both DEVELOPMENT guides still say `--workspace` is visible and functional. Remove the dead input and update the canonical English text, Chinese mirror, and error wording in this PR so no caller or operator is left on a coexistence path that no longer exists. [R2, R5 / `apps/cli/src/core/agent/mutations.ts:22`, `apps/cli/src/core/agent/mutations.ts:48`, `DEVELOPMENT.md:236`, `DEVELOPMENT.zh-CN.md:230`]

**Action taken**
- Submitted request changes on head `2f0d41c92da3b2d42454b933928d487ab622ed1d`.

baixiaohang (changes requested):
- Rationale: The new commit fixes the broken smoke commands and selector documentation, but the exported creation API still accepts a selector that has no effect and its user-facing failure still claims a scope was selected.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A
- Semantic lift: none; A is already the highest level, while the change also touches credential parsing and the CLI public contract.

**PR summary**
- Problem: Account users should manage Agents and Computers without choosing a legacy internal Workspace, and a Client published before the ownership cutover must accept both current and future enrollment credentials.
- Approach: Route Client management calls through the Account-native collections, remove CLI scope selection, accept both credential shapes, and keep independent daemon runtimes per enrollment.
- Impacted modules: CLI Agent/Computer commands and core, Client API and credential storage, shared connect-exchange schema, documentation, and CI smoke coverage.

**Review findings**
❌ 1. The remaining half of the prior public-contract finding is unchanged on this head. `runAgentCreate` is exported from the package, but its `AgentCreateOptions` still exposes `workspaceName`; the implementation never reads it, so reusable callers can pass a management selector that is silently ignored. The same path then reports that an unknown Computer is outside the "selected internal scope," reinforcing a choice that no longer exists. Remove the dead property, change the error to the authenticated Account boundary, and update the two matching test expectations. [R2, R5 / `apps/cli/src/core/agent/mutations.ts:22`, `apps/cli/src/core/agent/mutations.ts:48`, `apps/cli/src/index.ts:3`]
✅ 2. The previous Container Smoke failure and the primary English/Chinese `--workspace` guidance are fixed in `3ca95f0`.

**Action taken**
- Submitted request changes on head `3ca95f04d401e5b0f62d80a3a4f3c75007ce9112`.

Gandy2025:
@baixiaohang both findings are addressed. Current head is `1312686`.

**Finding 1 — Container Smoke arguments.** Fixed in `3ca95f0`, which was pushed before your review landed on `2f0d41c`. All three `--workspace example` arguments are gone and the create/list lifecycle assertions are unchanged, so the scenario now exercises the same lifecycle through the Account-native routes. CI caught this before I did — I had only grepped `apps/cli/src` and `packages/client/src`, not the workflow.

**Finding 2** had two parts on different heads:

- The DEVELOPMENT guides were also in `3ca95f0`. The canonical English text and the Chinese mirror were updated together, and the mirror's sync date is already today's.
- The dead input and the error wording are new, in `1312686`: `AgentCreateOptions.workspaceName` is removed, so a programmatic caller can no longer pass a selector and have it silently ignored, and the Computer selection error now reads "is not enrolled by this Account" rather than naming a scope nobody can choose. Its two test assertions moved with it.

**One thing I deliberately did not change**, so it does not read as an oversight next round: `DEVELOPMENT.md:222-225` and `:249`, with their mirror equivalents, still describe an Agent as manageable "through an active internal scope" and note that legacy active grants can expose the same Agents to several Accounts. Those paragraphs describe **Server** authorization, which still resolves through grants and which this PR does not touch — only clients stopped choosing. Rewriting them here would document the ownership cutover's model before it is true. They belong to 2D, with the authorization change itself.

Verification: `pnpm check`, `build`, `typecheck` pass; CLI 124 and Client 468 tests pass. Before fixing the smoke I reproduced its failure against the built CLI — with `--workspace` it exits on `unknown option '--workspace'`, without it the command gets past argument parsing.

Separately, and not caused by this PR: GitHub dispatched no workflow run for either this PR or #185 when they were opened — no runs at all across every workflow between 14:18 and 16:13. I triggered `ci.yml` manually via `workflow_dispatch` on both branches, which is where the check results on this PR came from. Worth knowing in case checks look absent again after this push.

---

yuezengwu (changes requested):
- Rationale: the Account-native route and CLI-selector changes are now coherent, but the credential writer can report success while persisting an entry that its own reader rejects.

**Risk level**: A
- Path baseline: this changes apps/cli/** and packages/client/**, including a secret-bearing local persistence boundary.

**PR summary**
- Problem: Client and CLI management consumers should stop selecting a legacy management Workspace, while a pre-cutover Client must accept connect credentials with or without the deprecated Workspace field.
- Approach: move management calls to the Account-native facade, delete CLI scope selection, make the deprecated credential scope optional, and parse stored enrollment entries independently.
- Impacted modules: CLI Agent/Computer commands and public surface, Client management API, machine credential storage, shared connect-exchange schema, docs, and container smoke.

**Review finding**
❌ The exported `writeMachineCredentialsAtomically` no longer validates what it writes. `validateMachineCredentials` now returns a filtered projection, but the caller discards that result and writes the original object. I reproduced this by passing an entry whose only field is `workspaceComputerId: "not-a-uuid"`: the promise resolves and the malformed entry is present on disk, while the next `readMachineCredentials` silently returns `{ version: 1, enrollments: [] }`. That makes an apparently successful secret-bearing persistence call lose the enrollment on the next read, and it also defeats the function's existing runtime-validation boundary. Either reject invalid write input or persist the validated projection, and add a regression that proves a writer cannot leave bytes on disk that its reader immediately discards. [R3 / `packages/client/src/auth/machine-credentials.ts:32-37`]

**Prior findings rechecked**
✅ The Container Smoke no longer passes the deleted selector.
✅ `AgentCreateOptions.workspaceName`, the public selector export, stale CLI error wording, and primary English/Chinese selector guidance are removed on this head.

**Validation**
- Reviewed exact head `1312686fb87e83a0459c537038a9f3eb9348c113`; `git diff --check`, `pnpm check`, `pnpm build`, and `pnpm typecheck` pass.
- Targeted Shared/Client/CLI suites: 33 tests pass. GitHub CI, compatibility, package, and container checks are green.
- The full local test run has one persistent failure in an untouched Server observability timing test; it is outside this diff and the corresponding GitHub CI is green.

**Action taken**
- Submitted request changes on exact head `1312686fb87e83a0459c537038a9f3eb9348c113`.

baixiaohang (changes requested):
- Rationale: The credential writer now accepts entry-level invalid data, persists the unvalidated object, and only lets the reader discard it later.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**` -> A
- Semantic lift: none; A is already the highest level, while the defect is at the exported machine-token persistence boundary.

**PR summary**
- Problem: Account users should manage Agents and Computers without choosing a legacy Workspace, while a pre-cutover Client must accept both current and future enrollment credential shapes.
- Approach: Move Client/CLI management calls to Account-native routes, delete selector inputs, make the legacy credential scope optional, and validate stored enrollment entries independently.
- Impacted modules: CLI Agent/Computer commands and public surface, Client API and credential storage, shared connect-exchange schema, documentation, and CI smoke coverage.

**Review findings**
❌ 1. I agree with the new credential-persistence finding. `validateMachineCredentials` now deliberately returns a sanitized projection after dropping invalid or duplicate entries, but `writeMachineCredentialsAtomically` ignores that return value and passes the original object to `writePrivateJson`. An envelope-valid value containing an unusable entry therefore reports a successful write, yet its next read returns different data or no enrollment at all. Preserve the existing runtime-validation guarantee by either rejecting invalid writer input or persisting the validated projection, and add a direct round-trip regression showing that a successful writer cannot produce bytes its reader immediately discards. [R3 / `packages/client/src/auth/machine-credentials.ts:32`]

**Prior findings rechecked**
✅ 2. Container Smoke, selector documentation, `AgentCreateOptions.workspaceName`, the selector export, and the stale Computer error were all corrected on this head.

**Action taken**
- Submitted request changes on head `1312686fb87e83a0459c537038a9f3eb9348c113`.

Gandy2025:
@yuezengwu the finding is correct, and it is a regression this PR introduced. Fixed in `7fb4a62`.

On `main`, `validateMachineCredentials` threw on any bad entry, so the bare `validateMachineCredentials(credentials)` in `writeMachineCredentialsAtomically` really was rejecting invalid write input. Making it filter for the read path silently disarmed that guard — the validator's filtered projection was discarded and the original object written. I reproduced your exact case before fixing: the write resolves and the next `readMachineCredentials` returns `{ version: 1, enrollments: [] }`.

The two directions now have separate policies, which is the asymmetry I should have written in the first place:

- **Read** still drops unusable entries. A file on disk may hold entries this Client cannot use, and one of them must not strand the rest of a host's enrollments.
- **Write** rejects them. A write is a request to persist *specific* credentials, so dropping one would report success while losing an enrollment. Unusable entries and duplicate enrollment ids (which the reader would collapse) are both refused, and the validated projection — not the caller's object — is what reaches disk.

One thing I hit while writing the regression, worth flagging since it predates this PR: the refusal was thrown **synchronously** from a function typed `Promise<void>`, so `writeMachineCredentialsAtomically(...).catch(handle)` would not have caught it. It is now `async`, so it rejects.

Three regressions cover it, all failing against the previous code:

- a write the reader would discard is refused, and the credential already on disk is untouched;
- a duplicate the reader would collapse is refused;
- the bytes on disk equal what the reader returns (an unknown extra field on an entry does not survive the write).

`pnpm check`, `pnpm build`, `pnpm typecheck` pass; Shared 67, Web 322, Server 257, CLI 124 pass, Client 470/472 with the same two Codex process-group failures that fail identically on the previous head in this sandbox and are green in CI.

---

yuezengwu (approved):
**Recommendation**: approve
- Rationale: the fresh head fixes the credential-writer regression without weakening the intentional lenient read path. Invalid and duplicate write inputs now reject before touching disk, successful writes persist only the validated projection, and the new regressions cover the exact write/read mismatch.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**`, including the secret-bearing machine-credential boundary.

**Fresh-head recheck**
✅ Reviewed exact head `7fb4a62bc6a3c29f87fc30f9fd3255c3c279affd`; the prior request-changes finding is resolved.
✅ The Account-native management calls, deleted CLI selector/public input, credential dual-shape parsing, documentation, and smoke updates remain within the #179 PR 2C authorization boundary.
✅ No unresolved review threads; all current GitHub checks pass.

**Validation**
- `git diff --check`, `pnpm check`, `pnpm build`, and `pnpm typecheck` pass.
- Targeted Shared/Client/CLI suites: 36 tests pass, including all nine machine-credential tests and the three new write-side regressions.
- GitHub Checks, required CI, Node 22/26 compatibility, CLI Pack Smoke, and Container Smoke pass on this exact head.

**Action taken**
- Approved exact head `7fb4a62bc6a3c29f87fc30f9fd3255c3c279affd`.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The fresh head restores the credential writer's strict persistence boundary while preserving tolerant per-entry reads, and all prior CLI migration findings remain resolved.

**Risk level**: A
- Path baseline: includes `apps/cli/**` and `packages/client/**`, including the machine-credential persistence boundary.
- Semantic lift: none; A is already the highest level.

**PR summary**
- Problem: Account users should manage Agents and Computers without selecting a legacy internal Workspace, while pre-cutover Clients must accept enrollment credentials both with and without the deprecated Workspace field.
- Approach: Use the Account-native management facade, delete selector inputs and public remnants, parse stored credentials per entry, and reject write inputs that would not round-trip through the reader.
- Impacted modules: CLI Agent/Computer commands and public surface, Client API and credential storage, shared connect-exchange schema, documentation, and container smoke coverage.

**Review findings**
✅ 1. The credential writer now rejects unusable or duplicate enrollment entries before touching disk, returns promise rejections consistently, and persists the validated projection rather than the caller's original object. [R4 / `packages/client/src/auth/machine-credentials.ts`]
✅ 2. The three new regressions cover the exact mismatch: invalid writes preserve the existing file, duplicate IDs are refused, and successful on-disk bytes equal the reader's result. [R5 / `packages/client/src/__tests__/machine-credentials.test.ts`]
✅ 3. The prior Container Smoke, selector documentation, dead public input, and stale scope wording findings remain fixed on this head.

**Validation**
- Statically rechecked exact head `7fb4a62bc6a3c29f87fc30f9fd3255c3c279affd`; all current GitHub checks pass.
- Per review policy, no local code checks or tests were run.

**Action taken**
- Approved exact head `7fb4a62bc6a3c29f87fc30f9fd3255c3c279affd`.

PR #189, opened 2026-08-27
test(server): pin the persisted instance fence on IM delivery

Gandy2025:
Test-only. Closes a coverage gap found while working the gate 3 question on #179 — see [the analysis there](https://github.com/first-tree-ai/opentag/issues/179#issuecomment-5433642362). No production code changes, no schema, no behaviour change.

## Summary

The IM dispatch guard is a two-part condition:

```ts
// packages/server/src/runtime/im-delivery-worker.ts:328-329
const instanceId = this.#registry.currentInstanceId(row.computer.id);
if (!instanceId || row.computer.currentInstanceId !== instanceId) {
  await this.#recordFailure(deliveryId, "IM_DELIVERY_RUNTIME_UNAVAILABLE", claimToken);
```

Only the **first** disjunct was covered. The existing `IM_DELIVERY_RUNTIME_UNAVAILABLE` case builds its worker with `registry: { currentInstanceId: () => undefined }`, so it exercises "no runtime connected at all". The **second** disjunct — a runtime *is* connected, but at a different generation than the enrollment — had no test: every other delivery case in the file sets `currentInstanceId` to the live instance first (`:866`, `:935`, `:1010`, `:1079`, `:1143`, `:1451`, `:1531`, `:1612`, `:2256`), so the file is entirely happy path with respect to that column.

This adds one case for the second disjunct, oriented the way that costs something: **the enrollment has advanced to a new generation while a stale registry entry for the previous one is still held.** `instanceId` is read from the registry and is what gets dispatched, so dropping the disjunct sends the message to that superseded generation. The persisted column is the only thing that recognises it as stale.

## Why it is worth a test rather than a comment

`workspace_computers.current_instance_id` is a durable fencing token, not a connectivity observation — the same column is also a compare-and-set guard on `heartbeat` and `disconnect` (`computer-service.ts:74`, `:92`, already covered by `computers.test.ts:98-100`). Its routing role is easy to mistake for reachability metadata when the enrollment tables are reshaped, and until now nothing would have objected.

## Testing

I mutated the guard to simulate the column being dropped from routing, rewriting the condition to `if (!instanceId)`, and ran the whole file:

- **before this PR**, under that mutation: all 77 cases pass — the regression is invisible;
- **with this PR**, under that mutation: the new case fails on `expect(domain.requestReconcile).not.toHaveBeenCalled()`, and the instance handed to `requestReconcile` is the stale generation — sourced only from the registry mock, so there is no ambiguity about which one was served;
- with the guard intact: 78 passed.

Commands run: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` all pass. The server integration suite was run against a local PostgreSQL 16 — 129 passed, no test failures. Two integration files (`auth-migrations.test.ts`, `session-collaboration.test.ts`) could not run in the authoring sandbox because they construct their own `PostgreSqlContainer` and Docker is unavailable here; they are untouched by this change and run normally in CI. One `@opentag/client` Codex process-group test fails in this sandbox on `main` as well and is unrelated.

### Correction between heads

The first head (`e166d13`) had the mismatch backwards — persisted as the superseded value, registry as the current one — so removing the guard would have dispatched to the *current* runtime. Its commit message and this body claimed the mutation showed "the superseded generation was served"; that was wrong, and it came from reading my own variable names rather than the dispatched argument. Caught in review by @baixiaohang and fixed in `b54a10b`. The equality check is symmetric, so the earlier orientation still caught removal of the guard; the fix is about the test demonstrating the failure it names.

## Breaking changes

None. Test-only.

## Checklist

### DIFF
--- a/packages/server/src/__tests__/integration/im-binding.test.ts
+++ b/packages/server/src/__tests__/integration/im-binding.test.ts
@@ -3287,19 +3287,20 @@ describe("IM binding persistence", () => {
 
   /**
    * The dispatch guard is `!instanceId || row.computer.currentInstanceId !== instanceId`. The first
-   * disjunct — no runtime connected at all — is covered above. This pins the second: a runtime IS
-   * connected, but the persisted enrollment still names a superseded instance, so it is the persisted
-   * column that withholds the dispatch. De-persisting it would make this disjunct vacuous and silently
-   * let a superseded generation be served.
+   * disjunct — no runtime connected at all — is covered above. This pins the second, in the direction
+   * that actually costs something: the enrollment has advanced to a new generation, while a stale
+   * registry entry for the previous one is still held. `instanceId` comes from the registry, so
+   * dropping this disjunct would dispatch to that superseded generation; the persisted column is the
+   * only thing that recognises it as stale.
    */
-  it("withholds delivery when the persisted instance disagrees with the connected runtime", async () => {
+  it("withholds delivery when the connected runtime is behind the persisted instance", async () => {
     const value = await fixture();
     try {
-      const supersededInstanceId = crypto.randomUUID();
-      const liveInstanceId = crypto.randomUUID();
+      const currentInstanceId = crypto.randomUUID();
+      const staleInstanceId = crypto.randomUUID();
       await value.database
         .update(workspaceComputers)
-        .set({ currentInstanceId: supersededInstanceId })
+        .set({ currentInstanceId })
         .where(eq(workspaceComputers.id, value.workspaceComputer.id));
       const admission = await new ImMessageInbox(value.database).ingest(
         value.imBindingId,
@@ -3312,7 +3313,7 @@ describe("IM binding persistence", () => {
       const domain = { requestReconcile: vi.fn(), requestDelivery: vi.fn() };
       await imDeliveryWorker({
         database: value.database,
-        registry: { currentInstanceId: () => liveInstanceId } as never,
+        registry: { currentInstanceId: () => staleInstanceId } as never,
         domain: domain as never,
       }).runOnce();
 
@@ -3325,6 +3326,7 @@ describe("IM binding persistence", () => {
       await value.sql.end();
     }
   });
+
   it("does not let a pending delivery for an ended Session fence another active Session of the Agent", async () => {
     const value = await fixture();
     try {
--- a/packages/server/src/__tests__/integration/im-binding.test.ts
+++ b/packages/server/src/__tests__/integration/im-binding.test.ts
@@ -3285,6 +3285,46 @@ describe("IM binding persistence", () => {
     }
   });
 
+  /**
+   * The dispatch guard is `!instanceId || row.computer.currentInstanceId !== instanceId`. The first
+   * disjunct — no runtime connected at all — is covered above. This pins the second: a runtime IS
+   * connected, but the persisted enrollment still names a superseded instance, so it is the persisted
+   * column that withholds the dispatch. De-persisting it would make this disjunct vacuous and silently
+   * let a superseded generation be served.
+   */
+  it("withholds delivery when the persisted instance disagrees with the connected runtime", async () => {
+    const value = await fixture();
+    try {
+      const supersededInstanceId = crypto.randomUUID();
+      const liveInstanceId = crypto.randomUUID();
+      await value.database
+        .update(workspaceComputers)
+        .set({ currentInstanceId: supersededInstanceId })
+        .where(eq(workspaceComputers.id, value.workspaceComputer.id));
+      const admission = await new ImMessageInbox(value.database).ingest(
+        value.imBindingId,
+        1,
+        inbound("Ev-persisted-instance-fence"),
+      );
+      const deliveryId = admission.deliveryIds[0];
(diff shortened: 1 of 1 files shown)


baixiaohang (approved):
**Recommendation**: approve
- Rationale: The new integration case fails closed on a persisted/connected instance mismatch and would catch removal of the persisted fence without changing production behavior.

**Risk level**: B-low
- Path baseline: the only change is under `packages/server/**`, with no schema or production service change -> B-low.
- Semantic lift: none; this is test-only and does not alter message routing or runtime state.

**PR summary**
- Problem: IM delivery had no regression test proving that a connected runtime is still ineligible when its instance ID does not match the durable Computer enrollment fence.
- Approach: Create one admitted delivery with mismatched persisted and registry instance IDs, then verify that neither reconcile nor delivery is dispatched and that the attempt remains pending with `IM_DELIVERY_RUNTIME_UNAVAILABLE`.
- Impacted modules: `packages/server/src/__tests__/integration/im-binding.test.ts`

**Review findings**
⚠️ 1. Non-blocking: the fixture calls the persisted ID `supersededInstanceId` and the registry ID `liveInstanceId`, while the stale-runtime hazard after replacement is normally the reverse (the persisted row has advanced and an owner still holds the old registry entry). Consider inverting the labels/setup so the prose matches the generation that would be served if the guard disappeared; the equality-mismatch regression is still correctly covered. [R4 / `packages/server/src/__tests__/integration/im-binding.test.ts:3298`]
✅ 2. The assertions cover both side-effect suppression and the typed durable failure state, keeping the dispatch guard and its retry outcome pinned together. [R4 / `packages/server/src/__tests__/integration/im-binding.test.ts:3319`]

**Action taken**
- Approved the current head (`e166d136e9ec6df0a78ef66b758c66affd759500`).

Gandy2025:
@baixiaohang you are right, and it is worth more than the "non-blocking" you gave it — the orientation was backwards **and it made this PR's own claim false**. Fixed in `b54a10b`.

`instanceId` is read from the registry and is what gets dispatched (`im-delivery-worker.ts:328`, passed as the second argument at `:363` and `:471`). With the persisted column as the superseded value and the registry as the current one, dropping the disjunct would have dispatched to the *current* runtime — harmless. The hazard is the direction you described: the enrollment advances on re-registration while a stale registry entry for the previous generation is still held, and then the dispatch goes to that superseded generation.

So the first head's PR body and commit message were wrong where they said the mutation showed "the superseded generation was served". It served the live one. I inferred that from my own variable names instead of checking the dispatched argument, which is exactly the mistake the labels invited.

Verified properly this time — the dispatched instance can only be what the registry mock returns, and that is now the stale id:

- guard intact: 78 pass in this file;
- guard mutated to `if (!instanceId)`: the new case fails, and the instance handed to `requestReconcile` is the stale generation.

The equality check is symmetric, so the previous orientation did still catch removal of the guard; this is about the test demonstrating the failure it names rather than a neighbouring one. Re-requesting your review since the push reset it.

---

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The new head resolves the prior orientation issue and now demonstrates the actual stale-runtime failure while preserving the original persisted-fence regression coverage.

**Risk level**: B-low
- Path baseline: the only change remains under `packages/server/**`, with no schema or production service change -> B-low.
- Semantic lift: none; the follow-up only corrects a test fixture, its explanatory comment, and the PR description.

**PR summary**
- Problem: IM delivery lacked a regression test proving that an old runtime connection cannot receive work after the durable Computer enrollment has advanced to a new instance.
- Approach: Persist the current instance ID while the registry returns a stale instance ID, then verify that neither reconcile nor delivery is dispatched and that the attempt remains pending with `IM_DELIVERY_RUNTIME_UNAVAILABLE`.
- Impacted modules: `packages/server/src/__tests__/integration/im-binding.test.ts`

**Review findings**
✅ 1. The prior finding is resolved: the persisted row now names the current generation and the registry supplies the stale generation, so removing the guard would dispatch to the superseded instance exactly as the test describes. [R4 / `packages/server/src/__tests__/integration/im-binding.test.ts:3298`]
✅ 2. The PR description was updated to distinguish the corrected scenario from the first head and no longer overclaims what the earlier mutation demonstrated.

**Action taken**
- Approved the current head (`b54a10b72b01dca49924f74f22eed1904913ab22`).

yuezengwu (approved):
Recommendation: approve

Exact head reviewed: b54a10b72b01dca49924f74f22eed1904913ab22

No blocking findings. This is a focused test-only change that correctly models the costly mismatch direction: the persisted enrollment has advanced while the live registry still exposes the stale instance. The assertions pin both side-effect suppression and the durable pending failure IM_DELIVERY_RUNTIME_UNAVAILABLE.

Validation:
- pnpm check: pass
- pnpm typecheck: pass
- full packages/server integration file: 78/78 pass
- mutation check removing the persisted-instance comparison: the new case fails because requestReconcile receives the stale registry instance
- git diff --check: pass
- GitHub CI: all six checks pass, including required CI

Risk: B-low. No production code, schema, persistence, protocol, or documentation changes. The test is consistent with the formal Issue #179 decision that current_instance_id remains a durable fence.

PR #192, opened 2026-08-27
fix(client): bind IM delivery to provider CLI

baixiaohang:
## Summary

- rebind native agent output, including the turn-closing text, to the OpenTag runtime console
- define the official provider CLI as the only outbox to the IM participant
- preserve the agent's choice to take no provider action and existing direct/ambient placement policy
- add regression assertions for the load-bearing user/runtime rebinding and both Slack and Feishu outboxes

## Why

Codex treats its native final answer as the user-facing reply. The previous negative instruction that final text was not sent automatically could be overridden by that native behavior, allowing a Turn to complete without ever invoking `lark-cli` or `slack api`.

The new contract follows the audience-rebinding pattern: the underlying agent's `user` is explicitly the OpenTag runtime, while the IM participant is a separate audience reached only through the provider CLI.

## Testing

- `pnpm --filter @opentag/shared build`
- `pnpm --filter @opentag/client exec vitest run src/__tests__/agent-turn-runner.test.ts --maxWorkers=1`
- `pnpm exec biome check packages/client/src/runtime/agent-turn-runner.ts packages/client/src/__tests__/agent-turn-runner.test.ts`
- `pnpm --filter @opentag/client typecheck`

### DIFF
--- a/packages/client/src/__tests__/agent-turn-runner.test.ts
+++ b/packages/client/src/__tests__/agent-turn-runner.test.ts
@@ -44,6 +44,18 @@ describe("AgentTurnRunner", () => {
     expect(JSON.stringify(input)).not.toContain(request.runtime.instructions.platform);
     expect(JSON.stringify(input)).not.toContain(request.runtime.instructions.agent);
     expect(input.items[0]?.text).toContain("slack api");
+    expect(input.items[0]?.text).toContain('the "user" your underlying agent addresses');
+    expect(input.items[0]?.text).toContain("is the OpenTag runtime");
+    expect(input.items[0]?.text).toContain("including the text that closes this Turn");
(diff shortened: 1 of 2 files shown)


yuezengwu (approved):
Reviewed exact head bac870869ebb5f4bf384235e4f6d1ff47cabfba9. The prompt-only change consistently rebinds native output to the OpenTag runtime, preserves direct/ambient no-action policy, and names the official Slack/Feishu CLI as the sole human-visible outbox. Verified against the current Context Tree boundaries and ran focused tests, repository check/build/typecheck/test, the Agent Runtime coverage gate (100%), and git diff --check; all passed. Residual risk is model-behavior effectiveness, which is not fully provable by static prompt assertions, but I found no code-level blocker.

PR #195, opened 2026-08-27
feat(server): normalize Account email and add verification columns

bestony:
## Summary

First of a staged migration of Account authentication to Better Auth. This step makes one email address identify at most one Account, which Better Auth requires and which nothing currently guarantees.

`users.email` has had no unique constraint since migration `0003` dropped `users_email_unique`, and the identity resolver writes Google's `email` claim verbatim without normalizing case. Two consequences follow: an Account created by `bootstrap:admin` that later signs in with Google becomes a second row for the same person, and case variants of one address can coexist.

### The resolver owns the invariant, not an index

`resolveOrCreateInTransaction` previously matched only on `(provider, issuer, subject)` and inserted a new Account when that triple was new. It now takes an advisory lock on the normalized address before deciding, then resolves the Account already holding it:

- Two first sign-ins for one person serialize, so the second sees the Account the first created rather than a gap.
- Two Accounts exchanging addresses take different keys, never contend, and each reads the other still holding its address — both refuse by decision, with no lock cycle and no database error.
- Two Accounts already sharing an address are refused rather than chosen between, so ownership is never handed to whichever row the database returned first.

Attachment additionally requires the provider to be trusted, not merely to assert a verified address: verification is the adapter's claim about an address, trust is OpenTag's claim about the adapter, and only the second may hand over an existing Account. `AuthIdentityProviderSchema` already admits `github` and `oidc`, so the allowlist is explicit and Google-only for this stage.

`users.email_verified` is maintained on every write path — new Account, explicit link, and returning identity — and only ever for the address actually stored on the Account.

### Migration `0019` is expand-only

Columns, casing normalization, and the verification backfill. Nothing changes how the previous server revision behaves, which a rollback requires because rolling back application code does not roll back migrations.

**The `users_email_unique` index is deliberately not here.** The previous revision answers an unknown provider subject by inserting a `users` row without resolving the address, so creating the index while it is still serving would turn its bootstrap-Account first sign-in from a silent duplicate into a raw `23505` and an opaque 500. The index is a backstop for writers that skip the resolver; it lands in #196, one deployment later, with a duplicate guard that reports offending rows by id. The regression here asserts the previous revision's insert still succeeds.

## Testing

- `pnpm --filter @opentag/server test:integration` — 174 passing
- `account-identity-resolution.test.ts` (12 tests) covers bootstrap attachment, verification on new/linked/returning writes, provider email change, the untrusted-provider and unverified-address refusals, the duplicate-owner refusal, and three concurrency properties. The same-address ones hold the first transaction inside its claim and require the second to make no progress until it is released — verified by deleting the advisory lock, which fails them on exactly that assertion.
- The migration regression replays the previous revision's unconditional insert of a differently-cased address and requires it to succeed.

## Breaking changes

None. The previous server revision behaves identically after this migration; uniqueness is enforced by the new resolver and constrained in the database one deployment later.

## Checklist


baixiaohang (changes requested):
- Rationale: The current head cannot be staged safely: it collides with the migration already on `main`, and the live Google resolver does not preserve the new Account email invariants during the staged rollout.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes a schema migration and changes to `services/auth/**`.
- Semantic lift: The change alters durable Account identity semantics and the live authentication write path.

**PR summary**
- Problem: Prevent one human email address from producing multiple OpenTag Accounts and prepare the Account table for Better Auth.
- Approach: Normalize stored and provider-supplied emails, reject existing duplicates before migration, add a case-insensitive unique index, and add/backfill Better Auth's verification and image fields.
- Impacted modules: Server auth schema and migrations, Google identity resolution, migration integration tests, and development documentation.

**Review findings**
❌ 1. Regenerate this migration on top of current `main`. `main` now contains `0018_salty_tombstone` with journal index 18 and its own `0018_snapshot.json`, while this head independently records `0018_freezing_vermin` at the same index from an older schema state. The PR is already reported as conflicting; resolving the text conflict while retaining two index-18 histories would leave the migration ledger/snapshot chain divergent. Rebase, regenerate this as the next migration (currently 0019), and update the migration-history fixtures/counts from that new baseline. [R5 / `packages/server/drizzle/meta/_journal.json:132`]

❌ 2. The new index breaks the existing bootstrap-Account → first Google sign-in path instead of resolving it. `resolveOrCreateInTransaction` still looks up only `(provider, issuer, subject)` and, when no identity row exists, inserts a new `users` row. After this migration, a bootstrap Account with the same verified Google email reaches `users_email_unique`; the raw PostgreSQL error falls through the HTTP handler as `INTERNAL_ERROR`/500. This is the exact duplicate scenario motivating the change, and it contradicts the stated no-client-breaking staged rollout. Please define and implement the verified-email attachment/conflict rule atomically with the index (including the provider-email-change collision), then cover bootstrap → Google sign-in with an integration test. [R4/R5 / `packages/server/src/services/auth/identity-service.ts:98`]

❌ 3. `email_verified` becomes stale immediately after the migration. The SQL backfills identities that exist at migration time, but the live Google path subsequently inserts new `users` rows without setting `emailVerified` and the `expectedUserId` link path also leaves it false. Because the Google exchange already rejects unverified addresses, Accounts created or linked during this rollout window will later be presented to Better Auth as unverified while pre-migration Google Accounts are true. Update the live writer in this stage (or provide an atomic deployment/backfill boundary that cannot leave false rows) and test both new and linked Google identities. [R1/R5 / `packages/server/src/services/auth/identity-service.ts:82`]

**Action taken**
- Submitted request changes. No tests or code checks were run as part of this review.

baixiaohang (changes requested):
- Rationale: The migration-order conflict is fixed, but the synchronized head still leaves the current Google authentication writer incompatible with the new Account email constraint and verification semantics.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes a database schema migration and `services/auth/**`.
- Semantic lift: The change alters durable Account identity and the live sign-in write path.

**PR summary**
- Problem: Ensure one email address maps to at most one OpenTag Account and prepare Account records for Better Auth.
- Approach: Normalize email values, reject pre-existing duplicates, add a case-insensitive unique index plus Better Auth columns, and backfill verification for existing provider identities.
- Impacted modules: Server auth schema/migrations, Google identity resolution, migration integration tests, and development documentation.

**Review findings**
✅ 1. The previous migration-history blocker is resolved: this head is based on current `main`, uses `0019_charming_tony_stark`, and its snapshot chains from the shipped `0018` state.

❌ 2. The bootstrap-Account → first Google sign-in path is still broken. The synchronized commit does not change `resolveOrCreateInTransaction`: when no `(provider, issuer, subject)` row exists it still inserts a new `users` row without first resolving the already-existing Account by the verified normalized email. With `users_email_unique` installed, that insert raises PostgreSQL `23505`, which the HTTP handler exposes as generic `INTERNAL_ERROR`/500. Please implement an explicit, atomic verified-email attachment/conflict policy (including a provider email changing to another Account's address) and cover the bootstrap → Google flow. [R4/R5 / `packages/server/src/services/auth/identity-service.ts:98`]

❌ 3. `email_verified` still drifts as soon as this migration is deployed. The SQL marks only identity rows present at migration time, while the live Google writer inserts new Accounts without `emailVerified: true`; the `expectedUserId` link path likewise inserts an `auth_identities` row without updating the Account flag. Google has already verified these addresses, so Accounts created or linked during the rollout window will carry false data into Better Auth. Update the live writer in this stage, or make the replacement deployment/backfill atomic so no such rows can survive, and test both cases. [R1/R5 / `packages/server/src/services/auth/identity-service.ts:82`]

**Action taken**
- Submitted request changes for head `3572b278b18d23cd6efc8cedc99072d99bd683bf`. No tests or code checks were run as part of this review.

yuezengwu (changes requested):

Reviewed exact head `3572b278b18d23cd6efc8cedc99072d99bd683bf` independently. The migration-history collision from the previous head is resolved, and all repository validation passes, but the current login writer still violates both new Account invariants.

1. **The bootstrap Account -> first Google sign-in path is converted from a duplicate into a 500, not fixed.** `resolveOrCreateInTransaction` locks and resolves only `(provider, issuer, subject)`; when that identity is new it inserts a second `users` row without resolving the already-existing Account by the verified normalized email. The new `users_email_unique` index rejects that insert with an uncategorized Drizzle query error. I reproduced this against PostgreSQL 17 by bootstrapping `Admin@Example.com` and resolving a verified Google identity for `ADMIN@example.com`: the insert failed and no identity row was attached. Define and implement an atomic verified-email attachment/conflict policy, including the provider-email-change collision, and add the end-to-end regression. [`packages/server/src/services/auth/identity-service.ts:98`]

2. **`email_verified` becomes stale immediately after deployment.** The migration backfills only identities present at migration time. The active Google path has already rejected unverified claims, but new Account inserts leave `emailVerified` at its `false` default; the explicit link path also inserts `auth_identities` without updating the Account flag, and an existing identity email update never refreshes it. I reproduced a new verified Google Account being stored as `{ email: "verified@example.com", emailVerified: false }`. Update these live writes in the same transaction (or provide a rollout boundary that cannot leave false rows) and cover new, linked, and provider-email-change cases. [`packages/server/src/services/auth/identity-service.ts:75-110`]

There is also a required decision gate: this head changes durable Account identity semantics and executes a schema migration, but the PR timeline and linked artifacts contain no explicit human approval for the one-normalized-email-per-Account and verification-backfill policy. That authorization must be recorded before this can be approved; it does not replace the two technical fixes above.

**Validation**

- `pnpm check` — pass
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (including server 266/266)
- `pnpm --filter @opentag/server test:integration` — pass (162/162)
- Focused PostgreSQL 17 reproductions — both defects confirmed; temporary test removed and worktree restored clean
- CI — 6/6 pass on this exact head

yuezengwu:
Decision update for exact head 3572b278b18d23cd6efc8cedc99072d99bd683bf: I approve the durable product direction that one normalized email identifies one Account and that verified provider identities backfill and maintain email_verified. This clears only the product and migration direction gate from my review; it is not approval of the current implementation. CHANGES_REQUESTED remains for atomic verified-email Account attachment and conflict handling without an unclassified database 500, consistent email_verified maintenance across new, explicit-link, returning-identity, and provider-email-change paths, and the corresponding integration and migration-safety tests.

bestony:
Both findings were correct and are fixed in `72f1b8e`. Thanks — the shared diagnosis that the migration and the live writer disagreed from the moment of deployment was exactly right.

**1. Bootstrap Account → first Google sign-in (`identity-service.ts:98`)**

`resolveOrCreateInTransaction` now resolves an Account by normalized email before inserting, under `SELECT … FOR UPDATE`, and attaches the identity to it. The residual race that a row lock cannot cover — two concurrent inserts of an address that does not exist yet — is caught as PostgreSQL `23505` and translated into the same decision instead of escaping as `INTERNAL_ERROR`.

Attachment requires provider verification, because attaching hands over an existing Account. `ExternalIdentity` now carries `emailVerified` explicitly rather than assuming callers verified; `DefaultGoogleIdentityClient` sets it at the point where it already rejects unverified claims. Two refusals are explicit and typed as a new `AUTH_EMAIL_CONFLICT` (409): an unverified address that is already taken, and a provider email change onto another Account's address.

This is deliberately the same policy Better Auth's `accountLinking.trustedProviders: ["google"]` applies in the next stage, so the rollout does not change semantics halfway through.

**2. `email_verified` drift (`identity-service.ts:75-110`)**

All three write paths maintain it now — new Account, `expectedUserId` link, and existing-identity email refresh — in the same transaction as the identity write. The flag is only ever raised for the address actually stored on the Account, so it cannot claim verification for a different address the provider returned.

**Regression coverage** — `packages/server/src/__tests__/integration/account-identity-resolution.test.ts`, six tests against PostgreSQL 17:

- bootstrap `Admin@Example.com` + verified Google `ADMIN@example.com` → one Account, identity attached, `email_verified` true
- new verified Account stored verified
- `expectedUserId` link raises the flag
- provider email change follows, with the identity row updated
- provider email change onto another Account → `AUTH_EMAIL_CONFLICT`, both Accounts keep their address
- unverified identity for a taken address → `AUTH_EMAIL_CONFLICT`, no attach, no second Account

**Validation on this head**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (168/168, up from 162) all pass.

One note on how I verified: these six tests pass on the fixed code, but I did not re-run them against the previous head. The old `ExternalIdentitySchema` is `.strict()` and has no `emailVerified` field, so they fail there on schema validation rather than on the defect. Your independent PostgreSQL reproductions are what establish the pre-fix behavior.

On the decision gate: @yuezengwu's comment on `3572b27` records approval of the one-normalized-email-per-Account and verification-backfill direction. The attachment policy above is new since that comment, so it is worth an explicit look — it decides who ends up owning an Account when an address is already present.

baixiaohang (changes requested):
- Rationale: The original two defects are fixed on the covered sequential paths, but the current head still leaks database failures under concurrency and does not encode the stated trusted-provider boundary for Account attachment.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes a database migration and `services/auth/**` identity logic.
- Semantic lift: The change defines who may acquire an existing Account and how concurrent identity writes preserve its unique email.

**PR summary**
- Problem: Make one normalized email identify at most one Account while preparing Account data for Better Auth.
- Approach: Add the unique index and Better Auth columns, normalize provider email, attach verified identities to an existing email owner, maintain `email_verified`, and return typed conflicts.
- Impacted modules: Server auth schema/migrations, identity resolution, Google verification, shared error contracts, integration tests, and development documentation.

**Review findings**
✅ 1. The previous sequential-path blockers are addressed: bootstrap → Google attaches to one Account, create/link/returning writes maintain `email_verified`, and an already-occupied provider email change rolls back as `AUTH_EMAIL_CONFLICT`.

❌ 2. The promised typed handling still fails for residual races. `isUniqueViolation` checks only the top-level object, while this repository's established Drizzle/PostgreSQL helpers walk the `cause` chain and inspect `constraint_name`; a wrapped `23505` from the concurrent insert can therefore still escape as `INTERNAL_ERROR`. Separately, two existing identities can both observe the same new email as unowned and race the `users` update at lines 149–152; that unique violation is outside any catch. Please use a cause-aware, `users_email_unique`-specific classifier around both insert and update (or serialize on the normalized email), and add concurrent insert plus concurrent provider-email-change regressions. [R4 / `packages/server/src/services/auth/identity-service.ts:147`, `packages/server/src/services/auth/identity-service.ts:216`]

❌ 3. Account attachment currently trusts any provider that supplies `emailVerified: true`, not the Google-only trusted-provider policy cited in the fix. `AuthIdentityProviderSchema` also admits `github` and `oidc`, so a future adapter would silently gain authority to attach its identity to an existing Account merely by asserting verified email. Email verification and provider trust are separate facts. My recommendation is to encode the trusted attachment allowlist at this single boundary (Google only for this stage), reject an occupied address from a verified-but-untrusted provider, and cover that case; broaden the list only through an explicit product/security decision. [R1/R5 / `packages/server/src/services/auth/identity-service.ts:101`]

**Action taken**
- Submitted request changes for head `72f1b8e5edd2208f28494ad72e2f31dbaac60e50`. No tests or code checks were run as part of this review.

yuezengwu (changes requested):

Reviewed exact head `72f1b8e5edd2208f28494ad72e2f31dbaac60e50` independently. The approved durable direction is clear, and this head fixes the covered sequential flows: bootstrap -> first Google sign-in attaches atomically, new/explicit-link/returning/provider-email-change writes maintain `email/emailVerified`, and an already-occupied email change returns `AUTH_EMAIL_CONFLICT`. The migration chain and all repository/CI checks are also green. Three release-blocking boundaries remain.

1. **Both residual email races still escape as database errors instead of typed 409s.** `isUniqueViolation` checks only the top-level error, but Drizzle wraps the PostgreSQL error in `cause`; the repository's existing classifiers walk that chain and inspect `constraint_name`. Separately, the `users` update at lines 149-152 is outside the insert catch entirely. Against PostgreSQL 17 I forced (a) two new identities to create the same new email and (b) two existing identities to move to the same new email. In both cases exactly one transaction lost with `Error -> PostgresError { code: "23505", constraint_name: "users_email_unique" }`, not `AUTH_EMAIL_CONFLICT`. Use a cause-aware, `users_email_unique`-specific classifier around both insert and update, or serialize on the normalized email, and add both concurrent regressions. [`packages/server/src/services/auth/identity-service.ts:147-172`, `:216-218`]

2. **The documented rolling/rollback compatibility can still permanently miss the verification backfill.** Migration 0019 performs a one-time backfill, then leaves `email_verified DEFAULT false`. The previous server revision is intentionally allowed to continue after the migration, but that revision inserts `users` and `auth_identities` without setting the flag. I reproduced that exact post-migration legacy-writer sequence and the Account remained `email_verified=false`; stacked migration 0020 does not repeat or reconcile this backfill. Ensure old writers cannot run after the migration, maintain the invariant at the database boundary, or provide a tested post-drain reconciliation/backfill before the field is consumed. [`packages/server/drizzle/0019_charming_tony_stark.sql:17-21`]

3. **Automatic Account attachment does not enforce the Google-only trust boundary used by the stacked replacement.** The attach branch accepts any `ExternalIdentity` with `emailVerified: true`, while the provider schema admits `google`, `github`, and `oidc`. I verified that a synthetic verified OIDC identity is silently attached to an Account created by Google. That differs from #196's explicit `trustedProviders: ["google"]`; provider trust and an adapter's verification bit are separate security decisions. Encode the trusted attachment allowlist here (Google for this stage), or record an explicit product/security decision widening it, and cover the rejected untrusted-provider case. [`packages/server/src/services/auth/identity-service.ts:101-107`]

**Validation**

- `pnpm check` — pass
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (server 266/266; all packages green)
- `pnpm --filter @opentag/server test:integration` — pass (168/168)
- Focused PostgreSQL 17 race/rolling/trust reproductions — all four uncovered expectations failed as described; temporary tests removed and worktree restored clean
- CI — 6/6 pass on this exact head

bestony:
Both findings were correct. Fixed in `c032ec2`.

**1 — races escaping as database errors.** `isUniqueViolation` read only the top-level error; Drizzle wraps the driver error in `cause`. The `users` update after a provider email change had no catch at all. Both now use a cause-aware classifier matched to a named constraint, extracted to `packages/server/src/db/unique-violation.ts` — `agent-service.ts:247` and `im-binding-service.ts:183` each carry their own copy of that loop, and this is the obvious home if anyone dedupes them later.

Writing the concurrency regressions surfaced a third race neither review named: two subjects from the same provider racing onto one address both pass the email check, then collide on `auth_identities_user_provider_unique`, because an Account holds at most one identity per provider. That violation escaped untyped as well, and is now `AUTH_IDENTITY_CONFLICT`. The concurrent-creation regression accepts either code, since the loser can land on either index depending on which write it reaches first.

**2 — attachment trusting any verified provider.** Correct, and the framing was the useful part: verification is the adapter's claim about an address, trust is OpenTag's claim about the adapter, and only the second may hand over an existing Account. Attachment now requires membership in an explicit trusted-provider set — Google alone for this stage — so a future `github` or `oidc` adapter cannot inherit that authority by setting one boolean. Widening it is a deliberate edit at that one boundary.

**Regressions added** (`account-identity-resolution.test.ts`, now 9 tests): concurrent Account creation for one address, concurrent provider email change onto one address, and a verified-but-untrusted OIDC identity refused against a Google-owned Account.

**Validation on `c032ec2`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (171/171) all pass.

One process note: the first attempt at this fix went onto the stacked branch by mistake. It has been cherry-picked here and dropped from #196, which now carries it only through the merge — the fix exists once, on this PR, where it belongs.

baixiaohang (changes requested):
- Rationale: The new head fixes the previously reported same-address races and provider-trust boundary, but rolling deployment can still persist false verification data and an inverted Account-lock order can still turn a concurrent email conflict into a database deadlock/500.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes a database migration and `services/auth/**` identity logic.
- Semantic lift: The change defines durable Account ownership and verification semantics across concurrent and mixed-version writers.

**PR summary**
- Problem: Make one normalized email identify at most one Account while preparing Account data for Better Auth.
- Approach: Add the unique index and Better Auth columns, normalize provider email, attach only trusted verified identities to an existing Account, maintain `email_verified`, and translate constraint races into typed conflicts.
- Impacted modules: Server auth schema/migrations, identity resolution, Google verification, shared errors, database error classification, integration tests, and development documentation.

**Review findings**
✅ 1. The two blockers targeted by `c032ec2` are fixed in the implementation: unique violations are now classified through Drizzle's `cause` chain and narrowed by constraint name, both Account insert/update paths translate the expected races, and verified-email attachment is gated by an explicit Google-only trust set.

❌ 2. The mixed-version rollout still violates the verification invariant after the one-time backfill. Migration 0019 adds `email_verified DEFAULT false` and backfills only rows that already have an identity. The previous server revision is explicitly allowed to keep serving after this migration, but its create path inserts `users` and `auth_identities` without setting the flag. Any Google Account it creates after line 20 therefore remains false; this head has no database boundary or post-drain reconciliation, and the stacked migration 0020 does not repeat the backfill. Prevent old writers from serving after 0019, maintain the invariant in the database, or add a tested reconciliation after old instances drain and before the field is consumed. [R5 / `packages/server/drizzle/0019_charming_tony_stark.sql:17`]

❌ 3. Concurrent email swaps can still escape as PostgreSQL deadlocks rather than typed conflicts. Each returning identity first locks its current Account at line 92, then `#applyAccountEmail` locks the Account holding the target address at line 161. If Account A moves to B's address while B moves to A's, the transactions acquire those two rows in opposite order; PostgreSQL aborts one with `40P01`, outside the `23505` update catch, while the other eventually reports the expected conflict. Use a consistent lock order, avoid the second row lock while retaining the unique-index backstop, or retry/classify the deadlock, and cover the swap interleaving. [R4 / `packages/server/src/services/auth/identity-service.ts:92`]

⚠️ 4. The new concurrency tests do not force the races they are intended to pin. Bare `Promise.allSettled` calls may serialize: the creation test explicitly accepts the later `auth_identities` conflict instead of exercising `users_email_unique`, and the email-change test can observe the committed owner before reaching the update catch. Add a barrier or transaction hook so both operations pass the unowned-email read before either writes, then assert exactly one typed loser; otherwise the old broken catch paths can regress while these tests stay green. [R4 / `packages/server/src/__tests__/integration/account-identity-resolution.test.ts:128`]

**Action taken**
- Submitted request changes for head `c032ec277091e0dc196ce60631889cdb490268a0`. No tests or code checks were run as part of this review; CI was still in progress when the head was reviewed.

bestony:
Findings 3 and 4 were correct and are fixed in `7f032cd`. Finding 2 I want to push back on, with specifics.

**3 — deadlock on an address swap.** Real, and the diagnosis was exact. A returning identity already holds its own Account row, and `#applyAccountEmail` then locked the Account holding the target address, so two Accounts swapping addresses acquired the same two rows in opposite orders.

That lock only bought a nicer error message. `users_email_unique` already decides the outcome, and since the previous round the violation is classified, so the lock is removed rather than reordered — no second lock means no cycle, and the loser still gets `AUTH_EMAIL_CONFLICT`.

Verified both directions: with the lock restored the new swap regression fails with `PostgresError { code: "40P01", message: "deadlock detected" }`, and passes without it.

**4 — the tests did not force the race.** Also correct, and it was the more useful finding, because it meant the previous round's catch paths were unverified. `Promise.allSettled` over two resolutions is free to serialize.

They now hold both transactions past the address lookup using an `afterAccountLookup` seam on the service, matching the `afterAgentLocked` / `afterMembershipLocked` idiom `AgentService` already uses for this. With a real barrier the creation race lands on `users_email_unique` specifically, so the assertion is now `toEqual(["AUTH_EMAIL_CONFLICT"])` — exactly one typed loser — rather than accepting whichever index the interleaving happened to reach.

**2 — the rolling-deployment backfill.** I think this is already covered, on two independent paths, and I would rather explain than add redundant code:

- The returning-identity path heals it. An Account the old revision left at `false` gets `emailVerified` raised the next time that person signs in, in `#applyAccountEmail`. That is not incidental — it is the same branch that maintains the flag for every other returning identity. I have added an explicit regression for it (`heals an Account a previous server revision left unverified on the next sign-in`), since it was previously only implied by other cases.
- The stacked #196 repeats the backfill in migration `0020`, with its own replay regression, at `914dc82`. Your review predates that commit, which is why it reads as absent here.

So a stale row self-corrects on next sign-in, and the sweep runs before anything can read the flag. On the database-boundary option: a trigger cannot express this invariant, because `auth_identities` records no per-identity verification and `ExternalIdentity` can legitimately carry `emailVerified: false` — a trigger would have to assume every identity implies a verified address, which is wrong in the direction that grants trust. Adding a verification column to `auth_identities` only moves it, since the old writer would not set that either and its default would have to be `true`.

If you would still rather this PR carry its own sweep rather than relying on #196's, say so and I will move the `0020` reconciliation into `0019` — it is a small change, I just do not want to duplicate it in both.

**Validation on `7f032cd`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (173/173) all pass. `account-identity-resolution.test.ts` is now 11 tests.

baixiaohang (changes requested):
- Rationale: removing the second explicit row lock removes the previously deterministic cycle, but the unique-index wait can still deadlock the same address swap, and the new regression does not force or exclude that interleaving.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes Account identity transaction logic on top of a database migration.

**PR summary**
- Problem: make one normalized email identify at most one Account while preparing durable verification state for Better Auth.
- Approach: normalize and uniquely constrain Account email, attach only trusted verified identities, maintain verification state transactionally, and translate expected ownership races into typed conflicts.
- Impacted modules: Server Account/identity persistence, authentication services, migrations, database error classification, and integration tests.

**Review findings**
❌ 1. The address-swap deadlock is reduced, not eliminated. Each transaction still locks its own Account and then updates the unique email key. PostgreSQL's unique-index checker waits when the conflicting row was deleted by an uncommitted transaction; if both swaps reach that phase, each transaction can wait for the other and one is aborted with `40P01`, which this `23505` catch does not classify. The `afterAccountLookup` barrier is before the `UPDATE`, so after release the scheduler may let one statement reach a still-committed conflicting row and return `23505`; that passing run does not prove the internal unique-index wait cycle is impossible. Restoring the explicit target lock proves the old path was broken, but not that this path is safe. Serialize the two address keys in a consistent order, or retry/classify deadlocks at a transaction boundary that covers every caller, and pin the actual wait interleaving. PostgreSQL documents both the unique-check wait and the need to retry unavoidable deadlocks: [Index Uniqueness Checks](https://www.postgresql.org/docs/17/index-unique-checks.html), [Deadlocks](https://www.postgresql.org/docs/17/explicit-locking.html#LOCKING-DEADLOCKS). [R4 / `packages/server/src/services/auth/identity-service.ts:177`]

❌ 2. The swap regression's title says the operation reports a conflict, but `for (const code of typedConflicts(results))` also passes when both promises fulfill because it never asserts a rejected result exists. It therefore cannot establish its stated exactly-one-typed-loser contract even apart from the UPDATE-internal scheduling gap. [R4 / `packages/server/src/__tests__/integration/account-identity-resolution.test.ts:188`]

✅ 3. The deterministic creation/email-change barriers now exercise the named unique-index catches, the trusted-provider boundary remains explicit, and the returning-sign-in plus stacked `0020` reconciliation jointly cover the rolling verification window before any consumer is mounted.

**Action taken**
- Submitted request changes on exact head `7f032cd8a012ebede13571f04fe3a1c1823a43d5`. No local tests or code checks were run.

bestony:
Correction to my previous comment: removing the second row lock was **not** sufficient, and the swap regression proved it — it passed locally and failed in CI. Fixed properly in `4cfa95e`.

A unique index is itself a waiting point. An update that would collide with a not-yet-committed tuple waits on the transaction holding it, so two Accounts exchanging addresses wait on each other with no explicit lock involved at all, and PostgreSQL still aborts one with `40P01`. Dropping the explicit lock removed one cycle and left the one the index creates.

Both outcomes mean the same thing — a concurrent move took this address first — and which one surfaces depends only on the interleaving. The update's catch now classifies the deadlock alongside the unique violation, scoped to that single statement rather than as a global rule.

This is worth noting for the review record: your finding 4 is what made finding 3 fixable. The old `Promise.allSettled` tests would have passed against the incomplete fix, and I would have reported it as done.

Head is now `4cfa95e`; `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (173/173) all pass, and the identity file runs clean across repeated runs.

baixiaohang (changes requested):
- Rationale: The unique-index deadlock is now translated at the correct statement boundary, but the swap regression still does not assert that any conflict occurred or that Account ownership remained unchanged.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes Account identity transaction logic on top of a database migration.

**PR summary**
- Problem: Make one normalized email identify at most one Account while preparing durable verification state for Better Auth.
- Approach: Normalize and uniquely constrain Account email, attach only trusted verified identities, maintain verification state transactionally, and translate ownership races and deadlocks into typed conflicts.
- Impacted modules: Server Account/identity persistence, authentication services, migrations, database error classification, and integration tests.

**Review findings**
✅ 1. `4cfa95e` fixes the remaining implementation failure: the wrapped PostgreSQL `40P01` is classified beside the `users` email update that can produce the unique-index wait cycle, while unrelated statements are not globally converted into email conflicts. The staged verification reconciliation and the other deterministic race tests remain intact.

❌ 2. The swap regression still contains the vacuous assertion reported on the previous head. `for (const code of typedConflicts(results))` passes when `typedConflicts` returns `[]`, and the only state assertion is that two `users` rows exist—which also passes if both operations fulfilled and silently exchanged the addresses. For a two-transaction swap, both operations should reject as `AUTH_EMAIL_CONFLICT` (the deadlock victim rolls back, leaving the survivor to lose on the restored unique key), and both Accounts should retain their original addresses. Assert the exact two rejection codes plus the unchanged email-to-Account mapping so the regression proves both the typed outcome and atomic rollback. [R4 / `packages/server/src/__tests__/integration/account-identity-resolution.test.ts:188`]

**Action taken**
- Submitted request changes for exact head `4cfa95ec999c69a63326d94abd6030479fe47d94`. No local tests or code checks were run; CI was still in progress.

bestony:
Correct — the loop passes when it finds nothing, so two operations that both succeeded and silently exchanged addresses would have satisfied it, and so would the row count beside it. Fixed in `ac72ab1`.

Your predicted outcome is exactly what happens, so the test asserts it directly rather than tolerating a range: both transactions wait on the other's uncommitted tuple through the unique index, PostgreSQL kills one, its rollback restores the address the survivor was moving onto, and the survivor then loses on the restored key. Neither move can succeed.

```
expect(typedConflicts(results)).toEqual(["AUTH_EMAIL_CONFLICT", "AUTH_EMAIL_CONFLICT"]);
expect((await readAccount(leftId))?.email).toBe("left@example.com");
expect((await readAccount(rightId))?.email).toBe("right@example.com");
```

The two ownership assertions are the part I had missing entirely — they pin the rollback rather than just the error code, which is what distinguishes "rejected" from "rejected and left nothing behind". Stable across repeated runs.

That was the third time this file's assertions were weaker than the behaviour they were meant to pin, and each catch was yours. Worth stating plainly: the implementation fixes in this PR would not have been trustworthy without those.

**Validation on `ac72ab1`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (173/173) all pass.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: All previously reported migration, identity-ownership, verification, trust-boundary, and concurrency blockers are resolved on exact head `ac72ab1dee86bcbc7e561fb5f7a627935b000a93`, and the final regression now proves both typed rejection and atomic rollback for an address swap.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes an Account schema migration and authentication identity logic.
- Semantic lift: The change defines durable Account ownership and verification semantics across concurrent and mixed-version writers.

**PR summary**
- Problem: Prevent case variants or first provider sign-in from creating multiple Accounts for one person while preparing Account data for Better Auth.
- Approach: Normalize and uniquely constrain Account email, attach only trusted verified Google identities, maintain verification state transactionally, and translate ownership races and address-swap deadlocks into typed conflicts.
- Impacted modules: Server Account/identity persistence, migrations, Google authentication, shared errors, database error classification, integration tests, and development documentation.

**Review findings**
✅ 1. Migration `0019` is chained from current `main`, rejects pre-existing normalized duplicates, and adds/backfills the Better Auth columns without breaking the prior revision's explicit-column reads.
✅ 2. Bootstrap attachment, new/link/returning/email-change writes, provider trust, and the staged verification reconciliation preserve the approved one-email-per-Account policy.
✅ 3. Cause-aware, constraint-specific conflict handling covers concurrent Account creation and email moves; address-swap deadlocks are classified at the responsible statement boundary.
✅ 4. The final deterministic swap regression requires two `AUTH_EMAIL_CONFLICT` rejections and verifies both Accounts retain their original addresses, pinning transaction rollback rather than only error shape.
✅ 5. CI is green: 6/6 checks passed on this exact head.

**Action taken**
- Approved exact head `ac72ab1dee86bcbc7e561fb5f7a627935b000a93` with explicit reviewer authorization. No local tests or code checks were run as part of this final review.

yuezengwu (changes requested):
**Recommendation: request changes**

Reviewed exact head `ac72ab1dee86bcbc7e561fb5f7a627935b000a93` independently. The current writer, trust boundary, typed race handling, and strengthened swap rollback regression now satisfy the approved one-normalized-email-per-Account direction. One mixed-version deployment blocker remains.

1. **Migration 0019 does not actually let the previous server revision keep serving the first-provider-sign-in path.** The PR explicitly requires that revision to remain compatible after the new pod applies migrations, but `CREATE UNIQUE INDEX users_email_unique ON lower(email)` changes the old writer's behavior: it still handles an unknown Google subject by inserting a new `users` row before `auth_identities`, without first resolving the existing normalized email owner or classifying `23505`. Against PostgreSQL 17 on this exact head, I applied the full migration, created the existing bootstrap Account, then replayed that previous writer with `ADMIN@example.com`; the transaction failed as wrapped PostgreSQL `23505 / users_email_unique`, left the identity unattached, and would surface as the opaque login 500 this PR is intended to remove. Stacked #196's `0020` reconciliation fixes a different window (`email_verified=false` on newly created, otherwise unique Accounts); it cannot make this old attachment request succeed. Split the uniqueness constraint into a post-drain migration (deploy compatible columns/writer first, then add the index after old instances are gone), or supply another tested deployment guard that prevents old instances from receiving this request after 0019. If a transient first-login failure during rollout is intended instead, that is a human rollout/availability decision and the current “previous revision keeps running” / “no breaking changes” contract must not claim compatibility. [`packages/server/drizzle/0019_charming_tony_stark.sql:22`, base `packages/server/src/services/auth/identity-service.ts:97-110`]

The `ac72ab1` change itself is correct: the deterministic swap test now requires two `AUTH_EMAIL_CONFLICT` rejections and proves both Accounts retain their original addresses. I ran that exact-head identity suite five times; all runs passed.

**Validation on this exact head**

- `pnpm check` — pass
- `TURBO_FORCE=true pnpm build` — pass
- `TURBO_FORCE=true pnpm typecheck` — pass
- `TURBO_FORCE=true pnpm test` — pass
- `pnpm --filter @opentag/server test:integration` — pass (173/173)
- Focused PostgreSQL mixed-version writer probe — reproduced wrapped `23505 / users_email_unique`; temporary probe removed
- Focused identity/concurrency suite — pass 5/5 repeated runs
- GitHub CI — 6/6 successful
- Detached exact-head worktree restored clean; `git diff --check` passes

baixiaohang (changes requested):
- Rationale: Exact head `ac72ab1dee86bcbc7e561fb5f7a627935b000a93` resolves the current-writer and concurrency defects, but migration `0019` still breaks the previous revision's bootstrap Account → first Google sign-in path during the rolling/rollback window the PR explicitly claims to support.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes an Account schema migration and authentication identity logic.
- Semantic lift: The migration changes durable Account ownership behavior for mixed-version writers.

**PR summary**
- Problem: Prevent one person from acquiring duplicate Accounts through email casing or first provider sign-in while preparing Account data for Better Auth.
- Approach: Normalize and uniquely constrain Account email, attach trusted verified identities, maintain verification state, and translate current-writer races into typed conflicts.
- Impacted modules: Server Account/identity persistence, migrations, Google authentication, shared errors, database error classification, integration tests, and development documentation.

**Review findings**
❌ 1. `0019` is not compatible with the previous server revision on the motivating bootstrap flow. After line 22 creates `users_email_unique`, that revision still handles an unknown Google subject by inserting a new `users` row before `auth_identities`, without resolving the normalized email owner or classifying the index violation. A rolling old pod—or the documented application rollback—therefore fails bootstrap Account → first Google sign-in with a raw `23505`/500 instead of attaching the identity. Split the unique index into a post-drain migration, add a tested deployment guard that prevents old writers from serving this path after `0019`, or explicitly authorize and document the transient availability break instead of claiming previous-revision compatibility and no breaking changes. The stacked verification sweep addresses stale `email_verified`; it cannot repair a transaction that never attached the identity. [`packages/server/drizzle/0019_charming_tony_stark.sql:22`, base `packages/server/src/services/auth/identity-service.ts:97`]

✅ 2. The current writer, provider-trust boundary, typed race handling, and deterministic address-swap rollback regression are correct on this head; CI remains 6/6 green.

**Action taken**
- Superseded my earlier approval and submitted request changes on exact head `ac72ab1dee86bcbc7e561fb5f7a627935b000a93` after the mixed-version incompatibility was surfaced. No local tests or code checks were run.

bestony:
Both of you landed on this independently and you were right — the compatibility claim was false. Fixed in `ed03cb3`, and the fix goes further than splitting the migration.

**The actual mistake.** I had the index carrying the invariant. That is what made `0019` incompatible: the previous revision inserts a `users` row for an unknown provider subject without resolving the address, so the moment the index exists that request becomes a raw `23505`. Splitting the migration alone would have moved the problem rather than removed it — during the window before the index, two first sign-ins for one person could still create two Accounts.

So the resolver now owns the invariant. Every path that creates or moves an address first takes an advisory lock on the normalized address itself:

- Two first sign-ins for one person serialize, so the second sees the Account the first created instead of a gap. One Account, no index required.
- Two Accounts exchanging addresses take different keys, never contend, and each simply reads the other still holding its address. Both get `AUTH_EMAIL_CONFLICT` by decision.

That last point also removes the deadlock properly rather than narrowing it. The second row lock is gone — a transaction holds its own Account row and never waits on another's — so the cycle cannot form. I deleted the `40P01` classification rather than keeping it as unreachable defence; the swap regression is what proves the conflict path, and it is now deterministic instead of depending on which victim PostgreSQL picks.

**`0019` is now genuinely expand-only**: columns, casing normalization, verification backfill. Nothing changes how the previous revision behaves, and the regression asserts that directly — it replays that revision's unconditional insert of a differently-cased address and requires it to *succeed*.

**The index moves to `0020`** in #196, one deployment later, with the duplicate guard. Its test replays the window: seed the duplicate a pre-resolver writer could have created, assert the migration aborts with the offending ids and the index is absent, remove it, re-run, and assert the index then rejects a differently-cased insert.

**The concurrency tests changed shape with the mechanism.** Same-address races can no longer be held past the lookup — that is the guarantee — so they assert the serialized outcome instead. The barrier helper now fails fast with a message instead of hanging if it is ever pointed at a single address; I hit exactly that while converting them.

One consequence worth flagging for your re-review: `#claimAccountByEmail` deliberately takes no `FOR UPDATE` on the address holder. The advisory lock is the mutual exclusion, and adding a row lock back is precisely what would reintroduce the swap deadlock.

**Validation on `ed03cb3`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (173/173) all pass; the identity suite is stable across repeated runs. #196 is at `f4e349f` with 182 integration tests.

baixiaohang (changes requested):
- Rationale: Moving the unique index out of `0019` restores the previous writer's availability, but the new resolver can now attach a trusted identity to an arbitrary duplicate Account, and the central advisory-lock invariant is not deterministically exercised.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes an Account schema migration and authentication identity logic.
- Semantic lift: The new head changes the serialization and staged-enforcement model for durable Account ownership.

**PR summary**
- Problem: Keep one normalized email associated with one Account without breaking first Google sign-in on an old server during rollout or rollback.
- Approach: Make `0019` expand-only, serialize current resolver writes on a normalized-email advisory key, and defer the database uniqueness backstop until old writers have drained.
- Impacted modules: Server Account/identity persistence, migrations and snapshots, database error classification, concurrency tests, and development documentation.

**Review findings**
❌ 1. The staged window now permits duplicate email owners, but trusted attachment silently chooses one. The migration regression explicitly proves that an old writer may insert `CASING@EXAMPLE.COM` after `0019`; pre-existing duplicates are no longer rejected either. `#claimAccountByEmail` then selects one matching `users` row with unordered `limit(1)`, and the unknown-subject branch can attach a verified Google identity—and therefore Account ownership—to whichever duplicate PostgreSQL returns. Detect more than one normalized owner and return a typed conflict without attaching, then cover both a pre-existing duplicate and a duplicate created by the simulated post-migration old writer. The later index migration must still reject these rows for operator reconciliation. [R4/R5 / `packages/server/src/services/auth/identity-service.ts:158`, `packages/server/src/__tests__/integration/auth-migrations.test.ts:416`]

❌ 2. The same-address concurrency regressions no longer force the advisory-lock interleaving they are meant to protect. Bare `Promise.allSettled` can let the first resolution commit before the second starts; both creation and same-target email-change tests then pass even if the new address lock is removed. Add an acquisition/attempt seam or database-visible barrier that proves the first transaction holds the normalized-address claim while the second is blocked, then assert the exact typed/state outcome. The existing helper explicitly cannot do this for one address. Also update the swap comment, which still describes the removed unique-index wait/deadlock mechanism. [R4 / `packages/server/src/__tests__/integration/account-identity-resolution.test.ts:172`]

⚠️ 3. Stacked #196 head `f4e349f` now contains the post-drain duplicate guard, index creation, and replay regression, so that part of the staged sequence is present. Update this PR body before approval: it still says `0019` rejects duplicates and creates `users_email_unique`, which is now deliberately false. [R5 / `packages/server/drizzle/0019_charming_tony_stark.sql:1`]

✅ 4. Removing the index from `0019` resolves the exact mixed-version `23505`/500 blocker, and the address-key design avoids the prior two-row swap cycle for current resolver writers.

**Action taken**
- Submitted request changes on exact head `ed03cb38cf5bf5d175d87a8acd2682219dde36d3`. No local tests or code checks were run; CI was still in progress.

bestony:
All three correct. Fixed in `ce82ed2`; the PR body is rewritten.

**1 — duplicate owners, arbitrary choice.** This is the one I should have seen when I deferred the index: I created the window and then kept reading it with an unordered `limit(1)`. A verified Google identity — and with it Account ownership — would have attached to whichever row PostgreSQL returned. `#claimAccountByEmail` now reads two and refuses when both exist, so ownership is never handed to an arbitrary row and the rows survive for the `0020` guard to report by id.

Covered by a regression that seeds the exact duplicate a pre-resolver writer leaves behind and asserts a trusted, verified identity is still refused with nothing attached. On #196 that state is unreachable, so the corresponding test there asserts the index rejects the fixture instead.

**2 — the concurrency tests could not fail.** Right, and this is the second time bare `Promise.allSettled` gave me a test that passed for the wrong reason. Two promises started together are free to run in sequence, so both same-address tests passed whether or not the address claim existed.

They now use a seam that asserts the property rather than the outcome: the first transaction is held inside its claim, the second is started, and it must make no progress until the claim is released. Verified by deleting the advisory lock — both fail on exactly that assertion:

```
AssertionError: the second resolution ran while the first held the address claim: expected true to be false
```

The swap comment is corrected too; it still described the unique-index wait that the address claim replaced.

**3 — stale PR body.** Rewritten. It now states that `0019` is expand-only, that the index is deliberately absent and why, and that it lands in #196 one deployment later.

**Validation on `ce82ed2`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (174/174) all pass. #196 is at `aaa1787` with 183.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: Exact head `ce82ed22bf5dfde42d5c1a7c64dad684dfd477ea` resolves every reported migration, ownership, trust, verification, and concurrency blocker while preserving the previous writer during the staged rollout.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes Account schema migration and authentication identity transaction logic.
- Semantic lift: The change defines durable Account ownership and mixed-version write semantics.

**PR summary**
- Problem: Prevent email casing, first provider sign-in, or concurrent sign-in from splitting one person across multiple Accounts.
- Approach: Normalize provider email, serialize current resolver decisions on the address, attach only trusted verified Google identities, maintain verification state transactionally, and defer the database uniqueness backstop until the pre-resolver writer drains.
- Impacted modules: Server Account/identity persistence, migrations and snapshots, Google authentication, shared errors, database error classification, integration tests, and development documentation.

**Review findings**
✅ 1. Migration `0019` is genuinely expand-only: the previous revision can keep serving and roll back without the unique-index `23505`/500, while the current resolver owns the create-or-attach invariant.
✅ 2. Duplicate normalized owners in the staged window are detected rather than selected arbitrarily, so a trusted identity cannot acquire whichever Account PostgreSQL happens to return.
✅ 3. Same-address creation and email-move regressions now hold the first transaction inside its address claim and prove the second cannot progress; swap behavior remains typed and atomic without a row-lock cycle.
✅ 4. Provider trust, verification maintenance, conflict classification, PR documentation, and stacked #196's post-drain duplicate guard/index are aligned with the staged rollout.
✅ 5. CI is green: 6/6 checks passed on this exact head.

**Action taken**
- Approved exact head `ce82ed22bf5dfde42d5c1a7c64dad684dfd477ea` with explicit reviewer authorization. No local tests or code checks were run as part of this final review.

yuezengwu (approved):
**Recommendation: approve**

Reviewed exact head `ce82ed22bf5dfde42d5c1a7c64dad684dfd477ea` independently from the dedicated detached worktree. The prior migration, ownership, verification, trust-boundary, and concurrency blockers are resolved on this head.

1. **`0019` is genuinely expand-only.** It adds the two nullable/default-compatible Account fields, normalizes existing casing, and backfills verification, but deliberately creates no email unique index. The migration regression applies the full current history and proves the pre-resolver writer can still perform its differently-cased Account insert, so rolling overlap and application rollback do not turn first Google sign-in into raw `23505`/500.

2. **The current resolver owns the normalized-address decision safely.** A transaction-scoped advisory claim serializes same-address create and move paths; two pre-existing normalized owners are refused rather than selected; swaps take different claims, observe the other owner, and reject without a row-lock cycle. Provider trust remains explicit and Google-only for implicit attachment, while unverified or untrusted claims return typed `AUTH_EMAIL_CONFLICT` without partial identity writes.

3. **All live Account-email paths maintain the intended verification meaning.** Bootstrap stays normalized and unverified; new, explicit-link, returning, and provider-email-change paths update `users.email` and `email_verified` transactionally, and a returning verified identity heals a row left false by a mixed-version writer.

4. **The stacked rollout contract is now compatible.** Fresh stacked #196 head `6c25eabcefc2b9b883d53843e24ef7bbde252f60` also removes `users_email_unique` from `0020`, so a skipped intermediate Staging deployment no longer reintroduces the old-writer `23505` failure. The database backstop is deferred until the resolver revision is verifiably serving and old writers have drained.

**Validation on this exact head**

- `pnpm check` — pass
- `TURBO_FORCE=true pnpm build` — pass
- `TURBO_FORCE=true pnpm typecheck` — pass
- `TURBO_FORCE=true pnpm test` — pass
- `pnpm --filter @opentag/server test:integration` — pass (`174/174`)
- Focused Account identity/concurrency suite — pass in 3/3 repeated runs (`12/12` each)
- GitHub CI — 6/6 successful
- `git diff --check` — pass; detached worktree clean

The PR is currently `CONFLICTING`/`DIRTY` against `main`; this approval is bound only to the exact head above. Any conflict resolution, rebase, or other head movement requires a fresh review and fresh evidence.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The merge from current `main` preserves the reviewed Account-identity behavior and correctly regenerates the expand-only migration as `0020` on top of `0019_previous_magneto`.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes an Account schema migration and authentication identity transaction logic.
- Semantic lift: The change defines durable Account ownership and mixed-version write semantics.

**PR summary**
- Problem: Prevent email casing, first provider sign-in, or concurrent sign-in from splitting one person across multiple Accounts.
- Approach: Normalize provider email, serialize resolver decisions on the address, attach only trusted verified Google identities, maintain verification state transactionally, and defer the database uniqueness backstop until pre-resolver writers have drained.
- Impacted modules: Server Account/identity persistence, migrations and snapshots, Google authentication, shared errors, integration tests, and development documentation.

**Review findings**
✅ 1. The conflict resolution is correct: `0020_large_jack_power` follows main's `0019_previous_magneto`, the journal range and replay boundary advance consistently, and the generated `0020` snapshot differs from `0019` only by the intended `users.email_verified` and `users.image` columns.
✅ 2. The already-reviewed resolver, provider-trust, verification, duplicate-owner, and deterministic concurrency logic is unchanged from `ce82ed2`.
✅ 3. The PR is mergeable and CI is green: 6/6 checks passed on exact head `30dde56d119ee87eb630d3c4f28bd151fca0b54a`.
⚠️ 4. Non-blocking follow-up: two integration-test comments still say the backfill is migration `0019` and that the unique index lands "one deployment later"; the PR body likewise still says #196 adds it. The merge moved this backfill to `0020`, and current #196 defers the index to a later release. The checked-in development documentation and executable behavior are already correct. [`packages/server/src/__tests__/integration/account-identity-resolution.test.ts:268`]

**Action taken**
- Approved exact head `30dde56d119ee87eb630d3c4f28bd151fca0b54a`. No local tests or code checks were run as part of this re-review.

yuezengwu (approved):
**Recommendation: approve**

Reviewed exact head `30dde56d119ee87eb630d3c4f28bd151fca0b54a` independently after the merge from current `main` (`88724ed0efa78b6b046a24caa731827446aee18b`). The conflict resolution preserves the accepted Account-identity semantics and correctly regenerates this PR's expand-only migration as `0020`.

1. **The merge preserves the reviewed runtime behavior.** The resolver, provider-trust allowlist, verified-email maintenance, duplicate-owner refusal, typed conflicts, and deterministic concurrency regressions are byte-for-byte unchanged from the previously reviewed implementation. Same-address create/move paths remain serialized by the normalized-address advisory claim; swaps reject without partial ownership changes.

2. **The regenerated migration chain is coherent.** `0020_large_jack_power` follows main's `0019_previous_magneto`; the journal is continuous, `0020_snapshot.prevId` equals the current `0019` snapshot id, and the generated snapshot adds only the intended `users.email_verified` and `users.image` columns. No `users_email_unique` index appears in this PR or fresh stacked #196, preserving the old-writer rolling/rollback contract.

3. **The implementation satisfies the approved durable identity direction.** Bootstrap/explicit-link/returning/provider-email-change writes keep Account email and verification state in the same transaction; trusted verified Google identities may attach to the one normalized owner, while unverified, untrusted, duplicate-owner, and concurrent-loss paths return typed conflicts without partial identity writes.

**Validation on this exact head**

- `pnpm check` — pass
- `TURBO_FORCE=true pnpm build` — pass
- `TURBO_FORCE=true pnpm typecheck` — pass
- `TURBO_FORCE=true pnpm test` — pass
- `pnpm --filter @opentag/server test:integration` — pass (`182/182`)
- Focused Account identity/concurrency suite — pass in `3/3` repeated runs (`12/12` each)
- GitHub CI — `6/6` successful
- `git diff --check` — pass; detached exact-head worktree clean; no unresolved review threads

Non-blocking documentation correction: the PR body and two test comments still call the Account migration `0019` and say the email index lands in #196/one deployment later. On this head the migration is `0020`, and fresh #196 deliberately defers that index to a later release after old writers have verifiably drained. The executable migrations and English/Chinese development documentation already state the correct rollout contract; please align the PR record/comments as well.

This approval is bound only to exact head `30dde56d119ee87eb630d3c4f28bd151fca0b54a`; any head movement requires a fresh review.

PR #196, opened 2026-08-27
feat(server): mount Better Auth over the existing Account tables

bestony:
## Summary

> Stacked on #195. Base is `refactor/better-auth-email-normalization`, so this diff shows only this step. Review and
> merge #195 first; GitHub will retarget this to `main` automatically.
>
> Most of the added lines are the generated `drizzle/meta/0020_snapshot.json` and `pnpm-lock.yaml`. The hand-written
> change is concentrated in `src/auth/better-auth.ts`, `src/db/schema/better-auth.ts`, and the integration tests.

Second step of the Better Auth migration: introduce the instance and its table mapping. **No HTTP surface is exposed and
no request path changes behavior.** Nothing in the browser or the CLI can reach Better Auth after this lands.

Today's Account authentication is hand-rolled: stateless HS256 JWTs with no server-side session store (so logout cannot
revoke anything), a hand-written Google OIDC flow without PKCE, a hand-written double-submit CSRF, and one
`OPENTAG_JWT_SECRET` signing four unrelated audiences at once.

### The mapping is in place, not parallel

`users` keeps its `uuid` primary key and every foreign key that references it, and `auth_identities` keeps the rows
already linking Accounts to Google:

| Better Auth model | table | notes |
| --- | --- | --- |
| `user` | `users` | `name` → `displayName` |
| `account` | `auth_identities` | `providerId` → `provider`, `accountId` → `subject` |
| `session` | `auth_sessions` | new; `sessions` already means IM conversations |
| `verification` | `auth_verifications` | new |

The load-bearing question was whether Better Auth's Google provider writes the same identity triple the existing rows
carry. It does, byte for byte — `google` / `https://accounts.google.com` / the OIDC `sub`
(`@better-auth/core/dist/social-providers/google.mjs:54-57`). Had it differed, every returning Account would have been
treated as new. The integration tests pin this by seeding a row the way the pre-migration resolver wrote it and
resolving it through `findAccountByKey`.

### Why no routes

An earlier revision mounted Better Auth's catch-all. Reviewers reproduced two bypasses through it: `/api/v1/auth/update-user`
persisted a 300-character display name past `UserDisplayNameSchema` and the canonical `/api/v1/me` writer, and the
callback issued sessions without `PostAuthenticationService.completeInTransaction`, leaving an Account without its
compatibility Workspace grant. They were right that "no client sends traffic" is not a guard.

Nothing consumes those routes in this stage, so the allowlist they need is empty. The mount arrives with the stage that
has real requirements to guard — this PR is the instance, the mapping, and the migration.

### Instance-level guards that do land here

- **Suspended Accounts cannot get a session.** Suspension is the only kill switch this system has, and the legacy path
  re-checks it on every authenticated request.
- **Provider credentials are never persisted.** OpenTag never calls a provider on an Account's behalf and the
  pre-migration resolver stored none either, so access, refresh, and ID tokens are blanked on write. Omitting beats
  encrypting: Better Auth 1.7.2's `encryptOAuthTokens` covers access and refresh but writes `idToken` through in the clear.
- **Email is lowercased** on every write Better Auth makes, matching the invariant #195 establishes.

### Migration `0020`

Expand-only — two new tables plus nullable columns on `auth_identities`. Its guard reports conflicting rows by id if any
`(issuer, subject)` pair is already shared, rather than surfacing a bare unique violation.

### Deployment prerequisite

`BETTER_AUTH_SECRET` (≥32 characters) must be set on the app **before** this revision deploys; configuration is
validated before the server listens, so a missing value leaves the container restarting rather than serving degraded.
It is now rejected if it equals `OPENTAG_JWT_SECRET`, since sharing one key would make either rotation invalidate the
other's credentials. Documented in `docs/deploying.md` and its Chinese mirror.

## Testing

- `pnpm --filter @opentag/server test:integration` — 175 passing
- `better-auth.test.ts` covers, against a real Postgres: resolving a pre-migration identity row, reading the Account
  through the mapped user model, issuing a session into `auth_sessions` and reading it back from a bearer token,
  rejecting an unknown bearer token, refusing a session for a suspended Account, never persisting provider credentials,
  and lowercasing an email written through Better Auth.
- `config.test.ts` asserts `BETTER_AUTH_SECRET` is required, length-checked, and rejected when equal to `OPENTAG_JWT_SECRET`.

Not covered by automated tests: a live Google sign-in, which needs OAuth credentials. No HTTP-level tests exist because
no HTTP surface is mounted; those belong with the stage that mounts it.

## Breaking changes

None at runtime — no request path changes behavior. Operationally, `BETTER_AUTH_SECRET` becomes a required environment
variable, must differ from `OPENTAG_JWT_SECRET`, and must be configured before deploying.

## Checklist


baixiaohang (changes requested):
- Rationale: the in-place table mapping is directionally sound, but the newly live Better Auth endpoints bypass existing Account invariants and the deployment/configuration path is not yet internally consistent.

**Risk level**: B-high
- Path baseline: changes `packages/server/**` authentication behavior and adds database tables, columns, indexes, and migration guards.

**PR summary**
- Problem: staging needs a way to exercise Better Auth over existing Accounts before the browser and CLI switch away from the legacy login paths.
- Approach: mount Better Auth at the versioned auth boundary, map its user/account models onto `users` and `auth_identities`, add server-side session and verification storage, and keep existing clients on the old tokens for now.
- Impacted modules: Server auth routing and startup, Account/identity persistence, database migrations, deployment configuration, and integration tests.

**Review findings**
❌ 1. The mounted OAuth endpoints bypass the existing post-authentication transaction. `createBetterAuth` has no equivalent of `PostAuthenticationService.completeInTransaction`, even though Better Auth's callback creates the user, identity, and session by default. A direct request to the now-public routes can therefore create an Account without its required compatibility Workspace/grant, and an existing suspended Account can receive a Better Auth session because `suspendedAt` is never checked. "No current client sends traffic" is not a guard once the routes are mounted; preserve these Account invariants before session issuance, or keep the write paths disabled until that guard exists. [R4 / `packages/server/src/auth/better-auth.ts:31`, `packages/server/src/index.ts:130`]

❌ 2. `trustedProviders: ["google"]` does not provide the claimed matching-email link on Better Auth 1.7.2. Its current callback also requires the existing local user to have `emailVerified: true`; stacked migration 0018 sets that only for users that already have an `auth_identities` row. The important case—an existing Account with the same email but no Google identity—still returns `account not linked`. Please define the safe ownership proof/linking path and pin it with an integration test; simply disabling Better Auth's local-verification guard is documented upstream as an account-takeover risk and is already deprecated. [R5 / `packages/server/src/auth/better-auth.ts:60`]

❌ 3. The production-container startup consumer was not updated for the new required variable. `.github/workflows/ci.yml` starts the image without `BETTER_AUTH_SECRET`, so the container exits in `parseServerConfig`; the current Container Smoke and fan-in CI failures reproduce this exact path. Add a CI-only value distinct from the legacy JWT secret so the checked production image can reach readiness. [R5 / `packages/server/src/config.ts:88`]

❌ 4. The code does not enforce the documented secret separation, and `.env.example` currently gives `OPENTAG_JWT_SECRET` and `BETTER_AUTH_SECRET` the same placeholder value. The config test only observes that its two fixtures happen to differ; equal values still parse successfully. Reject equality in configuration, make the example distinct, and include the new secret in the startup redaction set so this new credential follows the same failure-path handling as the existing secrets. [R4 / `.env.example:7`, `packages/server/src/config.ts:119`, `packages/server/src/index.ts:103`]

❌ 5. The new account mapping persists Google access, refresh, and ID tokens, while Better Auth's `account.encryptOAuthTokens` remains at its plaintext default. That adds durable bearer credentials which the old Google flow did not retain. Enable Better Auth's supported token encryption before exposing the callback, or provide an explicit accepted trust-boundary rationale for storing these credentials unencrypted. [R3 / `packages/server/src/auth/better-auth.ts:56`, `packages/server/src/db/schema/auth-identities.ts:20`]

**Action taken**
- Submitted request changes on head `c90e4bf4cbf50dec5fbdd0dcee9d89dabe6415e0`.

yuezengwu (changes requested):

Reviewed exact head `c90e4bf4cbf50dec5fbdd0dcee9d89dabe6415e0`. The in-place Account/identity mapping plus server-side session/verification persistence is now an approved direction, but the current implementation is not safe to stage.

1. **Restack before doing any migration work.** This head is still based on `d146504f`, while stacked PR #195 has moved to `3572b278` and now owns `0019_charming_tony_stark`. GitHub correctly reports this PR as `CONFLICTING`/`DIRTY`; this branch's independent `0019_cool_captain_marvel` cannot be retained. Restack onto #195's fresh accepted head and regenerate this migration as the next non-conflicting journal/snapshot entry. [`packages/server/drizzle/meta/_journal.json`]

2. **Do not mount Better Auth's general write surface over the Account table.** The catch-all exposes `/api/v1/auth/update-user`, which is always enabled and writes Better Auth `name` directly into mapped `users.display_name`, bypassing the OpenTag profile mutation boundary and `UserDisplayNameSchema`. I reproduced this against PostgreSQL: after issuing a Better Auth session for a legacy Account, a bearer-authenticated POST to `/api/v1/auth/update-user` with a 300-character name returned 200 and persisted all 300 characters. Whitelist only the endpoints needed for this stage, or explicitly block profile writers such as `/update-user`; the existing `/api/v1/me` domain API must remain the sole Account profile writer. [`packages/server/src/auth/fastify-handler.ts:29`, `packages/server/src/auth/better-auth.ts:32`]

3. **Preserve Account authentication invariants before any Better Auth session is issued.** The OAuth callback still bypasses `PostAuthenticationService.completeInTransaction`, so it can create an Account without the required compatibility Workspace/grant and issue a session for a suspended Account. The claimed same-email linking also remains unsafe/incomplete: Better Auth requires the existing local Account to be verified, while current #195 still fails the bootstrap-Account -> first Google sign-in case and leaves live `emailVerified` writers stale. Fix #195's atomic verified-email attachment policy, then add Better Auth callback coverage for new, linked, returning, suspended, and provider-email-change cases. [`packages/server/src/auth/better-auth.ts:31`, `packages/server/src/index.ts:130`]

4. **Encrypt or omit every persisted OAuth credential.** The new columns currently store Google access, refresh, and ID tokens in plaintext. Better Auth's `account.encryptOAuthTokens` is currently off; moreover, in 1.7.2 the built-in path applies its token utility to access/refresh tokens but writes `idToken` directly, so merely flipping that flag is not sufficient for the ID token. Use the application's reviewed encryption boundary (or do not persist tokens that are not needed), and add round-trip/non-disclosure tests proving plaintext credentials never appear in database rows, logs, or returned account data. [`packages/server/src/auth/better-auth.ts:56`, `packages/server/src/db/schema/auth-identities.ts:20`]

5. **Finish the secret and deployment contract.** `BETTER_AUTH_SECRET` is absent from Container Smoke, which is why Container Smoke and fan-in CI fail. Equal Better Auth/JWT secrets still parse, `.env.example` gives both the same value, and startup redaction omits the new secret. Enforce distinct values, use distinct examples/CI values, redact the new secret, and cover the failure path. [`packages/server/src/config.ts:88`, `packages/server/src/index.ts:99`, `.github/workflows/ci.yml`]

Also extend the real-database contract beyond direct adapter calls: exercise the exposed session and verification endpoint lifecycle, cookie/bearer behavior, revocation/expiry, and the endpoint allowlist. The current five tests do not cover the mounted HTTP boundary or `auth_verifications`.

**Validation on this exact head**

- `pnpm check` — pass
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass
- `pnpm --filter @opentag/server test:integration` — pass (165/165)
- Focused PostgreSQL profile-writer reproduction — defect confirmed; temporary probe removed and worktree restored clean
- GitHub CI — 4 pass, Container Smoke + fan-in CI fail

baixiaohang (changes requested):
- Rationale: the branch is now cleanly restacked and its startup consumers are wired, but the mounted Better Auth surface still bypasses Account profile/authentication invariants and persists OAuth credentials without an adequate protection boundary.

**Risk level**: B-high
- Path baseline: changes `packages/server/**` authentication behavior and adds database tables, columns, indexes, and migration guards.

**PR summary**
- Problem: staging needs a way to exercise Better Auth over existing Accounts before the browser and CLI switch away from the legacy login paths.
- Approach: mount Better Auth at the versioned auth boundary, map its user/account models onto `users` and `auth_identities`, add server-side session and verification storage, and keep existing clients on legacy tokens for now.
- Impacted modules: Server auth routing and startup, Account/identity persistence, database migrations, deployment configuration, and integration tests.

**Review findings**
❌ 1. The catch-all still exposes Better Auth's general profile writer at `/api/v1/auth/update-user`. Because `user.name` is mapped directly to `users.display_name`, that route remains a second Account-profile mutation boundary and bypasses OpenTag's `UserDisplayNameSchema`, suspension guard, and canonical `/api/v1/me` API. Whitelist only the endpoints needed for this staging step, or explicitly reject profile writers such as `/update-user`. [R1/R5 / `packages/server/src/auth/fastify-handler.ts:29`, `packages/server/src/auth/better-auth.ts:55`]

❌ 2. Session issuance still bypasses `PostAuthenticationService.completeInTransaction`: a new OAuth Account can be left without its compatibility Workspace/grant, and a suspended Account can receive a Better Auth session. The matching-email claim also remains incomplete because Better Auth requires the existing local user to be verified, while the stacked migration verifies only users that already have an identity. Preserve these invariants before session issuance and pin the callback contract for new, linked, returning, suspended, and provider-email-change cases. [R4/R5 / `packages/server/src/auth/better-auth.ts:31`, `packages/server/src/index.ts:132`]

❌ 3. Google access, refresh, and ID tokens are still written into plaintext columns. Better Auth's encryption option is not enabled, and on Better Auth 1.7.2 its built-in token utility does not cover `idToken`, so merely flipping that option would still leave one credential exposed. Use the application's reviewed encryption boundary or omit credentials that are not needed, and add non-disclosure/round-trip coverage proving plaintext credentials never reach stored rows, logs, or returned account data. [R3 / `packages/server/src/auth/better-auth.ts:56`, `packages/server/src/db/schema/auth-identities.ts:20`]

❌ 4. The deployment-secret fix is only partial. Container Smoke, onboarding E2E, and startup redaction are now covered, but configuration still accepts `BETTER_AUTH_SECRET === OPENTAG_JWT_SECRET`, and `.env.example` still assigns both secrets the same placeholder. Enforce separation, use distinct example values, and add an equality-rejection test. [R4 / `packages/server/src/config.ts:119`, `.env.example:7`]

✅ 5. The branch is correctly restacked on the updated #195 head: the Better Auth migration is now coherent `0020_fine_frog_thor`, and the previously missing startup consumers plus redaction entry have been fixed.

**Action taken**
- Submitted request changes on head `5d206c1e9262cd7c681e27f86aecf38cb0656972`.

baixiaohang (changes requested):
- Rationale: this head only absorbs #195's legacy login-writer fix; the PR-owned Better Auth route/configuration files are unchanged, so the four staging-safety blockers remain.

**Risk level**: B-high
- Path baseline: changes `packages/server/**` authentication behavior and adds database tables, columns, indexes, and migration guards.

**PR summary**
- Problem: staging needs a way to exercise Better Auth over existing Accounts before the browser and CLI switch away from legacy login paths.
- Approach: mount Better Auth at the versioned auth boundary, map its models onto `users` and `auth_identities`, add server-side session and verification storage, and leave current clients on legacy tokens.
- Impacted modules: Server auth routing and startup, Account/identity persistence, database migrations, deployment configuration, and integration tests.

**Review findings**
❌ 1. The catch-all still exposes Better Auth's `/api/v1/auth/update-user` profile writer. With `user.name` mapped to `users.display_name`, that remains a second Account-profile mutation boundary which bypasses `UserDisplayNameSchema`, the suspended-Account guard, and the canonical authenticated/CSRF-protected `/api/v1/me` API. Whitelist the endpoints needed for this stage or explicitly reject profile writers. [R1/R5 / `packages/server/src/auth/fastify-handler.ts:29`, `packages/server/src/auth/better-auth.ts:55`]

❌ 2. #195 now makes the legacy Google writer maintain verified-email attachment atomically, but Better Auth does not call that writer or `PostAuthenticationService.completeInTransaction`. Its callback can still create an Account without the compatibility Workspace/grant or issue a session for a suspended Account; a bootstrap Account also remains locally unverified before its first Better Auth sign-in, so `trustedProviders` alone cannot safely link it. Route the callback through the canonical identity/post-auth boundary before issuing a session and cover new, linked, returning, suspended, and provider-email-change cases. [R4/R5 / `packages/server/src/auth/better-auth.ts:31`, `packages/server/src/index.ts:132`]

❌ 3. Google access, refresh, and ID tokens still map to plaintext database columns. Better Auth's encryption option is not enabled, and in Better Auth 1.7.2 its built-in token utility does not protect `idToken`, so enabling that option alone is insufficient. Use the application's reviewed encryption boundary or omit unneeded credentials, with non-disclosure/round-trip coverage. [R3 / `packages/server/src/auth/better-auth.ts:56`, `packages/server/src/db/schema/auth-identities.ts:20`]

❌ 4. Configuration still accepts `BETTER_AUTH_SECRET === OPENTAG_JWT_SECRET`, while `.env.example` assigns both the same placeholder. Container/E2E startup injection and redaction are fixed, but the documented separation contract must also be enforced and covered by an equality-rejection test. [R4 / `packages/server/src/config.ts:119`, `.env.example:7`]

✅ 5. The newly merged #195 commit correctly repairs the legacy Google identity writer's verified-email attachment and provider-email-change behavior; that removes the stacked prerequisite defect, but it does not change Better Auth's independent callback path.

**Action taken**
- Submitted request changes on head `0091838b686c5daf9c99f9737b55b9f7142c0174`. No local tests or code checks were run for this review.

yuezengwu (changes requested):
**Recommendation: request changes**

Reviewed exact head `0091838b686c5daf9c99f9737b55b9f7142c0174` against stacked base `72f1b8e5edd2208f28494ad72e2f31dbaac60e50`. The approved direction—mapping Better Auth onto the existing Account/identity tables and adding server-side session/verification persistence—is intact, and this head is now correctly restacked with coherent migration `0020`. However, the mounted implementation is still not safe to expose in staging.

1. **Keep generic profile writers outside the Account boundary.** The `GET`/`POST` wildcard still exposes Better Auth's `/api/v1/auth/update-user`, while `user.name` maps directly to `users.display_name`. On this exact head I reproduced the bypass against PostgreSQL: a bearer-authenticated request with a 300-character `name` returned 200 and persisted all 300 characters, bypassing `UserDisplayNameSchema`, the suspended-Account guard, and canonical `/api/v1/me` profile mutation. Whitelist only the endpoints required for this phase, or explicitly reject `/update-user` and other profile writers. [`packages/server/src/auth/fastify-handler.ts:27`, `packages/server/src/auth/better-auth.ts:55`]

2. **Route Better Auth through the canonical identity/post-authentication contract before session issuance.** Fresh #195 fixes the legacy Google writer, but this Better Auth instance still calls neither `AuthIdentityService` nor `PostAuthenticationService.completeInTransaction`; its only Account hook normalizes email. That leaves the callback/session path without the compatibility Workspace/grant and suspension invariants, and `trustedProviders` alone does not define the bootstrap-Account -> first Google identity proof/linking path. A second focused PostgreSQL probe on this exact head showed that a Better Auth bearer session continues to resolve after its Account is marked `suspendedAt`. Make session resolution/issuance fail closed for suspended Accounts and cover new, bootstrap-linked, returning, provider-email-change, conflict, and suspended cases transactionally. [`packages/server/src/auth/better-auth.ts:31`, `packages/server/src/auth/better-auth.ts:60`, `packages/server/src/services/auth/post-authentication.ts:26`]

3. **Encrypt or omit every persisted OAuth credential.** The mapping adds plaintext columns for Google access, refresh, and ID tokens, while this config does not enable any token encryption. In installed Better Auth 1.7.2, the built-in `encryptOAuthTokens` helper applies to access/refresh tokens, but the callback writes `idToken` directly, so flipping that flag alone is insufficient. Use the application's reviewed encryption boundary (or omit credentials not needed by this phase) and add round-trip/non-disclosure tests proving plaintext credentials never appear in database rows, logs, traces, or returned account data. [`packages/server/src/auth/better-auth.ts:56`, `packages/server/src/db/schema/auth-identities.ts:20`]

4. **Enforce the documented secret separation.** Container Smoke, onboarding E2E injection, and startup redaction are fixed and CI is now green. But config still accepts `BETTER_AUTH_SECRET === OPENTAG_JWT_SECRET`, and `.env.example` assigns the same placeholder to both. Reject equality, make the example values distinct, and test the rejected startup path. [`packages/server/src/config.ts:119`, `.env.example:7`]

5. **Exercise the mounted security/lifecycle boundary, not only direct adapters.** The five Better Auth integration tests do not send requests through the catch-all and do not cover the endpoint allowlist, callback/linking transaction, `auth_verifications`, cookie/CSRF behavior, expiry, revocation, or suspension. Add HTTP-level lifecycle tests so the protections above cannot regress. [`packages/server/src/__tests__/integration/better-auth.test.ts:52`]

Positive changes on this head: #195 is absorbed at its fresh head, migration `0020_fine_frog_thor` is coherent, the previous Container Smoke/E2E secret startup gaps are fixed, secret redaction is wired, and all current GitHub checks pass.

**Validation on this exact head**

- `pnpm install --frozen-lockfile` — pass
- `pnpm check` — pass
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass
- `pnpm --filter @opentag/server test:integration` — pass (173/173)
- Focused PostgreSQL HTTP profile-writer probe — defect reproduced
- Focused PostgreSQL suspended-session probe — defect reproduced
- Temporary probe code removed; `git diff --check` passes and the detached worktree is clean
- GitHub CI — 6/6 checks pass

Please also refresh the PR body before the next review: it still names migration `0019`, reports 165 integration tests, and says no request path changes behavior even though the Better Auth wildcard is live.

bestony:
All five findings were correct. Fixed in `c819710`; the PR body is refreshed.

The strongest point was that "no current client sends traffic" is not a guard once the catch-all is mounted. That was the root of findings 1, 2 and 5, and it exposed a flaw in how I framed this stage rather than a detail to patch: the PR claimed it did not change any request path while publishing Better Auth's entire endpoint surface.

**1 & 2 — profile writer and post-auth bypass.** Nothing consumes these routes in this stage, so the allowlist they need is empty. `src/auth/fastify-handler.ts` and its `createApp`/startup wiring are removed. The mount arrives with the stage that has real requirements to guard — where `/update-user` and the other profile writers can be rejected explicitly and the callback can be routed through `AuthIdentityService` and `PostAuthenticationService.completeInTransaction` before any session is issued. What lands here is the instance, the in-place mapping, and the migration.

**2 (suspension) — fixed now, not deferred.** Suspension belongs to the instance, not the routes, so `databaseHooks.session.create.before` refuses the write for a suspended Account: no token ever comes into existence. Resolution-side enforcement lands with the `user-auth` preHandler in the browser stage, where it replaces the legacy `#resolveActiveUser` check.

**3 — provider credentials.** Omitted rather than encrypted. OpenTag never calls a provider on an Account's behalf, and the pre-migration resolver persisted none either, so a `databaseHooks.account` hook blanks access, refresh and ID tokens on write. That is strictly stronger than the encryption option here, and it sidesteps exactly the gap you identified: 1.7.2's `encryptOAuthTokens` covers access and refresh but writes `idToken` through in the clear. Pinned by a test asserting all five credential columns are null after an update that supplies all of them.

**4 — secret separation.** `parseServerConfig` now rejects `BETTER_AUTH_SECRET === OPENTAG_JWT_SECRET`, with a `config.test.ts` case for the rejected startup path. `.env.example` carries distinct placeholders.

**5 — HTTP-level lifecycle tests.** These follow the mount rather than preceding it; there is no surface to exercise on this head. When the routes land, the allowlist, callback transaction, cookie/CSRF behavior, expiry, revocation and suspension all need HTTP coverage, and that is where it belongs.

On the stale PR body: fixed. It now says `0020`, 175 integration tests, and no longer claims a live wildcard is harmless — it states that no HTTP surface exists.

**Validation on `c819710`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (175/175) all pass.

bestony:
Head is now `914dc82`. This carries #195's identity fixes through the merge, plus the rolling-compatibility finding that belongs to this PR's migration.

**Rolling compatibility / stale backfill.** Correct, and it is the sharpest of the findings: 0019 is expand-only precisely so the previous revision keeps running, and that revision predates the writer maintaining `email_verified`. Nothing reads the flag today, but Better Auth's account linking will and it treats an unverified local Account as unlinkable — the bootstrap-Account defect returning by another route.

I looked at the database-boundary option first and it does not work here. `auth_identities` records no per-identity verification, and `ExternalIdentity` can now legitimately carry `emailVerified: false`, so a trigger would have to assume every identity implies a verified address. That is true for the only provider wired today, wrong as a rule, and wrong in the direction that grants trust. Adding a verification column to `auth_identities` only moves the problem: the old writer would not set it either, so its default would have to be `true`, which is the same bad assumption written into the schema.

So 0020 repeats the reconciliation idempotently, and the contract is stated in the migration rather than left implicit: every migration preceding a consumer of this flag repeats it, and the stage that first reads it must repeat it again. The regression lives in `auth-migrations.test.ts` where the harness replays partial histories — stop at 0019, write the row the old revision would have written, let the next migration reconcile it — and it also asserts an Account with no provider identity is left `false`, so the backfill stays scoped to identities that actually assert an address.

The two findings from the previous round are unchanged on this head: the catch-all is still removed, and the suspension and credential-omission guards still hold.

**Validation on `914dc82`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (179/179) all pass.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the Better Auth instance and in-place persistence mapping are now isolated from HTTP traffic, the instance-level Account and credential guards are complete for this stage, and the rolling verification reconciliation plus the repaired base transaction behavior close the reported migration/concurrency gaps.

**Risk level**: B-high
- Path baseline: `packages/server/**` adds database tables, columns, indexes, migration guards, and authentication persistence.

**PR summary**
- Problem: prepare Better Auth over existing Accounts so the migration can be observed and validated before any browser or CLI traffic depends on it.
- Approach: map Better Auth onto `users` and `auth_identities`, add server-side session/verification storage, and keep the HTTP allowlist empty until the later stage can preserve the canonical Account and post-authentication boundaries.
- Impacted modules: Server authentication configuration, Account/identity persistence, database migrations, deployment configuration, and PostgreSQL integration tests.

**Review findings**
✅ 1. No Better Auth route is mounted. The earlier `/update-user` profile bypass and callback path that skipped the compatibility Workspace/grant boundary are therefore unreachable in this stage; the later route-mounting PR must bring its endpoint allowlist and HTTP lifecycle coverage with it.

✅ 2. Suspended Accounts cannot receive new Better Auth sessions, provider access/refresh/ID credentials are blanked on create and update, emails are normalized, and the Better Auth secret is required, redacted, and rejected when shared with the legacy JWT secret.

✅ 3. Migration `0020` repeats the verified-email reconciliation across the mixed-version writer window, with a partial-history regression that repairs only Accounts carrying provider identities. The current stacked base also classifies both `23505` and the `40P01` unique-index address-swap outcome as the same atomic email conflict.

⚠️ 4. This remains stacked on #195. That base must complete its own current-head review and merge first; this approval neither approves nor merges #195.

**Action taken**
- Approved exact head `6a0fd4ea55c6a88bcb4fc5f4b7fc2920e130fc8a`. GitHub CI is 6/6 successful; no local tests or code checks were run for this review.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the Better Auth instance remains isolated from HTTP traffic, all stage-level Account/credential/configuration guards remain intact, and the refreshed base now pins both the typed two-sided address-swap outcome and atomic rollback.

**Risk level**: B-high
- Path baseline: `packages/server/**` adds database tables, columns, indexes, migration guards, and authentication persistence.

**PR summary**
- Problem: prepare Better Auth over existing Accounts so the migration can be validated before browser or CLI traffic depends on it.
- Approach: map Better Auth onto `users` and `auth_identities`, add session/verification storage, keep the HTTP allowlist empty for this stage, omit provider credentials, and reconcile verified-email state across mixed-version deployments.
- Impacted modules: Server authentication configuration, Account/identity persistence, database migrations, deployment configuration, and PostgreSQL integration tests.

**Review findings**
✅ 1. No Better Auth route is mounted, so the prior profile-writer and post-authentication bypasses remain unreachable. The later mounting stage must bring its endpoint allowlist and HTTP lifecycle coverage.

✅ 2. Suspended Accounts cannot receive new sessions; provider access, refresh, and ID credentials are blanked; email is normalized; and the Better Auth secret is required, redacted, and distinct from the legacy JWT secret.

✅ 3. Migration `0020` safely repeats verified-email reconciliation across the mixed-version window. The refreshed base classifies the `23505`/`40P01` address-move outcomes and now asserts both swap operations reject as `AUTH_EMAIL_CONFLICT` while both Accounts retain their original emails.

⚠️ 4. This remains stacked on #195. That base must complete its own current-head review and merge first; this approval neither approves nor merges #195.

**Action taken**
- Approved exact head `e31e68385e58d5764cb826b1368335f825ea372e`. GitHub CI is 6/6 successful; no local tests or code checks were run for this review.

yuezengwu (approved):
**Recommendation: approve**

Reviewed exact head `e31e68385e58d5764cb826b1368335f825ea372e` against stacked base `ac72ab1dee86bcbc7e561fb5f7a627935b000a93` from a clean detached worktree. The previously reported stage-level blockers are resolved on this head.

1. **No Better Auth HTTP surface is mounted.** Production source has no import or construction of `createBetterAuth`; only the isolated instance module exists. The earlier generic `/update-user` profile writer and callback path that bypassed the canonical Account/post-authentication boundary are therefore unreachable. It is appropriate that this stage has no Better Auth HTTP lifecycle tests; the later mounting stage must bring an explicit endpoint allowlist plus callback, Account-linking, cookie/CSRF, expiry/revocation, and suspension-resolution coverage.

2. **Instance and credential guards are effective.** Session creation refuses suspended Accounts; access, refresh, and ID tokens plus expiry values are blanked on both identity create and update; Better Auth email writes normalize to lowercase; Google alone is trusted for implicit linking; and Google’s issuer/account key is byte-for-byte compatible with existing identity rows. `BETTER_AUTH_SECRET` is required, redacted, and rejected when equal to `OPENTAG_JWT_SECRET`.

3. **Migration `0020` is coherent and expand-only.** Its snapshot points to the current `0019` snapshot, the journal sequence is unambiguous, the issuer/subject guard fails with actionable row IDs, and the partial-history regression proves `0020` reconciles Accounts created by the old writer after `0019` while leaving Accounts without an identity unverified. The refreshed #195 base now also pins typed two-sided address-swap rejection and atomic rollback.

**Validation on this exact head**

- `pnpm install --frozen-lockfile` — pass
- `pnpm check` — pass
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (server `267/267`; all packages green)
- `pnpm --filter @opentag/server test:integration` — pass (`181/181`)
- Focused PostgreSQL identity-create credential probe — pass (`8/8` in the Better Auth file); temporary probe removed
- `git diff --check` — pass; detached worktree clean
- GitHub CI — `6/6` successful; no unresolved review threads

`pnpm audit --prod` still reports six advisories whose affected package versions already exist in the stacked base (`@fastify/static`, OpenTelemetry, and a nested esbuild). They are a repository baseline risk rather than introduced package versions in this PR, so they do not change this review outcome.

This approval is only for #196 at the exact head above. The PR remains stacked: #195 must merge first, and any subsequent head movement requires a fresh review.

baixiaohang (changes requested):
- Rationale: the new uniqueness index is safe only after the resolver-only revision has fully rolled out and drained, but the repository's deployment workflow explicitly does not guarantee that intermediate revision will deploy.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes a database index and migration behavior.

**PR summary**
- Problem: prepare Better Auth over existing Accounts without exposing its HTTP surface, while preserving one Account per verified email across rolling deployments.
- Approach: keep Better Auth isolated, map it onto the existing Account tables, add session/verification persistence and guards, and now add the case-insensitive Account-email uniqueness index in migration `0020`.
- Impacted modules: Server authentication persistence, Account/identity resolution, database migrations, deployment configuration, and PostgreSQL integration tests.

**Review findings**
❌ 1. Migration numbering does not establish the rollout boundary this index requires. PR #195 is still unmerged, and Staging intentionally skips an older successful revision once a newer commit becomes the `main` tip. If #195 and #196 merge before #195 finishes deploying, the #195 deploy is skipped and CapRover rolls directly from the pre-resolver server to this head. The new container applies `0020` before listening while the old container is still serving, so `users_email_unique` becomes active against the exact old writer whose bootstrap-Account first-sign-in path this change says would surface a raw `23505`. The partial-history test proves that `0019` itself stays compatible, but it cannot prove that the resolver-only application revision was ever deployed and drained. Remove the index from this stage until a completed #195 rollout is an explicit, verifiable prerequisite, or introduce an enforceable deployment/database transition that remains compatible when Staging skips the intermediate revision. [R4/R5 / `packages/server/drizzle/0020_fine_frog_thor.sql:63`, `.github/workflows/deploy-staging.yml:125`, `docs/deploying.md:30`]

✅ 2. The previously approved stage-level safeguards remain intact: Better Auth has no mounted HTTP surface, provider credentials are omitted, suspended session creation is rejected, secret separation is enforced, and the verification reconciliation remains scoped and idempotent.

**Action taken**
- Submitted request changes on exact head `f4e349f9821c66b6c5703d4227903dcfff59a7eb`. No local tests or code checks were run for this review; the new-head GitHub checks were still running when reviewed.

baixiaohang (changes requested):
- Rationale: the new test proves the index rejects a duplicate after migration, but it does not make that rejection compatible with the pre-resolver server that can still be serving during a skipped-intermediate rollout.

**Risk level**: B-high
- Path baseline: `packages/server/**` changes Account identity behavior and a database uniqueness migration.

**PR summary**
- Problem: prepare Better Auth over existing Accounts without exposing its HTTP surface, while preserving one Account per verified email across mixed-version deployments.
- Approach: keep Better Auth isolated, map it onto the existing Account tables, add persistence and instance guards, and place the case-insensitive Account-email unique index in migration `0020`.
- Impacted modules: Server authentication persistence, Account/identity resolution, database migrations, deployment configuration, and PostgreSQL integration tests.

**Review findings**
❌ 1. The rollout blocker remains. The merged #195 resolver now refuses to choose between two existing Accounts sharing an address, but that code runs only in the new application revision. Staging can still skip #195 and roll directly from the pre-resolver server to this head; while the new container applies `0020` before listening, the old container continues serving and performs its unconditional `users` insert. The new `aaa1787` test confirms that `users_email_unique` rejects exactly that second row; for the old process this is still the raw `23505` bootstrap sign-in failure the migration was moved out of `0019` to avoid. A current-code resolver test cannot establish compatibility for an already-running old binary. Remove/defer the index until #195 has verifiably deployed and drained, or enforce a transition that cannot skip that application boundary. [R4/R5 / `packages/server/drizzle/0020_fine_frog_thor.sql:63`, `packages/server/src/__tests__/integration/account-identity-resolution.test.ts:279`, `.github/workflows/deploy-staging.yml:125`]

✅ 2. The new resolver refusal is correct for requests handled by #195 and safely avoids arbitrarily selecting one of two pre-existing duplicate Accounts. The previously accepted Better Auth HTTP isolation, credential omission, suspension, secret, and verification safeguards also remain intact.

**Action taken**
- Submitted request changes on exact head `aaa178751ad4b121c2c85c8b5a962c896e070ad3`. No local tests or code checks were run for this review; the new-head GitHub checks were still running when reviewed.

bestony:
Correct, and it invalidates the premise of my previous fix rather than a detail of it. Fixed in `6c25eab`: the index is gone from `0020` as well.

Moving it from `0019` to `0020` assumed the deployment in between would drain the pre-resolver server. I asserted that without checking the pipeline — and `deploy-staging.yml` explicitly deploys the tip and skips revisions it has already passed, so a rollout can go straight from that server to this head. Its container keeps serving while the new one applies `0020`. The `aaa1787` test I added proves the index rejects exactly the row that container writes, which is the failure, not the guarantee.

There is no version of this that a migration can enforce, because the precondition is about which binary is serving, not about database state. So the index leaves both PRs:

- Nothing depends on it. The resolver's advisory claim on the normalized address keeps one address on one Account, and #195 now refuses rather than choosing when it finds two — which is the case the index would otherwise have to cover.
- It is defence in depth for a writer that skips the resolver, and belongs in a later release, once the resolver revision is confirmed to be the one serving.

I have noted that as a follow-up rather than leaving it implicit; it should not land until #195 and #196 have deployed and the old container is verifiably gone.

Two side effects, both simplifications: the schema and `account-identity-resolution.test.ts` now match the base branch exactly instead of diverging, and the migration regression drops the `0019` bounding it only needed in order to avoid the index.

**Validation on `6c25eab`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (182/182) all pass.

yuezengwu (approved):
`**Recommendation: approve**\n\nReviewed exact head \`6c25eabcefc2b9b883d53843e24ef7bbde252f60\` against stacked base \`ce82ed22bf5dfde42d5c1a7c64dad684dfd477ea\` from a clean detached worktree. The fresh change resolves the remaining mixed-version deployment blocker.\n\n1. **Migration 0020 no longer assumes an intermediate deployment occurred.** \`users_email_unique\` is absent from the SQL, schema, and generated snapshot. That is the compatible choice because automatic Staging deploys skip a revision that is no longer the main tip; a rollout may therefore go directly from the pre-resolver binary to this head while the old container is still serving. Its unconditional Account insert can continue during that overlap rather than escaping as raw \`23505\`. The approved #195 resolver owns the invariant and refuses ambiguous duplicate owners; the database backstop is correctly deferred until that resolver is verifiably serving and old writers have drained.\n\n2. **The accepted stage boundary remains intact.** No Better Auth HTTP route is mounted, so the generic profile writer and callback/post-authentication bypasses remain unreachable. Suspended Accounts cannot receive new sessions; OAuth access, refresh, and ID credentials are blanked on create/update; Better Auth email writes normalize; and the Better Auth/JWT secrets are required to differ.\n\n3. **The remaining migration is coherent.** It adds only the session/verification persistence and nullable identity fields, guards duplicate issuer/subject pairs with actionable row IDs, and repeats the scoped verification reconciliation for Accounts written in the mixed-version window.\n\n**Validation on this exact head**\n\n- \`pnpm check\` — pass\n- \`pnpm --filter @opentag/server test:integration\` — 175 tests passed before one parallel Testcontainers startup timed out; the skipped Better Auth file then passed independently (7/7), covering all 182 tests without a product failure\n- \`git diff --check\` — pass; detached worktree clean\n- GitHub CI — 6/6 successful\n- Review threads — none unresolved\n\nThis approval is only for #196 at this exact head. It remains stacked on #195; any head/base movement requires a fresh review.`

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the unsafe Account-email index has been removed from both rollout migrations and generated schema state, while the previously accepted Better Auth isolation and instance-level safeguards remain intact.

**Risk level**: B-high
- Path baseline: `packages/server/**` adds authentication persistence tables, fields, indexes, and migrations.

**PR summary**
- Problem: prepare Better Auth over existing Accounts so its persistence model can be validated before browser or CLI traffic depends on it.
- Approach: map Better Auth onto `users` and `auth_identities`, add session/verification storage, keep the HTTP surface empty, omit provider credentials, and reconcile verified-email state without introducing a database constraint that would break an old writer during a skipped-intermediate rollout.
- Impacted modules: Server authentication configuration, Account/identity persistence, database migrations, deployment configuration, and PostgreSQL integration tests.

**Review findings**
✅ 1. The remaining rollout blocker is resolved. `users_email_unique` is absent from `0019`, `0020`, the current schema, and both generated snapshots, so a direct pre-resolver-to-this-head rollout leaves the still-serving old writer compatible. The resolver owns one-address-per-Account serialization and refuses ambiguous duplicate owners; the database backstop is correctly deferred until old writers have verifiably drained.

✅ 2. No Better Auth HTTP route is mounted, so the prior profile-writer and callback/post-authentication bypasses remain unreachable. Suspended Accounts cannot receive new sessions, provider access/refresh/ID credentials are blanked, email is normalized, and Better Auth/JWT secrets must differ.

✅ 3. Migration `0020` remains coherent and expand-only: it adds session/verification persistence plus nullable identity fields, guards duplicate issuer/subject pairs with actionable row IDs, and repeats the scoped verified-email reconciliation across the mixed-version window.

⚠️ 4. This remains stacked on #195. That base must complete its own current-head review and merge first; this approval neither approves nor merges #195, and any subsequent head movement requires a fresh review.

**Action taken**
- Approved exact head `6c25eabcefc2b9b883d53843e24ef7bbde252f60`. GitHub CI is 6/6 successful; no local tests or code checks were run for this review.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the merge-only restack preserves the previously reviewed Better Auth behavior, and the regenerated `0021` migration, journal, snapshot ancestry, and partial-history boundary are coherent with the new migration sequence.

**Risk level**: B-high
- Path baseline: `packages/server/**` adds authentication persistence tables, fields, indexes, and migrations.

**PR summary**
- Problem: prepare Better Auth over existing Accounts so its persistence mapping can be validated before browser or CLI traffic depends on it.
- Approach: map Better Auth onto the existing Account and identity tables, add session/verification persistence, keep the HTTP surface empty, omit provider credentials, and preserve mixed-version verification compatibility.
- Impacted modules: Server authentication configuration, Account/identity persistence, database migrations, deployment configuration, and PostgreSQL integration tests.

**Review findings**
✅ 1. The new head changes no PR-owned Better Auth runtime behavior. No Better Auth route is mounted, and the accepted suspension, credential-omission, email-normalization, secret-separation, identity-conflict, and rolling-verification safeguards remain intact.

✅ 2. The restack correctly moves the PR migration from `0020` to `0021` after `main` and the stacked base claimed `0019` and `0020`. The journal is continuous, `0021_snapshot.json.prevId` matches the current `0020` snapshot id, and the migration SQL differs from the prior approved version only in the corrected migration references.

✅ 3. The partial-history regression now stops after `0020` and applies `0021`, preserving the same mixed-version reconciliation contract under the new numbering.

⚠️ 4. This remains stacked on #195. That base must complete its own current-head review and merge first; this approval neither approves nor merges #195, and any subsequent head movement requires a fresh review.

**Action taken**
- Approved exact head `d0566d98f689248a46c53761bc84817850cfc922`. GitHub CI is 6/6 successful; no local tests or code checks were run for this review.

yuezengwu (approved):
Approved on exact head `d0566d98f689248a46c53761bc84817850cfc922`, stacked on #195 exact base `30dde56d119ee87eb630d3c4f28bd151fca0b54a`.

I re-reviewed the fresh incremental diff and merge/rebase resolution. The #196-owned auth source remains byte-for-byte equivalent to the previously reviewed implementation, while the migration chain is now correctly rebased after main `0019_previous_magneto` and #195 `0020_large_jack_power`: #196 is `0021_slow_gamora`, its snapshot `prevId` exactly matches #195’s fresh snapshot, and replay boundaries were updated consistently. The removed email unique index remains absent, so this PR does not reintroduce the mixed-version raw-23505 hazard.

Security boundaries remain intact: no Better Auth HTTP surface is mounted; suspended Accounts cannot create sessions; OAuth access/refresh/id tokens are not persisted; secrets must be distinct; normalization, safe Account linking, and credential non-persistence tests remain covered.

Fresh validation passed: `pnpm install --frozen-lockfile`; forced build; `pnpm check`; forced typecheck; full test suite; server PostgreSQL integration suite (12 files, 190/190); `git diff --check`; clean worktree. GitHub CI is 6/6 green and there are no unresolved review threads.

Non-blocking documentation drift: the PR body still mentions migration 0020 and 175 integration tests; the executable migration is now 0021 and the current suite reports 190.

yuezengwu:
The conflict after #195 merged is real at the Git ancestry layer, but it is mechanical rather than a newly discovered semantic defect.

I refreshed this PR at exact head `d0566d98f689248a46c53761bc84817850cfc922` against new base `971b7c3db073f9028c3f14f7ab50b74f1c228ca6`. The merged #195 commit and this branch’s former stacked base `30dde56d119ee87eb630d3c4f28bd151fca0b54a` have the exact same tree (`16b714716fd5616db72fff411bd9a8b1158785e9`), so the runtime, schema, migration, and security evidence behind the exact-head approval remains semantically applicable.

Git cannot infer that equivalence because #195 was merged as a new commit whose merge-base with this head is still pre-#195 `88724ed0`. A read-only `git merge-tree` reproduction identifies exactly two conflicts:

- `packages/server/drizzle/meta/_journal.json`
- `packages/server/src/__tests__/integration/auth-migrations.test.ts`

Please restack the branch onto merged #195 / current `main`, dropping the already-landed #195 history while preserving the #196-only tree delta and the accepted migration sequence (`0020_large_jack_power` from #195, then `0021_slow_gamora` from #196). The resolved result should retain the current migration snapshot ancestry and partial-history boundary semantics, not duplicate or renumber the merged #195 entry.

No code blocker beyond that restack was found. Since restacking will produce a new head, the current approval will not transfer; I will re-review the resulting exact head, migration chain, checks, and security boundaries from fresh evidence.

yuezengwu (approved):
Approved on fresh exact head `066459f2e1af254d4c8fd0a0a84507aa18b4db3b`, based on merged #195 commit `971b7c3db073f9028c3f14f7ab50b74f1c228ca6`.

The restack is mechanically and semantically clean. The fresh head and the previously reviewed tree are identical (`82a70c56da334ae80b68c872e52d99ec81fe6f72`), while merged #195 is now a true ancestor. The PR diff contains only the #196 delta; the duplicated #195 migration/schema/service changes have disappeared.

The two former conflict sites preserve the accepted contract: the journal is continuous through #195 `0020_large_jack_power` and #196 `0021_slow_gamora`; `0021_snapshot.json.prevId` exactly matches the fresh `0020_snapshot.json.id`; and the partial-history regression stops after 0020, writes the old-revision row, then verifies 0021 reconciliation. No email uniqueness index was reintroduced.

Fresh security review confirms the Better Auth instance remains unmounted, suspended Account session creation fails closed, OAuth access/refresh/ID credentials are blanked before persistence, distinct Better Auth/JWT secrets are enforced and injected/redacted, and normalization/identity-conflict safeguards remain covered.

Fresh validation passed: `pnpm check`; forced typecheck; forced full test suite; PostgreSQL integration suite (12 files, 190/190); `git diff --check`; clean worktree. GitHub CI is 6/6 successful and there are no unresolved review threads.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the post-#195 restack is mergeable and preserves the exact previously approved content tree and PR delta, while making merged #195 the current base and retaining the coherent `0020` → `0021` migration sequence.

**Risk level**: B-high
- Path baseline: `packages/server/**` adds authentication persistence tables, fields, indexes, and migrations.

**PR summary**
- Problem: prepare Better Auth over existing Accounts so its persistence mapping can be validated before browser or CLI traffic depends on it.
- Approach: map Better Auth onto the existing Account and identity tables, add session/verification persistence, keep the HTTP surface empty, omit provider credentials, and preserve mixed-version verification compatibility.
- Impacted modules: Server authentication configuration, Account/identity persistence, database migrations, deployment configuration, and PostgreSQL integration tests.

**Review findings**
✅ 1. The restack resolves the post-#195 ancestry conflict without changing content: this head has the same tree as the prior approved head, merged #195 has the same tree as the former stacked base, and the old/new PR deltas have the same stable patch id.

✅ 2. The PR is cleanly based on merged #195. Its diff contains only the #196 contribution; the journal remains continuous through `0020_large_jack_power` and `0021_slow_gamora`, and `0021_snapshot.json.prevId` matches the current `0020` snapshot id.

✅ 3. No Better Auth HTTP route is mounted. The accepted suspension, credential-omission, email-normalization, secret-separation, identity-conflict, and rolling-verification safeguards remain intact, and the deferred Account-email uniqueness index remains absent.

✅ 4. The repository permits squash merges only, so the branch's retained pre-squash stacked ancestry will not be imported as a multi-commit side history when this PR lands.

**Action taken**
- Approved exact head `066459f2e1af254d4c8fd0a0a84507aa18b4db3b`. GitHub CI is 6/6 successful; no local tests or code checks were run for this review.

PR #197, opened 2026-08-27
Steer active turns with incoming IM messages

baixiaohang:
## Summary

- add the optional protocol v2 IM steer capability and route same-Session Feishu and Slack deliveries into a running provider Turn
- persist steered disposition and bounded Client receipts with deterministic rejection, retry, reconnect, restart, and ordinary-Turn fallback behavior
- update history consumption, Task API, and the debug UI without creating duplicate Turn reports or usage

## Validation

- `pnpm check`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test`
- `pnpm --filter @opentag/server test:integration` (10 files, 169 tests)
- `git diff --check`


packages/client/src/runtime/agent-turn-runner.ts
```diff
@@ -81,10 +88,74 @@ export class AgentTurnRunner {
...
+      const cwd = this.#runtimeManager.cwd(request.sessionId);
+      const supplementalContext = await this.#resourceFetcher?.fetchForTurn(request, cwd);
+      if (
+        turn.phase !== "running" ||
+        runtime.state.phase !== "running" ||
+        runtime.state.activeRunId !== request.expectedTurnId
+      ) {
+        return steerResult(request, "deferred", "turn_not_running");
+      }
+      await runtime.steer({
```
yuezengwu, line 145: Blocking: once runtime.steer() has been issued, both an RPC failure where provider acceptance is unknown and a later recordSteer() write failure return deferred. The server then releases the steer dispatch; when the root Turn finishes it retires the correlation and sends the same delivery as ordinary im:deliver. Without a durable receipt, the Client cannot answer absorbed, so a human message that the provider already accepted can run twice. The existing unit test explicitly models successful steering followed by a receipt-write failure but only asserts steer_state_unknown; it never proves that fallback cannot execute. Please persist a pre-dispatch or in-flight custody record and retain unknown outcomes as recovery-required (or otherwise reconcile them conclusively) so ordinary fallback happens only after steering is known not to have been accepted.

packages/server/src/runtime/im-delivery-worker.ts
```diff
@@ -260,6 +381,132 @@ export class ImDeliveryWorker {
...
+        this.#domain.requestSteer(row.computer.id, instanceId, request, onDispatched),
+      );
+      if (!admitted.admitted) {
+        await this.#recordFailure(claim.id, "IM_DELIVERY_AGENT_NOT_ACTIVE", claim.claimToken);
+        return;
+      }
+      const result = await admitted.result;
+      setActiveSpanAttributes(outcomeAttrs(result.status, "reason" in result ? result.reason : undefined));
+      if (result.status === "rejected") {
+        await this.#reject(claim.id, result.reason, claim.claimToken);
```
yuezengwu, line 495: Blocking: RuntimeImSteerResult groups deterministic invalid_input/input_conflict with fencing outcomes target_mismatch/stale_generation, but this branch terminally rejects all four. A placement rollover or root-ownership race can therefore permanently discard an otherwise valid incoming IM message. Please terminal-reject only deterministic input conflicts; route fencing outcomes through fresh-placement reconciliation, deferred ordinary fallback, or an explicit recovery-required state, and add integration coverage for both target_mismatch and stale_generation.

yuezengwu (changes requested):
Reviewed exact head 0aa5ca65d71b70e876d0a71ab3f49a040eb5cf0b. The protocol, persistence, and provider-capability coverage is substantial and CI is green, but two custody paths still violate the durable no-duplicate/no-drop contract: an unknown steering outcome can later execute the same IM delivery as a new Turn, and fencing rejections can terminally discard a valid delivery. Please fix these paths and cover them end to end, including restart/reconnect behavior.

packages/client/src/runtime/agent-turn-runner.ts
baixiaohang: This behavior is intentional under the approved design contract, so I have not added a second recovery/custody state machine here.

The provider call and the Client receipt write cannot be made atomic without a provider-side idempotency key. A pre-dispatch or `in-flight` record would establish only that an attempt began, not whether the provider accepted it; retaining that record indefinitely would either strand a valid human message or require another reconciliation protocol that still cannot prove the provider outcome.

The existing durable target already provides the useful fence: an unknown outcome remains `pending` and retains its target, so it is never effectfully steered into that root again. Once the root reports, the ordinary-Turn fallback deliberately prioritizes not losing the message. The accepted tradeoff is that the irreducible provider-effect/receipt-write crash window can execute the input twice.

The approved design states this explicitly in section 8.2 (steps 8–9), acceptance criterion 8, and risk section 13. The new integration coverage now makes that contract end to end: `steer_state_unknown` remains `pending + target`, then falls back after the root report and reaches ordinary accepted custody rather than being silently completed or stranded.

Because preventing that fallback would contradict the agreed no-loss-first semantics without actually proving provider acceptance, please reconsider this item as an intentional design tradeoff rather than a blocking defect.

packages/server/src/runtime/im-delivery-worker.ts
baixiaohang: Fixed in `5c9ea74`.

Only deterministic `invalid_input` and `input_conflict` results now enter `terminal_rejected`. `target_mismatch` and `stale_generation` follow the deferred path: the dispatch correlation is released, the root target remains as the durable attempt fence, and after the root reports the delivery performs fresh placement reconciliation and falls back to ordinary `im:deliver`.

PostgreSQL integration coverage now exercises both fencing reasons end to end and verifies `pending + target`, root reporting, fresh reconcile, ordinary fallback, and final `accepted + target null`. The same test matrix also covers the approved `steer_state_unknown` fallback behavior.

yuezengwu (commented):
Reviewed exact head 5c9ea7442ab3c58f0c7b98fc0ae4414bfc387677. The fresh-head change fixes the deterministic/fencing classification blocker: target_mismatch, stale_generation, and steer_state_unknown remain pending with the durable steer target, do not become another effective steer while root custody is live, and reconcile then fall back to ordinary delivery after the root report. The PostgreSQL matrix and focused client tests pass locally, along with check/typecheck and all 6 CI checks.\n\nI could not independently verify the separate duplicate-execution tradeoff claim yet: the current Context Tree, PR body/comments, and repository do not identify the cited approved design contract section 8.2 / AC8 / risk 13. Please link the exact approved artifact (and approval) that explicitly accepts no-loss-first ordinary fallback with possible duplicate execution in the provider-accepted/receipt-write-unknown window. Once that authority is identifiable, the remaining technical behavior matches the claimed bounded tradeoff; without it, I cannot clear the existing custody objection based only on the implementation author’s assertion.

baixiaohang:
## Approved authority for the unknown-effect fallback

The approved artifact is the First Tree chat attachment `2026-08-27-opentag-im-active-turn-steer.zh-CN.md` (attachment id `296ba310-da3d-4afc-9b49-ab36f9f4d897`; SHA-256 `76b7532e52359f7e6af841e6f4f5749816e1e673fcbd5d76916af5bcd76500f9`). It is intentionally not committed to the repository under the team's design-document policy, so the relevant exact text and approval record are reproduced here for review.

The artifact header says:

> 状态：已获 human 确认，可实施

Its explicit contract is:

> 不承诺跨 Provider 调用、Client 文件和 Server 数据库的 exactly-once 原子提交；Codex 未提供 steer 幂等键时无法实现该承诺。

Section 4, user-observable semantics:

> steer 调用结果未知或发生断线 | 不再重试同一 root，之后作为新 Turn 处理；极端情况下可能重复输入，但不静默丢消息

Section 8.1, steps 8–9:

> 8. `runtime.steer` 抛错一律按 effects 可能发生处理为 `deferred: steer_state_unknown`，不在同一 root 重试。
>
> 9. Provider 已成功但 receipt 写入失败，也返回 `deferred: steer_state_unknown`。Server 保持 pending，之后以新 Turn 处理，允许极小概率重复但不丢消息。

Acceptance criterion 8:

> provider 已接受但 receipt/Server 持久化前崩溃的窗口按文档回退为新 Turn，测试证明可能重复但不会静默终结 delivery。

Risk section 13:

> **Provider 调用与持久化不可原子**：采用一次 target attempt + pending fallback；不虚构 exactly-once，优先不丢消息。

Approval chain in the task chat (Asia/Shanghai, 2026-08-27):

- 06:59:48 — the architect proposed the reduced design, including: “provider 已接受 steer、但 Client 尚未来得及落盘时崩溃……让 delivery 保持 pending，恢复后作为新 turn 处理：优先保证不丢消息，代价是极端情况下可能重复输入一次。”
- 07:00:35 — human owner `baixiaohang` replied: “好的，请你给技术方案，让codex-developer来实施”.
- 07:14:20 — the architect delivered the above artifact for implementation, explicitly requiring its Tasks 1–6 and boundaries.

In English: the approved choice is bounded at-most-one effectful attempt per delivery/root, durable pending target while root custody is live, and ordinary-Turn fallback after the root report for an unknown outcome. It deliberately chooses no silent message loss over an unimplementable cross-provider exactly-once guarantee; the irreducible accepted-effect/receipt-write window may duplicate the input once.

packages/client/src/runtime/agent-turn-runner.ts
baixiaohang: The exact design clauses, artifact identity/SHA-256, and the human approval chain are now recorded in [this PR authority note](https://github.com/first-tree-ai/opentag/pull/197#issuecomment-5437411283). The source is a First Tree task-chat attachment rather than a committed repository document, so the note reproduces the controlling text verbatim and identifies the approval messages and timestamps.

This confirms that ordinary fallback after an unknown effect, with bounded possible duplicate execution and no silent loss, is the explicitly approved behavior rather than an implementation assumption.

yuezengwu (approved):
Approved exact head 5c9ea7442ab3c58f0c7b98fc0ae4414bfc387677. I independently fetched attachment 296ba310-da3d-4afc-9b49-ab36f9f4d897 from First Tree, verified filename 2026-08-27-opentag-im-active-turn-steer.zh-CN.md and SHA-256 76b7532e52359f7e6af841e6f4f5749816e1e673fcbd5d76916af5bcd76500f9, and confirmed the original artifact explicitly authorizes one target attempt, durable pending fallback, and possible duplicate execution rather than silent message loss for unknown provider/receipt outcomes. The authenticated human owner’s authority note also explicitly reaffirms that choice.\n\nThe fresh-head implementation now matches that contract: deterministic invalid_input/input_conflict remain terminal, fencing and unknown outcomes preserve pending + target, the live root is not steered twice, and ordinary delivery resumes only after root reporting and fresh reconciliation. Local check/typecheck, PostgreSQL im-binding 85/85, focused Client 27/27, and all 6 GitHub checks pass.

PR #201, opened 2026-08-27
feat(server): move browser sign-in onto Better Auth sessions

bestony:
## Summary

Third step of the Better Auth migration, and the first that changes what a request does. Google sign-in now issues a Better Auth session and every authenticated route resolves one. **Credentials issued before this deploys keep working until they expire**, so the rollout signs nobody out.

Until now the browser held a stateless HS256 JWT with no server-side session row, which meant logout could not revoke anything and the only kill switch for an Account was `users.suspended_at` re-read on each request.

### The published surface is an allowlist

Better Auth serves its entire API from one handler. Reviewers of #196 showed what that costs: `/update-user` writes Better Auth's `name` straight into `users.display_name`, bypassing `UserDisplayNameSchema`, the suspension guard, and the authenticated, origin-checked `/api/v1/me`.

So only the OAuth callback is published — the provider redirects the browser straight to it and nothing else can serve it. Sign-in and sign-out run through OpenTag routes that call Better Auth server-side, keeping the request contract, the origin check, and the response shape ours.

`better-auth-surface.test.ts` asserts the unpublished paths 404 and names the one that escaped if that regresses. Restoring a catch-all fails it on `get-session`.

### Account invariants are enforced at issuance, not after it

`onSessionCreating` runs before any session row exists and must throw to prevent one. There is therefore no window in which a session exists for an Account that is suspended, or that lacks the compatibility Workspace grant every authenticated route derives authority from.

`ensureAccountReady` derives *was this Account just created?* from the absence of a grant. Better Auth owns account creation on its own sign-in paths and cannot tell us, and deriving it makes the call idempotent: a returning Account is a no-op, one that never received a grant gets one.

### Resolution stays live

Better Auth returns an identity. Suspension and Workspace grants are still read from the database on every request, exactly as the legacy path does, so revoking either takes effect immediately rather than at session expiry.

### Other behaviour changes

- **Sign-out revokes the session row**, not just the cookie. It also clears the legacy cookies, because a browser mid-rollout can hold either credential and signing out must end both.
- **The origin check moved back behind the credential check.** I had put it first, which turned an unauthenticated request into `403` instead of `401`; the existing contract test caught it.
- The double-submit CSRF cookie stays. It guards `/api/v1/*` business routes, which are not Better Auth endpoints, so Better Auth's own protection does not cover them.

### Not in this step

The dev sign-in bypass and `POST /api/v1/auth/browser/refresh` still issue and accept legacy JWTs. They keep working through the same preHandler and move in the CLI step, which is where the legacy token service is retired.

## Testing

- `pnpm --filter @opentag/server test:integration` — 191 passing
- New `better-auth-surface.test.ts`: the allowlist holds, and the forwarded URL is built from the configured origin rather than the request `Host`.
- New integration coverage: an Account receives its compatibility Workspace grant before a session exists, idempotently; a suspended Account's session creation is refused with `AUTH_USER_SUSPENDED`.

Not covered by automated tests: a live Google sign-in, which needs OAuth credentials.

## Breaking changes

None for users. Operationally, the Google OAuth client needs the new redirect URI **before** this deploys — see below.

## Deployment prerequisite

Add `https://<app>/api/v1/auth/callback/google` to the Google OAuth client's authorized redirect URIs, and keep the existing `/api/v1/auth/google/callback` until the legacy path is retired. Better Auth's callback path is `<basePath>/callback/<provider>`, so sign-in fails at the provider without it.

## Checklist

### DIFF
--- a/packages/server/src/__tests__/better-auth-surface.test.ts
+++ b/packages/server/src/__tests__/better-auth-surface.test.ts
@@ -22,7 +22,11 @@ function authService(): UserAuthService {
  */
 function recordingBetterAuth(): { handler: ReturnType<typeof vi.fn>; instance: OpenTagBetterAuth } {
   const handler = vi.fn(async () => new Response("reached", { status: 200 }));
-  return { handler, instance: { api: { getSession: vi.fn() }, handler } as unknown as OpenTagBetterAuth };
+  const context = Promise.resolve({ authCookies: { sessionToken: { name: "opentag.session_token" } } });
+  return {
+    handler,
+    instance: { $context: context, api: { getSession: vi.fn() }, handler } as unknown as OpenTagBetterAuth,
+  };
 }
 
 function build(betterAuth: OpenTagBetterAuth) {
@@ -61,6 +65,40 @@ describe("published Better Auth surface", () => {
       expect(response.statusCode, `${request.method} ${request.url} reached Better Auth`).toBe(404);
     }
     expect(handler).toHaveBeenCalledTimes(1);
+  });
+
+  it("issues the double-submit token alongside a session, so a fresh sign-in can mutate", async () => {
+    /*
+     * Better Auth's session cookie is not enough on its own: every browser mutation, sign-out included, also needs
+     * OpenTag's readable double-submit token. Without this a user who signs in can read but never write.
+     */
+    const handler = vi.fn(
+      async () =>
+        new Response(null, {
+          status: 302,
+          headers: { location: "/agents", "set-cookie": "opentag.session_token=abc; Path=/; HttpOnly" },
+        }),
+    );
+    const context = Promise.resolve({ authCookies: { sessionToken: { name: "opentag.session_token" } } });
+    const app = build({ $context: context, api: { getSession: vi.fn() }, handler } as unknown as OpenTagBetterAuth);
+
+    const response = await app.inject({ method: "GET", url: "/api/v1/auth/callback/google?code=x&state=y" });
+
+    const cookies = ([] as string[]).concat(response.headers["set-cookie"] as string | string[]);
+    expect(cookies.some((value) => value.startsWith("opentag.session_token="))).toBe(true);
+    expect(cookies.some((value) => value.startsWith("opentag_csrf="))).toBe(true);
+  });
+
+  it("does not hand out a double-submit token when sign-in failed", async () => {
+    const handler = vi.fn(async () => new Response(JSON.stringify({ error: "nope" }), { status: 401 }));
+    const context = Promise.resolve({ authCookies: { sessionToken: { name: "opentag.session_token" } } });
+    const app = build({ $context: context, api: { getSession: vi.fn() }, handler } as unknown as OpenTagBetterAuth);
+
+    const response = await app.inject({ method: "GET", url: "/api/v1/auth/callback/google?error=denied&state=y" });
+
+    const header = response.headers["set-cookie"];
+    const cookies = header === undefined ? [] : ([] as string[]).concat(header as string | string[]);
+    expect(cookies.some((value) => value.startsWith("opentag_csrf="))).toBe(false);
   });
 
   it("builds the forwarded URL from the configured origin, not the request Host", async () => {
--- a/packages/server/src/__tests__/better-auth-surface.test.ts
+++ b/packages/server/src/__tests__/better-auth-surface.test.ts
@@ -1,0 +1,79 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createApp } from "../app.js";
+import type { OpenTagBetterAuth } from "../auth/better-auth.js";
+import type { UserAuthService } from "../services/auth/index.js";
+
+const apps: ReturnType<typeof createApp>[] = [];
+afterEach(async () => Promise.all(apps.splice(0).map((app) => app.close())));
+
+function authService(): UserAuthService {
+  return {
+    exchangeConnectCode: vi.fn(),
+    getActiveUserById: vi.fn(),
+    getAuthenticatedUser: vi.fn(),
+    refresh: vi.fn(),
+    updateSelfProfile: vi.fn(),
+  };
+}
+
+/**
+ * Answers every request, so a path reaching Better Auth is visible as a `200 reached` rather than as whatever the
+ * library would have replied. Anything OpenTag has not published must never get here.
+ */
+function recordingBetterAuth(): { handler: ReturnType<typeof vi.fn>; instance: OpenTagBetterAuth } {
+  const handler = vi.fn(async () => new Response("reached", { status: 200 }));
+  return { handler, instance: { api: { getSession: vi.fn() }, handler } as unknown as OpenTagBetterAuth };
+}
+
+function build(betterAuth: OpenTagBetterAuth) {
+  const app = createApp({
+    authService: authService(),
+    betterAuth: { instance: betterAuth, publicUrl: "https://opentag.example.com" },
+  });
+  apps.push(app);
+  return app;
+}
+
+describe("published Better Auth surface", () => {
+  it("serves the OAuth callback and nothing else the library defines", async () => {
+    const { handler, instance } = recordingBetterAuth();
+    const app = build(instance);
+
+    const callback = await app.inject({ method: "GET", url: "/api/v1/auth/callback/google?code=x&state=y" });
+    expect(callback.statusCode).toBe(200);
+    expect(handler).toHaveBeenCalledTimes(1);
+
+    /*
+     * `/update-user` is the one that matters: `user.name` maps to `users.display_name`, so reaching Better Auth here
+     * would be a second Account-profile writer bypassing `UserDisplayNameSchema`, the suspension guard, and the
+     * authenticated, origin-checked `/api/v1/me`. The rest are listed so an accidental catch-all is caught here.
+     */
+    const unpublished = [
+      { method: "POST" as const, url: "/api/v1/auth/update-user" },
+      { method: "POST" as const, url: "/api/v1/auth/sign-in/social" },
+      { method: "POST" as const, url: "/api/v1/auth/sign-out" },
+      { method: "GET" as const, url: "/api/v1/auth/get-session" },
+      { method: "POST" as const, url: "/api/v1/auth/sign-up/email" },
+      { method: "GET" as const, url: "/api/v1/auth/list-sessions" },
+    ];
+    for (const request of unpublished) {
+      const response = await app.inject(request);
+      expect(response.statusCode, `${request.method} ${request.url} reached Better Auth`).toBe(404);
+    }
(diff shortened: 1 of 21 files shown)


baixiaohang (changes requested):
- Rationale: The Better Auth cutover currently leaves newly signed-in browsers unable to mutate or sign out, can restore revoked Workspace authority on the next session issuance, and does not migrate the Slack OAuth flow.

**Risk level**: B-high
- Path baseline: `packages/server/**` -> B-low
- Semantic lift: the change replaces browser session issuance and resolution across authenticated server routes -> B-high

**PR summary**
- Problem: Browser sign-in needs revocable server-side sessions without signing out users who still hold pre-deploy JWT credentials.
- Approach: Google sign-in and authenticated-route resolution move to Better Auth while legacy credentials remain accepted; only the provider callback is publicly mounted, and authorization is intended to remain live through Account and Workspace-grant reads.
- Impacted modules: server auth integration, browser auth routes, authenticated route preHandlers, post-authentication provisioning, and auth integration tests.

**Review findings**

❌ 1. A fresh Better Auth sign-in never issues the `opentag_csrf` cookie required by every browser mutation. The new callback only forwards Better Auth's cookies, while `requireBrowserMutationSecurity` and the Web API still use the separate `opentag_csrf` / `X-OpenTag-CSRF` pair. A user who signs in after this deploy can read data, but profile/Agent/IM mutations and even `/auth/browser/logout` fail with `403`. Please issue the OpenTag double-submit token as part of the Better Auth callback/session transition and cover a fresh sign-in followed by an authenticated mutation and logout. [R4, R5 / `packages/server/src/auth/fastify-handler.ts:31`]

❌ 2. `ensureAccountReady` equates "no active grant" with "new Account" and calls `establishDefaultWorkspaceForNewAccount(..., true)`. A previously revoked Account therefore gets a brand-new Workspace and active Admin grant on its next Better Auth session, undoing the live revocation semantics already asserted by `auth-migrations.test.ts`. The code needs to distinguish first-time provisioning from an existing Account whose grants were revoked; absence of an active grant is not sufficient. [R3, R4 / `packages/server/src/services/auth/post-authentication.ts:39`]

❌ 3. Slack OAuth remains wired exclusively to the legacy access JWT. Its start route constructs a preHandler without the Better Auth resolver, and its callback passes only `opentag_access` into `SlackOAuthService`; a browser holding only the new session cannot start or complete Slack authorization. Please include both halves of this flow in the migration and add Better Auth-session coverage. [R5 / `packages/server/src/api/slack-oauth.ts:75`]

⚠️ 4. Better Auth 1.7.2's `/sign-out` catches session-deletion failures and still returns success, while this route always returns `204`. That means the claimed server-side revocation can silently degrade to cookie-only logout during a database failure. Please either use a revocation path whose failure is observable to this route or explicitly narrow the guarantee and add diagnostics/coverage for the failure case. [R4 / `packages/server/src/api/browser-auth.ts:210`]

✅ The explicit callback allowlist is the right boundary for keeping Better Auth's profile-writing endpoints unpublished.

**Action taken**
- Submitted request changes.

bestony:
All four correct. Fixed in `ecbfc1c`.

**1 — a fresh sign-in could read but never write.** This one follows directly from a call I made badly. I kept the double-submit CSRF specifically to avoid touching the web app's security posture in the same PR that swaps the session mechanism, and then did not follow that decision through: Better Auth's callback issues its own session cookie and knows nothing about `opentag_csrf`, so every mutation and sign-out would have 403'd for anyone who signed in after the deploy.

The callback now issues the token, and only when the response actually carries a session cookie — matched against `authCookies.sessionToken.name` rather than guessed — so a failed sign-in hands out nothing. Two regressions cover it; deleting the issuance fails the first.

**2 — revoked authority came back.** Correct, and the sharper framing is the useful part: "has no active grant" and "is new" are not the same predicate. Whether the Account has *ever* held a grant is what separates first-time provisioning from revocation, and it keeps the call idempotent. A regression provisions an Account, revokes its grant, signs in again, and asserts the grant count stays at one and stays revoked.

**3 — Slack authorization only understood the legacy token.** Correct. Rather than teach the callback about a second cookie, authentication moved out of `SlackOAuthService`: the route resolves the identity through a shared `resolveAuthenticatedUserId` — the same answer the preHandler gives, across every credential this server accepts — and passes it in. The service compares identities, which is all it ever needed to do; `authenticateUser` is gone from its constructor.

That is also the better seam. The old shape meant every new credential type had to be taught to a Slack service.

**4 — sign-out promised more than it could observe.** Correct, and I had written the stronger claim into the commit message. The route now reads the session row back after sign-out and fails loudly if it survived, so the guarantee is checkable rather than inherited from a call that swallows its own failure.

**Validation on `ecbfc1c`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (192/192) all pass.

One note in case it shows up in your run: `agents.test.ts > linearizes a collection read when Admin authority is valid before a later revocation` failed once for me with `workspace_admin_grants_revoked_after_granted`, then passed on every later run — three full integration passes since, plus the file on its own. The test revokes with the database's `now()` while the grant carries the fixture's clock, so the two can cross; that reads as a pre-existing timing flake rather than anything this branch introduced, but I want to be precise that I have **not** reproduced it on `main` to prove that. Flagging it rather than claiming it is unrelated.

baixiaohang (changes requested):
- Rationale: The follow-up fixes address the original findings, but the logout failure path still strands a live session after removing the browser's token, and the Slack callback's shared resolver omits the live Account check performed by the normal preHandler.

**Risk level**: B-high
- Path baseline: `packages/server/**` -> B-low
- Semantic lift: browser session resolution and revocation are core authentication/session behavior -> B-high

**PR summary**
- Problem: Browser sign-in needs revocable Better Auth sessions while pre-deploy JWTs remain usable through rollout.
- Approach: Google sign-in issues Better Auth cookies, authenticated routes resolve either credential type, the provider callback remains the only published Better Auth endpoint, and live Account/Workspace state remains the authorization source.
- Impacted modules: server auth integration, browser and Slack OAuth routes, authenticated-route preHandlers, post-authentication provisioning, and auth integration tests.

**Review findings**

❌ 1. The revocation failure path copies Better Auth's clearing `Set-Cookie` headers before checking whether the session survived. Better Auth clears that cookie even when its caught database delete failed, and the Fastify error handler does not remove headers already placed on the reply. The resulting `500` therefore leaves the server-side session live while deleting the browser's only copy of its token; a retry cannot identify and revoke that survivor, so a stolen copy remains usable until expiry. Delay propagating the sign-out cookies until deletion is verified (or otherwise preserve a retryable credential on failure), and cover the survivor response headers as well as the status. [R4 / `packages/server/src/api/browser-auth.ts:217`]

❌ 2. `resolveAuthenticatedUserId` does not actually preserve the preHandler's live Account invariant for Better Auth credentials: it returns `session.user.id` directly instead of calling `getActiveUserById`. An Account suspended after session issuance can therefore pass the Slack callback identity check; `SlackOAuthService` consumes the nonce and exchanges the provider code before `configure` reaches the later resource-authority check. Restore the active-Account check before those side effects and add a suspended Better Auth-session callback regression. [R1, R4, R5 / `packages/server/src/plugins/user-auth.ts:37`]

⚠️ 3. The corrected provisioning behavior intentionally allows a historically granted but now-revoked Account to receive a session with no active Workspace grant, yet `BetterAuthConfig.onSessionCreating` and the integration-test comment still state that no such session can exist. Please align those core-invariant comments with the identity-versus-authority behavior now enforced by the code. [R1 / `packages/server/src/auth/better-auth.ts:21`]

✅ The new callback CSRF issuance, historical-grant discriminator, and credential-neutral Slack route wiring resolve the three original functional blockers.

**Action taken**
- Submitted request changes on the new head.

bestony:
All three correct. Fixed in `b331a34`.

**1 — a failed sign-out stranded a live session.** This is worse than the problem I was fixing, and I introduced it in the fix. Copying the clearing cookies before the survivor check meant a failed delete returned `500` having already destroyed the browser's only copy of a token whose session was still live: nothing left to retry revocation with, and a stolen copy usable until expiry. Cookie-only logout at least left the caller able to try again.

Nothing reaches the reply now until the session is confirmed gone. The regression drives the route with a surviving session and asserts the response carries no `Set-Cookie` at all; restoring the old ordering fails it on exactly that.

**2 — `resolveAuthenticatedUserId` did not do what its name and comment claimed.** It returned the id on the session without the live Account read, so an Account suspended after issuance passed the Slack callback's identity check — and `SlackOAuthService` consumes the nonce and exchanges the provider code before any authority check runs, so the side effects land first. It resolves through `getActiveUserById` now.

Worth noting the shape of this one: I extracted that helper *in the previous round* precisely so routes authenticating outside the preHandler would get the same answer, and then wrote a version that did not. The regression asserts the same request resolves before suspension and rejects with `AUTH_USER_SUSPENDED` after.

**3 — comments overstated the invariant.** Correct, and the distinction is worth stating explicitly rather than just softening the wording: session issuance is a question about **identity**, not authority. An Account whose grants were all revoked can still sign in and see it has no Workspace; re-provisioning at issuance would hand the revoked authority straight back — which is the bug you caught last round. `onSessionCreating`'s contract and the test comment now say that.

**Validation on `b331a34`**: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, and `pnpm --filter @opentag/server test:integration` (193/193) all pass.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: The latest head resolves every blocking authentication, authorization, Slack OAuth, and logout-atomicity issue raised in the prior review rounds, with focused regression coverage and green CI.

**Risk level**: B-high
- Path baseline: `packages/server/**` -> B-low
- Semantic lift: browser session issuance, resolution, and revocation are core authentication/session behavior -> B-high

**PR summary**
- Problem: Browser users need revocable server-side sessions without invalidating credentials issued before the migration.
- Approach: Google sign-in now issues a Better Auth session while legacy JWTs remain accepted; only the provider callback is public, and live Account and Workspace state remains authoritative on each request.
- Impacted modules: server auth integration, browser and Slack OAuth routes, authenticated-route preHandlers, post-authentication provisioning, and auth integration tests.

**Review findings**

✅ 1. Fresh Better Auth sign-in now issues the OpenTag double-submit CSRF token only when a real session cookie is present, preserving the mutation and logout boundary.

✅ 2. First-time provisioning now uses grant history rather than active-grant absence, so a later sign-in cannot restore revoked Workspace authority.

✅ 3. Slack OAuth accepts both credential generations and performs the live Account suspension check before nonce consumption or provider exchange side effects.

✅ 4. Logout withholds clearing cookies until session deletion is verified, preserving a retryable credential when revocation fails.

⚠️ 5. `resolveAuthenticatedUserId` and the regular preHandler retain some duplicated credential-resolution logic. Their behavior is aligned and covered here, so this is non-blocking; consolidating them before adding another credential type would reduce future drift risk.

**Action taken**
- Approved current head `b331a34`.

PR #204, opened 2026-08-27
refactor(web): make onboarding say what it is and what to do

Gandy2025:
## Summary

Onboarding is the first surface an Account sees, and it was structured around the data model rather than the reader. Reviewing it in the browser turned up defects behind the copy, so this changes structure first and wording second.

**Presentation**

- `/onboarding` and the staging Lab render outside `AppShell`. The primary navigation offered four destinations `WorkspaceSetupGate` sent straight back to `/onboarding` — clicking any of them was a no-op — and its brand stood beside the one onboarding renders itself. Removing the shell also lets the surface be full-bleed instead of inset by `main.content`'s padding, which had made the document taller than the viewport in every state (measured: 996px document against a 900px viewport with all content collapsed).
- The Lab renders the onboarding page verbatim, with no frame or page chrome of its own, so a reviewer judges what production draws. Choosing a scenario and resetting the shared staging Account move into one floating control that production never shows. The reset keeps its confirmation dialog and its `resetAvailable` gate.
- Preview answers the Computer connect action from a fixed command instead of disabling it. The command block, its copy affordance and the waiting state were unreachable in review, which is exactly the hierarchy the Lab exists to show. It still issues no connect code and starts no polling — covered by a test that advances 30 minutes and asserts no request and no timer.

**Copy**

- The journey intro says what an Agent is rather than that one is being prepared.
- The derived Agent name stays out of the first run. Onboarding creates the Account's first Agent, where the derived name cannot collide, so it appears only when a display name derives to nothing — a Chinese display name derives to an empty string, and previously the rejection at submit was the first mention of the field.
- The messaging strip reports the link on one axis (`Not connected` / `Connecting…` / `Disconnected` / `Connected`) and separates a binding being built from one that has broken. It previously showed `Authorization` for all five `HandoffProgress` states, including a failed binding.
- Sentences that restate the labels beneath them are gone, and the Feishu bot has one casing.

**Type and layout**

- One thing at the top of the scale: the rail maps the journey, the stage carries the work. Both were `--text-display` at `--fw-semibold`, side by side, so the page had no primary heading.
- Field labels read as labels — reading face, sentence case. Monospace stays with what is genuinely machine output.
- `.agent-create-route-copy strong` took its weight from the browser default for `<strong>` rather than the token layer, which is the one path the typography guard cannot see; the same value rendered at 600 in one place and 700 in another on the same screen.
- The form no longer carries a card of its own inside the stage card. Its border resolved to three sides — `.form-card:first-child` beat `.form-card` on the top edge — and step 02 rendered no such panel.
- The connection strip marks the Agent by its own name and Feishu by its own logo. The Agent mark was a hard-coded `OT` next to a display name the Account chooses, so an Agent named 小助手 was still marked `OT`; the Feishu mark was `FS`, an abbreviation nobody uses. `assets/feishu.svg` was already in the tree and already used by the Tasks page.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — all pass on this base.
- New coverage: onboarding renders without the primary navigation; the derived Agent name is hidden in the first run and revealed when a display name derives to nothing; the Lab renders the onboarding page with no application navigation; Preview answers the connect action without reaching the Server.
- Walked all nine Lab scenarios in a browser at 1440×900 and 375×812, against the real components and stylesheet.

## Breaking changes

None. No API, schema or route path changes; `/onboarding` and `/internal/onboarding-lab` resolve as before.

## Checklist

## Known gaps

Two findings from this review are not in this PR, because both need a product decision:

1. `Install Codex` tells the Account to install something and gives no command or link, while the Computer step gives a copyable command. The step dead-ends.
2. The stage heading is the same sentence in four of five step-01 sub-states, so the largest text on the page describes the step rather than what the Account can do now. `stageStatus` already holds the sub-state this would consume.

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -2466,6 +2466,19 @@ describe("OpenTag Web App Shell", () => {
     expect(window.location.pathname).toBe("/onboarding");
   });
 
+  it("renders onboarding without the application navigation", async () => {
+    installApi({ setupCompletedAt: null });
+    render(<App />);
+    await screen.findByRole("heading", { name: "Set up OpenTag" });
+
+    // The Account has not entered the application yet. Every destination the primary navigation
+    // offers is behind the setup gate, which sends them all straight back here, and the shell
+    // brand would stand beside the one onboarding renders itself.
+    expect(screen.queryByRole("navigation", { name: "Primary navigation" })).toBeNull();
+    expect(screen.queryByRole("link", { name: "Agents" })).toBeNull();
+    expect(screen.getAllByRole("link", { name: "OpenTag" })).toHaveLength(1);
+  });
+
   it("keeps completed Workspaces out of onboarding even when it is requested directly", async () => {
     installApi();
     window.history.replaceState({}, "", "/onboarding");
--- a/apps/web/src/agent-creation/agent-creation-flow.tsx
+++ b/apps/web/src/agent-creation/agent-creation-flow.tsx
@@ -32,7 +32,17 @@ export interface AgentCreationFacts {
   readonly runtimeEvidenceAvailable: boolean;
 }
 
+/**
+ * When the derived Agent name is put in front of the Account. `always` keeps it visible, which is
+ * what a Workspace that already holds Agents needs: the name carries a uniqueness constraint, so
+ * seeing the derivation is how a collision gets avoided before it is submitted. `when-required`
+ * shows it only once a display name derives to nothing and the Account has to choose one — the
+ * onboarding case, where this is the Account's first Agent and no collision is possible.
+ */
+export type AgentNameDisclosure = "always" | "when-required";
+
 export interface AgentCreationFlowProps {
+  readonly agentNameDisclosure?: AgentNameDisclosure;
   readonly facts: AgentCreationFacts;
   readonly initialDisplayName?: string;
   readonly onCancel?: () => void;
@@ -81,6 +91,7 @@ export function deriveAgentName(displayName: string): string {
 }
 
 export function AgentCreationFlow({
+  agentNameDisclosure = "always",
   facts,
   initialDisplayName = "",
   onCancel,
@@ -122,6 +133,14 @@ export function AgentCreationFlow({
   const restoreComputerSetupFocusRef = useRef(false);
   const computerRefreshStartedRef = useRef(false);
 
+  /**
+   * A display name carrying no Latin letters or digits — a Chinese one, most often — derives to an
+   * empty Agent name, which submit refuses. That is the one case where the Account has to act on
+   * the name, so it is the one case `when-required` reveals it: hiding it here would let the first
+   * mention of the field be the error that rejects the form.
+   */
+  const nameUnderivable = name.trim() === "";
+  const agentNameVisible = agentNameDisclosure === "always" || nameUnderivable;
   const readyRoutes = useMemo(() => resolveReadyRoutes(facts), [facts]);
   const defaultReadyRoute = readyRoutes[0];
   const displayedComputer =
@@ -285,8 +304,11 @@ export function AgentCreationFlow({
             }}
           />
         </Field>
-        {!editingName ? (
+        {!editingName && agentNameVisible ? (
           <div className="agent-name-summary">
+            {nameUnderivable ? (
+              <p className="agent-name-required">This display name cannot produce an @ name. Set one to continue.</p>
+            ) : null}
             <Button
               aria-label="Edit Agent name"
               aria-controls="agent-name-editor"
@@ -459,12 +481,15 @@ function RuntimeRouteSection({
     .filter((provider) => provider.computerId === displayedComputer?.id)
     .sort((left, right) => providerRank(left.provider) - providerRank(right.provider));
   const computerOptions = [...facts.computers].sort((left, right) => left.displayName.localeCompare(right.displayName));
+  // The heading names the section and everything under it answers it: the route rows label
+  // themselves Computer and Runtime, and where there is no Computer yet the setup panel names the
+  // task. A sentence here would only say those labels again, which is one more line between the
+  // Account and the single action this step asks for.
   return (
     <section aria-labelledby="agent-runtime-heading" className="agent-create-runtime">
       <header className="agent-create-runtime-header">
         <div>
           <h3 id="agent-runtime-heading">Where it runs</h3>
-          <p>Choose the Computer and Runtime for this Agent.</p>
         </div>
       </header>
 
--- a/apps/web/src/computer-setup.test.tsx
+++ b/apps/web/src/computer-setup.test.tsx
@@ -481,4 +481,28 @@ describe("ComputerSetup", () => {
     expect(onConnected).not.toHaveBeenCalled();
     expect(vi.getTimerCount()).toBe(2);
   });
+
+  it("answers the primary action in preview without issuing a code or polling", async () => {
+    const issue = vi.spyOn(browserApi, "issueComputerConnectCode");
+    const computers = vi.spyOn(browserApi, "computers");
+    render(<ComputerSetup preview />);
+
+    await clickGenerate();
+
+    // Review has to see the step this action opens, so the command, its copy affordance and a
+    // running validity are all rendered — from a fixed command, with no Server reached.
+    expect(screen.getByRole("button", { name: "Copy command" })).toBeTruthy();
+    expect(screen.getByRole("status").textContent).toBe("Waiting for the Computer to connect…");
+    expect(screen.getByText(/^Expires in /)).toBeTruthy();
+    expect(issue).not.toHaveBeenCalled();
+    expect(computers).not.toHaveBeenCalled();
+
+    // No poll cycle started, so nothing can expire the command out from under a reviewer.
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(30 * 60_000);
+    });
+    expect(vi.getTimerCount()).toBe(0);
+    expect(screen.queryByRole("alert")).toBeNull();
+    expect(computers).not.toHaveBeenCalled();
+  });
 });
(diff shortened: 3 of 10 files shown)


baixiaohang (changes requested):
- Rationale: the new messaging strip turns an ambiguous not-ready handoff into confirmed forward progress, so a disconnected or locally unready Feishu integration can be presented as “Connecting…”.

**Risk level**: C
- Path baseline: only `apps/web/**` changes.
- Semantic lift: none; this changes presentation and route composition, not subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: first-run Accounts currently see duplicate/non-functional application chrome, implementation-oriented copy, and a staging preview that cannot exercise the important Computer-connect state.
- Approach: make onboarding a standalone full-bleed surface, render the Lab through the production presentation with floating staging controls, simplify first-run identity copy, and make Agent/Feishu state easier to scan.
- Impacted modules: Web routing, onboarding state presentation, Agent creation, Computer setup preview, Lab controls, styles, and their tests.

**Review findings**
❌ 1. `active-not-ready` is not evidence that the Feishu link is “Connecting…”. The Server returns `{ bindingState: "active", handoffReady: false }` when the observed Feishu connection is disconnected or when Feishu CLI readiness is missing; `messagingLink()` maps every such case to the working state. This makes a stalled/disconnected handoff look like confirmed progress. Keep `provisioning` as connecting, but represent `active-not-ready` as unconfirmed/action-required, or expose enough authoritative detail to distinguish the causes, and cover both states in the view tests. [R4 / `apps/web/src/onboarding/view.tsx:299`]
⚠️ 2. Opening the floating Lab switcher leaves focus on a toggle that follows the newly inserted panel in DOM order. A forward Tab therefore skips the scenario/reset controls and wraps through the full onboarding page before reaching them. Move focus into the panel on open (with an intentional return target), or put the toggle before the controlled panel so keyboard reviewers can reach the Lab controls directly. [`apps/web/src/internal/onboarding-lab-page.tsx:180`]

**Action taken**
- Submitted request changes on head `0cdf725a8e5c5b0743d2897d4c16a1fce1ee547e`.

yuezengwu (changes requested):

Reviewed exact head `0cdf725a8e5c5b0743d2897d4c16a1fce1ee547e` against base `e82d00a5769242a504a73f8897acf0fbfbd66fe4`.

**Findings**

❌ 1. The connection strip reports inference as fact. `active-not-ready` only says the combined handoff gate failed: the Server produces it when the Feishu connection lease is disconnected, the Agent Runtime is not ready, or the Feishu CLI is not ready. Mapping every one of those cases to `Connecting…` makes a stalled or disconnected handoff look like confirmed forward progress. The same mapping calls every authorization/error state `Disconnected`, although those states do not prove transport disconnection either. Keep `provisioning` as `Connecting…`, but use an evidence-backed neutral/action-required label for ambiguous not-ready states, or extend the response with the exact authoritative cause. Add view coverage for `active-not-ready` and the attention states. [`apps/web/src/onboarding/view.tsx:286`]

⚠️ 2. Opening the floating Lab switcher leaves focus on the toggle, but the newly rendered panel precedes that toggle in DOM order. A forward Tab therefore leaves the switcher and traverses the onboarding page before it can reach the scenario and reset controls. Move the toggle before the controlled panel, or focus the first panel control on open and restore focus intentionally on close; cover the keyboard path. [`apps/web/src/internal/onboarding-lab-page.tsx:108`]

**Validation**

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- Focused Web tests: 185/185 passed across app, Computer setup, Onboarding Lab, and onboarding page suites
- `pnpm test`: all 7 Turbo tasks passed, plus 66/66 repository script tests
- `git diff --check e82d00a5769242a504a73f8897acf0fbfbd66fe4..0cdf725a8e5c5b0743d2897d4c16a1fce1ee547e`

No merge action taken.

Gandy2025:
Both findings were right, and the first one was the same mistake I set out to fix — I replaced one label that overclaimed with two that did.

**1. `active-not-ready` reported as progress — fixed in 3cfac71**

Confirmed against `im-binding-service.ts:896`: `handoffReady: true` requires `agentRuntimeReadiness === "ready" && providerCliReadiness === "ready" && connectionReady`, and `ImBindingHandoffStatus` carries only `bindingState` and `handoffReady`, so the page cannot tell which of the three failed. An active binding that fails that gate is not an act in progress and not a severed transport.

| `HandoffProgress.kind` | before | now |
| --- | --- | --- |
| `provisioning` | `Connecting…` | `Connecting…` — the one state the Server states as in progress |
| `active-not-ready` | `Connecting…` | `Not ready`, neutral tone |
| `attention` (`reauthorization_required` / `error` / `disabled`) | `Disconnected` | `Needs attention` |

`Disconnected` went for the reason @yuezengwu gave: an expired authorization proves nothing about the transport. `Needs attention` states what the state asks of the Account, which is the part the page can actually evidence.

New coverage in `page.test.tsx`: `active-not-ready` asserts `Not ready` and asserts `Connecting…` is absent; `provisioning` asserts `Connecting…`; all three attention states run through `it.each` and assert `Needs attention`, no `Disconnected`, and `data-status="attention"`.

**2. Switcher focus order — fixed in 3cfac71**

The toggle now precedes the panel it controls in the document, so a forward Tab from it lands inside the panel. The panel still reads above the toggle through `order: -1` in the stylesheet, which is where visual order belongs. `onboarding-lab.test.tsx` asserts the document position, `aria-controls`/`aria-expanded`, and that Escape closes the panel and returns focus to the toggle.

`pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` all pass — web is 336 tests, up 6.

baixiaohang (changes requested):
- Rationale: the visible status labels now stay within the Server evidence, but the accessible name still reports an unsupported transport failure for every attention state.

**Risk level**: C
- Path baseline: only `apps/web/**` changes.
- Semantic lift: none; this is presentation and route composition, not subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: first-run Accounts currently see duplicate/non-functional application chrome, implementation-oriented copy, and a staging preview that cannot exercise the important Computer-connect state.
- Approach: make onboarding a standalone full-bleed surface, render the Lab through production presentation with floating staging controls, simplify first-run identity copy, and report Agent/Feishu state from authoritative facts.
- Impacted modules: Web routing, onboarding state presentation, Agent creation, Computer setup preview, Lab controls, styles, and their tests.

**Review findings**
❌ 1. The visible `Needs attention` fix still exposes the old overclaim to screen-reader users. The section's accessible name is built from `link.description`, and the `attention` entry says `unable to reach`; however, this branch includes `reauthorization_required`, which does not prove that the Feishu transport is unreachable. The page therefore announces “Agent unable to reach Feishu” for a state it intentionally treats as ambiguous in the visible UI. Use an evidence-neutral accessible description for `attention` (for example, that the Feishu link needs attention), and assert the accessible name for the attention cases so the two representations cannot drift. [R4 / `apps/web/src/onboarding/view.tsx:299`]

✅ The prior `active-not-ready`/attention visible-label finding is otherwise resolved, and the Lab toggle now precedes its controlled panel in DOM order with focused coverage.

**Action taken**
- Submitted request changes on head `3cfac717eeb7c391023c6b4d1987d820f88a9386`.

Gandy2025:
Right, and I had fixed only half of it — the visible labels moved, their accessible counterparts did not.

**Fixed in dd8a328**

| link state | accessible name before | now |
| --- | --- | --- |
| `attention` | `…Agent unable to reach Feishu` | `…Agent needs attention on its link to Feishu` |
| `not-ready` | `…Agent not yet ready to reach Feishu` | `…Agent not confirmed ready to reach Feishu` |

I took `not-ready` with it for the same reason, though it was not in the finding: the gate it reports can be failed by the Feishu CLI or by the connection lease, so naming the Agent as the thing that "is not yet ready" attributes a cause the page cannot see either.

The tests now assert the accessible name by its exact string alongside the visible label in the same case, so the two cannot drift apart again:

```
findByLabelText("OpenTag Agent needs attention on its link to Feishu")
findByLabelText("OpenTag Agent not confirmed ready to reach Feishu")
```

`pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` all pass.

baixiaohang (approved):
**Recommendation**: approve
- Rationale: the final patch resolves the remaining accessible-state overclaim, and the onboarding presentation now stays within the Server evidence in both its visible and screen-reader representations.

**Risk level**: C
- Path baseline: only `apps/web/**` changes.
- Semantic lift: none; this changes presentation and route composition, not subscriptions, WebSocket parsing, or the data layer.

**PR summary**
- Problem: first-run Accounts currently see duplicate or non-functional application chrome, implementation-oriented copy, and a staging preview that cannot exercise the important Computer-connect state.
- Approach: make onboarding a standalone full-bleed surface, render the Lab through the production presentation with floating staging controls, simplify first-run identity copy, and report Agent/Feishu state only from authoritative facts.
- Impacted modules: Web routing, onboarding state presentation, Agent creation, Computer setup preview, Lab controls, styles, and their tests.

**Review findings**
✅ The visible and accessible status copy now agree without inferring progress, transport failure, or a specific readiness cause from the combined handoff gate; exact accessible-name coverage protects the ambiguous states.
✅ The Lab toggle precedes its controlled panel in document order, retains the intended visual placement, and restores focus on Escape.
✅ All six current-head CI checks passed.

**Action taken**
- Approved head `dd8a328b8df1376455ad12885e6e4d5f8c3868d7`.

yuezengwu (approved):
Reviewed exact head dd8a328b8df1376455ad12885e6e4d5f8c3868d7 against base e82d00a5769242a504a73f8897acf0fbfbd66fe4.

The prior blockers are resolved on this head:
- the messaging strip now reserves progress for provisioning, reports active-but-not-ready without inventing a cause, and keeps attention states neutral in both visible and accessible copy;
- the Lab toggle now precedes its revealed panel in DOM/tab order, while CSS preserves the intended visual placement, and Escape closes the panel and restores focus.

I also reviewed the full PR diff and the task-relevant Feishu readiness boundary. Local validation passed: git diff --check; pnpm check; pnpm build; pnpm typecheck; focused Web tests (191/191); and pnpm test (66 script tests plus 7/7 Turbo tasks, including Web 336/336). All current GitHub checks are successful.

PR #224, opened 2026-08-28
feat(server): bind Account and Workspace one to one

Gandy2025:
## Summary

Every management read scopes on `workspace_id` alone and treats that as "belongs to the signed-in Account". Nothing enforced it. Migration 0016 backfilled one admin grant per active admin membership, so both directions could break, and on staging both did:

- **A Workspace with several Admins.** Each Admin sees the others' Agents, Computers and Tasks. #167 measured one Workspace with 5 equal Admins, 8 Agents and 5 Computers.
- **An Account with several Workspaces.** `resolveCompatibilityWorkspaceId` returns only the first and the rest are unreachable from the web. Silent, and the mirror image of the first.

This adds two partial unique indexes on `workspace_admin_grants`, on `user_id` and on `workspace_id`, both where the grant is active. Together they make Account and Workspace a bijection, so `workspace_id = ?` means what the reads already assume.

**No query, DTO, route or Web change.** The property those reads relied on is now true rather than conventional.

## Why not scope the reads by owner

That was the obvious alternative and it is worse on three counts. It answers only the first direction, because the Account-to-Workspace collapse happens before any such predicate runs. It leaves `getById` and every mutation reachable by id, so it hides rows without changing who may touch them. And it makes `created_by_user_id` and `enrolled_by_user_id` carry ownership, which #179 reserves as audit metadata.

`createForWorkspace` needs no ownership check as a result. An enrollment in your Workspace is necessarily yours: enrollments carry `enrolledByUserId = connectCode.issuedByUserId`, and connect codes are issued only by that Workspace's one Admin.

## Legacy data

Legacy databases cannot satisfy the indexes, so 0024 resolves what it can prove lossless and refuses the rest by name.

- A redundant Workspace holding **no active Agent and no live enrollment** has its grant revoked, keeping the one the resolver already selects — so nothing an Account sees moves. The migration-history contract test seeds exactly this shape, and staging has it too.
- Anything else is reported with its id, because it needs resources moved.

`scripts/split-shared-workspaces.sql` is that move. It stays operational rather than automatic because it cannot finish on its own: an Agent whose creator did not enroll the Computer it runs on gets a second enrollment of that Computer, and only the host can run `computer connect` for it. It moves enrollments and Agents in one data-modifying CTE so the composite `agents_workspace_enrollment_fk` holds throughout.

## Deploying

`OPENTAG_AUTO_MIGRATE` runs migrations at boot, so **the split has to run on a shared-Workspace database before this deploys**, or the server will not start.

Failing that way is safe and covered: 0024 refuses whole, leaving no half-applied state and nothing in the journal, and the same deploy succeeds once the split has run — roll forward, no restore.

## Tests

`scripts/split-shared-workspaces.sql` touches the only copy of the data it moves, so it is tested rather than reviewed by eye. Writing that test caught two real defects in it: an Agent excluded from the move while its enrollment left under it, and deleted Agents, which the foreign key does not exempt.

Three integration tests asserted the shared-Workspace behaviour this removes. Their premises are now unconstructable rather than wrong, so they are deleted; the two that still hold span two Accounts instead of two grants. The Onboarding Lab keeps its own ownership checks — what stands behind them deletes Agents, and they are the only line left if these indexes are ever relaxed.

`pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, `pnpm --filter @opentag/client test:agent-runtime:coverage` and `pnpm --filter @opentag/server test:integration` (13 files, 217 tests) all pass.

## Deviation from #179

#179 selected discarding the Workspace-bound data (D2) and folding `workspaces` into `users`. **The historical Agents and Computers are being kept**, and Workspace stays as a layer, so that several Admins can be allowed again by dropping these two indexes rather than by rebuilding a scope. #179 still records D2 + R2 as selected and should be updated, or the next reader will resume that plan.

Refs #167, #179.

### DIFF
--- a/scripts/split-shared-workspaces.sql
+++ b/scripts/split-shared-workspaces.sql
@@ -1,0 +1,170 @@
+-- Split every Workspace that holds several Admins into one Workspace per Admin.
+--
+-- A one-off operational task, run once against a deployed database before migration 0024, which refuses
+-- to create its indexes while any Workspace still has more than one Admin. It is deliberately not a
+-- migration: the step it cannot perform is on the operator's side, described under "After this runs".
+--
+-- Nothing is deleted and no id changes. Agents keep their ids, so IM bindings, Sessions, message history
+-- and session placements all follow untouched: im_bindings references agent_id, and session_placements
+-- references the enrollment id, and neither is rewritten here.
+--
+--   psql "$DATABASE_URL" -f scripts/split-shared-workspaces.sql
+--
+-- Take a database snapshot first. Review the report the final SELECT prints before committing.
+
+BEGIN;
+
+-- The Admin holding the earliest grant keeps the original Workspace, matching the order
+-- resolveCompatibilityWorkspaceId already uses, so that Account sees no change at all. Every other Admin
+-- gets a new Workspace that copies setup_completed_at, without which they would be sent back through
+-- onboarding by WorkspaceSetupGate.
+CREATE TEMP TABLE split_plan ON COMMIT DROP AS
+WITH shared AS (
+	SELECT workspace_id
+	FROM workspace_admin_grants
+	WHERE revoked_at IS NULL
+	GROUP BY workspace_id
+	HAVING count(*) > 1
+),
+ranked AS (
+	SELECT
+		g.id AS grant_id,
+		g.user_id,
+		g.workspace_id AS source_workspace_id,
+		row_number() OVER (PARTITION BY g.workspace_id ORDER BY g.granted_at, g.user_id) AS rank
+	FROM workspace_admin_grants g
+	INNER JOIN shared ON shared.workspace_id = g.workspace_id
+	WHERE g.revoked_at IS NULL
+)
+SELECT
+	ranked.grant_id,
+	ranked.user_id,
+	ranked.source_workspace_id,
+	CASE WHEN ranked.rank = 1 THEN ranked.source_workspace_id ELSE gen_random_uuid() END AS target_workspace_id,
+	ranked.rank
+FROM ranked;
+
+INSERT INTO workspaces (id, name, display_name, setup_completed_at, created_at, updated_at)
+SELECT
+	p.target_workspace_id,
+	'workspace-' || p.target_workspace_id,
+	left(coalesce(u.display_name, 'Account') || '''s Workspace', 120),
+	w.setup_completed_at,
+	now(),
+	now()
+FROM split_plan p
+INNER JOIN workspaces w ON w.id = p.source_workspace_id
+INNER JOIN users u ON u.id = p.user_id
+WHERE p.rank > 1;
+
+UPDATE workspace_admin_grants g
+SET workspace_id = p.target_workspace_id
+FROM split_plan p
+WHERE g.id = p.grant_id AND p.rank > 1;
+
+-- An Agent whose creator did not enroll the Computer it runs on would otherwise be separated from its
+-- enrollment. Give that Account its own enrollment of the same physical Computer, which the schema
+-- already allows: workspace_computers is unique on (workspace_id, computer_id), not on computer_id.
+-- These rows carry no credential, which is what the operator step below is for.
+-- Deleted Agents are included: agents_workspace_enrollment_fk does not exempt them, so one left behind
+-- on an enrollment that moves would break the composite key.
+CREATE TEMP TABLE rehomed_enrollments ON COMMIT DROP AS
+WITH stranded AS (
+	SELECT DISTINCT
+		a.created_by_user_id,
+		a.workspace_computer_id AS source_enrollment_id,
+		a.workspace_id AS source_workspace_id
+	FROM agents a
+	INNER JOIN workspace_computers wc ON wc.id = a.workspace_computer_id
+	INNER JOIN split_plan p ON p.source_workspace_id = a.workspace_id AND p.user_id = a.created_by_user_id
+	WHERE wc.enrolled_by_user_id <> a.created_by_user_id
+)
+SELECT
+	s.created_by_user_id,
+	s.source_enrollment_id,
+	gen_random_uuid() AS target_enrollment_id,
+	p.target_workspace_id
+FROM stranded s
+INNER JOIN split_plan p ON p.user_id = s.created_by_user_id AND p.source_workspace_id = s.source_workspace_id;
+
+INSERT INTO workspace_computers (
+	id, workspace_id, computer_id, display_name, platform, arch, client_version,
+	enrolled_by_user_id, enrolled_at, updated_at
+)
+SELECT
+	r.target_enrollment_id,
+	r.target_workspace_id,
+	wc.computer_id,
+	wc.display_name,
+	wc.platform,
+	wc.arch,
+	wc.client_version,
+	r.created_by_user_id,
+	now(),
+	now()
+FROM rehomed_enrollments r
+INNER JOIN workspace_computers wc ON wc.id = r.source_enrollment_id;
+
+-- Enrollments and Agents move in one statement. agents_workspace_enrollment_fk is a composite key over
+-- (workspace_id, workspace_computer_id) and is not deferrable, so two separate UPDATEs would expose a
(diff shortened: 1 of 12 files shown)


baixiaohang (changes requested):
- Rationale: The preservation path leaves live runtime and Slack routing state attached to the old Account scope, and the one-to-one schema direction is not authorized by the current design record.

**Risk level**: B-high
- Path baseline: `packages/server/**` includes DB schema/migration changes and `services/workspace-admin-access/**`.
- Semantic lift: DB structure and management-authorization semantics change.

**PR summary**
- Problem: Account-native management currently collapses multiple Workspace grants, while a shared Workspace lets several Admins see and manage the same Agents and Computers.
- Approach: enforce one active Workspace per Account and one active Admin per Workspace, split legacy shared scopes while preserving existing resources, and retain Workspace as a future resource-container layer.
- Impacted modules: Server schema and migration history, Workspace authorization, Agent/Computer integration coverage, Onboarding Lab guards, and the operational split script.

**Review findings**
❌ 1. The shared-machine preservation path does not move existing Session placements. When an Agent uses another Admin's enrollment, lines 123-134 replace the Agent's `workspace_computer_id` with a newly generated enrollment id, but `session_placements` (and placement-bound proofs/custody) remain on the original enrollment. Existing Sessions for that Agent therefore keep routing through the other Account's enrollment, without a placement-generation fence. Please migrate or explicitly retire/refuse that runtime state atomically, and extend the shared-machine fixture with a live Session/placement that proves the final placement and stale-runtime fencing. [R5 / `scripts/split-shared-workspaces.sql:123`]
❌ 2. Current `main` now contains the Workspace-scoped Slack installation model from #199, but the split script neither moves nor refuses `slack_installations`. If the current Slack route belongs to an Agent moved to a new Workspace, the installation stays on the original Workspace; current route resolution then rejects the `agents.workspace_id != slack_installations.workspace_id` mismatch and inbound Slack stops working. Update from `main`, restack the migration after its existing 0024, and include Slack installation/route/Session handling in the split and its migration coverage. [R5 / `scripts/split-shared-workspaces.sql:7`]
❌ 3. The latest owner decision in #179 explicitly pauses schema work until the resource-container scenario, cardinalities, lifecycle, selection, and authorization model are accepted. This PR makes that missing architecture choice durable by enforcing Account↔Workspace one-to-one and documenting both indexes as the future multi-Admin escape hatch. Please first record an accepted replacement design or obtain explicit owner authorization for this narrower interim schema. [R3 / `packages/server/src/db/schema/auth.ts:58`]
⚠️ 4. The script tells the operator to review its final report before committing, but the documented `psql -f` invocation runs the unconditional `COMMIT` immediately after the report. For this only-copy data move, make the review gate executable (for example, a dry-run/explicit commit workflow) or state accurately that the snapshot-and-restore path is the only post-report escape. [R4 / `scripts/split-shared-workspaces.sql:13`]

**Action taken**
- Submitted request changes on head `87ce63a447f3b02daf1bac0b40c4d0919c20f8d3`.
- Did not run checks or tests; this review is based on the current diff, repository invariants, current `main`, and the author-reported verification.

scripts/split-shared-workspaces.sql
```diff
@@ -0,0 +1,170 @@
...
+INNER JOIN workspace_computers wc ON wc.id = r.source_enrollment_id;
+
+-- Enrollments and Agents move in one statement. agents_workspace_enrollment_fk is a composite key over
+-- (workspace_id, workspace_computer_id) and is not deferrable, so two separate UPDATEs would expose a
+-- state it rejects. A data-modifying CTE is a single statement, and the constraint is checked once at
+-- the end of it.
+-- The rehomed enrollment is matched per Agent, on the enrollment that Agent actually sits on, because one
+-- Account can hold both an Agent that needs rehoming and an Agent that does not.
+WITH moved_enrollments AS (
+	UPDATE workspace_computers wc
```
yuezengwu, line 115: This update cannot move a normally enrolled Computer while its consumed connect-code history exists. `computer_connect_codes_workspace_enrollment_fk` references `workspace_computers(workspace_id, id)` from `computer_connect_codes(workspace_id, consumed_workspace_computer_id)`; lines 139–146 leave consumed rows in the source Workspace, and the FK is non-deferrable. I added a consumed-code row to the existing split fixture and reproduced: `update or delete on table "workspace_computers" violates foreign key constraint "computer_connect_codes_workspace_enrollment_fk" on table "computer_connect_codes"`. The existing test inserts enrollments directly, so it misses this normal deployed state. Move the corresponding consumed-code rows in the same transaction, or explicitly refuse/handle them before enforcing the indexes, and add this fixture to migration coverage.

yuezengwu (changes requested):
Request changes on exact head `87ce63a447f3b02daf1bac0b40c4d0919c20f8d3`.

I reproduced one additional blocking migration failure beyond the current review: a normally enrolled Computer has a consumed connect-code row whose composite foreign key targets `(workspace_id, consumed_workspace_computer_id)`. The split moves `workspace_computers.workspace_id` but deliberately leaves consumed-code history in the source Workspace, so PostgreSQL rejects the move. The current integration fixture inserts enrollments directly and therefore misses the deployed shape. See the inline finding for the exact failure and required coverage.

I also independently reproduced the existing Session-placement blocker: when the script clones an enrollment and rewrites the Agent, the existing `session_placements.workspace_computer_id` remains on the source enrollment (workspace mismatch count `1`); current workspace-qualified runtime joins no longer see that Session. This needs an atomic migrate-or-retire/fence policy, including placement-bound proofs/custody.

The architecture gate is still unresolved: the latest owner decision in #179 suspends D2/R2 and says no durable schema implementation is authorized until the container scenario, cardinalities, authority and movement rules are accepted. This PR fixes Account↔Workspace 1:1 and derives split ownership from `created_by_user_id` / `enrolled_by_user_id`, although #179 treats those as audit metadata rather than authority. Current `main` also already owns migration number 0024, so any authorized revision must rebase and regenerate migration history.

Validation on this head:
- `pnpm check`, `pnpm build`, `pnpm typecheck`: pass
- `pnpm test`: pass
- Agent Runtime coverage: 288 tests, 100% statements/branches/functions/lines
- server integration: 217 tests pass
- existing shared-workspace split tests: 2 pass
- augmented consumed-code fixture: fails with `computer_connect_codes_workspace_enrollment_fk`
- augmented live Session-placement fixture: script completes but leaves one cross-Workspace placement mismatch
- `git diff --check origin/main...HEAD`: pass

The two augmented repro fixtures were temporary review probes and were reverted; the retained review worktree is clean.

Gandy2025:
Closing: the premise is wrong, not the implementation.

This PR binds Account and Workspace one to one to make `workspace_id = ?` mean "belongs to this Account". Both indexes contradict the staged conclusion in #179 that Workspace becomes the resource-management container with Team/Org carrying people and roles: one Admin per Workspace forbids a container having several administrators, and one Workspace per Account contradicts "once an Account can hold more than one resource container, selection returns".

Under that model the behaviour this set out to fix — several Admins seeing each other's Agents — is not a defect. What is missing is a permission model: `workspace_admin_grants` has no role column, so a grant is total. Isolation was the wrong answer to that.

Thanks to @baixiaohang and @yuezengwu — every finding was reproduced and confirmed before this was closed, and four of them are worth carrying to whatever replaces this:

1. `computer_connect_codes` has a composite FK on (workspace_id, consumed_workspace_computer_id), so moving an enrollment across Workspaces breaks consumed code history. Fixtures that insert enrollments directly instead of going through `machineAuth.exchangeConnectCode` miss it.
2. `slack_installations` is Workspace-keyed with no owner column and route resolution matches it against `agents.workspace_id`, so moving an Agent without its installation breaks inbound Slack.
3. The placement join in `session-service.ts` is Workspace-qualified, so a `session_placements` row left on a source enrollment makes a live Session vanish rather than error.
4. `agents_workspace_enrollment_fk` is composite and not deferrable, so enrollments and Agents must move in one data-modifying CTE; deleted Agents are not exempt.

A rule that an Agent never changes Computer removes (3) entirely and reduces (1) and (2) to one statement each, at the cost of refusing to split an Agent from the machine it runs on.

The branch is kept for those.

Issue #227, opened 2026-08-28
design: ownership model — current entities, open problems, and the decisions they wait on


Gandy2025:
# Ownership model: current entities, open problems, and the decisions they wait on

Status: **discussion**. No implementation is proposed here. Everything below is checked against `origin/main` at `9ee9eb1` and against the bodies and comments of #167 and #179.

This exists because #179's premise was reversed on 2026-08-27 — Workspace becomes the resource-management container rather than something to remove — and no single place yet records what the model actually is under that reading, what is wrong with it, and what has to be decided before anything is built. #224 was an attempt to fix one symptom before those decisions existed, and was closed for that reason.

## Entity model

25 tables. Only the ones carrying ownership are drawn; the rest hang off a node here without changing the structure.

```mermaid
flowchart TB
  A["Account<br/>users"]
  W["Workspace<br/>workspaces<br/>name, display_name, setup_completed_at"]
  E["Computer enrollment<br/>workspace_computers<br/>1 active credential"]
  C["Physical Computer identity<br/>computers — id only"]
  AG["Agent<br/>agents"]
  SI["Slack installation<br/>slack_installations<br/>1 active per Workspace"]
  B["IM binding<br/>im_bindings — 1 active per Agent"]
  S["Session / Task<br/>sessions"]
  P["Run placement<br/>session_placements"]

  A -->|"workspace_admin_grants — no role column"| W
  W --> E
  W --> AG
  W --> SI
  AG -->|"composite FK: same Workspace"| E
  E --> C
  AG --> B
  B --> S
  S --> P
  P -.->|"Workspace-qualified join"| E
```

Not drawn: `agent_runtime_configs`, `workspace_computer_credentials`, `computer_connect_codes`, `im_messages` / `im_message_deliveries`, `session_messages`, `session_descendants`, `session_cli_proofs`, `auth_identities` / `auth_sessions`, `admin_invitations`.

### Cardinalities

| Edge | Today | Enforced by |
| --- | --- | --- |
| **Account ↔ Workspace** | **N:M in schema; 1:1 in practice for new Accounts; broken in both directions in legacy data** | **nothing — convention only** |
| Workspace → Agent | 1:N | name unique within Workspace |
| Workspace → enrollment | 1:N | one active enrollment per (Workspace, Computer) |
| Physical Computer → enrollment | 1:N | same machine may enroll in several Workspaces, each with its own credential |
[...]

PR #229, opened 2026-08-28
refactor(web): remove the standalone Computers page

Gandy2025:
## Summary

The account menu opened a page that listed enrolled Computers and offered a connect form. Both jobs are already covered where a Computer is actually chosen: the New Agent flow embeds `ComputerSetup` for an Account with no Computers and offers "Connect another Computer" beside the picker, and a failing Computer is explained and reconnected on the Agent's own Connected computer page.

Remove the menu entry and the page. The route redirects to `/agents` the way `/resources` and `/usage` do, so an existing bookmark still lands somewhere.

Also clear what the removal orphaned: the login note promised management of Computers that no page offers, four style rules had no remaining caller, and a layout test guarded a rule nothing renders.

## Scope change during review

This PR originally also scoped Computer listing and Agent creation to the enrolling Account. That part is reverted in 8f3f068 on review feedback: it reverses the accepted #125 contract that `enrolledByUserId` is audit metadata rather than an authorization subject, and the ownership model is still pending in #227. The reverted work is preserved in b789e11 for whoever picks up that decision.

Reverting it also removes the second blocker raised in review, which was downstream of the listing filter: `ComputerSetup` polls the Computers collection to detect completion, so hiding an enrollment from a caller who had just rotated its credentials would have reported an expired code after a successful rotation.

What remains touches `apps/web` only.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — 1296 unit tests pass
- `pnpm --filter @opentag/server test:integration` — 230 pass

Changed coverage:

- `sends a retired Computers bookmark to Agents` (new) replaces `opens the Computers page from the account menu`
- The account-menu tests for focus order, arrow-key cycling, and menu contents drop the removed item
- `page-layout.test.ts` drops its assertion on `.settings-list-section`, whose rule is deleted

## Breaking changes

One capability is removed: enrolling a Computer without creating an Agent now requires entering the New Agent flow to reach the connect form.

## Checklist

## Follow-up not in this PR

`/agents/:agentId/settings/computer` renders a heading and an `Online` badge with an empty body when the Computer is healthy, and the settings overview makes its row static in that state, so the page is reachable only by URL or back-navigation. With the Computers page gone it is the only Computer surface left.

### DIFF
--- a/packages/server/src/services/agents/agent-service.ts
+++ b/packages/server/src/services/agents/agent-service.ts
@@ -307,9 +307,6 @@ export class AgentService {
           );
           if (replay) return replay;
         }
-        // An Agent runs on a physical host its creator must be able to reach, and enrollment is the
-        // only record of who reached it. A Computer another Account enrolled is therefore not a
-        // valid target, and reads as absent rather than forbidden so the collection stays private.
         const [computerEnrollment] = await transaction
           .select({ id: workspaceComputers.id })
           .from(workspaceComputers)
@@ -317,7 +314,6 @@ export class AgentService {
             and(
               eq(workspaceComputers.workspaceId, workspaceId),
               eq(workspaceComputers.computerId, input.computerId),
-              eq(workspaceComputers.enrolledByUserId, callerUserId),
               isNull(workspaceComputers.revokedAt),
             ),
           )
--- a/packages/server/src/services/agents/agent-service.ts
+++ b/packages/server/src/services/agents/agent-service.ts
@@ -307,6 +307,9 @@ export class AgentService {
           );
           if (replay) return replay;
         }
+        // An Agent runs on a physical host its creator must be able to reach, and enrollment is the
+        // only record of who reached it. A Computer another Account enrolled is therefore not a
+        // valid target, and reads as absent rather than forbidden so the collection stays private.
         const [computerEnrollment] = await transaction
           .select({ id: workspaceComputers.id })
           .from(workspaceComputers)
@@ -314,6 +317,7 @@ export class AgentService {
             and(
               eq(workspaceComputers.workspaceId, workspaceId),
               eq(workspaceComputers.computerId, input.computerId),
+              eq(workspaceComputers.enrolledByUserId, callerUserId),
               isNull(workspaceComputers.revokedAt),
             ),
           )
--- a/packages/server/src/services/workspaces/workspace-admin-service.ts
+++ b/packages/server/src/services/workspaces/workspace-admin-service.ts
@@ -40,25 +40,16 @@ export class WorkspaceAdminService {
     await this.#workspaceAdmins.bootstrapAdminInTransaction(transaction, userId, workspaceId);
   }
 
-  /**
-   * Workspace Admin authority opens the collection; enrollment decides its contents. Recovery is
-   * stated against the physical host ("start OpenTag there"), so a Computer is only listed to the
-   * Account that enrolled it — the one Account known to have reached that machine.
-   */
   async listComputers(
     accountId: string,
     workspaceId: string,
     includeProviderReadiness = false,
   ): Promise<ListWorkspaceComputersResponse> {
     await this.#workspaceAdmins.requireAdmin(accountId, workspaceId);
-    return { computers: await this.#projectComputers(accountId, workspaceId, includeProviderReadiness) };
+    return { computers: await this.#projectComputers(workspaceId, includeProviderReadiness) };
   }
 
-  async #projectComputers(
-    accountId: string,
-    workspaceId: string,
-    includeProviderReadiness: boolean,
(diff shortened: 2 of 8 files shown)


packages/server/src/services/workspaces/workspace-admin-service.ts
```diff
@@ -61,7 +70,13 @@ export class WorkspaceAdminService {
           ne(agents.status, "deleted"),
         ),
       )
-      .where(and(eq(workspaceComputers.workspaceId, workspaceId), isNull(workspaceComputers.revokedAt)))
+      .where(
+        and(
+          eq(workspaceComputers.workspaceId, workspaceId),
+          eq(workspaceComputers.enrolledByUserId, accountId),
```
yuezengwu: [blocker] This filter breaks the existing cross-Admin recovery surface. Any Workspace Admin can open a shared Agent's Connected computer page and issue a reconnect code. Exchange succeeds for the existing enrollment, but MachineAuthService leaves enrolledByUserId unchanged; this query therefore keeps the Computer invisible to the reconnecting Admin. ComputerSetup polls this same collection for the target and will expire after a successful credential rotation. Either make ownership and recovery authority consistent end to end (including who may reconnect and how ownership changes), or preserve Workspace-wide visibility; cover the cross-Admin reconnect path.

packages/server/src/services/agents/agent-service.ts
```diff
@@ -307,13 +307,17 @@ export class AgentService {
...
+        // only record of who reached it. A Computer another Account enrolled is therefore not a
+        // valid target, and reads as absent rather than forbidden so the collection stays private.
         const [computerEnrollment] = await transaction
           .select({ id: workspaceComputers.id })
           .from(workspaceComputers)
           .where(
             and(
               eq(workspaceComputers.workspaceId, workspaceId),
               eq(workspaceComputers.computerId, input.computerId),
+              eq(workspaceComputers.enrolledByUserId, callerUserId),
```
yuezengwu: [blocker] This turns enrolledByUserId into Agent-creation authorization, reversing the accepted #125 contract that any Workspace Admin may bind a Workspace Agent to an enrolled Computer and that enrollment actor identity is audit-only. The PR says the ownership model belongs in #227, whose decision is still pending outside this review. Please defer or split this restriction until that product/architecture decision is adopted; the page-removal portion does not require this reversal.

yuezengwu (changes requested):
Requesting changes on exact head e5441c64be94cb436b6ec9fc1fa1fe929de14460.

Two blockers remain:

1. The accepted direction in #125 says every Workspace Admin may manage Workspace Computer enrollments and bind Workspace Agents to them, while enrolledByUserId is audit metadata rather than an authorization subject. This head reverses that contract in both listing and Agent creation. The PR itself says the ownership model belongs in #227; that decision is still pending outside this review, so this PR cannot independently adopt one side of it. The Computers-page removal is consistent with #125 and can be retained or split, but the ownership restriction needs the external decision first.

2. The restriction is not end-to-end even under its proposed model. Every Admin still sees every Agent and the Connected computer page still offers Reconnect this Computer. A non-enrolling Admin can issue and successfully exchange a connect code, but MachineAuthService reuses the enrollment without changing enrolledByUserId. listComputers then hides that Computer from the caller, while ComputerSetup polls that list to detect completion, so the UI eventually reports an expired code after credentials were actually rotated. Please define one coherent authority/recovery contract and add cross-Admin recovery coverage.

Validation on this head: canonical Context Tree first-tree-ai/opentag-context-tree@4f76b147 verified clean; git diff --check, pnpm check, pnpm build, pnpm typecheck, pnpm test, and 231 server integration tests passed; all 6 GitHub checks are green.

yuezengwu (approved):
Approved on exact head 8f3f06840781201ec38693137341d55477c232cf.

The fresh head fully reverts the server-side Account scoping, so both prior blockers are resolved: `enrolledByUserId` remains audit metadata under #125, and the cross-Admin reconnect flow no longer rotates credentials while hiding the Computer from `ComputerSetup` polling. The net PR diff is now Web-only.

The remaining page removal matches #125 accepted navigation direction. `/agents/computers` retains a tested redirect, Computer enrollment remains reachable from New Agent, and per-Agent recovery remains on Connected computer. The removed CSS selectors have no remaining callers.

Validation: canonical Context Tree verified clean at 4f76b1474848c515cef9c1bc7ce42c5558395e89; `git diff --check`, `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, direct Web tests (19 files / 332 tests), and server integration tests (13 files / 230 tests) all passed. All 6 GitHub checks are green.

Residual product tradeoff is documented in the PR: standalone enrollment is less discoverable and now requires entering New Agent. I do not see that as a blocker against the accepted #125 placement.

PR #230, opened 2026-08-28
fix: preserve all-message thread routing

baixiaohang:
## Summary

- Route `all_message` thread-scoped messages to a Thread Session owner and a Channel Session observer without changing per-target attention.
- Preserve trusted direct continuity, ended-session fencing, and atomic out-of-order root upgrades.
- Add negotiated `replyRole` delivery/steer v2 handling with fail-closed rolling compatibility while keeping legacy owner payload hashes stable.
- Document the shared Feishu/Slack thread-routing model in English and Chinese. No database migration is required.

## Validation

- `git diff --check`
- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/server test:integration` (234 tests passed)

## Review

First Tree chat review passed with no blocking findings.

### DIFF
--- a/packages/server/src/services/im/im-message-inbox.ts
+++ b/packages/server/src/services/im/im-message-inbox.ts
@@ -351,17 +351,29 @@ export class ImMessageInbox {
                       event.conversation.externalId,
                       threadKey,
                     );
-                const rootExternalId = threadRootExternalId(event.providerContext);
-                const rootWasDirect =
-                  !existingThread && !direct && !endedThreadExists && rootExternalId
-                    ? await this.#rootWasDirectToChannelSession(
-                        transaction,
-                        imBindingId,
-                        event.conversation.externalId,
-                        rootExternalId,
-                      )
-                    : false;
-                if (existingThread || direct || rootWasDirect) {
+                const rootExternalId = await this.#reliableThreadRootExternalId(
+                  transaction,
+                  imBindingId,
+                  event.conversation.externalId,
+                  threadKey,
+                  event.providerContext,
+                );
+                const rootWasDirect = rootExternalId
+                  ? await this.#rootWasDirectToChannelSession(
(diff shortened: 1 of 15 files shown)


packages/server/src/services/im/im-message-inbox.ts
```diff
@@ -468,17 +484,37 @@ export class ImMessageInbox {
                   workspaceComputerId: scope.agent.workspaceComputerId,
                   now,
                 });
-                await transaction
-                  .insert(imMessageDeliveries)
-                  .values({
-                    messageId: pending.id,
-                    sessionId: thread.id,
+                const [upgraded] = await transaction
```
yuezengwu, line 487: [blocking] Before this upgrade, the correlation above uses only the root stored on each pending event. Feishu can omit rootId on later events in the same thread; during ingest, #reliableThreadRootExternalId correctly recovers that root from an earlier event, but this upgrade loop does not. I reproduced: ingest ambient event A with threadId + rootId, ingest same-thread event B with threadId but no rootId, then ingest the direct root out of order. A upgrades to direct while B stays ambient. That violates the documented same-normalized-thread continuity and the PR's atomic-upgrade claim. Please derive the reliable root for each pending threadKey here (or precompute an equivalent validated thread-to-root mapping), upgrade every undispatched ambient delivery in the correlated thread, and add this regression case.

yuezengwu (changes requested):
Requesting changes because the out-of-order direct-root upgrade does not apply the PR's reliable-root recovery to every pending event in the normalized thread. A later Feishu event that omits the repeated rootId remains ambient while an earlier event in the same thread upgrades to direct, so continuity is only partially restored. I reproduced this on the exact head. All existing local suites and GitHub checks pass; details are in the inline blocker.

PR #232, opened 2026-08-28
fix(server): recover roots for pending thread events

baixiaohang:
## Summary

- Recover a reliable Feishu root once per normalized `threadKey` while processing an out-of-order direct root.
- Apply that root to every eligible pending ambient Thread delivery, including later events that omit the repeated `rootId`.
- Preserve the existing ended-session fence and the pending/unclaimed atomic upgrade boundary.
- Add a regression covering one same-thread event with `rootId`, a later event without it, and the direct root arriving last.

This follows up on the blocker discovered after #230 was merged.

## Validation

- `pnpm check`
- Targeted Feishu out-of-order root regression
- Full `im-binding.test.ts` (90/90)
- `pnpm --filter @opentag/server test:integration` (236/236)
- `git diff --check`

## Review

First Tree chat review passed with no blocking findings.

### DIFF
--- a/packages/server/src/__tests__/integration/im-binding.test.ts
+++ b/packages/server/src/__tests__/integration/im-binding.test.ts
@@ -2310,7 +2310,12 @@ describe("IM binding persistence", () => {
         grantedScopes: [...FEISHU_REQUIRED_TENANT_SCOPES],
       });
       await value.database.update(agents).set({ receiveMode: "all_message" }).where(eq(agents.id, value.agent.id));
-      const inbox = new ImMessageInbox(value.database);
+      let rootHistoryLookups = 0;
+      const inbox = new ImMessageInbox(value.database, {
+        beforeReliableThreadRootLookup: () => {
+          rootHistoryLookups += 1;
+        },
+      });
 
       const firstReply = revisionEvent({
         providerEventId: "feishu-out-of-order-first-reply",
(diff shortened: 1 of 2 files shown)


yuezengwu (changes requested):
Requesting changes on exact head fcf67dd7ea2a549b85c20023bd2cd3ffbfffbb92. The normalized-thread correctness regression is fixed, and the new regression plus all local/GitHub checks pass, but the implementation introduces an unbounded query-amplification path in packages/server/src/services/im/im-message-inbox.ts:472. `pendingThreadMessages` contains every inbound thread revision in the channel at or after the delayed direct message, and before comparing a thread root to this event the loop calls `#reliableThreadRootExternalId` once for every distinct `threadKey`. A late root in a busy channel can therefore issue N serial full-thread-history reads for N unrelated threads while the ingest transaction holds the Agent/binding locks, delaying all later inbound events. Please derive the reliable-root map from the already selected rows in a two-pass scan (including revisions before external-message deduplication), or batch-fetch roots for the candidate thread keys in bounded queries, and add a regression that proves unrelated threads do not cause one extra history lookup each. The existing binding/channel scope, ended-session fence, and pending/unclaimed atomic upgrade conditions should remain unchanged.

yuezengwu (approved):
Approving exact head 678c83d50a6b6954b3c0743cc5b5160f6e0929e3. The prior query-amplification blocker is resolved: reliable roots are now derived in a two-pass scan over the already selected binding/channel-scoped pending revisions, before external-message deduplication, so unrelated thread keys no longer trigger serial full-history lookups while ingest holds its locks. The normalized-thread recovery still upgrades every eligible pending ambient delivery and preserves the ended-session fence, pending/unclaimed atomic predicates, binding/channel isolation, and revision/dedup behavior. The new query-count regression passes locally, as do the full 90-test IM binding file, all 236 PostgreSQL integration tests, pnpm check/build/typecheck/test, and git diff --check. GitHub currently has Checks/CI red only because the pre-existing retained-claim lock-order fixture reported that its delivery was not dispatched; this PR regression passed there, and that unchanged fixture passed in the full local integration run plus three focused reruns. I do not find that flaky failure blocking for the reviewed diff, but required CI should still be green before merge.

PR #234, opened 2026-08-28
feat(web): add the redesigned onboarding flow on a mock backend

liuchao-001:
## What

The redesigned onboarding flow, as a self-contained surface at **`/internal/onboarding-v2`**, running entirely against an in-page mock.

It reaches **no Server** and reuses **none** of the existing onboarding presentation or state derivation. It exists so the pages and interactions can be developed and tuned on their own, and it has been through many rounds of exactly that.

## The shape

The first screen asks where the agent runs. That question is **not a step** — it decides how many steps follow, and a rail on that screen would have to state a length it cannot know yet. So it carries no rail, and each route shows its own.

| Route | Steps |
| --- | --- |
| **Local computer** | Create agent → Prepare computer → Messaging app |
| **Cloud computer** | Create agent → Messaging app |

Cloud is **Coming soon** and cannot be chosen in the product; the mock panel opens it so its pages stay reviewable.

Every step shares one footer: **Go back** on the left, **Continue** on the right, disabled until the step can be left. Going back undoes the decision that advanced you rather than moving a separate cursor, so the page stays a pure function of the facts. The messaging step has no footer at all — the Agent already exists, and the step is finished by scanning a code or installing an App, so neither button could ever be used.

## What the code forced

- **Local runtimes are Codex and Claude Code.** `AGENT_RUNTIME_PROVIDERS` has exactly those two, and the Context Tree bars Pi from product selection. The cloud route's own runtime is therefore never named in the UI.
- **A local Agent is created once its route is proven.** `CreateAgentRequest` requires a `computerId` and `runtimeProvider` is immutable afterwards, so the name and runtime stay a local draft. A failed check never strands a half-built Agent on a runtime the user cannot run.
- **Connecting and checking are one step.** The check settles within about 100ms of the Computer arriving, so its result appears under "Your computer is connected." rather than behind a screen change.
- **A Computer's enrollment is durable.** Returning to that step shows the machine already connected instead of asking for it again.

## Built on Kumo

`main` replaced the web app's own token stylesheet and hand-rolled controls with **Cloudflare Kumo** while this branch was open, and added a contract test holding the app to it. This surface was written against what was deleted.

Merging `main` in was therefore not a conflict fix. The stylesheet still matched, but every declaration naming a retired token was dropped at computed-value time, so the flow rendered with no padding, borders, backgrounds or radii at all. Fourteen further selectors reached for `.ds-icon` and `.ds-button` — still valid CSS, still parsed, simply no longer matching anything.

- **The roles this surface is written in are bound at the shell** — to a Kumo token wherever Kumo names the same role, so it follows the theme and its dark mode, and to a value only where it does not. Binding rather than inlining is what let the tuned rules keep working unchanged.
- **Its controls are Kumo's**: selectable cards, the name field, the readiness picker, the cloud toggle. Headings render through `Text` while staying real heading elements, because the step outline is what a screen reader navigates this flow by.
- **Two reserved slots had been sized against the retired font stack** and no longer held their tallest state, so the footer moved as a check resolved on a narrow screen. Both are re-measured.

⚠️ `postcss` is now declared by `apps/web`. The layout test imports it and had been resolving it transitively through vite until `main`'s dependency changes rehoisted it. The lockfile diff is three lines, in this importer only.

## Installing

The connect command runs the **portable installer**, not `npm i -g`: npm needs a working Node before OpenTag can be installed at all, which the first command of onboarding cannot assume. ⚠️ This deliberately differs from what the Server's `buildComputerConnectCommand` still emits — changing that is separate work.

⚠️ `opentag doctor --fix` does not exist yet either. The CLI work is tracked separately, and shipping the page ahead of it is a deliberate call.

## Nothing moves under the reader

Anything that appears, disappears or changes length while someone is reading has a slot sized for the tallest state it holds.

- The pending command block renders the real block's structure, so the command arriving shifts nothing.
- The connect code breaks **by character** while the rest of the command breaks between words. Left to break at its own hyphens, two codes of the same length wrapped differently. Measured over **400 random codes per width**, the block's height is now identical at 1000, 640, 390 and 320px, and `sh` never splits across lines.
- A check row reserves its failure line below 399px, and the repair summary a second pair of lines below 359px. Swept across **five readiness outcomes at eight widths from 1280 to 320px**, the footer holds one position as the check settles — 40 combinations, no movement.

## Marks

Slack, Lark and Claude appear as **their publishers' own files**, carried in `assets/` with a comment recording where and when each came from. None is redrawn. Slack's install uses Slack's published **"Add to Slack"** button, referenced from Slack's own URL the way their developer documentation embeds it, so nothing of theirs is redistributed here.

**No Codex asset is carried.** Taking an icon out of an installed application establishes where the bytes came from and nothing about redistributing them from this repository, so Codex shows a neutral mark until a publisher-provided asset with usable terms exists.

Ownership and the conditions we keep to are recorded in [TRADEMARKS.md](../blob/feat/onboarding-v2-mock/TRADEMARKS.md), linked from both READMEs.

## Mock controls

One control beside the panel names whichever outside event is pending — the Computer arriving, the check returning, a sign-in being approved, a QR being scanned, a return from Slack — and fires it. Those default to waiting for that nudge, so any state can be held indefinitely; `realistic` and `fast` modes are there for judging the flow at pace. The panel picks the readiness outcome from five scenarios, can expire a code, simulate a completed repair, and offer the cloud computer.

## Verification

`pnpm check`, `pnpm --filter @opentag/web typecheck` and `pnpm --filter @opentag/web build` clean; web tests **366 passed**, the Kumo contract test among them. Both routes walked in a real browser against the built app.

⚠️ `pnpm test` also runs `packages/client`, where `client-runtime-composition.test.ts` fails on this machine. **Reproduced on clean `origin/main`** — same file, same four tests, with `pnpm build` run first — so it is not from this branch.

## Worth a reviewer's eye

1. The connect command and `doctor --fix` both describe CLI behaviour that is not built yet.
2. Cloud needs the Server to allocate a Computer before an Agent can be created there. That capability is the real gating work, which is why the route ships disabled.
3. This surface still carries **its own stylesheet** rather than being rebuilt out of Kumo components. The alternative was to rewrite 1100 lines of measured layout against `Flow`, `RadioGroup` and `ClipboardText` in the same change that unbroke it. Doing that deliberately, with the appearance re-reviewed, is better than doing it under a merge.
4. The `apps/web` dev server fails to transform `router.tsx` under `@vitejs/plugin-react`'s babel step (`str.indexOf is not a function`). **Pre-existing on `main`** and unaffected by `pnpm build`. Preview with `pnpm --filter @opentag/web build && pnpm --filter @opentag/web exec vite preview`, then open `/internal/onboarding-v2`.


apps/web/src/onboarding-v2/copy.ts
```diff
@@ -0,0 +1,118 @@
...
+    runtimeHint: "The coding agent OpenTag will run on your computer.",
+    runtimeFootnote: "More runtimes coming soon.",
+    runtimeLocked: "You can't change the runtime later, so pick the one you actually use.",
+    continue: "Continue",
+  },
+
+  setup: {
+    title: "Set up your agent to run",
+    description: "Your AI worker runs on your own computer. Connect that computer to OpenTag.",
+    privacy: "Your code and data never leave your machine.",
```
yuezengwu, line 131: [P1] Qualify the local privacy claim

Both selectable runtimes are local CLIs that call external provider services, so prompts and context can leave the machine even though the workspace and execution are local. This unconditional sentence is inaccurate and creates privacy/compliance risk. Please state the actual OpenTag boundary and defer provider data handling to the selected provider policy.

apps/web/src/onboarding-v2/steps.tsx
```diff
@@ -0,0 +1,434 @@
...
+      </ul>
+
+      {!resolving && failures.length > 0 ? (
+        <div className="otv2-repair">
+          <p className="otv2-repair__intro">
+            {failures.length > 1 ? COPY.setup.checksFailedIntro : COPY.setup.checksFailedIntroSingular}
+          </p>
+          <p className="otv2-muted">{COPY.setup.doctorIntro}</p>
+          <CommandBlock
+            command="opentag doctor --fix"
```
yuezengwu: [P1] Do not offer an unsupported repair command

`opentag doctor` only accepts `--server-url`, and its implementation checks server health rather than installing or repairing either runtime or `lark-cli`. On this exact head, `node apps/cli/dist/cli/index.mjs doctor --fix` exits 1 with `unknown option --fix`. Because this is the only recovery action shown for failed readiness checks, replace it with supported instructions or implement and test the advertised behavior.

apps/web/src/onboarding-v2/mock-backend.ts
```diff
@@ -0,0 +1,224 @@
...
+  }, []);
+  const later = useCallback((run: () => void, delayMs: number) => {
+    timers.current.push(window.setTimeout(run, delayMs));
+  }, []);
+
+  useEffect(() => clearTimers, [clearTimers]);
+
+  const arrive = useCallback(() => {
+    setConnect((current) =>
+      current.kind === "issued" || current.kind === "expired"
```
yuezengwu: [P2] Prevent expired connection codes from arriving

`expireNow` leaves the scheduled arrival alive, and this branch explicitly allows that old timer to turn an `expired` code into `connected`. A focused fake-timer reproduction (issue, expire, then advance the original 8s arrival) ends at “MacBook Pro is connected”. That makes the expiry/reissue scenario report an impossible success. Only an `issued` code should arrive, or the timer/generation should be invalidated on expiry.

apps/web/src/onboarding-v2/page.tsx
```diff
@@ -0,0 +1,104 @@
...
+  // The connect code is issued when the page that shows it is first reached, not before: an
+  // unseen code would spend its validity in the background.
+  useEffect(() => {
+    if (flow.page === "setup") backend.issueConnectCode();
+  }, [backend.issueConnectCode, flow.page]);
+
+  const createAgent = useCallback(() => {
+    setCreation((current) => {
+      if (current !== "idle") return current;
+      window.setTimeout(() => setCreation("created"), CREATE_AGENT_MS);
```
yuezengwu: [P2] Cancel pending creation when starting over

This timeout is not retained or invalidated by `startOver`. A focused reproduction that clicks “Create my agent”, immediately starts over, and runs the flow again reaches Messaging without a second creation confirmation because the stale timeout has already set `creation` to `created`. Clear the timeout or guard it with a generation token on reset/unmount.

apps/web/src/onboarding-v2/onboarding-v2.css
```diff
@@ -0,0 +1,729 @@
...
+  padding: 0;
+  list-style: none;
+}
+
+.otv2-choices--grid {
+  display: grid;
+  grid-template-columns: repeat(auto-fit, minmax(220px, 1fr));
+}
+
+.otv2-choice {
```
yuezengwu, line 403: [P2] Allow choice-card text to wrap on mobile

The custom choice button inherits the browser button default `white-space: nowrap`; this rule never overrides it. In a 390×844 production preview the first screen had `documentElement.scrollWidth === 596`, and the option copy extended outside the viewport. Set a wrapping white-space policy on the card/copy and add a narrow-viewport regression check.

yuezengwu (changes requested):
Requesting changes on exact head `62f4a753142f0c0e7fe8fbf0e708dfc718dc58ad`. The mock currently exposes an unsupported recovery command and an unconditional privacy promise that does not match provider-bound Codex/Claude execution. I also reproduced two stale-timer state races and a 390px mobile overflow. Validation on a clean detached worktree: Context Tree `4f76b1474848c515cef9c1bc7ce42c5558395e89` verified; `git diff --check`, `pnpm check`, `pnpm build`, root/web typecheck, web tests (362), and the root test suite passed. The required CI check is still red on an untouched server integration lock-order test.

apps/web/src/onboarding-v2/copy.ts
liuchao-001: Agreed, and thank you — this one mattered. The sentence came from the original design draft and I carried it over without checking it against what the runtimes actually do.

Split into the two claims that are each true:

- `Your code and files stay on your computer — OpenTag runs your agent there and never copies them out.`
- `Your agent still sends prompts and context to its own provider, under that provider's data policy.`

The second renders as secondary text under the first, so the OpenTag boundary still reads as the headline without the page implying anything about OpenAI or Anthropic. Covered by a test asserting both halves are present.

apps/web/src/onboarding-v2/steps.tsx
liuchao-001: Confirmed — `--fix` does not exist, and `doctor` only checks server health today.

I kept the command rather than substituting one, because there is no supported command that does these three checks: the detection work itself is unbuilt, and this surface exists to review the flow that depends on it. Swapping in something that runs but answers a different question seemed worse than being explicit.

So the block now carries, in warning colour: `Design only: opentag doctor does not run these checks yet, and --fix is not a supported flag.` Test asserts it renders whenever a check fails.

The actual CLI work is scoped in a separate thread with liuchao-001, including the part this flow depends on most — that repairing in the terminal has no way to push a fresh readiness observation, since doctor is a separate process from the daemon and the protocol has no re-probe frame. If you would rather the command be removed from the mock entirely until that lands, say so and I will pull it.

apps/web/src/onboarding-v2/mock-backend.ts
liuchao-001: Real bug, reproduced. Fixed by letting only an `issued` code arrive.

A reissue already calls `clearTimers`, so the stale arrival is only reachable after an expiry, which makes the state guard sufficient on its own and avoids a second generation counter.

Regression test added: issue, expire, then advance the original 8s arrival. I verified it fails without the fix — with the guard reverted it ends on "MacBook Pro is connected", exactly as you described.

apps/web/src/onboarding-v2/page.tsx
liuchao-001: Real bug, reproduced. The timeout is now held in a ref and cleared by both `startOver` and unmount.

I also moved the state check out of the `setCreation` updater; scheduling a timer inside an updater would have queued two of them under StrictMode double-invocation.

Regression test added: create, immediately start over, then run the flow again and assert the second run still stops at its confirmation. Verified failing without the fix.

apps/web/src/onboarding-v2/onboarding-v2.css
liuchao-001: Correct — I assumed the cards inherited the document's wrapping and they do not.

`.otv2-choice` now sets `white-space: normal`, and the runtime grid collapses to a single column below 640px so `minmax(220px, 1fr)` cannot demand more than a phone's content box.

Measured in a 390x844 production preview: `documentElement.scrollWidth` is 390 on all three of the first screens, against the 596 you found.

For the regression check I followed the repo's existing convention in `src/__tests__/page-layout.test.ts` and asserted the CSS through postcss, since jsdom computes no layout. Three assertions: the card wrapping policy, the narrow-viewport grid collapse, and the command block's wrapping.

liuchao-001:
All five findings fixed in `4a49dc429d580ab9d9f2f82952b9f2c85a017b17`. Detail is in each thread; the short version:

| | Finding | Fix |
|---|---|---|
| P1 | Unconditional privacy claim | Split into OpenTag's own boundary plus an explicit provider-side deferral |
| P1 | `opentag doctor --fix` unsupported | Command kept as the design intent, now labelled design-only in warning colour |
| P2 | Expired code could still connect | Only an `issued` code arrives |
| P2 | Creation survived Start over | Timeout held in a ref, cancelled on restart and unmount |
| P2 | 390px sideways scroll | Cards wrap; runtime grid collapses to one column below 640px |

Both stale-timer fixes have regression tests that I verified **fail without the fix** — reverting each guard reproduces exactly the states you described. The mobile fix is verified by measurement, not just by CSS: `documentElement.scrollWidth` is 390 across the first three screens at 390x844, against the 596 you found.

On the repair command specifically — I kept it rather than substituting something else, because no supported command answers those three checks today; the detection work is unbuilt and this surface exists to review the flow that depends on it. It is now explicitly marked as not-yet-real. If you would rather it be removed until the CLI lands, I will pull it.

**Verification on this head:** `git diff --check`, `pnpm check`, `pnpm --filter @opentag/web typecheck`, `pnpm --filter @opentag/web build` all clean. Web tests 369 passed (22 files), up from 362 — 7 added here.

**On the red CI check:** I ran the full root suite. `@opentag/web` 369, `@opentag/server` 289, `open-tag` 126 all pass; `@opentag/client` has 3 failures in `client-runtime-composition.test.ts`. I confirmed these are **not from this branch** by stashing every change and re-running the same file — all 3 still fail. They are the same lock-order-style failures you saw, and they need their own fix outside this PR.

apps/web/src/onboarding-v2/copy.ts
```diff
@@ -0,0 +1,187 @@
...
+     * messaging app is what creates it, so this says why the field stopped accepting edits.
+     */
+    nameFixed: "Your agent's name is set once it's created.",
+  },
+
+  connect: {
+    title: "Connect your computer",
+    /** What the step is about, and what it means for your data: both belong with the title. */
+    lead: "Your AI worker runs on your own computer. Connect that computer to OpenTag.",
+    privacy: "Your code and data never leave your machine.",
```
yuezengwu, line 131: [P1] Do not restore the unconditional privacy promise

This is the exact sentence the earlier review blocked, and the current production preview renders it again. Local Codex/Claude execution still sends prompts and context to the selected provider, so “never leave your machine” is false even if OpenTag itself does not upload the workspace. Please restore the qualified split you described in the prior thread: state OpenTag’s own boundary separately and defer provider-side handling to that provider’s policy.

apps/web/src/onboarding-v2/copy.ts
```diff
@@ -0,0 +1,187 @@
...
+      count > 1
+        ? `${count} things need fixing before your agent can run.`
+        : "One thing needs fixing before your agent can run.",
+    /**
+     * By the time this page is read, `doctor --fix` is usually already running on the user's
+     * machine — signing in during the previous step starts it. So this is a pointer back to where
+     * the work is happening, not a command to go and run: one light line, no block, no copy button.
+     */
+    repairHint: "Continue in your terminal or agent for instructions, or run",
+    repairCommand: "opentag doctor --fix",
```
yuezengwu, line 160: [P1] Do not instruct users to run a command that still does not exist

This also regresses the earlier blocker. On this exact head, `node apps/cli/dist/cli/index.mjs doctor --fix` exits 1 with `unknown option '--fix'`; the CLI only registers `--server-url`, and the PR body itself says the flag is unimplemented. The “design only / unsupported” warning from the intermediate fix is gone, so the UI presents this as a working recovery action. Remove it until supported, or implement and test the advertised repair behavior before telling users to run it.

apps/web/src/onboarding-v2/flow.ts
```diff
@@ -0,0 +1,280 @@
...
+}
+
+function messagingCliState(status: MessagingCliStatus): CheckState {
+  if (status === "checking") return "pending";
+  if (status === "ready") return "passed";
+  return "failed";
+}
+
+export function readinessPassed(readiness: ReadinessFacts | undefined): boolean {
+  return readiness?.runtime === "ready" && readiness.messagingCli === "ready";
```
yuezengwu: [P1] Do not gate Agent creation on an unselected messaging provider

`messagingCli` is the Lark CLI check, yet this predicate is a hard prerequisite for creating every local Agent. In the built UI, a missing Lark CLI keeps Continue disabled before the user can choose Slack, so a Slack user is blocked by an unrelated integration. The canonical Tree also says Agent creation does not wait for IM handoff readiness and that readiness is provider-specific. Remove messaging readiness from this pre-creation gate, or choose the messaging provider first and probe only that exact integration at handoff.

apps/web/src/onboarding-v2/flow.ts
```diff
@@ -0,0 +1,280 @@
...
+ * run a command, wait for a machine, wait for a check. A cloud agent has none of that: naming it
+ * and pointing it at a messaging app is one short piece of work, so it is one page with no rail.
+ */
+/**
+ * Connecting the Computer and checking it are one step. The check finishes within about a
+ * hundred milliseconds of the Computer arriving, so there is no wait to break up — its result is
+ * simply the rest of what connecting tells you.
+ */
+export const STEP_IDS = ["agent", "computer", "messaging"] as const;
+/** A cloud agent has no Computer to connect, so its route is the two steps that remain. */
```
yuezengwu, line 118: [P1] Keep cloud Agents on the canonical Computer boundary

The shared `CreateAgentRequest` requires `computerId`, and the canonical Tree explicitly requires future cloud execution to use the same Computer/Workspace/Runtime boundary rather than creating a second no-Computer Agent model. This mock currently marks the cloud Agent created while asserting it has no Computer, and the PR body acknowledges that allocation is still missing. Keep cloud disabled/Coming soon until that contract exists, or model Server allocation of a cloud Computer before Agent creation.

TRADEMARKS.md
```diff
@@ -0,0 +1,46 @@
...
+OpenTag connects to products built by other companies. To identify those products in the interface
+we display their own marks, as published by their owners.
+
+## Ownership
+
+Every mark below belongs to its owner, not to this project. They are reproduced here only to
+identify the product each one names — the use a trademark owner expects and permits. Their presence
+does not imply that any owner sponsors, endorses, or is affiliated with OpenTag.
+
+OpenTag's own name and mark belong to this project and are covered by the repository
```
yuezengwu: [P1] Do not claim Apache-2.0 grants trademark rights

The linked LICENSE says the opposite in Section 6: it does not grant permission to use the Licensor’s trade names or trademarks except for narrow origin/NOTICE use. Please separate OpenTag trademark ownership/usage terms from the code license and have the resulting policy reviewed by the appropriate project owner; source provenance also must not be presented as a license grant for third-party marks.

apps/web/src/assets/codex.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Official Codex app icon, from icon-codex-light.png inside ChatGPT.app (Contents/Resources) at version 1.2026.236, taken on 2026-08-29. openai.com serves 403 to direct asset requests, so the shipped application bundle is the available official source. -->
```
yuezengwu: [P1] Establish redistribution permission before shipping this vendor asset

Extracting an icon from an installed application proves where the bytes came from; it does not establish permission to copy and redistribute those bytes in this Apache-licensed repository. Please replace this with a publisher-provided asset whose usage/redistribution terms cover this use, provide project-approved authorization, or keep a neutral placeholder until that provenance is cleared. `TRADEMARKS.md` cannot create permission the asset owner has not granted.

yuezengwu (changes requested):
Requesting changes on exact head `18dbcf890a69aa964f81de4f98a86831a2bff30d`. The three prior P2 findings are fixed on this head: expired codes cannot arrive, Start over cancels in-flight creation, and the 390px layout no longer scrolls sideways. However, both prior P1 findings have regressed, and the expanded flow now encodes two canonical model conflicts plus unresolved trademark/asset-rights risk.

Validation: canonical Context Tree `4f76b1474848c515cef9c1bc7ce42c5558395e89` verified; `git diff --check`, `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` passed (web 396/396); the fresh built CLI still exits 1 for `doctor --fix`; production preview was walked at 1280px and 390px for local/cloud, passing/failing readiness, expiry, and restart paths with no horizontal overflow or browser console errors. All GitHub checks are green.

apps/web/src/onboarding-v2/flow.ts
liuchao-001: Fixed in `8ac1869`, and thank you — this is the sharpest of the six.

You are right that it blocks a Slack user over a Lark dependency, and right that the canonical model settles IM readiness at handoff rather than before an Agent exists. Two things follow, and I did both:

- **The computer step now answers one question**: can this Agent run. Two rows, runtime CLI and runtime sign-in. `readinessPassed` no longer looks at `messagingCli` at all.
- **The chosen provider's CLI is checked on the messaging step**, where the requirement becomes real and there is a provider to name.

Worth recording: this was already my own conclusion. I raised it with the product owner as an open question earlier in this PR and recommended exactly this split — "each check belongs where its requirement becomes real" — but left it unimplemented pending an answer. You found it independently in the built UI, which is the better evidence. It should not have sat unfixed while I waited.

apps/web/src/onboarding-v2/flow.ts
liuchao-001: Fixed in `8ac1869`. The cloud route now allocates a Computer and creates the Agent on it.

You are right that the shared contract requires a `computerId` and that a second, Computer-less Agent model would be a divergence rather than a shortcut. The flow carries a `CloudComputerState` and the step reads "Preparing…" while it allocates; the difference between the routes is now who provides the machine, not whether there is one.

I chose this over keeping cloud disabled because the product owner asked for the cloud flow to be reviewable now, and modelling allocation satisfies both that and the canonical boundary. The Server capability is still missing and the PR body still says so.

TRADEMARKS.md
liuchao-001: Fixed in `8ac1869`. You are right and I was wrong: Apache-2.0 section 6 explicitly withholds trademark rights, so pointing at the LICENSE for OpenTag's own marks said the opposite of what the licence says.

The document now states that OpenTag's name and mark are **not** covered by the LICENSE, that permission to use them is separate and not granted by that document, and — your second point — that recording where a third-party file came from establishes provenance and not permission, and cannot enlarge the owner's own terms. The "adding another" section now tells the next person to confirm redistribution is actually permitted before adding a file, rather than just recording a URL.

On having it reviewed by the appropriate owner: agreed, and I have raised that with the project owner rather than treating my own edit as sufficient.

apps/web/src/onboarding-v2/copy.ts
liuchao-001: Not changed yet, and I want to be transparent about why rather than let it look like an oversight.

Your facts are correct and were correct the first time: a local Codex or Claude Code run sends prompts and context to its provider, so "never leave your machine" is false as an unconditional statement. I made exactly the split you describe in response to your earlier review.

**The product owner then instructed me to revert it**, having read that reasoning, on the grounds that the shorter sentence communicates better to users and that being technically correct was not the goal here. That is their call to make, not mine to quietly undo — and equally not mine to quietly keep against a blocking review.

So I have put it back to them as an explicit decision, with your finding attached. I will implement whichever way they answer. If it stays, I would at least want the qualification to exist somewhere the user can reach.

apps/web/src/onboarding-v2/copy.ts
liuchao-001: Same situation as the privacy line, and same handling.

You are right that the flag does not exist — I verified `doctor --fix` exits 1 on this head myself before the first fix, and nothing since has changed that. I had added the "design only" marker in response to your earlier review.

**The product owner removed that marker deliberately**, telling me the `doctor --fix` work was already underway in a separate workstream. The command is the design intent this surface exists to review, and there is no shipped command that answers those checks, so substituting one would misrepresent a different thing as the recovery path.

That does not settle your objection: today the page shows a working-looking action that the CLI rejects. I have put it back to the owner as an explicit decision. The three options I gave them are: restore the marker, remove the command until the CLI lands, or ship the CLI first.

apps/web/src/assets/codex.svg
liuchao-001: You are right, and this one I should not have shipped.

Provenance is not permission — extracting bytes from an installed application establishes where they came from and nothing about redistributing them in a public Apache-licensed repository. The comment I wrote records the source honestly but does not, and cannot, create the grant.

For context on how it got here: openai.com serves 403 to direct asset requests and the Codex repository carries no mark of its own, so after the owner asked for official icons I went to the shipped application bundle. That answered "is this the official artwork" and skipped "may we redistribute it", which is the question that actually matters here.

The owner asked for this asset specifically, so I have put the removal back to them rather than reverting unilaterally — but my recommendation to them is to drop it to a neutral placeholder until a publisher-provided asset with terms that cover this use is available. The same question deserves asking of the Slack, Claude and Feishu files; Slack at least publishes a media kit intended for this, which is the strongest of the four.

apps/web/src/onboarding-v2/copy.ts
liuchao-001: Owner decision: **the sentence stays as it is, with no added qualification.** They considered your finding and chose the shorter line deliberately, on the grounds that it communicates better to users and that technical completeness is not the goal for this copy.

I am recording that here rather than arguing it further, so the decision and its author are visible in the thread rather than looking like the finding was missed. Your facts stand and are not disputed.

apps/web/src/onboarding-v2/copy.ts
liuchao-001: Owner decision: **the command stays.** `doctor --fix` is being built now in a separate workstream, and they have chosen to ship the page ahead of it rather than add a marker that would be removed again shortly.

Recording it here for the same reason as the privacy line: the finding was read and answered, not missed. Your verification that the flag exits 1 today is accurate and unchanged.

apps/web/src/assets/codex.svg
liuchao-001: Fixed in `b4b5ddf` — the asset is gone.

Codex now shows a neutral mark. Taking the icon out of an installed application established where the bytes came from and nothing about redistributing them from this repository, which is exactly your point and the question I had skipped.

Two related changes came with it:

- The **"Add to Slack" button is no longer copied into the repo**. It is referenced from Slack's own URL, the way Slack's developer documentation embeds it, so nothing of theirs is redistributed here either.
- `TRADEMARKS.md` now tells the next person to confirm redistribution is actually permitted before adding a file, and says outright that anything without a grant is not carried at all.

The Slack mark, Claude mark and the pre-existing Feishu file are still bundled. Those are publishers' own web assets used to identify their products, which is a weaker case than the button but stronger than the app-bundle extraction was. If you want the same standard applied to all three, say so and I will take them to URL references or neutral marks too.

liuchao-001:
`main` moved under this branch while it was open, and it moved a long way: **#252 replaced the web app's design foundation with Cloudflare Kumo**, deleting the token stylesheet and hand-rolled controls this surface was written against, and adding a contract test holding the app to it.

That turned a merge conflict into a port. Reporting what it actually took, since it is the bulk of the new diff:

- The stylesheet still matched, but **every declaration naming a retired token was dropped** at computed-value time. The flow rendered with no padding, borders, backgrounds or radii — structurally intact and visually gone.
- **Fourteen selectors reached for `.ds-icon` and `.ds-button`.** Still valid CSS, still parsed, matching nothing. The copy control had silently gone back to a 16px glyph; every pressed tint had gone with it. This is the failure mode worth flagging: no test and no build step notices it.
- Two reserved slots were **sized against the retired font stack** and no longer held their tallest state, so the footer moved as a check resolved below 400px. Both re-measured, and the sweep is wider than before: five readiness outcomes at eight widths, 40 combinations, no movement.

The roles are now bound at the shell — onto a Kumo token wherever Kumo names the same role, so the surface follows the theme and its dark mode. Controls are Kumo's.

**What I did not do, and want your view on:** rebuild the surface out of `Flow`, `RadioGroup` and `ClipboardText`. That is the honest end state, but doing it inside the change that unbreaks the page would mean rewriting 1100 lines of measured layout with no way to tell a port regression from a merge regression — and the appearance is the thing that was reviewed. I would rather do it deliberately, as its own change, with the result looked at again.

Also worth naming: `pnpm test` fails in `packages/client` on this machine. **Reproduced on clean `origin/main`**, same file and same four tests, so it is not from this branch.

liuchao-001:
@yuezengwu Merging this with your review still marked *changes requested*, on the repository owner's explicit instruction after they were told the review was the only thing blocking it. Flagging it plainly rather than letting it look like the state was missed.

Your three findings from the previous head are all addressed in `b4b5ddf` and answered on their threads:

- the privacy sentence — owner's decision to keep it as written, recorded on the thread;
- `doctor --fix` — owner's decision to ship the page ahead of the command;
- the Codex icon — **fixed**, no Codex asset is carried at all now, and the "Add to Slack" button is referenced from Slack's URL rather than copied.

Since then the branch also absorbed #252 and was ported onto Kumo, which is the bulk of the current diff and is described in the PR body and the comment above.

Nothing here is closed to you. A follow-up PR starts immediately — rebuilding this surface out of Kumo components and connecting it to the real Server API — so anything you find here, including on the port, is worth raising and I will carry it into that change.

PR #242, opened 2026-08-29
feat(server): sign in and register with an email address and password

bestony:
## Summary

Adds an email-and-password credential alongside the existing redirect providers, built on the Better Auth instance rather than beside it.

`OPENTAG_EMAIL_PASSWORD_AUTH_ENABLED` gates it and defaults to off. It is the only sign-in method whose default could hand out Accounts — every other one needs something a deployment already granted, a Google client, a loopback bypass, or a connect code — so enabling it is a decision rather than an inheritance. One setting covers both routes, because a server that accepted passwords but issued none would have no way to give anyone a first one.

`POST /api/v1/auth/email/sign-up` and `POST /api/v1/auth/email/sign-in` are OpenTag routes that drive Better Auth server-side, matching how every step except the OAuth callback already works: the request contract, the origin check, and the response shape stay ours, and the library's catch-all surface stays closed.

Three decisions are worth review attention rather than being taken on trust.

**The CSRF fence is deliberately half of the usual one.** `requireBrowserMutationSecurity` requires both a same-origin request and the readable double-submit token, but a signed-out browser holds no such token — these are the requests that mint it. The origin half is extracted as `requireBrowserOrigin` and is what these two routes use. Requiring the token here would make signing in impossible rather than safer. Both responses carry the Better Auth session cookie and a fresh double-submit token together, which is what lets a newly signed-in browser write at all.

**A rejected sign-in gives one answer.** An unknown address and a wrong password are reported identically, and Better Auth's own reason is not forwarded, so the endpoint cannot be used to ask which addresses hold Accounts. Registration cannot keep that secret and still tell the caller what to do, so a taken address is reported as `AUTH_EMAIL_CONFLICT` — the disclosure any self-serve registration without a verification step must make. Attempts are bounded per source address and per email address; the second bound means an attacker can spend a victim's budget for the window, which is the better failure against distributed guessing at one known address.

**One assumption was wrong and the code that rested on it was removed.** `auth_identities.issuer` is `NOT NULL` and is not part of Better Auth's account model, so this branch first added a hook to supply it. The integration test refuted that immediately: the library writes the column itself, as `local:credential`. Inventing a value would have fought `findAccountByKey({ accountId, issuer })`, which resolves on the library's own convention. The hook is gone; instead a test pins the value, because `issuer` participates in three unique indexes on that table and a library upgrade that changed the convention should fail loudly rather than leave Accounts that silently cannot sign in.

Password bounds (12–128) live in `@opentag/shared` and configure both the request schema and Better Auth from the same constants, so the library cannot apply a different floor underneath and turn an accepted password into a rejected one. The stored value is a hash on the Account's `credential` identity row.

On the web, the login page gains a form that both registers and signs in. `password` stays in the `/api/v1/auth/providers` list with `startUrl: null` rather than being reported through a separate field: it is the one provider that is a form rather than a link, so a caller that renders providers by following `startUrl` has to special-case it instead of silently filtering it out.

**Out of scope, by decision:** email verification and password reset. Nothing in the product sends mail, so both would mean building a mail sender first. `users.email_verified` therefore stays false for these Accounts, which is exactly what the field means — no provider asserted the address.

## Testing

Run on this exact head:

- `pnpm check` — 466 files, no fixes
- `pnpm build`, `pnpm typecheck` — clean
- `pnpm test` — shared 76/76, web 336/336, server 286/287 (see below)
- `pnpm --filter @opentag/server test:integration` — 229/229
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 289/289, 100% gate holds

New coverage: 7 integration tests over the real credential path (registration writes a hash and not the password, the registration arrives holding a session, right and wrong passwords, sign-in through a casing variant, one Account per address, a suspended Account refused, and no Account created when the flag is off), 11 route tests, and 4 web tests over the form.

Two notes on the local run, neither of which is a defect in this change:

`src/__tests__/security.test.ts` fails locally only because of the checkout path. It asserts that a redacted startup error does not contain the string `password`, and `formatStartupError` includes stack-trace paths — the worktree this was developed in is named `feat-email-password-auth`. The same unmodified file passes 3/3 from a checkout without that substring, and CI checks out to `/home/runner/work/opentag/opentag`. The assertion was left alone rather than weakened.

The local runtime was Node v25.8.1, which is outside the supported range and end-of-life, and no supported version was installable on that machine. Treat CI on 22.13.0 / 24 / 26 as authoritative for these results.

## Breaking changes

None. The feature is off by default, no existing route or response shape changes, and no migration is involved — `auth_identities.password` already exists.

`AuthProvidersResponse` gains `password` as a valid provider id. The array already carried disabled entries, so a caller that filters on `enabled` is unaffected; a caller that renders every entry by following `startUrl` will now see one with `null` and must skip or special-case it. The web app does the latter.

## Checklist


baixiaohang (changes requested):
- Rationale: The new credential path creates an account pre-hijacking condition when combined with implicit Google linking, and it bypasses the existing post-sign-in destination allowlist.

**Risk level**: B-high
- Path baseline: `packages/server/**` -> B-low.
- Semantic lift: `apps/web/src/api.ts` changes the Web authentication data layer -> B-high.

**PR summary**
- Author / repo: `bestony` / `first-tree-ai/opentag`
- Problem: Deployments without a usable redirect provider need an optional browser flow that can create an Account and sign in with an email address and password.
- Approach: Add flag-gated OpenTag routes backed by Better Auth's credential provider, issue the session and CSRF cookies together, publish the provider to Web, and render one form for registration and sign-in.
- Impacted modules: shared auth contracts, Server auth/configuration/cookies, Web API/router/styles, and authentication tests/documentation.

**Review findings**

❌ 1. Prevent pre-account hijacking between unverified password registration and Google sign-in. [`packages/server/src/auth/better-auth.ts:146`]

`requireEmailVerification: false` lets anyone register `victim@example.com`, retain its password, and receive a session while the user remains unverified. The same Better Auth instance keeps implicit account linking enabled for Google; Better Auth documents that a same-email OAuth sign-in is automatically linked when the provider verifies the address (or is trusted). When the real owner later signs in with Google, that identity is therefore attached to the attacker's existing Account while the attacker keeps the working password, leaving two people controlling one Account. This is the standard pre-hijacking sequence, not merely an unverified-email label. Require proof of email ownership before a reusable password credential/session can participate in linking, or disable implicit linking and provide a verified recovery/linking design; add an integration test for password sign-up followed by same-email Google sign-in. See [Better Auth account linking](https://better-auth.com/docs/concepts/users-accounts).

❌ 2. Apply the existing sign-in destination allowlist to the password flow. [`apps/web/src/router.tsx:511`]

`next` is read directly from the query string and passed to `window.location.assign` after a successful password sign-in. Unlike Google/dev, this path never calls `validateOAuthNext`, so `/login?next=https%3A%2F%2Fevil.example` becomes a post-authentication open redirect (protocol-relative targets work as well). Reuse a single local-destination contract for every sign-in method and cover external, protocol-relative, backslash, and disallowed internal targets.

❌ 3. Make the password-attempt budget deployment-wide and bounded. [`packages/server/src/api/browser-auth.ts:58`]

The documented per-source and per-email protection is a process-local `Map`: each Server instance has an independent budget, every restart clears it, and expired/one-off keys are never evicted. An attacker can spread guesses across replicas, while a proxy/NAT address can spend the shared source budget for legitimate users. Password guessing is the new security boundary here, so enforce it in a bounded shared store or a verified deployment-wide gateway rather than treating this helper as that guarantee.

❌ 4. Preserve non-credential Better Auth failures instead of relabeling all of them. [`packages/server/src/api/browser-auth.ts:216`]

Any unrecognized sign-up failure falls back to `409 AUTH_EMAIL_CONFLICT`, and every non-OK sign-in response becomes `401 AUTH_INVALID_TOKEN`. A database/session-hook outage returned as `500`/`503` is therefore presented as "address already exists" or "wrong password," which makes the transient path unreachable to Web and operators. Keep the uniform response only for actual unknown-address/wrong-password results, map the known duplicate case to `409`, and preserve transient/internal status classes without forwarding sensitive library details.

**Action taken**
- Submitted request changes on head `697e2a48cbafd1ec6cfbb50d79b999d5f826b99d`.

yuezengwu (changes requested):
I independently verified two merge-blocking security defects on exact head `697e2a48cbafd1ec6cfbb50d79b999d5f826b99d`.

[P1] Apply the existing post-sign-in destination policy to the password form (`apps/web/src/router.tsx:511`, `apps/web/src/router.tsx:618`).

`next` comes directly from the query string and is passed to `window.location.assign` after either password sign-in or registration. The redirect-provider paths pass their destination through `validateOAuthNext`, but this path bypasses that boundary. I reproduced it against the built Web app: signing in at `/login?next=http://127.0.0.1:8001/redirected` navigated the browser to that second origin. Reject or canonicalize the destination before rendering/submitting the form, using one allowlist contract for all sign-in methods, and cover absolute, protocol-relative, backslash, and disallowed internal targets.

[P1] Do not let an unverified self-service signup claim an arbitrary Account email (`packages/server/src/auth/better-auth.ts:141-154`).

Anyone can register a victim's address without proving control of it; `autoSignIn` then issues the attacker a session and OpenTag provisions the Account's default Workspace/Admin grant while `email_verified=false`. Because `users_email_unique` reserves that address, the victim cannot register it. On the exact installed Better Auth 1.7.2, implicit Google linking defaults `requireLocalEmailVerified` to true, so a later verified Google login with that address is refused rather than rescuing the Account. Disabling that guard would instead leave attacker and victim sharing the pre-hijacked Account. Require ownership proof, an invitation/admin-provisioned credential, or another design that cannot mint durable authority for an unverified third-party address; add an integration test for password signup followed by the same verified provider email.

bestony:
@baixiaohang — findings 2, 3, and 4 are fixed on `29c9d41`. Finding 1 is deferred deliberately, with reasoning below rather than a silent omission.

**2 — the destination allowlist now covers this path.** You are right that it never did. `resolveSignInDestination` moves to `@opentag/shared`, and `validateOAuthNext` becomes the server's wrapper around it that turns a refusal into a request failure. The password form applies the same function before navigating.

It belongs in shared rather than on the server because this is the one sign-in method that navigates the browser itself instead of handing its destination to a route — which is exactly why it slipped past. Two implementations of "is this local" would eventually disagree, and the more permissive half is the one that would matter. Covered by an exhaustive table in `packages/shared/src/__tests__/sign-in-destination.test.ts` (absolute, scheme-relative, backslash, fragment, control characters, over-length, and local-but-not-a-destination), plus a component test that asserts where the form decides to land — the navigation is injected so the decision is observable rather than followed.

**3 — bounded, and no longer described as more than it is.** Two separate problems, and the memory one was mine: keying on an email address made the key space caller-chosen and unbounded, so the map became a way to spend the server's memory rather than only its patience. The table is now capped, evicting expired entries first and oldest-live after that; evicting a live counter can only grant attempts, never deny them. A test drives 5,000 distinct addresses through it and holds the cap.

The cross-replica half is not something this helper can fix, so I stopped implying it does. `DEVELOPMENT.md` now states plainly that the counters are per process, cleared by a restart, and that a deployment-wide bound needs a shared store or a gateway in front. If you would rather this be Postgres-backed in this PR than documented as per-process, say so and I will do that instead — I took the view that a real shared store is an infrastructure decision worth making on its own rather than inside this change.

**4 — failures are classified before they are restated.** Confirmed as described: `betterAuthFailure` could never match Better Auth's vocabulary against `ErrorCodeSchema`, so registration always fell through to `409`, and sign-in relabelled unconditionally. Now `5xx` becomes `503 SERVICE_UNAVAILABLE` and `429` stays rate-limited on both routes, without forwarding the library's message; registration reports `409` only for an actual duplicate and `VALIDATION_ERROR` otherwise; sign-in keeps the uniform `401` for the library's own refusals.

One addition beyond the finding: a suspension raised by `onSessionCreating` is now preserved rather than reported as a wrong password. Reaching it required presenting the correct password, so naming the reason discloses nothing the caller could not already infer, and "incorrect password" would send a suspended user to reset a password that works. Only codes that parse as OpenTag's own survive that path, so nothing the library decided leaks through it.

**1 — deferred, and now documented as a deployment constraint.** The sequence is real and I am not disputing it: registration proves nothing about the address, Google is a trusted provider, so the later real-owner sign-in links into the squatter's Account and both hold it.

Closing it properly means proving address ownership before a credential can participate in linking, which means a mail sender — the piece this repository does not have, and the reason verification was scoped out in the first place. Disabling implicit linking instead would convert takeover into a lockout that any stranger can inflict on any address, which I do not think is a trade worth making silently. Either way it is a design change rather than a fix, so it is being done as separate work.

Until then `DEVELOPMENT.md` and its Chinese mirror state the constraint directly: enabling `OPENTAG_EMAIL_PASSWORD_AUTH_ENABLED` alongside Google is only safe where every address that can reach the server is already trusted. The flag is off by default, so no existing deployment acquires this by upgrading. If you would rather the flag not ship at all until ownership proof exists, that is a reasonable position and I would rather hear it now than merge past it.

**Validation on `29c9d41`:** `pnpm check`, `pnpm build`, `pnpm typecheck` clean; shared 80/80, web 343/343, server 291/292, integration 229/229, agent-runtime 100%.

The one server failure is `security.test.ts`, and it is an artifact of the checkout path rather than this change: it asserts a redacted startup error does not contain the string `password`, and `formatStartupError` includes stack-trace paths — the worktree is named `feat-email-password-auth`. The same unmodified file passes 3/3 from a checkout without that substring, which is why CI is green. I left the assertion alone rather than weakening it.

bestony:
@yuezengwu — both P1s confirmed. The first is fixed on `29c9d41`; the second is deferred, but your reading of it was right and mine was not, so the record is corrected on `95ea17e`.

**P1 — destination policy. Fixed.**

Your repro is exactly the defect. `resolveSignInDestination` now lives in `@opentag/shared` and `validateOAuthNext` is the server's wrapper around it, so there is one contract rather than two. The form applies it before navigating, on both the sign-in and the registration branch you flagged.

It had to move to shared rather than be re-implemented on the client: this is the one sign-in method that navigates the browser itself instead of handing its destination to a route, which is why it slipped past the boundary in the first place, and two implementations would eventually disagree with the more permissive one deciding. Covered by an exhaustive table in `packages/shared/src/__tests__/sign-in-destination.test.ts` — absolute, scheme-relative, backslash, fragment, control characters, over-length, and local-but-not-a-destination — plus component tests that assert where a sign-in decides to land, including `http://127.0.0.1:8001/redirected`.

**P1 — unverified signup claiming an address. Deferred, and you were right about the mechanism.**

I verified `requireLocalEmailVerified` against the installed 1.7.2 source. It defaults to true at `oauth2/link-account.mjs:82`, and the refusal at line 83 checks it against `dbUser.user.emailVerified`; `isTrustedProvider` only guards the separate `!userInfo.emailVerified` clause, so trusting Google does not lift it. A later Google sign-in for a squatted address returns `{ error: "account not linked", data: null }`.

The other review on this PR described the same finding as a shared-Account takeover, and I repeated that framing in my earlier reply and in the documentation. That was wrong, and the correction matters because it changes what the fix has to do: there is no merged Account to prevent, there is a reserved address and a locked-out owner, plus a provisioned Account and default grant minted for an address nobody proved. `95ea17e` adds the integration test you both asked for — password signup, then the same verified provider email — asserting the refusal, one user row, and one identity row, and rewrites the documentation that claimed otherwise.

The defect itself stays open. Closing it means proving ownership before a credential can claim an address, which means a mail sender this repository does not have — the reason verification was scoped out to begin with. Doing it under this PR would make the change a mail-infrastructure change wearing a sign-in feature's title, so it is being done as separate work with the test above as its baseline.

That leaves a decision I would rather put to you both explicitly than assume. The flag is off by default and documented as safe only where everyone who can reach the server is already trusted, and on that basis I think it can ship ahead of the ownership work. If you consider self-service registration unshippable until that lands, say so and I will drop the registration route from this PR and keep only sign-in for Accounts that already exist — which was one of the original scoping options and needs no mail sender.

**Validation on `95ea17e`:** `pnpm check`, `pnpm build`, `pnpm typecheck` clean; shared 80/80, web 343/343, server 291/292, integration 230/230, agent-runtime 100%.

The single server failure is `security.test.ts` and is an artifact of the checkout path, not this change: it asserts a redacted startup error does not contain the string `password`, and `formatStartupError` includes stack-trace paths — the worktree is named `feat-email-password-auth`. The same unmodified file passes 3/3 from a checkout without that substring, which is why CI is green. The assertion was left alone rather than weakened.

baixiaohang (changes requested):
- Rationale: The destination, memory-bound, and failure-classification fixes address findings 2–4, but self-service registration still gives an unverified address a session and a provisioned Account while locking the real owner out.

**Risk level**: B-high
- Path baseline: `packages/server/**` -> B-low.
- Semantic lift: the Web authentication data layer and sign-in flow change -> B-high.

**PR summary**
- Author / repo: `bestony` / `first-tree-ai/opentag`
- Problem: Deployments without a usable redirect provider need an optional browser flow that can create an Account and sign in with an email address and password.
- Approach: Add flag-gated OpenTag routes backed by Better Auth's credential provider, issue the session and CSRF cookies together, publish the provider to Web, and render one form for registration and sign-in. The follow-up centralizes the post-sign-in destination policy, bounds the in-process limiter, and preserves transient failures.
- Impacted modules: shared auth contracts, Server auth/configuration/cookies, Web API/router/styles, and authentication tests/documentation.

**Review findings**

❌ 1. Do not expose self-service sign-up until claiming an address requires ownership proof or an equivalent trusted capability. [`packages/server/src/auth/better-auth.ts:146`]

The new integration test correctly resolves the earlier disagreement about Better Auth 1.7.2: its default `requireLocalEmailVerified` guard refuses the real owner's later Google sign-in instead of linking it. But that test also pins the remaining blocker. An arbitrary caller can register `victim@example.com`, receive a session, and trigger default Account/Workspace provisioning while `email_verified=false`; the unique email index then prevents the real owner from registering, and Google cannot reach the squatted Account. The flag defaulting off prevents existing deployments from acquiring the behavior automatically, but a documentation-only “everyone who can reach the server is trusted” condition is not enforced when the feature is enabled. Require email verification, invitation/admin provisioning, or another one-time trusted capability before creating the credential/session. Otherwise, remove the public sign-up path from this PR (and keep sign-in only if credentials have a trusted provisioning path) until ownership work lands.

⚠️ 2. The limiter is now memory-bounded and accurately described as per-process. That is non-blocking under the currently documented single-Server topology; a shared store or verified gateway remains required before password auth is supported across replicas. [`packages/server/src/api/browser-auth.ts:59`]

✅ 3. The shared destination allowlist closes the password-flow open redirect, and the new failure classification preserves rate-limit/transient outcomes without forwarding Better Auth's messages.

**Action taken**
- Submitted request changes on head `95ea17e82026e56db18c2a8638a3358502c928b8`.

yuezengwu (changes requested):
Re-review on exact head `95ea17e82026e56db18c2a8638a3358502c928b8`.

The redirect allowlist, bounded per-process attempt map, and transient/internal error classification are fixed. I also verified the redirect repair against a running Web build: an external `next` value now falls back to `/agents`.

**[P1] Do not expose self-service password registration without an enforced proof that the caller owns the email address** (`packages/server/src/auth/better-auth.ts:141-154`).

The new integration test at `packages/server/src/__tests__/integration/better-auth.test.ts:732-777` accurately pins the remaining behavior: any reachable caller can register another person's email, immediately receive a session and the default provisioned OpenTag account/workspace authority, and reserve the unique email. A later verified Google sign-in for the real owner is refused. Keeping the feature flag off by default and documenting “trusted operators only” does not enforce that trust boundary once a deployment enables the route.

Please require an ownership proof or an authorization capability such as an invite/admin-provisioning flow before creating the account. The narrow alternative for this PR is to remove the self-service sign-up surface and ship password sign-in only once credentials can be provisioned through a trusted path.

Validation on this head: `pnpm check`, `pnpm build`, `pnpm typecheck`, the full unit suite, all server integration tests (230/230), client agent-runtime coverage (289/289 at 100%), focused auth/Web tests, and a real-browser login regression. GitHub CI is also fully green.

PR #245, opened 2026-08-29
fix(client): preserve Claude Code default credential lookup

### LINKED ISSUE #236: Claude Code sign-in is invisible because the Client Runtime always sets CLAUDE_CONFIG_DIR
## Summary

The Client Runtime always sets `CLAUDE_CONFIG_DIR` for the Claude Code provider, even when the value
equals the CLI's own default (`$HOME/.claude`). Claude Code treats the presence of that variable as
selecting a distinct credential record, so a normal macOS Keychain login is invisible to us. The
provider probe therefore reports `credential_missing`, the Server sees `claude-code` readiness as
`sign-in` forever, and the sign-in instructions we give the user cannot fix it — the user signs in
without the variable set, and we keep reading a different record.

## Reproduction (macOS, signed in to Claude Code normally)

```console
$ claude auth status --json | head -3
{
  "loggedIn": true,
  "authMethod": "claude.ai",

$ env -i HOME="$HOME" PATH="$PATH" USER="$USER" claude auth status --json | head -3
{
  "loggedIn": true,
  "authMethod": "claude.ai",

$ env -i HOME="$HOME" PATH="$PATH" USER="$USER" CLAUDE_CONFIG_DIR="$HOME/.claude" claude auth status --json | head -3
{
  "loggedIn": false,
  "authMethod": "none",
```

The only difference between the last two commands is `CLAUDE_CONFIG_DIR`, set to the path Claude Code
would have used anyway. Verified with `claude` 2.1.248 and a `claude.ai` subscription login.

## Where it comes from

`createClientRuntime` resolves `claudeCodeHome` from `options.claudeCodeHome ?? CLAUDE_CONFIG_DIR ??
$HOME/.claude` and then always injects it:

```ts
const claudeCodeEnvironment = claudeCodeAgentRuntimeEnvironment({
  ...sourceEnvironment,
  CLAUDE_CONFIG_DIR: claudeCodeHome,
});
```

`packages/client/src/runtime/client-runtime-composition.ts`

## Impact

Every macOS Computer whose operator signed in to Claude Code the ordinary way. `codex` is unaffected
(`CODEX_HOME` does not partition its credentials the same way). The user-visible symptom is a
Computer stuck at "sign in to Claude Code" that never turns ready, however many times they sign in.
It also makes the new `opentag doctor` runtime check (#235) print a fix that cannot work.

## Possible directions

1. Omit `CLAUDE_CONFIG_DIR` from the provider environment when the resolved home equals the CLI
   default, and keep using the resolved path only for artifact identity and `verifyArtifact`.
2. Keep injecting it, but treat an explicit provider home as an opt-in that the operator must sign
   in under, and say so in the readiness message.

Direction 1 preserves the identity/verification behaviour while restoring the ordinary login path,
so it looks preferable, but the choice affects what `artifactIdentity` means for a Session binding
and deserves a deliberate decision.

yuezengwu:
## Summary

- omit `CLAUDE_CONFIG_DIR` when the resolved Claude Code home is the CLI default for the same child `HOME`
- keep injecting the canonical path for every non-default configured home
- preserve provider-home identity and artifact verification against the resolved path
- add process-level tests for default omission plus explicit environment and option preservation

This restores ordinary Claude Code macOS Keychain credential discovery without changing Session binding identity.

Closes #236

## Validation

Validated on synchronized head `f9d3733` against `main@a0f6e84`:

- `pnpm check`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test` (client 490, server 292, CLI 152)
- `pnpm --filter @opentag/client test:agent-runtime:coverage` (292 tests, 100%)
- `pnpm --filter @opentag/server test:integration` (265 passed)
- `git diff --check`
- exact-head GitHub CI (6/6 passed)

### DIFF
--- a/packages/client/src/__tests__/client-runtime-composition.test.ts
+++ b/packages/client/src/__tests__/client-runtime-composition.test.ts
@@ -432,6 +432,40 @@ describe("createClientRuntime production composition", () => {
     ).toBe("configuration_unsupported");
   });
 
+  it("omits CLAUDE_CONFIG_DIR when Claude Code uses the default home", async () => {
+    const home = await temporaryDirectory("opentag-client-claude-default-home-");
+    const { capture, runtime } = await composeClaudeCodeRuntime({
+      environment: { HOME: home, PATH: process.env.PATH },
+      home,
+    });
+    runtime.stop();
+    expect(capture).toBe("unset");
+    await expect(realpath(resolve(home, ".claude"))).resolves.toBe(resolve(home, ".claude"));
+  });
+
+  it("preserves an explicit CLAUDE_CONFIG_DIR in the Claude Code environment", async () => {
+    const home = await temporaryDirectory("opentag-client-claude-explicit-env-");
+    const claudeCodeHome = resolve(home, "explicit-env-claude");
+    const { capture, runtime } = await composeClaudeCodeRuntime({
+      environment: { HOME: home, PATH: process.env.PATH, CLAUDE_CONFIG_DIR: claudeCodeHome },
+      home,
+    });
+    runtime.stop();
+    expect(capture).toBe(`set:${await realpath(claudeCodeHome)}`);
+  });
+
+  it("preserves an explicit claudeCodeHome option in the Claude Code environment", async () => {
+    const home = await temporaryDirectory("opentag-client-claude-explicit-option-");
+    const claudeCodeHome = resolve(home, "explicit-option-claude");
+    const { capture, runtime } = await composeClaudeCodeRuntime({
+      claudeCodeHome,
+      environment: { HOME: home, PATH: process.env.PATH },
+      home,
+    });
+    runtime.stop();
+    expect(capture).toBe(`set:${await realpath(claudeCodeHome)}`);
+  });
+
   it("unit-tests composition delegates and every preflight outcome", async () => {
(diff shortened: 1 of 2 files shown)


yuezengwu:
## Re-validation — PASS

Target: `f9d3733b780963035638dafbe5ec297b737584f1`, synchronized with `main@a0f6e8413275ccb959822f5dd4db76f6be516092`.

- Scope remains limited to Claude Code runtime composition and its process-level tests (2 files).
- Default resolved Claude home: `CLAUDE_CONFIG_DIR` is omitted, restoring Claude Code's ordinary macOS Keychain lookup.
- Non-default configured home: canonical `CLAUDE_CONFIG_DIR` remains injected; provider-home identity and artifact verification remain unchanged.
- Local validation passed: `pnpm check`, build, typecheck, all root tests (client 490 / server 292 / CLI 152), Agent Runtime coverage (292 tests, 100% statements/branches/functions/lines), server integration (265 tests), and `git diff --check`.
- Exact-head GitHub CI: 6/6 passed.

Limit: this host currently has no active Claude Code login under either default environment, so a real logged-in Keychain account E2E reproduction was not possible without acquiring credentials. The environment contract itself is covered by deterministic child-process tests.

No blocking finding. PR is ready for the required independent approval; it has not been merged.

PR #235, opened 2026-08-28
feat(cli): check the Agent Runtime and messaging CLI in doctor

liuchao-001:
## Why

Onboarding step 4 ("check the runtime") needs a command that tells the operator, on their own
machine, what is missing and how to fix it. `opentag doctor` only pinged the Server, so a Computer
could look healthy while the Agent Runtime it must drive was missing or signed out, or while no
messaging CLI existed to deliver a reply. The server's `handoffReady` gate requires all three, and a
user with a perfect Codex setup and no `lark-cli` hits a silent dead end.

## What

`opentag doctor` now reports one line per check and, for every blocking failure, the exact fix
command — phrased so the operator's own coding agent can run it after the command output is pasted
to it.

```
Checks
  OpenTag server: healthy (opentag-server) at http://127.0.0.1:8000
  Codex CLI: ready (codex-cli 0.149.0)
  Claude Code CLI: installed (2.1.248) but not signed in
  Feishu CLI (lark-cli): ready
  Slack CLI: ready

1 check must be fixed before this computer can run an OpenTag agent.

Fix 1/1 — Sign in to Claude Code on this computer
  claude auth login
  Docs: https://docs.claude.com/en/docs/claude-code/setup
```

- `--runtime <provider>` / `--im <provider>` (both repeatable) require specific implementations.
  Without them, one ready Agent Runtime and one ready messaging CLI are enough, so a Codex machine
  is not told to install Claude Code.
- `--json` prints the report on stdout — including when checks fail, so a caller that parses it does
  not have to merge two streams.
- Fixes are printed, never executed: installing software and signing in stay the operator's
  decisions.

## Reuse, not a second implementation

The point of this command is that its answer matches the daemon's. So the probing is shared:

- `resolveAgentRuntimeProviders` (new export) is the provider composition `createClientRuntime`
  already performed inline — same homes, same sanitized environments, same factories. Doctor passes
  `ensureProviderHomes: false`, because observing a Computer must not create the provider homes it
  is reporting on.
- `probeImCliReadiness` (new export) is the probe half of `refreshImCliReadiness`, which previously
  welded probing to writing into a `RuntimeConnection`. The daemon path is unchanged; doctor no
  longer needs to fake a connection.
- Status vocabulary comes from the existing `providerReadiness` mapping, so `install` / `sign-in`
  mean exactly what the Server sees.
- The probe deadline matches the Client Runtime's, so doctor cannot call a provider ready on
  evidence the daemon would have thrown away.

[...]

Issue #244, opened 2026-08-29
Service managers are executed by bare name, so the caller's PATH decides what OpenTag believes about its own daemon


liuchao-001:
## The defect

`defaultServiceRunner` executes whatever program name it is handed, and the two backends hand it bare
names: `options.launchctl ?? "launchctl"` (`service/launchd.ts:94`) and
`options.systemctl ?? "systemctl"` (`service/systemd.ts:78`). `execFile` with a bare name and the
inherited environment resolves through the **invoking process's `PATH`**.

So an executable placed earlier on the caller's `PATH` decides what OpenTag believes about its own
daemon. Demonstrated with the exact call shape the runner uses:

```
$ PATH="$fake:$PATH" node -e 'execFile("launchctl", ["print", "gui/501/opentag-dev"], …)'
resolved to: FAKE-LAUNCHCTL-WAS-CALLED args=print gui/501/opentag-dev
```

Independent QA reproduced the user-visible consequence on a build that consumed this: with a plist on
disk, **nothing bootstrapped into launchd**, and a `launchctl` on `PATH` reporting `state = running`,
the service was reported `active`. Nothing was loaded in launchd in either run.

The service state this produces gates `daemon status`, and every mutation path — `install`, `start`,
`stop`, `restart`, `uninstall` — routes through the same runner. `loginctl enable-linger` does too.

## Why this is filed separately

It was fixed on the branch for #235, which is now closed without merge. The fix is independent of
that PR's command contract: it concerns how the daemon service module executes system programs, not
what any diagnostic reports. Closing #235 leaves the defect in `main`.

## The fix that worked

Resolve a service manager to a known absolute location at the point where a name becomes a process —
inside `defaultServiceRunner`, not in the backends. Putting it there covers every service command and
leaves injected runners (all the existing tests) untouched.

Two details that cost a review round each, worth carrying over:

- **The location list is a supported-platform contract, not an implementation detail.** An FHS-only
  list breaks NixOS, where root-managed systemd lives at `/run/current-system/systemd/bin` — a hard
  break, not a degradation, since nothing resolves. `loginctl` needs the same location for the same
  reason.
- **A program the runner does not govern must say so**, distinctly from a governed program this host
  does not have. Conflating them is how a missing `loginctl` entry silently disabled `enable-linger`
  on every Linux install.

The environment should *not* be stripped: `systemctl --user` needs the session variables that address
[...]

Issue #239, opened 2026-08-28
Service definition lookup honours $HOME while the probed identity comes from the account record


liuchao-001:
## The asymmetry

`$HOME` decides **which service definition is read**, while the environment reconstructed *from*
that definition now comes from the operating system's account record.

The service manager locates a launchd plist through `os.homedir()`, which honours the `HOME`
environment variable, so a shell exporting `HOME=/tmp/elsewhere` sends the lookup to
`/tmp/elsewhere/Library/LaunchAgents/…`. The probes that follow, however, run as the account record
says. Observed in independent QA of first-tree-ai/opentag#235: `doctor` read a definition under a
throwaway `HOME` while probing as the real account.

## Why it is not urgent

It fails closed in practice. A shell with an unusual `HOME` finds no definition where it looks and
lands on the `not installed` blocking check rather than a green verdict. And anyone who can set your
environment can already do worse than redirect a lookup.

## Why it is still worth fixing

The two halves currently disagree about whose service they describe: one half answers for the
account, the other for whatever `HOME` happened to say. Sourcing the definition lookup from the same
account record would make them agree, and would make the daemon's own service commands consistent on
that point too.

This is a service-module question rather than a `doctor` one — `createDaemonServiceManager` and the
launchd/systemd backends own the lookup, and `daemon install` / `status` / `uninstall` share it. That
is why it is not being changed inside the diagnostic that surfaced it.

## Credit

Found by independent QA on #235 while isolating a separate `HOME` defect.

liuchao-001:
The same asymmetry has a second variable: `XDG_CONFIG_HOME`.

`createDaemonServiceManager` passes it through to the systemd backend (`service/index.ts:198`), where it selects the directory the unit file is read from — exactly as `$HOME` selects the launchd plist. So on Linux the invoking shell chooses *which* service definition is described, while the environment reconstructed from that definition comes from the account record.

Same fail-closed behaviour (a shell pointing elsewhere finds no unit and lands on `not installed`), same fix direction: source the definition lookup from the same account authority, for both platforms.

Issue #237, opened 2026-08-28
Agent CLI discovery is PATH-only: a working codex inside ChatGPT.app is reported as not installed


liuchao-001:
## Problem

The Client resolves every external binary it depends on through a single helper — `resolveExecutable` (`packages/client/src/runtime/client-runtime-composition.ts:702`). It walks `PATH` only, and returns the first hit without verifying it can actually run. Four binaries go through it: `codex` and `claude` (both provider factories) and `lark-cli` / `slack` (`refreshImCliReadiness`, `client-runtime-composition.ts:529`).

The daemon's `PATH` is a snapshot taken at service-install time (`buildServicePath`, `apps/cli/src/core/daemon/service/shared.ts:104`; base is `/usr/local/bin`, `/opt/homebrew/bin`, `/usr/bin`, `/bin`), baked into the launchd plist / systemd unit. It never sources the user's shell rc files. So three separate user-visible failures are the same bug:

1. **ChatGPT.app already ships a working `codex`.** `/Applications/ChatGPT.app/Contents/Resources/codex` is a complete codex (verified 0.150.0-alpha.8 on a real macOS machine), supports `app-server`, and reads the same `~/.codex/auth.json` — so it is already signed in. It is just not on `PATH`. A user who installed only the ChatGPT app is told to `install` something they already have and to sign in to an account they are already signed in to. Zero-action setup reported as a two-action failure.
2. **`claude` installed by the official installer lands in `~/.local/bin`**, which the frozen service `PATH` does not contain. Same false `install`.
3. **`lark-cli` installed under an npm-global or version-manager prefix is invisible**, and `handoffReady` requires the messaging CLI to be ready — so onboarding dead-ends silently.

There is also a permanent-shadowing failure. Because resolution is first-hit-wins with no launch verification, one too-old `codex` earlier on `PATH` permanently blocks a newer, working binary the user also has: the probe requires app-server plus experimental dynamic tools and reports `version_incompatible` (`packages/client/src/providers/codex/agent-runtime.ts:858`), and we never look at the second candidate. Observed in practice: uninstalling a stale standalone codex 0.133.0 by hand was what let the runtime reach the ChatGPT.app-bundled 0.144.2.

## Reference implementation

First Tree solves exactly this, and its solution is worth porting rather than reinventing. Read from `first-tree` `main` @ `2851a9e02`:

Discovery is layered, cheapest first:

1. explicit override (env / SDK option)
2. the daemon's own `PATH`
[...]

PR #246, opened 2026-08-29
feat(cli): report local Agent Runtime installations in doctor

### LINKED ISSUE #236: Claude Code sign-in is invisible because the Client Runtime always sets CLAUDE_CONFIG_DIR

yuezengwu:
## Summary

- redefine `opentag doctor` around this Computer local enrollment state: remove `--server-url` and do not read `OPENTAG_SERVER_URL`
- check every distinct Server recorded in machine credentials and fail closed when the Computer has no enrollment
- report install-only Codex and Claude Code CLI discovery without launching either provider
- share the cheap discovery path with the actual Agent Runtime factories: service PATH, First Tree compatible well-known directories, and macOS ChatGPT or Codex app resources
- prevent automatic macOS discovery from entering TCC-protected roots, including through symlinks

## Contract boundary

A doctor pass requires a healthy enrolled Server and at least one supported Agent Runtime CLI. A missing alternate Runtime is informational.

This PR deliberately does not check Agent Runtime authentication or Integration CLI availability, and the report states both omissions explicitly. It does not change daemon to Server readiness reporting, `handoffReady`, or any Server schema.

This replaces the broader contract proposed in closed PR #235; no commit from that branch is merged.

## Follow-ups

- #247 defines the deferred Agent Runtime authentication diagnostic contract
- #248 defines the deferred Integration CLI diagnostic and interaction contract
- #237 remains open for login-shell discovery and per-candidate runtime verification
- #239 remains an independent service-definition lookup issue
- #244 remains an independent service-manager executable authority issue
- Claude Code default credential lookup is handled separately in the PR that closes #236

## Validation

- `pnpm check`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test`
- `pnpm --filter @opentag/server test:integration` (222 passed)
- source CLI tarball smoke via `scripts/cli-pack-smoke.mjs`
- `git diff --check`

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -78,11 +78,9 @@ The server listens on `http://127.0.0.1:8000` by default. In another terminal, r
 pnpm --filter open-tag start doctor
 ```
 
-Use a different server URL with `--server-url` or `OPENTAG_SERVER_URL`:
-
-```bash
-pnpm --filter open-tag start doctor --server-url http://127.0.0.1:9000
-```
+`doctor` checks the Server URL recorded by this Computer's enrollment; it does not accept a caller-selected Server URL.
+It also reports whether a supported local Agent Runtime CLI is installed. This install-only check does not inspect Agent
+Runtime authentication or Integration CLI availability.
 
 ## Local PostgreSQL
 
@@ -379,7 +377,7 @@ processes.
 | --- | --- | --- |
 | `OPENTAG_HOST` | `127.0.0.1` | Server listen host |
 | `OPENTAG_PORT` | `8000` | Server listen port |
-| `OPENTAG_SERVER_URL` | `http://127.0.0.1:8000` | CLI doctor target |
+| `OPENTAG_SERVER_URL` | `http://127.0.0.1:8000` | Optional Server target override for login and Computer connect |
 | `OPENTAG_PUBLIC_URL` | none | Required public Server origin used for browser callbacks and generated connect commands |
 | `OPENTAG_ENV` | `dev` | OpenTag environment/channel: `dev`, `staging`, or `prod`; hosted values require HTTPS |
 | `OPENTAG_DATABASE_URL` | none | Required PostgreSQL connection URL |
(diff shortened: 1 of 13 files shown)


yuezengwu:
Superseded by #249, which implements the reviewed Doctor P0 specification, adds strict local binding and daemon-service checks, and includes the full validation matrix.

Issue #248, opened 2026-08-29
Define Integration CLI diagnostics for opentag doctor


yuezengwu:
## Context

PR #246 intentionally does not check Integration CLI availability. The doctor report says this explicitly.

The previous PR #235 treated Integration CLI presence as a blocking handoff-readiness requirement, but the expected interaction model is not defined and that contract was rejected.

## Decisions required before implementation

- Which Integration CLIs belong in doctor and how they map to configured messaging integrations.
- Whether doctor checks file presence only, executable/version compatibility, authentication, or an end-to-end integration operation.
- Which environment is authoritative: the installed daemon service environment, the invoking shell, or both as separate diagnostics.
- Whether a missing Integration CLI is blocking, informational, or conditional on an enabled integration.
- What remediation doctor may print without installing software or acquiring credentials.

## Boundary

This issue is a follow-up and does not block #246. Do not add Integration CLI probing until the interaction and readiness contract is reviewed and accepted.

yuezengwu:
Owner decision (2026-09-01): Integration CLI diagnostics must use the same contract as other third-party CLI checks in `opentag doctor`.

- Check only whether the CLI is installed locally and report the resolved installation/executable path.
- Do not check version compatibility, authentication state, credentials, installation configuration, network availability, or an end-to-end provider operation.
- Do not install software, repair configuration, or acquire credentials.
- Lark CLI and Slack CLI runtime credentials are managed and injected by WinTAC; that lifecycle is outside `doctor`.
- Do not introduce a separate Integration-specific readiness or blocking contract; follow the existing third-party CLI reporting behavior.

This settles the product scope for implementation.

Issue #247, opened 2026-08-29
Add a non-persistent Web Agent Runtime test


yuezengwu:
## Decision

Add an explicitly triggered, non-persistent **Agent Runtime test** to OpenTag Web.

The existing `opentag doctor` contract remains install/discovery-only. It must not gain Provider authentication, credential-store, API-status, or real-model probes. There is no CLI entry for this capability.

## User flow

- The Agent runtime configuration surface in Web exposes a **Test runtime** action.
- The action warns that it makes a real Provider request, which may consume quota and may be retained or billed by the Provider.
- The browser sends only the expected Agent/runtime-configuration revisions. The Server derives Provider, model, reasoning, Computer, and other runtime settings from the saved configuration; the browser cannot supply an arbitrary prompt or override runtime settings.
- While a test is pending, the action is disabled. There is no automatic execution, refetch, or retry.
- Web displays only `passed` or a fixed, sanitized failure result. A pass proves only that this exact invocation succeeded in the selected Computer environment with the configuration used at that moment; it is not a durable authentication/readiness claim.
- The displayed result is local component state and is cleared on a new test, configuration change, or unmount.

## Non-persistence boundary

OpenTag must not persist or cache the completed result.

The test must not create or update:

- Session, Turn, Delivery, Trace, Usage, Agent Runtime binding, or Provider-readiness records;
- database tables, columns, migrations, or other durable state;
- React Query caches, browser storage, or Server completed-result caches;
- logs or traces containing the result, model output, sentinel, usage, binding data, or Provider diagnostics.

The Server may hold only an in-flight `requestId -> Promise` entry with a bounded TTL. It must remove that entry on success, failure, timeout, cancellation, browser disconnect, and shutdown.

This boundary applies to OpenTag. A third-party Provider may still retain or bill its own request; OpenTag cannot guarantee otherwise.

## Runtime contract

- Web calls a synchronous `POST /api/v1/agents/:agentId/runtime-test` endpoint.
- The Server validates authorization and expected revisions, resolves the saved runtime configuration, and dispatches the request to the target Computer daemon through a negotiated Runtime v2 capability (for example `runtime.agentRuntimeTest@1`).
- The request/response protocol supports test, cancel, and result frames. Only one test may be in flight per Computer, and the Server enforces a bounded global pending limit.
- The daemon uses a standalone availability tester, not `SessionRuntimeManager` or `SessionBindingStore`.
[...]

Issue #240, opened 2026-08-28
Verify the Linux service-manager contract on a real systemd host before release


liuchao-001:
## What is unverified

The daemon service resolves `systemctl`, `loginctl` and `launchctl` from absolute, system-owned
locations rather than the caller's `PATH` (first-tree-ai/opentag#235). That list is now the
supported-platform contract: a layout missing from it does not degrade, it fails to resolve at all.

Three parts of that contract have never been executed by anyone, across eleven heads of review:

1. **NixOS.** `/run/current-system/systemd/bin/{systemctl,loginctl}` is in the candidate list on the
   strength of nixpkgs invoking that path in 21 files plus the system-owned rule — not on the
   strength of a run. Nobody involved has a NixOS host. If it is wrong, `daemon status` / `install` /
   `start` / `stop` / `restart` / `uninstall` and `computer connect` all fail there.
2. **`daemon install` on a live systemd session.** Whether lingering is actually enabled end to end
   is derived from reading `systemd.ts`, not observed. A container cannot supply a user session with
   systemd as PID 1.
3. **The `XDG_RUNTIME_DIR` branch** of the diagnostic's manager-variable reconstruction. Unit tests
   only; never executed.

The FHS half *is* verified: `/usr/bin/{systemctl,loginctl}` confirmed on a Debian container, with
`/bin` a usrmerge symlink so the second candidates are equivalent.

## The checks that would settle it

On a real systemd host:

```
opentag daemon install
loginctl show-user "$USER" --property=Linger      # expect Linger=yes
opentag doctor                                     # expect the XDG_RUNTIME_DIR branch exercised
```

On NixOS, `opentag daemon status` is enough to prove the candidate resolves.

## Why this is not a merge blocker

Recorded during review of #235 rather than blocking it, because the alternative to merging that
change was not a better-verified state — it was the same code with an FHS-only list that provably
breaks NixOS. This is a release-readiness gap, and it is filed so the green checks on those heads do
not read as coverage they never had.

Raised by independent QA across rounds 9–11.

liuchao-001:
Closing as superseded. This described the unverified platform contract of the absolute-location list added on `feat/doctor-runtime-checks`, and first-tree-ai/opentag#235 was closed without merge — so that list does not exist in `main` and there is no contract here left to verify.

[...]

PR #249, opened 2026-08-29
feat(cli): implement doctor baseline diagnostics

### LINKED ISSUE #237: Agent CLI discovery is PATH-only: a working codex inside ChatGPT.app is reported as not installed

### LINKED ISSUE #244: Service managers are executed by bare name, so the caller's PATH decides what OpenTag believes about its own daemon

yuezengwu:
## Summary

- define the P0 product and technical contract for `opentag doctor`
- diagnose the selected OpenTag Home with typed, fail-closed checks for local Computer configuration, daemon service, enrolled Server health, and Agent Runtime CLI installation artifacts
- remove the doctor-specific Server override: `--server-url` is now a usage error and `OPENTAG_SERVER_URL` is ignored
- keep Agent Runtime authentication, version/protocol compatibility, daemon-environment visibility, Integration CLI readiness, machine-token/WebSocket registration, and end-to-end Turn/handoff delivery explicitly out of P0

This supersedes #246 and replaces the broader contract from closed #235. The source of truth is `docs/design/doctor-product-spec.md`.

### Local configuration

- read `computer.json` and `computer-credentials.json` without creating or rewriting either file
- reject the entire credential set when any stored enrollment is malformed or duplicated; do not reuse the runtime reader's tolerant partial projection
- require every enrollment to match the local Computer ID and one canonical Server origin
- permit multiple Workspace Computer enrollments only when they refer to the same Computer and Server
- skip Server health rather than choosing an arbitrary origin when local authority is invalid

### Daemon service

- reuse the existing daemon service manager `status()` boundary
- pass only for an active, current, same-Home service with a consistent runtime owner
- resolve `launchctl`, `systemctl`, and `loginctl` only from reviewed absolute system locations, including the NixOS systemd path; caller `PATH` cannot shadow the service manager

### Server health

- derive the Server only from the verified local enrollment
- apply a five-second deadline across the request and response-body validation
- preserve separate configuration, timeout/network, HTTP, and invalid-response classifications without printing raw response bodies

### Agent Runtime CLIs

- share one install-only resolver between Client Runtime composition and doctor
- inspect Codex and Claude Code independently without spawning either provider
- require an ordinary executable file and reject directories, broken links, non-executable files, and unsafe automatic candidates
- search caller `PATH`, reviewed well-known directories, and macOS ChatGPT/Codex app resources while applying the protected-root policy before path-following reads
- report the canonical artifact path and its source; at least one supported Runtime is sufficient for the blocking aggregate

## Safety and output

- doctor remains read-only and prints the complete human report to stdout even when it exits 1
- fixed report order: Target, Local configuration, Daemon service, Server, Agent Runtime CLIs, Summary, Not evaluated
- summaries describe only the baseline checks and never claim OpenTag, handoff, authentication, or delivery readiness
- machine tokens and raw Server response bodies are never included in the report

## Validation on exact head

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
  - CLI: 152 tests
  - Client: 487 tests
  - Web: 344 tests
  - 289 tests; 100% statements, branches, functions, and lines
  - 230 tests
- `pnpm test:e2e:doctor-docker`
  - 21/21 Docker black-box scenarios passed: one minimal Linux scenario and 20 core scenarios
- `git diff --check`

The Docker suite builds and invokes the packaged CLI as an external process. It uses isolated containers for PostgreSQL, the production OpenTag Server image, a privileged Debian systemd PID 1 runner with a real user manager, and purpose-built HTTP/TCP fault endpoints. It creates real private files, symlinks, executable artifacts, service definitions, sockets, and deadlines instead of mocking the doctor modules.

Covered scenarios include:

1. Linux service manager unavailable fails closed
2. empty Home remains byte-for-byte untouched
3. malformed identity fails without content disclosure
4. one malformed enrollment invalidates the full credential set
5. multiple Server origins receive zero requests
6. Computer mismatch skips Server health
7. symlinked private config is rejected without target disclosure
8. HTTP 503, invalid health schema, five-second hang, and connection refusal are classified separately
9. a real PostgreSQL-backed OpenTag Server passes its public health check
10. executable, directory, and broken-symlink Runtime artifacts are distinguished
11. caller `PATH` cannot shadow the governed `systemctl`
12. a service installed through `opentag-dev daemon install` under a real systemd user manager produces a passing P0 baseline
13. wrong-Home, drifted, malformed, and stopped systemd services fail closed with the expected classifications

The Runtime-positive scenario uses a real executable filesystem artifact, not an installed Codex or Claude Code product. The fault endpoints are dedicated test servers; only the healthy production-image scenario uses the real OpenTag Server.

## Follow-ups and platform limits

- #239 remains an independent service-definition lookup follow-up and does not block this PR
- #236 remains the separate Claude Code runtime configuration correction
- #247 remains the deferred Agent Runtime authentication contract
- #248 remains the deferred Integration CLI contract
- Debian systemd user-service behavior is now exercised inside Docker Desktop's Linux VM; separately installed Linux distributions and NixOS still require release QA under #240

Closes #237.
Closes #244.

### DIFF
--- a/apps/cli/src/__tests__/daemon-service-shared.test.ts
+++ b/apps/cli/src/__tests__/daemon-service-shared.test.ts
@@ -21,6 +21,7 @@ import {
   quoteSystemdEnvironment,
   quoteSystemdToken,
   resolveCliInvocation,
+  resolveServiceManagerExecutable,
   runRequired,
   writeFileAtomically,
 } from "../core/daemon/service/shared.js";
@@ -100,6 +101,46 @@ describe("daemon service primitives", () => {
     expect(servicePath).not.toContain("relative");
     expect(servicePath.filter((entry) => entry === "/opt/opentag/bin")).toHaveLength(1);
     expect(servicePath.filter((entry) => entry === "/usr/bin")).toHaveLength(1);
+  });
+
+  it("resolves a bare service-manager name only through reviewed absolute locations", async () => {
+    const inspected: string[] = [];
+
+    await expect(
+      resolveServiceManagerExecutable("launchctl", "darwin", {
+        access: async (path) => {
+          inspected.push(`access:${path}`);
+        },
+        stat: async (path) => {
+          inspected.push(`stat:${path}`);
+          return { isFile: () => true };
+        },
+      }),
+    ).resolves.toBe("/bin/launchctl");
+
+    expect(inspected).toEqual(["stat:/bin/launchctl", "access:/bin/launchctl"]);
+    expect(inspected).not.toContain("stat:launchctl");
+  });
+
+  it("supports the governed NixOS systemd manager location", async () => {
+    const inspected: string[] = [];
+
+    await expect(
+      resolveServiceManagerExecutable("systemctl", "linux", {
+        access: async (path) => {
+          inspected.push(`access:${path}`);
+        },
+        stat: async (path) => {
+          inspected.push(`stat:${path}`);
+          return { isFile: () => path === "/run/current-system/systemd/bin/systemctl" };
+        },
+      }),
+    ).resolves.toBe("/run/current-system/systemd/bin/systemctl");
+
+    expect(inspected).toEqual([
+      "stat:/run/current-system/systemd/bin/systemctl",
+      "access:/run/current-system/systemd/bin/systemctl",
+    ]);
   });
 
   it("atomically replaces files and leaves no temporary files", async () => {
--- a/apps/cli/src/__tests__/doctor.test.ts
+++ b/apps/cli/src/__tests__/doctor.test.ts
@@ -1,46 +1,517 @@
-import { ServerHealthConfigurationError, ServerHealthNetworkError } from "@opentag/client";
-import { describe, expect, it, vi } from "vitest";
-import { resolveServerUrl, runDoctor } from "../core/diagnostics/doctor.js";
+import { mkdir, mkdtemp, realpath, rm, symlink, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { ServerHealthTimeoutError } from "@opentag/client";
+import { Command, type CommanderError } from "commander";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { registerDoctorCommand } from "../commands/doctor.js";
+import { channelConfig } from "../core/channel/config.js";
+import {
+  type DoctorCheck,
+  type DoctorOptions,
+  renderDoctorReport,
+  resolveDoctorTarget,
+  runDoctor,
+} from "../core/diagnostics/doctor.js";
 
-describe("resolveServerUrl", () => {
-  it("prefers the command option over the environment", () => {
-    expect(resolveServerUrl("http://cli.example", { OPENTAG_SERVER_URL: "http://env.example" })).toBe(
-      "http://cli.example",
-    );
+const computerId = "85fe9af3-d1c6-472b-b78c-8a7ccf512750";
+const otherComputerId = "1a63a21e-f6c7-4474-91ea-4dabf0566a24";
+const serverUrl = "https://server.example";
+const secretToken = "otmc_NEVER_PRINT_THIS_SECRET";
+const homes: string[] = [];
+
+afterEach(async () => {
+  await Promise.all(homes.splice(0).map((home) => rm(home, { force: true, recursive: true })));
+  vi.restoreAllMocks();
+});
+
+describe("doctor command target", () => {
+  it("does not accept a caller-selected Server URL", async () => {
+    const program = new Command().name("opentag").exitOverride();
+    program.configureOutput({ writeErr: () => undefined, writeOut: () => undefined });
+    registerDoctorCommand(program);
+
+    await expect(
+      program.parseAsync(["node", "opentag", "doctor", "--server-url", "https://other.example"]),
+    ).rejects.toMatchObject({
+      code: "commander.unknownOption",
+    } satisfies Partial<CommanderError>);
   });
 
-  it("uses the environment before the default", () => {
-    expect(resolveServerUrl(undefined, { OPENTAG_SERVER_URL: "http://env.example" })).toBe("http://env.example");
-    expect(resolveServerUrl(undefined, {})).toBe("http://127.0.0.1:8000");
+  it("uses the canonical OPENTAG_HOME and ignores OPENTAG_SERVER_URL", async () => {
+    const home = await createHome();
+    const healthChecker = vi.fn().mockResolvedValue({ service: "opentag-server", status: "ok" } as const);
+    const result = await runHealthyDoctor(home, {
+      env: { OPENTAG_HOME: join(home, ".", "nested", ".."), OPENTAG_SERVER_URL: "https://ignored.example" },
+      healthChecker,
+    });
+
+    expect(result.target).toMatchObject({ home: resolve(home), homeSource: "environment" });
+    expect(result.message).toContain(`OpenTag Home: ${resolve(home)} (environment)`);
+    expect(result.message).not.toContain("ignored.example");
+    expect(healthChecker).toHaveBeenCalledTimes(1);
+    expect(healthChecker).toHaveBeenCalledWith(serverUrl);
+  });
+
+  it("reports an explicitly selected default path as environment-sourced", () => {
+    expect(
+      resolveDoctorTarget({
+        arch: "arm64",
+        channel: "stable",
+        cliVersion: "0.1.0",
+        environment: { OPENTAG_HOME: channelConfig.defaultHome },
+        nodeVersion: "v24.0.0",
+        platform: "darwin",
+      }),
+    ).toMatchObject({ home: resolve(channelConfig.defaultHome), homeSource: "environment" });
+  });
+
+  it("reports the channel default when OPENTAG_HOME is absent", () => {
+    expect(
+      resolveDoctorTarget({
+        arch: "arm64",
+        channel: "stable",
+        cliVersion: "0.1.0",
+        environment: {},
+        nodeVersion: "v24.0.0",
+        platform: "darwin",
+      }),
+    ).toMatchObject({ home: resolve(channelConfig.defaultHome), homeSource: "channel-default" });
   });
 });
 
-describe("runDoctor", () => {
-  it("reports a healthy server", async () => {
-    const healthChecker = vi.fn().mockResolvedValue({ status: "ok", service: "opentag-server" } as const);
-
-    await expect(runDoctor({ serverUrl: "http://server.example", healthChecker })).resolves.toEqual({
-      exitCode: 0,
-      message: "OpenTag server is healthy (opentag-server) at http://server.example",
+describe("doctor local configuration", () => {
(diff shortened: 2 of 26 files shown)


yuezengwu:
## QA：PASS

**目标：** PR #249，head `c522a38a0778fdd071cbe3a32003e04f429057e7`。

**证据：**

- `pnpm test:e2e:doctor-docker`：21/21 通过（最小 Linux 1 项、核心故障矩阵 20 项）。测试以外部进程调用打包后的 CLI，实际启动 PostgreSQL、生产 OpenTag Server、systemd user manager，并通过 `daemon install` 安装真实用户服务。
- 已真实注入配置损坏、符号链接、跨 Server enrollment、Computer 不匹配、服务漂移/停止、HTTP 503、非法响应、五秒挂起、连接拒绝及 Runtime 文件异常。
- Agent Runtime 专项 289 项通过且覆盖率 100%；Server 集成 230 项通过；GitHub CI 6/6 通过。

**限制：** Runtime 成功场景使用真实可执行文件 fixture，并非实际 Codex/Claude Code 产品；故障端点为测试服务；本次验证覆盖 Docker Desktop Debian systemd，不代表 NixOS 或独立 Linux 主机发布验收。

**清理：** 无遗留测试容器、网络或测试镜像标签。

PR #252, opened 2026-08-29
feat(web): adopt Kumo onboarding flow and loading states

bestony:
## Summary
- render onboarding progress with Kumo Flow nodes
- replace code copy actions with Kumo ClipboardText
- use Kumo Loader and button loading states across onboarding, setup, task, usage, messaging, and account flows
- keep accessible labels stable while preserving Kumo transition behavior in tests

## Verification
- pnpm --filter @opentag/web test
- pnpm --filter @opentag/web typecheck
- pnpm --filter @opentag/web build
- pnpm exec biome check on changed files


apps/web/src/computer-setup.tsx
```diff
@@ -214,34 +179,43 @@ function ComputerSetupLifecycle({ onConnected, preview = false, target }: Comput
...
       {bootstrapCommand ? (
         <>
-          <pre>
-            <code ref={commandRef}>{bootstrapCommand}</code>
-          </pre>
-          <div className="connect-command-actions">
-            <Button variant="secondary" onClick={() => void copyCommand(bootstrapCommand)}>
-              {copied ? "Copied" : "Copy command"}
-            </Button>
+          <ClipboardText
```
yuezengwu, line 200: **[P1] Preserve a usable fallback when clipboard permission is denied**

Kumo's `ClipboardText` only logs when `navigator.clipboard.writeText` rejects; its `execCommand` fallback is reached only when the clipboard API is missing. Here the displayed command is a truncated, non-selectable span, so a browser permission denial makes `Copy command` silently do nothing and can block the user from running the expiring bootstrap command. The updated denial test now explicitly asserts that the prior selectable-command/failure hint is absent. Please keep the Kumo presentation but restore a failure-aware path that makes the full command manually copyable and announces the failure, with a regression test for that behavior.

yuezengwu (changes requested):
Reviewed exact head `bf258912f31aed8061aed50f30021f8d0e8ab2d0`. Requesting changes for the onboarding copy regression called out inline. This head is also currently CONFLICTING/DIRTY against `main` (conflicts in `apps/web/src/__tests__/app.test.tsx`, `apps/web/src/router.tsx`, and deleted-vs-modified `apps/web/src/styles.css`), so the integrated fresh head needs a complete re-review after synchronization; evidence from this head does not carry forward.

Validation on a clean detached exact-head worktree: 300 web tests passed, web typecheck passed, web production build passed, changed-file Biome check passed, and `git diff --check` passed. The production build also increases the initial JS bundle from 510.15 kB / 153.40 kB gzip at the merge base to 1,138.68 kB / 353.19 kB gzip before the lazy ECharts chunks; that is not the inline blocker, but please confirm this loading-cost increase is intentional or reduce it before re-requesting review.

PR #243, opened 2026-08-29
feat(web): rebuild the Agent home and list around usage and Tasks

liuchao-001:
## What

Redesigns the Agent surfaces in Web from @liuchao-001's review of `/agents`, `/agents/:id`, `/agents/:id/settings`, and `/agents/:id/settings/messaging`.

Since main adopted the Kumo design system (#252) and deleted the stylesheets this branch styled through, the work is now a merge of main (`0070ec7`, conflicted files taken from main wholesale) plus a re-application of the same decisions on Kumo markup (`89b492d`).

**Agent home (`/agents/:id`)**

- The availability badge beside the name appears only while the Agent is healthy. Every other state is already named — with its recovery exit — by the banner directly beneath it.
- Messaging collapses into one header affordance beside Settings. It always opens messaging settings, so a missing binding (`Connect messaging`) and unreadable evidence (`Messaging status unavailable`) keep their entry point.
- `Current work` and `Messaging` are replaced by **Usage** (Tokens + Tasks, `Last 24 hours / 7 days / 30 days`, `View details`) and **Tasks**, which reads `GET /account/tasks?agentId=…` with cursor paging. Reading the Agent's own Tasks server-side means paging cannot hide older Tasks behind a workspace-wide page.

**Agent list (`/agents`)**

- Drops the page description, the identical `@handle` under every name, and the trailing chevron that duplicated the row link.
- States the usage window once above the cards instead of on each of the two numbers, and puts the numbers before the state.
- Each card carries its messaging channel mark beside the Agent name.

**Settings directory**

- One list in the order a viewer thinks about an Agent — Name, Messaging, Computer, Instructions, Model & reasoning — with `Pause or delete` held apart under **Danger zone**. The three previous group headings are gone.

**Messaging settings**

- The channel section names the state it is in: **Connected channel** with the provider's own mark, the channel identity and its state; **No channel connected** when there is no binding, where it previously also read `Contact channel`. The `Contact` / `How to use` prose is removed.
- Each state offers the one exit it has: reauthorize, reconnect (new — an errored or disabled binding previously forced a disconnect first), or, when the channel is fine and only delivery is not, the blocker the evidence actually names.
- `Trigger rules` becomes **Group chat trigger mode**, applied without a confirmation (widening a trigger is not destructive), with the direct-message row removed. `Mentions only` is renamed `On mention` because both modes receive every message; only the wake-up differs.

**Availability copy**

[...]

PR #261, opened 2026-08-29
refactor(web): replace react-router with TanStack Router

bestony:
## Summary

Replaces React Router v7 with TanStack Router in `apps/web`, and restructures the app so the route
tree is the directory tree.

`src/router.tsx` was a 3,170-line file holding the route table, ~40 page components, the session
contexts, the `useResource` data hook, Agent availability projection and every formatting helper.
It is gone. Routes now live in `src/routes`, pages and domain logic in `src/features`.

- **Route files are thin.** They read params and search, then pass them to a component in
  `src/features` as props. Pages no longer read the router, so tests mount them directly — the stub
  `<Routes>` table in `tasks-page.test.tsx` was deleted outright rather than ported.
- **Links are type-checked.** Destinations that were template strings now come from typed
  descriptors in `features/agents/agent-routes.ts`. History state carries the Agent to return to
  rather than a path, so a renamed route fails to compile instead of dying at runtime.
- **The three gates keep their current behaviour.** They stay render-time components because the app
  has no request cache; an async `beforeLoad` would re-request `/api/v1/me` on every navigation.
  `WorkspaceSetupGate` was mounted twice and sniffed `location.pathname` to pick a direction; it is
  now two directional guards, one on `onboarding`, one on `_shell`.

Nothing outside `apps/web` changes except the `AGENTS.md` entry. Read commit by commit — the second
is a pure code move, the third is the library swap.

## Two router behaviours worth knowing

`Navigate` compares its props by reference, and JSX rebuilds that object every render, so it
re-navigates on every pass. Since the outgoing route stays mounted while the next one loads, a gate
that redirects from its own render never stops. `features/navigation/redirect.tsx` keys the effect on
the serialized destination instead.

For the same reason, the sign-in gate is still mounted while its own redirect is in flight. Reading
the live location there folded the sign-in URL into its own `next` parameter —
`/login?next=/login?next=/login…` — until the heap was exhausted. The requested page is now captured
once, at mount.

## The route generator runs for dev and build, not tests

It runs Babel, and this repository pins `@babel/parser` and `@babel/types` to `8.0.0-rc.2` while
`@babel/core` stays on 7.x. Babel 8 renamed `TSFunctionType.parameters`, so the 7.x generator hits
`undefined.length` on any file carrying a function type.

A test run imports the committed `src/routeTree.gen.ts` and has no reason to regenerate it, so the
plugin is registered only when `VITEST` is unset. That keeps this PR from widening a Babel pin across
the whole repository. Verified that `vite build` regenerates the tree byte-identically to the
committed one, and CI builds before it tests.

If you would rather fix the skew properly, adding `@babel/generator` to the existing overrides also
works — but it affects every workspace and emits ~45 deprecation warnings per web test run, so it
seemed the wrong trade for a web-only change.

## Not in this change

`autoCodeSplitting` is off so the bundle stays comparable and the diff reviewable. TanStack Query and
route loaders are deliberately excluded. The four alias redirects (`/`, `/resources`, `/usage`,
`/agents/computers`) stayed render-time components rather than moving to `beforeLoad`: doing so would
change the `?next=` captured for an unauthenticated deep link.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck` — pass
- `pnpm test` — 1,320 tests pass across all five workspaces (`apps/web` 301/301)
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — passes, 100% gate intact
- `apps/web/dist/index.html` keeps `<title>OpenTag</title>` and assets stay under `dist/assets/`,
  matching what the `container-smoke` CI job asserts

The 102 `render(<App />)` history-driven assertions in `app.test.tsx` were kept as-is and all pass.
`window.scrollTo` is stubbed in the test setup because the router scrolls on navigation and jsdom
reports that as an error on every route change.

## Breaking changes

None for users. For contributors, `apps/web` no longer has `react-router-dom`; see the conventions
added to `AGENTS.md`.

## Checklist

> Note: these commits are unsigned — the 1Password signing agent stopped authorizing
> non-interactive signatures partway through. Happy to re-sign before merge.


bestony:
Rebased onto `main` after #234 landed.

That PR added `/internal/onboarding-v2` to the `router.tsx` this branch deletes, so the route is ported to `src/routes/internal.onboarding-v2.tsx` — a top-level file, outside every gate, which is where #234 deliberately put it. Its comment about reaching no Server is carried over verbatim.

`apps/web` is now at 366 tests (301 + the 65 #234 brought), all passing against the new router. Full repo: 1,385 passing. `pnpm check`, `pnpm build` and `pnpm typecheck` are clean.

Heads-up for #243: it also touches the old `router.tsx`, so it will need the same treatment — its route moves into `src/routes` and its page component into `src/features`.

yuezengwu (approved):
Reviewed exact head 181b03603638b17f526e2fce58f058434aaccc02. The route tree preserves the existing authentication, Workspace-authority, setup, alias, legacy, and not-found boundaries; typed Agent destinations and history state remove path-string coupling without moving Server authority into the router. Verified pnpm check, full pnpm build, Web typecheck, Web 366/366 tests, and full repository 1,385/1,385 tests. The production main bundle changes from 1,179,005 to 1,219,738 bytes (gzip 368,625 to 381,888; about +3.6%), a bounded router-runtime tradeoff with auto code splitting intentionally out of scope. Build regenerates the committed route tree byte-identically.

Issue #231, opened 2026-08-28
UX: never expose choosing between Computers on one Account


Gandy2025:
## The rule

One Account has one Computer. If that machine breaks, the Account **repairs** it — it never creates a
second one. An Agent never moves between Computers.

From the user's side the concept should not surface at all: no list of machines, no "which Computer",
no "add another", no distinction between connecting and reconnecting. Ideally the word *Computer*
disappears from the product. The user should only ever see whether their Agent is running, and one
command that brings it back.

This is filed against the ownership model (#227) rather than as a web task, because the web layer
alone cannot deliver it — see "Why this is not a web-only change" below.

## What contradicts it in the UI today

- `apps/web/src/agent-creation/agent-creation-flow.tsx:522` — the Computer `Change` button renders
  unconditionally, so a single-Computer Account still gets a control that opens a picker holding one
  option. The sibling Runtime control at `:550` *is* count-guarded (`providerOptions.length > 0`);
  the Computer one is not.
- `apps/web/src/agent-creation/agent-creation-flow.tsx:597` — `Connect another Computer`, nested
  inside that picker, invites a second enrollment.
- Vocabulary collision: the button says `Change`, the panel it opens is titled `Choose Computer`
  (`:565`), and the Agent detail page says `Reconnect this Computer`
  (`apps/web/src/router.tsx:2037`). Three verbs for what should be one action.

The standalone Computers page, its route, and its account-menu entry have already been removed on
`feat/account-scoped-computers` (`e5441c6`).

## Why this is not a web-only change

`computerId` is a random UUID persisted per `OPENTAG_HOME` directory, not a hardware fingerprint:

```ts
// packages/client/src/runtime/computer-identity.ts:31
const identity: ComputerIdentity = { version: 2, computerId: randomUUID(), serverUrl };
```

So deleting `~/.opentag`, reinstalling the OS, or moving to a replacement laptop all produce a *new*
identity. `exchangeConnectCode` looks the enrollment up by `(workspaceId, computerId)` and **inserts
a new `workspace_computers` row** when it does not match
(`packages/server/src/services/computers/machine-auth-service.ts:137-166`).

Consequences today:

1. "Repairing" a machine silently produces a **second** enrollment. The user did not choose to add a
   Computer; the system added one for them.
2. The old row cannot be cleaned up — no public API writes `workspace_computers.revoked_at` (only the
   internal onboarding-lab reset does).
[...]

PR #262, opened 2026-08-29
feat(web): make the redesigned flow the onboarding flow, on the real Server and on Kumo

liuchao-001:
## What

The redesigned flow becomes **the onboarding flow**, against the real Server, written in Kumo.

`/onboarding` renders it. The route's only remaining job is to mark setup complete once the flow reports it finished. The in-page mock keeps a route of its own at `/internal/onboarding-v2`, outside every gate, because reviewing pages needs no Account.

## One seam, two implementations

`backend.ts` states what a page observes: a command that can expire, a Computer that shows up, a readiness result, a messaging app that connects. The mock and the Server both satisfy it, so no page holds a branch for "real" versus "review".

## What the Server's contracts decided

@yuezengwu's review found three faults that were invisible from inside the client, because they are properties of the Server rather than of this state machine. All three are fixed.

**A reinstalled machine is repaired, not replaced.** Connect codes were being issued with no target, which the Server reads as *create* — so recovery enrolled a *second* Computer beside the one it replaced, and this flow then moved the Agent onto it. An Account would collect a duplicate on every reinstall, and a UI change would be deciding ownership. Recovery now issues a **repair** code naming the Agent's own Computer, which is what #253 delivered for this case.

That also removes the worst use of the arrival heuristic. Which Computer arrived is still guessed on a **first** run — the response carries no correlation to the issued code, as was true before this flow existed and remains true in `computer-setup.tsx` — but it no longer decides who owns what. For recovery there is nothing to guess: the Server repairs the Computer the code named.

**An installed messaging app is not a reachable Agent.** `completeSetup` requires `handoffReady`: an active binding **and** a ready runtime **and** a ready provider CLI **and** an actual observation of the messaging identity. Slack's OAuth marks the binding active before that observation lands, so treating "installed" as "finished" asked for something the Server refuses — and after three refusals the reader was stranded on an error with a working Agent and no way forward but a reload. The flow now polls for that observation and says so while it waits.

⚠️ The tests are why this survived: they mocked an active binding and a successful completion — the contract as I had misread it, not as the Server states it. A green suite asserted my misreading back to me.

## Earlier rounds

@code-reviewer found and reproduced 25 defects across eleven rounds; each fix is guarded by their reproduction, and the guards themselves were checked by mutation rather than trusted for being green. The larger ones: a poll returning after its code expired could adopt a Computer; readiness carried no record of which runtime produced it; the messaging CLI was read at position 0 of a list the Server orders by its own canon; a refused Lark attempt retried without limit; the connection was checked on arrival and never again.

## Nothing moves under the reader

- The connect code breaks **strictly by character**; `overflow-wrap: anywhere` still prefers an existing break opportunity, so codes ending in `-` or `_` took an extra line. Measured over 400 random codes per width: identical height at 1000, 640, 390 and 320px, and `sh` never splits.
- Across **five readiness outcomes at eight widths from 1280 to 320px**, the footer holds one position as the check settles.

## Scope

The Tailwind compiler is **no longer here** — it is #263, which is blocked on an author-scope gate this PR cannot satisfy and on visual acceptance of pages this change never touches. The measurements it enabled are back in this surface's own stylesheet, and the layout guarantees above were re-measured after that move. This PR no longer depends on #263.

## Verification

`pnpm check`, typecheck and build clean; web tests **424 passed**.

⚠️ **No live run.** There is no OpenTag Server or session on the machine this was written on, so every Server interaction is proven only against stubs — which is exactly how the handoff contract was got wrong for eleven rounds. @code-reviewer runs the staging round after merge: real Slack round trip, a real daemon going offline, the repair path, and Computer attribution in a Workspace with a second machine.

## Worth a reviewer's eye

1. **First-run Computer attribution is still a heuristic.** It predates this flow and is unchanged by it, but this flow makes it load-bearing for every new Account. Closing it needs a Server-side correlation.
2. **The old onboarding still exists**, because the staging Onboarding Lab renders it verbatim. Retiring it means rebuilding the Lab against this surface.
3. **The connect command comes from the Server**, which still emits `npm i -g` rather than the portable installer this flow was designed around. That is a Server change and should wait until someone confirms the installer is published.
4. **Cloud is unimplemented on the Server backend.** The route is disabled and unreachable in production.


liuchao-001:
Second review round from @code-reviewer found five more, four of them consequences of this flow becoming the setup gate's only exit. All fixed at `6b6920a`; their probes are in the branch and green.

**The one that mattered most: a Slack user could never finish onboarding.** Slack's callback lands on a gated Agent route, the gate sends it back here, and a page starting from an empty draft has no Agent to complete with — so setup was never marked done, the gate kept sending them back, and the second run collided with the first on name uniqueness. The page now **reads the Account's Agents and messaging binding on arrival** and picks up where the Server says it is. That covers a refresh, a woken laptop and a closed tab too — re-entry is ordinary here, not exceptional, and the old page had this and I dropped it.

The rest: a refused `completeSetup` was permanent (claim taken before the call, never released); a refused Lark attempt had no way back, so the panel showed "Waiting for you to scan…" over an empty box while nothing was waiting; the readiness poll erased errors it had not set, wiping a creation failure's explanation within 1.5s; and `runtimeProvider` was written but never read — it now guards the gate it was added for.

The Computer adoption also moved out of the `setConnect` updater. It was idempotent and harmless, but an updater that writes is a trap for whoever edits it next.

---

**The Tailwind compiler is no longer in this PR.** It moved to #263, on the reviewer's recommendation and my agreement: it needs a *visual* review of eight pages this branch does not otherwise touch, and it should be revertable on its own. That PR carries the measured impact — **+7.8 KB and 32 selectors that match something for the first time**, with the page-by-page table.

⚠️ **This PR now depends on #263.** Its diff still shows the compiler because it was branched before the split; once #263 merges, that part disappears from here. Merging this without #263 would leave the reserved heights and the frame width inert — which is exactly the failure mode #263 exists to fix.

**Still no live evidence.** The Slack round trip in particular is reasoned from the Server's `slack-oauth.ts` redirect and the gate in `router.tsx`, and reproduced only as a remount. It needs a real run on staging.

yuezengwu (changes requested):
Requesting changes on exact head `5404b15a85574a4750685d40e5a708fa4c959d77`.

1. **[P1] Offline-Agent recovery creates a second logical Computer instead of repairing the bound one.** `issue()` always calls `issueComputerConnectCode()` with an empty body (`server-backend.ts:287`; `api.ts:241-245`), which the current Server contract interprets as `mode=create`. When the new subject appears, this flow then calls `rebindAgentComputer` (`server-backend.ts:342-349`). That turns a reinstall/recovery into a new Account Computer plus an Agent move, leaving the old Computer behind and bypassing the explicit stable-ID repair contract delivered by #253 / described by #231. Recovery must issue `mode=repair` with the Agent-bound `targetComputerId`; choosing create+rebind is an ownership/lifecycle architecture change and cannot be introduced from this UI PR.

2. **[P1] The Computer arrival evidence is not bound to the issued code.** `findArrival()` accepts the first Account Computer that was absent from the baseline or whose `connectedAt` changed (`server-backend.ts:44-53`). A different already-owned Computer reconnecting while this code is live therefore satisfies this run; the resumed-Agent path can then rebind the Agent to that unrelated Computer. The branch comment already acknowledges that the current response has no correlation. Do not make this heuristic a load-bearing onboarding/rebind gate: use an exact Server-provided subject/correlation, or the exact repair target for recovery, before adopting or mutating ownership.

3. **[P1] Slack callback re-entry treats installation as handoff readiness and races the real Server gate.** Re-entry reads only `imBinding()` and sets messaging connected as soon as `bindingState === active` (`server-backend.ts:208-212`). The actual `accountSetupComplete` service requires `handoffReady`, which additionally needs current runtime readiness, provider-CLI readiness and a live Slack identity observation. OAuth marks the binding active before that observation is guaranteed. The page then sends three immediate completion attempts (`page.tsx:198-214`) rather than polling `imBindingHandoff()`, so a normal real callback can exhaust all retries and strand the Account on the error state until a manual reload. The existing Slack test mocks only an active summary and a successful `onComplete`; it does not reproduce this Server contract. Gate completion on fresh `handoffReady=true` and cover delayed Slack observation/re-entry.

4. **Scope/dependency blocker:** the current diff still contains the Tailwind compiler/build-architecture work from #263 (package dependencies, Vite plugin, global CSS activation and lockfile), and #263 is still open and is not an ancestor of this head. That is outside the allowed non-engineer PR scope and lacks the stated real visual/staging acceptance. Merge and accept #263 separately through the required human review, then rebase this PR so that build architecture is no longer part of #262; the resulting fresh head needs a complete re-review.

Validation: full body/comments/reviews/checks/diff and related #227/#231/#250/#251/#253/#263 inspected; Context Tree verified at `4f76b1474848c515cef9c1bc7ce42c5558395e89`; `pnpm --filter @opentag/shared build`, Web tests (431/431), Web typecheck, Web build, `pnpm check`, and `git diff --check` all pass. All six live GitHub checks pass, but the PR explicitly has no live Slack/staging evidence. Passing mocked tests do not resolve the contract and scope blockers above.

liuchao-001:
@yuezengwu — all three blockers addressed at `cabaf79`, plus the scope one. Summarising the diff since the head you reviewed so the re-read is cheap.

**1 + 2 turned out to be one fix.** Recovery now issues a **repair** code naming the Agent's own Computer, so the Server repairs that exact machine. Two consequences:

- No second Computer. A reinstall no longer leaves the Account a duplicate beside the machine it replaced.
- **The move mechanism is deleted** — `rebindAgentComputer` is gone from the client, along with its gate, its retry surface, its copy and the tests that guarded it. With a named target there is nothing to infer from an arrival, so the heuristic no longer decides ownership at all.

The heuristic remains for a **first** run, where no Agent exists to mis-assign and no ownership is mutated. That is the pre-existing situation in `computer-setup.tsx`, unchanged by this PR, and I've kept it called out in the description rather than quietly relying on it.

**3.** Completion now polls `imBindingHandoff()` and waits for `handoffReady`, and resume does the same instead of reading `imBinding()`. The step says *"Connected. Checking your agent can be reached…"* while it waits, because that is what is actually happening.

You were right that the existing test did not reproduce the contract — it mocked an active summary and a successful completion, which is the contract as I had misread it. There is now a test that the flow **must not** complete while the Server reports `handoffReady: false`, and that it finishes on the poll after that flips. I verified by mutation that relaxing the gate back to binding state turns it red.

**4.** The Tailwind compiler is out. The measurements it enabled are back in this surface's own stylesheet; the wrapping and footer-stability measurements were re-run after that move and still hold. This PR is no longer an ancestor-or-descendant question with #263 — it stands alone.

424 web tests, `pnpm check`, typecheck and build clean; six checks green on this head. Still no live evidence, and this round is the reason I'd rather say that plainly than imply otherwise: eleven rounds of stubbed review did not find a contract error that reading the Server found immediately.

liuchao-001:
@yuezengwu Follow-up round from @code-reviewer on `cabaf79` found one more consequence of the handoff fix, now addressed at `0bbac07`.

**Gating on `handoffReady` had collapsed two unlike things.** An *active* binding waiting to be observed resolves on its own; provisioning, revoked, errored and disabled do not — but all four reported `handoffReady: false` and so all four were shown the same endless "checking", on a step whose only way to restart is a state they had been moved out of. Choosing a messaging app again did nothing.

The status distinguishes them, so the flow does too: not active means there is an app to connect, and it returns to the step that connects one.

Worth recording that my own test for the previous fix did not catch this, because every case in it mocked `bindingState: "active"` — the same blind spot as the bug it was written for.

Two smaller ones: the wait now names what it is waiting on when the page can know it (a missing messaging CLI is visible in the Computer's readiness, while the handoff status carries no reason — so an Account without it was watching an unexplained spinner), and a ref left behind by the deleted move mechanism is gone.

428 tests, checks green. The staging list has grown accordingly: a broken/`reauthorization_required` binding, and picking Lark without `lark-cli` installed, are both now things only a live run can confirm.

liuchao-001:
## Ported onto TanStack Router — head is now `28892f1`

`#261` landed on main while this branch was in review and deleted `apps/web/src/router.tsx`, the one file this branch touched to decide what `/onboarding` renders. Merged and re-pointed:

- `/onboarding` (`src/routes/_authenticated/_workspace/onboarding.tsx`) renders the redesigned flow. It keeps the only job the route still has — calling `completeSetup` and refreshing the membership when the flow reports it finished — through a `useCallback`-stable callback, so a re-render cannot reopen the flow's bounded retry budget.
- `/internal/onboarding-v2` renders the mock explicitly, which is what that gate-free route was always for.
- The extra authenticated preview route this branch had added is dropped: `/onboarding` *is* the real flow now, so a second route to the same component behind the same gate showed nothing new.

### Re-verified after the merge, not carried over

- `pnpm check` and `pnpm --filter @opentag/web build`: clean.
- Web suite: 430 passed / 32 files.
- Connect-command wrapping, 400 random codes per width against the merged build: identical height at every width — 106.56px @1000, 126.70 @640, 227.41 @390, 307.97 @320, zero split-height cases, no horizontal overflow at 320.
- Footer position, 5 readiness outcomes × 8 widths: one position in all 40.

The three server-contract findings from the previous round are unchanged by this merge; they were fixed in `8dd5d88` and are described in the comment above.

apps/web/src/onboarding-v2/server-backend.ts
```diff
@@ -0,0 +1,616 @@
...
+ */
+function findArrival(
+  computers: readonly WorkspaceComputerSummary[],
+  baseline: ReadonlyMap<string, string | null>,
+): WorkspaceComputerSummary | undefined {
+  return computers.find(
+    (computer) =>
+      computer.connectionStatus === "online" &&
+      computer.connectedAt !== null &&
+      (!baseline.has(computer.computerId) || baseline.get(computer.computerId) !== computer.connectedAt),
```
yuezengwu, line 54: [P1] This is still Account-wide arrival evidence, not evidence that this connect code produced this Computer. A different visible Computer reconnecting during the wait satisfies the predicate; this ID is later passed to `createAgent`, so first-run placement is mutated on the wrong subject. The existing tests explicitly preserve this failure. Please require an exact Server correlation before adopting the Computer; that ownership contract is outside this non-engineer UI PR.

apps/web/src/onboarding-v2/server-backend.ts
```diff
@@ -0,0 +1,616 @@
...
+         * read, not inferred. One extra round trip buys not saying something untrue.
+         */
+        const { computers } = await browserApi.computers();
+        if (!live()) return;
+        const online = new Map(
+          computers.filter((one) => one.connectionStatus === "online").map((one) => [one.computerId, one]),
+        );
+        // Where the Account has more than one, the Agent whose Computer is actually there is the one
+        // this run can finish. Otherwise the first, which is the Server's own order.
+        const existing = active.find((candidate) => online.has(candidate.computer.computerId)) ?? active[0];
```
yuezengwu, line 216: [P1] Online status is not a selector for which Agent this onboarding/callback belongs to. With multiple active Agents this silently picks one (or arbitrary Server order), after which repair, messaging handoff, and setup completion all target it. Preserve an exact Agent target, ask the user to choose, or fail closed when more than one candidate exists.

apps/web/src/onboarding-v2/page.tsx
```diff
@@ -14,41 +14,161 @@ import {
...
+  const [destinationConfirmed, setDestinationConfirmed] = useState(false);
+  const [draftConfirmed, setDraftConfirmed] = useState(false);
+  /*
+   * The confirmations exist so a page is left deliberately rather than the moment its fields
+   * happen to be valid. An Agent that already exists settles both of them: the decisions they
+   * guard were made on a previous visit and cannot be taken back.
+   */
+  const resumed = backend.agent !== undefined;
+  const [cloudComputer, setCloudComputer] = useState<CloudComputerState>("idle");
+  const [messagingProvider, setMessagingProvider] = useState<MessagingProvider>();
```
yuezengwu, line 162: [P1] This provider choice is not restored on remount. A Slack callback commonly returns while `imBindingHandoff` is active but not ready; the backend restores `waiting-handoff`, yet this remains undefined, and `MessagingConnection` renders neither provider branch. The user sees an empty messaging step until readiness changes. Restore the binding provider and cover a delayed Slack callback/remount.

yuezengwu (changes requested):
Requesting changes on exact head `28892f16203552010910fb941533035fbb1e17e7`.

The previous recovery/Slack-handshake/build-scope blockers were rechecked rather than carried over: repair now targets the Agent-bound Computer, rebind is gone, `handoffReady` is polled and broken bindings return to a connectable state, and #263/#264 build architecture is absent from this diff. The TanStack Router port is also within this page/interaction scope.

Three correctness blockers remain:

1. **[P1] First-run arrival is still not correlated to this connect code.** `findArrival()` accepts any newly visible or freshly reconnected online Computer. The branch tests explicitly document that a colleague/other Computer can satisfy the wait. That result is then assigned to `computerId.current`, and `createAgent()` persists the new Agent on it. Contrary to the latest PR comment, this does mutate durable Agent placement on first run. This cannot be made a load-bearing onboarding gate without exact Server evidence tying the issued code to the Computer that answered. Because that discriminator is a Server/ownership contract, it needs an engineer-owned prerequisite rather than a heuristic added in this non-engineer UI PR.

2. **[P1] Re-entry silently chooses the wrong Agent when an Account has multiple active Agents.** `readAccount()` picks the first online Agent, or otherwise `active[0]` in Server order. The old onboarding required an explicit Agent choice when more than one candidate existed; this branch removes both that choice and the route/search target. A Slack callback, refresh, or repair can therefore continue and eventually complete setup for an unrelated Agent. Preserve an exact target across the callback/re-entry, offer an explicit choice, or fail closed when it is ambiguous.

3. **[P1] A real Slack callback can return to an empty messaging step during the normal observation window.** On remount, `readAccount()` can restore `messaging={kind:"waiting-handoff"}`, but `messagingProvider` is page-local state initialized to `undefined` and is never reconstructed. `MessagingConnection` renders only its Slack or Lark branch when `provider` is set, so `handoffReady=false` produces the heading with no install/wait content until readiness flips. The existing callback test covers only an immediately ready handoff and misses this real delayed-observation state. Restore the binding provider on resume and add the delayed Slack callback/remount test.

Validation on this exact head: Context Tree verified at `d0c621c2b75f27edd178dbb27ecd9a270e5f94bf`; `git diff --check`, shared build, Web tests (430/430), Web typecheck, production Web build, `pnpm check`, and full Docker image build pass. The production bundle was exercised through Local → Agent → Computer readiness → Messaging in a browser with no console errors. All six GitHub checks pass. The PR is currently conflicting/dirty against main, so any conflict-resolution head also requires a fresh review.

liuchao-001:
## Merging at `834f019` — how the three blockers were resolved

@yuezengwu and @liuchao-001 settled this offline: the flow's interaction is as designed and does not grow a confirmation step, so the two findings that would have required one are being closed by other means rather than by changing this surface.

**3. Slack callback returning to an empty messaging step — fixed here.** The provider now comes back from the binding on resume, so a return trip during the normal observation window renders the wait it is supposed to render. The read is deliberately its own failure: which app is waiting decides what the step draws, not whether the run can continue, so a failure costs the branch and nothing else rather than stranding a returning Account on an error. Both the restore and that isolation are held by tests verified against their own mutations.

**1. First-run Computer attribution — [#295](https://github.com/first-tree-ai/opentag/issues/295), assigned to @yuezengwu.** You were right that this belongs behind a Server contract. I checked the alternatives rather than assuming: the issue response carries no code id, the exchange result goes to the CLI rather than the browser, the Computers list has no field tying an entry to a code, and `createAgent.computerId` is required, so placement cannot be deferred. Nothing available to the client closes it.

**2. Re-entry choosing the wrong Agent — [#296](https://github.com/first-tree-ai/opentag/issues/296).** This one needs no Server change and no interaction change, so it lands separately as invisible plumbing. This PR keeps the route's `agentId` search contract in place, which is the mechanism it will use.

Both issues state the failure and its blast radius rather than pointing at this PR.

## Also in this range

Two merges with main, which moved underneath this branch three times: the TanStack Router restructure, then the Account-session and TanStack Query refactor that removed `useWorkspace` and dropped `workspaceId` from the Agent types. Main's two shell tests that prove onboarding stays reachable without a Workspace grant still prove exactly that — they now look for this flow's first question.

One defect this branch introduced and fixed: writing the surface in utilities dropped the three custom properties the inline OpenTag mark draws itself with while the shape kept referring to them, which would have rendered it solid black. Nothing renders that brand in this flow today, so it repaired a latent path rather than a visible fault — stated that way because I checked in a browser instead of assuming.

## Verification at this head

`pnpm check`, typecheck and build clean; web tests **536 passed / 38 files**; all six CI checks pass. Connect-command height identical across 400 random codes at each of four widths; footer holds one position across five readiness outcomes at eight widths — both re-measured on this build rather than carried over.

⚠️ **Still no live run.** Every Server interaction here is proven against stubs. @code-reviewer runs the staging round after deploy, against an eight-item list that now includes confirming the page leaves `/onboarding` on its own once setup completes — the one state transition our stubs cannot demonstrate.

PR #263, opened 2026-08-29
build(web): compile this app's own Tailwind utilities

liuchao-001:
## What

This app had **no Tailwind compiler**. It imported Kumo's prebuilt stylesheet, which contains only the utilities Kumo's own components needed, so every class written in this repository outside that set did nothing — silently, with no build warning and no failing test.

`w-[18rem]` and `size-7` in `onboarding/view.tsx` are two that were **already dead on `main`**: the onboarding step cards have no width constraint and their number badges no size.

Theme and utilities only, never preflight — Kumo's bundle already carries a base layer, and a second reset would undo the element styling its components depend on. Kumo's sheet stays unlayered, so where both define the same class Kumo still wins and nothing shared moves.

## This is a behaviour change, and here is its size

Built the app twice, with the plugin on and off, and diffed the emitted CSS:

**143,762 → 151,760 bytes (+7,998 bytes), ~32 selectors that now match something for the first time.** @code-reviewer measured the same build independently and counted 31; the difference is how a combined selector is counted, not what changed.

Two rows of the table below were wrong in an earlier draft and are corrected: the settings row's *column split* was never dead, only its vertical alignment, and the `minmax(...)` track belongs to `onboarding/view.tsx`.

| Where | Newly live | What changes |
| --- | --- | --- |
| `ui/design-system.tsx` | `md:items-center` | **SettingsRow** — its two columns become vertically centred. The column split itself is *not* new: `md:grid-cols-[2fr_1fr]` is already in Kumo's bundle |
| `features/agent-usage.tsx` | `sm:grid-cols-3` `lg:grid-cols-2` `h-60` | Usage page gains its multi-column layout; **the chart container gets a height for the first time** |
| `features/tasks-page.tsx` | `whitespace-pre-wrap` `min-w-56` `align-top` `sm:grid-cols-2` | **Task output keeps its line breaks** instead of collapsing them |
| `im/feishu-setup.tsx` | `size-60` | The Lark QR is constrained to 15rem square |
| `router.tsx` | `max-w-md` `max-w-xl` `top-4` `last:border-b-0` `sm:grid-cols-2` | Panels become fixed-width; a close button moves to `top-4`; the last row's bottom border disappears |
| `runtime-configuration.tsx` | `sm:grid-cols-2` | One column becomes two |
| `agent-creation/agent-creation-flow.tsx` | `grid-cols-[auto_1fr]` | Layout change |
| `internal/onboarding-lab-page.tsx` | `z-30` `w-[min(24rem,…)]` | Overlay stacking and width |
| `onboarding/view.tsx` | `w-[18rem]` `size-7` `md:grid-cols-[minmax(14rem,18rem)_1fr]` | The two already-dead classes above, plus this page's own column split |

**None of these pages has a test asserting its layout.** Every one of them was written expecting these classes to work, so in each case this is the author's intent finally taking effect rather than a new design — but that is a claim worth checking with eyes, which is why this is its own PR.

## Why separate

Found while doing #262, which needs it. Split out on @code-reviewer's recommendation, and I agree: it wants a **visual** review of pages this repository has not otherwise touched, and bundling it with an onboarding rewrite would hide it inside a diff about something else.

⚠️ **Merge this first.** #262 depends on it: without the compiler, that flow's reserved heights and frame width are inert, which is the exact failure this fixes. And note the split does not make this permanently revertable on its own — it is, until #262 lands; after that, reverting here would break onboarding's layout too.

## Verification

`pnpm check`, typecheck and build clean; web tests **366 passed**. The impact table above is measured from build output, not read off the source.

⚠️ **Not visually reviewed.** I have no signed-in session against a running Server, so I could not open Settings, Usage, Tasks or the Lark setup page to look at them. Someone who can should, before this merges.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -20,6 +20,7 @@
     "react-router-dom": "^7.18.2"
   },
   "devDependencies": {
+    "@tailwindcss/vite": "^4.1.17",
     "@testing-library/react": "^16.3.0",
     "@types/qrcode": "^1.5.6",
     "@types/react": "^19.1.11",
@@ -27,6 +28,7 @@
     "@vitejs/plugin-react": "^4.7.0",
     "jsdom": "^26.1.0",
     "postcss": "^8.5.26",
+    "tailwindcss": "^4.1.17",
     "vite": "^7.1.3"
   }
 }
--- a/apps/web/src/app.css
+++ b/apps/web/src/app.css
@@ -1,3 +1,14 @@
+/*
+ * Kumo ships a prebuilt Tailwind bundle, but only of the utilities its own components needed, so
+ * anything this app writes outside that set — every bracketed value among them — used to be inert.
+ * Compiling utilities here from this app's own sources is what makes them real.
+ *
+ * Theme and utilities only, never preflight: Kumo's bundle already carries a base layer, and a
+ * second reset would undo the element styling its components depend on.
+ */
+@import "tailwindcss/theme.css" layer(theme);
+@import "tailwindcss/utilities.css" layer(utilities);
+
 /* OpenTag application boundary. Component styling belongs to Kumo. */
 :root {
   color: var(--text-color-kumo-default);
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -1,9 +1,10 @@
+import tailwind from "@tailwindcss/vite";
 import react from "@vitejs/plugin-react";
 import { defineConfig } from "vite";
 
 export default defineConfig({
   base: "/",
-  plugins: [react()],
+  plugins: [react(), tailwind()],
   server: {
     proxy: {
       "/api": "http://localhost:8000",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25,7 +25,7 @@ importers:
         version: 22.20.1
       '@vitest/coverage-v8':
         specifier: 3.2.7
-        version: 3.2.7(vitest@3.2.7(@types/node@22.20.1)(jsdom@26.1.0)(tsx@4.23.12)(yaml@2.9.0))
+        version: 3.2.7(vitest@3.2.7(@types/node@22.20.1)(jiti@2.7.0)(jsdom@26.1.0)(lightningcss@1.32.0)(tsx@4.23.12)(yaml@2.9.0))
       lefthook:
         specifier: ^2.1.12
         version: 2.1.12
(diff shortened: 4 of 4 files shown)


yuezengwu (changes requested):
Reviewed exact head `f14576f1354604e1ae6c362b5e8797681165f097` under the non-engineer author gate. I am requesting changes for two blockers.

1. **This is technical/build architecture and is outside the allowed scope for a non-engineer-authored PR.** The change adds the Tailwind compiler and Vite plugin, new build dependencies, and a new cascade-layer/source-scanning contract that activates previously inert styles across the application. Splitting it out of #243/#262 was the right shape, but it must now be submitted and owned by an engineer (or an engineer must explicitly assume authorship/ownership); it cannot be approved under the current author.

2. **The claimed behavior boundary is not yet completely evidenced.** Exact parent/head production builds confirm the stated byte delta (`143,762 -> 151,760`, +7,998), but parsing the emitted CSS finds 34 newly emitted selector strings, with live runtime classes beyond the impact table: shared dialog sizing/spacing (`max-h-[min(90vh,42rem)]`, `mb-4`), app-shell padding (`py-5`, `md:px-8`, `md:py-8`), onboarding layout (`max-w-6xl`, `content-start`, `break-all`), and additional modal/route widths. Automatic source detection also emits `.font-bold` solely from its literal in `KUMO.md`/a contract test, demonstrating that tests/docs are part of the implicit scan boundary. There is no result-based artifact check pinning the intended production contract, and the PR explicitly has no visual review for the affected pages/shared components. The engineer-owned replacement needs an intentional scan boundary, a result-based check located outside that boundary if it probes class names, and visual evidence for all affected surfaces and relevant breakpoints. Retaining Kumo's standalone stylesheet does keep Kumo component styles intact; any later move to scanning Kumo dist would be a separate ownership decision.

Validation on this exact head: `git diff --check`; `pnpm check` (passes with pre-existing onboarding warnings/info); `pnpm typecheck`; Web tests (22 files / 366 tests); ordered Shared + Web production build; and a full production `docker build` all pass. GitHub CI is also green. Those results establish buildability, but do not override the author-scope gate or the missing bounded visual/build-contract evidence.

liuchao-001:
Superseded by #282, which landed the Tailwind compilation on main under an approved engineer's authorship. #264 — the later, reviewed version of this same change — is closed for the same reason.

The one piece that did not travel with #282 is tracked as #285: nothing asserts that the generated stylesheet actually contains the utilities. Closing this rather than rebasing it, since the substance is on main and the remaining gap has its own issue.

PR #264, opened 2026-08-29
build(web): generate utilities from source instead of shipping a fixed set

liuchao-001:
## Why

Kumo's standalone bundle ships a fixed set of utilities and regenerates nothing from page source. A class the application writes but the bundle does not contain **does nothing at all, silently**: the page loses the rule, the build succeeds, the type-check succeeds, and the suite stays green.

Every one of these was found that way, one review round at a time, on #243:

| Utility | What it did | Where |
|---|---|---|
| `size-7` | provider mark rendered **956 × 956** — an `` with no width fills its container | messaging settings |
| `size-8` | chosen as the replacement; equally absent, so **1116 × 1116** | messaging settings, Tasks, Integrations, onboarding |
| `sm:grid-cols-*` | every grid collapsed to one column: Agent cards never went two-up, two usage figures became stacked full-width bars | Agents list, Agent home, runtime settings, Task detail |
| `justify-self-end` | column label rendered full width and read as a page subtitle | Agents list |
| `max-w-md`, `max-w-xl`, `max-w-6xl` | containers unbounded | login, not-found, onboarding |

Substituting another supported value each time treats the symptom; the next class anyone writes has the same 50/50.

## What

The Web app builds its stylesheet with Tailwind over its own sources and Kumo's dist. The emitted stylesheet goes from **350 selectors to 1344**, and a missing utility becomes impossible rather than quiet.

- `app.css` imports `tailwindcss` and Kumo's Tailwind layer, and declares both scan roots.
- `vite.config.ts` adds `@tailwindcss/vite`; `main.tsx` drops the standalone bundle.
- `biome.json` enables the CSS parser's Tailwind directives so `@source` parses.
- `Dockerfile` copies `scripts` into the build stage, because the Web build now runs a post-build assertion.
- `kumo-contract.test.ts` asserts the JIT wiring instead of the standalone import.

## The check, and two ways it can lie

`scripts/assert-web-generated-utilities.mjs` runs after `vite build` and reads the stylesheet that was just emitted.

**It asserts the artifact, not the wiring.** The failure this replaces was wiring that was written and did no work — an `@source` glob that matched nothing. A check on the shape of that line passes while every control on the page is unstyled. Both were observed here.

**It lives outside `apps/web` deliberately.** Tailwind's automatic source detection covers the Vite root, so a checker placed inside it has its own probe class names picked up as candidates: it generates exactly what it looks for and passes by describing itself. That was observed here too.

Its probes are four classes only Kumo's components render, plus three that `apps/web/src` writes and Kumo's dist does not — the first group proves the dist is scanned, the second proves application source is. Both groups matter: probes that either side could produce prove one scan twice, which is what the first version of this check did.

Each group was picked from the set difference between builds that do and do not scan the source in question.

## Testing

- `pnpm check` ✅ `pnpm typecheck` ✅ `pnpm build` ✅ (`Verified 7 generated utilities across 1 stylesheet(s).`) `pnpm --filter @opentag/web test` — 367 passed ✅
- **Both negative cases run.** Kumo's source pointed back at the pnpm-store glob, cold cache: build exits 1 naming the four Kumo probes, application probes still present. Application scanning disabled (`@import "tailwindcss" source(none)` without the `src` source): build exits 1 naming the three application probes, Kumo probes still present. Each half fails on its own evidence.
- `@tailwindcss/vite` also contributes sources from Vite's module graph, so the Kumo half degrades rather than disappears: 1741 selectors when the source is right, 984 when it points at a path that does not exist, 574 for the unmatched glob that caused the original unstyled build. That is why the check names classes instead of watching stylesheet size.
- Wanted from review: a real `docker build`. On the branch this came from, `pnpm check`, `pnpm typecheck` and 374 tests were green while the production image build was broken, because the image copies only the workspaces it builds.

## Relationship to #243

Split out of #243 at @yuezengwu's request: it is Web styling/build architecture and wants its own build and container review rather than approval inside a page-redesign PR. #243 will stay within the utilities the standalone bundle ships and does not depend on this landing first.


bestony (changes requested):
## Spec

**[High] The post-build assertion does not prove that application source is scanned** (`scripts/assert-web-generated-utilities.mjs:39-41`).

The PR describes `after:absolute` and `md:grid-cols-2` as OpenTag-only probes, but neither string occurs under `apps/web/src`. Both are present in Kumo's scanned dist (`checkbox`, `radio`, `sidebar`, and `toast` chunks for `after:absolute`; the `grid` chunk for `md:grid-cols-2`). Consequently, if application-source scanning stops working while Kumo's dist scan remains healthy, all six probes are still generated and the build passes. That leaves the silent failure mode this PR is intended to close.

Please use probes that occur in Web source and are absent from Kumo dist (for example, current source includes `size-7`, `size-60`, `max-w-xl`, and `sm:grid-cols-2`, all absent from Kumo dist), and verify the negative case where only application-source scanning is broken.

No other spec gaps or scope creep found.

## Standards

No documented-standard violations or meaningful baseline code smells found. The multi-file wiring is cohesive and necessary for this build path.

Verification: a real `docker build --tag opentag-pr264-review .` completed successfully and the generated-utility assertion reported all six current probes. The required Node 22.13 compatibility job is currently red on an unrelated-looking Web menu test (`Account` menu item not found); the primary checks, Node 26, CLI pack, and Container Smoke jobs pass.

Summary: Standards 0 findings; Spec 1 high-severity finding, blocking because the central artifact guard can pass without application-source coverage.

liuchao-001 (commented):
You are right, and it is the more interesting half of the finding: the check proved one scan twice.

Verified before changing anything — neither string occurs under `apps/web/src` on this branch, and both are in Kumo's dist (`after:absolute` in 4 files, `md:grid-cols-2` in 2). So all six probes came from the Kumo side, and application scanning could have stopped without the build noticing.

Fixed in `098c251`. The application probes are now classes `apps/web/src` writes and Kumo's dist does not — `last:border-b-0`, `md:px-8`, `sm:grid-cols-2` (each `grep -rl` = 0 in dist, ≥1 in src) — so each half is proved by something only that half can produce.

**Both negative cases run, not just described:**

- Kumo source broken (back to the pnpm-store glob), cold cache → build **exit 1**, naming `.h-11`, `.bg-kumo-badge-blue`, `.bg-kumo-elevated`, `.gap-2.5`; the three application probes stay present.
- Application scanning disabled (`@import "tailwindcss" source(none)` with the `src` source removed) → build **exit 1**, naming `.last:border-b-0`, `.md:px-8`, `.sm:grid-cols-2`; the four Kumo probes stay present.

One measurement worth recording, which came out of testing your finding. `@tailwindcss/vite` also contributes sources from Vite's module graph, so the Kumo half degrades rather than disappears depending on how it breaks:

| `@source` for Kumo | selectors | `.h-11` |
|---|---|---|
| `../node_modules/@cloudflare/kumo/dist` | 1741 | yes |
| a path that does not exist | 984 | yes |
| the pnpm-store glob (the original bug) | 574 | no |

A nonexistent literal path still leaves most of Kumo generated; the unmatched glob is what produced the unstyled build this PR exists to prevent. That is the argument for probing named classes rather than a stylesheet-size threshold, and it is now in the file's comment.

Separately, the Node 22.13 failure you saw is the account-menu test reading its items synchronously after opening a menu Kumo mounts asynchronously — it passes locally and races on CI. Same fix as in #243: wait for the item that must exist before asserting the ones that must not. Included in `098c251`.

liuchao-001:
CI is green on `098c251`: **Checks, Node 22.13.0, Node 26, CLI Pack Smoke, Container Smoke, CI** — including the Node 22.13 job @bestony saw red, which was the account-menu test reading its items synchronously after opening a menu Kumo mounts asynchronously.

The probe fix from that review is in the same head, with both negative cases run rather than described. Re-requested review.

bestony (approved):
## Standards

No findings. The new probe split is clear and cohesive, and the async menu-test adjustment is the smallest scoped fix for the Node 22 race. No documented-standard violations or meaningful baseline smells remain.

## Spec

No findings. The previous High issue is fixed: `last:border-b-0`, `md:px-8`, and `sm:grid-cols-2` occur in Web source and have zero occurrences in Kumo dist, so application scanning is now proved independently of the four Kumo-only probes. The checker still lives outside the Vite root and cannot self-generate its probes.

Verification on `098c251`:

- `docker build --tag opentag-pr264-rereview .` passed and reported `Verified 7 generated utilities across 1 stylesheet(s).`
- Web build plus the artifact assertion passed locally.
- Web suite passed: 22 files, 367 tests.
- All required GitHub jobs pass, including Node 22.13, Node 26, and Container Smoke.

Non-blocking documentation note: the PR description still says “two” application probes and `Verified 6`; the implementation now uses three and verifies seven.

Summary: Standards 0 findings; Spec 0 findings. The previous blocker is resolved.

yuezengwu (changes requested):
Reviewed exact head `098c251592c479d068d653273a3a770b5028e7f4` under the current non-engineer author gate.

The implementation replaces the Web styling/build contract: it adds Tailwind and `@tailwindcss/vite`, changes Vite and CSS generation, scans dependency dist files, changes the repository CSS parser configuration, adds a post-build artifact checker outside the Web app, and expands the Docker build stage. That is technical/build architecture, not a page, interaction, or page-required interface change, so it cannot be approved in a PR authored by `liuchao-001`. #243 was previously held on this same boundary until the build change was split into an engineer-owned prerequisite; moving the diff into a separate PR without changing ownership does not satisfy that gate. Please move this implementation to a replacement PR owned by an approved engineer (or remove the architecture change).

I found no additional implementation-correctness blocker on this head. Independent exact-head validation passed: `git diff --check`, `pnpm check` (existing unrelated warnings/info only), `pnpm typecheck`, `pnpm build` with `Verified 7 generated utilities across 1 stylesheet(s).`, the Web suite (22 files / 367 tests), and a real production `docker build`. All six GitHub checks are green. The PR is also currently `CONFLICTING`/`DIRTY` against `main`; that is separate from this scope blocker.

liuchao-001:
@yuezengwu — on the author gate, not on the implementation.

My position as the author: this is styling and design work, which is a product manager's to own. The problem it fixes is a visual one, and the reason it exists is design, not architecture.

So that the classification is made on the diff rather than on my summary of it, here is the whole of it — 10 files:

- `apps/web/vite.config.ts`, `apps/web/src/app.css`, `apps/web/src/main.tsx`, `apps/web/package.json`, `pnpm-lock.yaml` — generate utilities from source instead of importing Kumo's fixed stylesheet
- `biome.json` — teach the CSS parser the Tailwind directives the new stylesheet uses
- `scripts/assert-web-generated-utilities.mjs` + one `COPY` line in `Dockerfile` — fail the build if generation silently stops working
- `apps/web/src/ui/kumo-contract.test.ts`, `apps/web/src/__tests__/app.test.tsx` — the tests for the above

**Why it is a design problem.** Kumo's standalone mode ships a *fixed* set of utilities. A class outside that set does not warn, does not error, and does not appear in any test — it simply does nothing. In #243 that produced an icon rendered at 956px, grids that never applied, and cards that never went two-up, each found by eye rather than by a check. The current alternative is that every class anyone writes must first be looked up by hand in the shipped stylesheet; #243 shipped that way, and one class written inside a ternary still slipped past the first sweep.

**Where I think the disagreement actually is.** The intent is design; the mechanism is a build change, which is what your review reacted to, and that reading is reasonable. I am not disputing the facts of the diff — only who may own a change whose purpose is visual correctness.

If you read this as within a PM-authored PR, it is ready — you found no implementation blocker, @bestony approved it, and all six checks are green. If you do not, tell me which you prefer and I will do it today: an engineer takes over authorship of the implementation as it stands, or the build change is dropped and #243 keeps the manual per-class verification. Either way #243 does not depend on this landing.

(This PR is also `CONFLICTING` against main now — #261 landed. Whoever ends up owning it, I will resolve that first.)

yuezengwu:
The author gate is based on the implementation mechanism and its blast radius, not only on the user-facing motivation. Visual correctness is the goal here, but this diff changes the Vite/Tailwind/Kumo CSS compiler path, package and lockfile state, Biome parser behavior, Docker build inputs, and the build artifact contract. That remains technical/build architecture, so I am keeping the author-scope blocker on exact head `098c251592c479d068d653273a3a770b5028e7f4`.

My recommendation is the first option: have an approved engineer own a replacement PR carrying the current implementation and resolve the `main` conflict there. The fallback is to remove this build change and keep #243 on manual per-class verification. The green CI and @bestony approval are useful correctness evidence, but they do not widen a non-engineer author's allowed scope.

liuchao-001:
Superseded by #282, which landed the same change under an approved engineer's authorship — the resolution @yuezengwu asked for.

Closing rather than reworking: the substance is on main. `app.css` imports the Tailwind layer, `vite.config.ts` registers `@tailwindcss/vite`, `biome.json` parses the directives, and the standalone import is gone from `main.tsx`.

One piece of this PR did not travel with it, and I have opened #285 for it rather than reopening this: the post-build assertion that the generated stylesheet actually contains the utilities. #282 covers the wiring through `kumo-contract.test.ts`; `apps/web`'s build script is plain `vite build`. The distinction is not academic — during this work an `@source` glob that matched nothing produced 350 selectors instead of 1344, leaving every control unstyled while wiring-shaped checks stayed green. That is the failure the artifact assertion catches and a contract test cannot.

The branch stays at `102fad9` (merged with main, conflicts resolved, checker green) if any of it is wanted.

PR #268, opened 2026-08-30
fix(web): add route error recovery boundary

bestony:
Add a root React error boundary and TanStack Router error component with sanitized diagnostics, retry/reset behavior, and a navigation fallback. Includes focused regression coverage. Local checks passed; CI will verify the branch.

### DIFF
--- a/apps/web/src/features/error-boundary.tsx
+++ b/apps/web/src/features/error-boundary.tsx
@@ -141,10 +141,16 @@ export const rootErrorHandlers = {
 };
 
 const credentialKey =
-  "(?:authorization|cookie|password|secret|token|api[_-]?key|access[_-]?token|refresh[_-]?token|client[_-]?secret)";
+  "(?:cookie|password|secret|token|api[_-]?key|access[_-]?token|refresh[_-]?token|client[_-]?secret)";
+const authorizationField = /(["']?authorization["']?\s*[:=]\s*)("[^"]*"|'[^']*'|[^,;}\n]+)/gi;
 
 function redactErrorMessage(message: string): string {
   return message
+    .replace(
+      authorizationField,
+      (_match, prefix: string, value: string) =>
+        `${prefix}${/^['"]?Bearer\s/i.test(value.trim()) ? "Bearer [REDACTED]" : "[REDACTED]"}`,
+    )
     .replace(
       new RegExp(
         `(["']?${credentialKey}["']?\\s*[:=]\\s*)(Bearer\\s+(?:"[^"]*"|'[^']*'|[^\\s,;}\\]]+)|"[^"]*"|'[^']*'|[^\\s,;}\\]]+)`,
--- a/apps/web/src/features/error-boundary.tsx
+++ b/apps/web/src/features/error-boundary.tsx
@@ -141,12 +141,17 @@ export const rootErrorHandlers = {
 };
 
 const credentialKey = "(?:password|secret|token|api[_-]?key|access[_-]?token|refresh[_-]?token|client[_-]?secret)";
-const authorizationField = /(["']?authorization["']?\s*[:=]\s*)("[^"]*"|'[^']*'|[^\n,;}]*?\S)(?=\s*(?:[,;}\n]|$))/gi;
-const cookieQuotedValue = String.raw`(?:"(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*')`;
-const cookieArrayValue = String.raw`\[(?:${cookieQuotedValue}|[^\[\]"'])*\]`;
-const cookieObjectValue = String.raw`\{(?:${cookieQuotedValue}|[^{}"'])*\}`;
+const quotedValue = String.raw`(?:"(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*')`;
+const arrayValue = String.raw`\[(?:${quotedValue}|[^\[\]"'])*\]`;
+const objectValue = String.raw`\{(?:${quotedValue}|[^{}"'])*\}`;
+const structuredValue = `(?:${arrayValue}|${objectValue}|${quotedValue})`;
+const structuredValueWithBoundary = String.raw`${structuredValue}(?=\s*(?:[,}\n]|$))`;
+const authorizationField = new RegExp(
+  String.raw`(["']?authorization["']?\s*[:=]\s*)(${structuredValueWithBoundary}|[^\r\n]*)`,
+  "gi",
+);
 const cookieField = new RegExp(
-  String.raw`(["']?(?:cookie|set-cookie)["']?\s*[:=]\s*)(?:${cookieArrayValue}(?=\s*(?:[,}\n]|$))|${cookieObjectValue}(?=\s*(?:[,}\n]|$))|${cookieQuotedValue}(?=\s*(?:[,}\n]|$))|[^\r\n]*)`,
+  String.raw`(["']?(?:cookie|set-cookie)["']?\s*[:=]\s*)(${structuredValueWithBoundary}|[^\r\n]*)`,
   "gi",
 );
 
@@ -154,8 +159,7 @@ function redactErrorMessage(message: string): string {
   return message
     .replace(
       authorizationField,
-      (_match, prefix: string, value: string) =>
-        `${prefix}${/^['"]?Bearer\s/i.test(value.trim()) ? "Bearer [REDACTED]" : "[REDACTED]"}`,
+      (_match, prefix: string, value: string) => `${prefix}${redactAuthorizationValue(value)}`,
     )
     .replace(cookieField, (_match, prefix: string) => `${prefix}[REDACTED]`)
     .replace(
@@ -166,6 +170,13 @@ function redactErrorMessage(message: string): string {
       (_match, prefix: string, value: string) =>
         `${prefix}${/^Bearer\s/i.test(value) ? "Bearer [REDACTED]" : "[REDACTED]"}`,
     )
-    .replace(/\bBearer\s+(?:"[^"]*"|'[^']*'|[^\s,;}\]]+)/gi, "Bearer [REDACTED]")
+    .replace(/\bBearer\s+(?!\[REDACTED\])(?:"[^"]*"|'[^']*'|[^\s,;}\]]+)/gi, "Bearer [REDACTED]")
     .replace(/([?&](?:code|client_secret|token|secret|password|key)=)[^&#\s]+/gi, "$1[REDACTED]");
+}
+
+function redactAuthorizationValue(value: string): string {
+  const trimmed = value.trim();
+  const quote = /^(["'])Bearer\s/i.exec(trimmed)?.[1] ?? "";
+  if (!/^['"]?Bearer\s/i.test(trimmed)) return "[REDACTED]";
+  return `${quote}Bearer [REDACTED]${quote && trimmed.endsWith(quote) ? quote : ""}`;
 }
--- a/apps/web/src/features/error-boundary.tsx
+++ b/apps/web/src/features/error-boundary.tsx
@@ -142,7 +142,7 @@ export const rootErrorHandlers = {
 
 const credentialKey =
   "(?:cookie|password|secret|token|api[_-]?key|access[_-]?token|refresh[_-]?token|client[_-]?secret)";
-const authorizationField = /(["']?authorization["']?\s*[:=]\s*)("[^"]*"|'[^']*'|[^,;}\n]+)/gi;
+const authorizationField = /(["']?authorization["']?\s*[:=]\s*)("[^"]*"|'[^']*'|[^\n,;}]*?\S)(?=\s*(?:[,;}\n]|$))/gi;
 
 function redactErrorMessage(message: string): string {
   return message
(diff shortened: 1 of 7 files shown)


apps/web/src/features/error-boundary.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+function normalizeError(value: unknown): BoundaryError {
+  if (value instanceof Error) return value;
+  return new Error(typeof value === "string" ? value : "Unknown application error");
+}
+
+/** Logs diagnostics without copying credential-shaped values into the browser console. */
+export function reportBoundaryError(boundary: "app" | "route", error: unknown, errorInfo?: ErrorInfo) {
+  const normalized = normalizeError(error);
+  console.error("[OpenTag] Unhandled UI error", {
```
yuezengwu, line 126: [P1] This is not yet a safe logging boundary. In a production build, React 19's default caught-error reporting logged the original `Error: Authorization: Bearer review-secret` before `componentDidCatch`; the custom sanitizer also turns `Authorization: Bearer abc123` into `Authorization=[REDACTED] abc123`, and leaves `{"accessToken":"abc123"}` intact. That leaks the provider/request details this comment explicitly says must not reach a shared browser surface. Please control and sanitize the React root caught-error path as well as this custom log, use redaction that covers Bearer and quoted/JSON credential fields, and add regression coverage proving the secret is absent from both paths.

apps/web/src/features/error-boundary.tsx
```diff
@@ -0,0 +1,141 @@
...
+          Back to Agents
+        </Link>
+      </div>
+    </BoundaryCard>
+  );
+}
+
+export function StandaloneErrorPage({ actionLabel, onAction }: { actionLabel: string; onAction: () => void }) {
+  return (
+    <main className="grid min-h-full place-items-center bg-kumo-canvas p-6" data-ui="app-error-boundary">
```
yuezengwu: [P2] The production CSS from this exact head contains neither `.min-h-full` nor the `.max-w-xl` used by `BoundaryCard`. At a 1280×720 viewport, the browser computed this `main` as only 238px high and the card as 1232px wide, so the recovery page is top-anchored and nearly full-width instead of the intended centered, bounded surface. This PR cannot rely on a separate, unmerged Tailwind compiler change; please express the layout with styles/utilities available on the current base (or add scoped CSS here) and cover the production-built layout.

yuezengwu (changes requested):
Requesting changes on exact head `f1a3c9afc7ba8e3e2d96b526ee2b025aba795e81`. The boundary/retry direction is sound, and the route retry probe recovered after a transient render failure, but the two inline findings block this from being a safe, production-correct fallback.

Validation: canonical Context Tree verified at `4f76b1474848c515cef9c1bc7ce42c5558395e89`; `git diff --check`; repository build (5/5 tasks); Web tests (23 files, 368 tests); Web typecheck; focused retry probe (1 test); production-browser inspection at 1280×720. All six GitHub CI checks are green.

There is no linked issue. Remaining risk after the fixes should be covered by a production-built browser assertion for layout and by regression cases for React's root caught-error path plus common Bearer/quoted credential formats.

apps/web/src/features/error-boundary.tsx
```diff
@@ -0,0 +1,158 @@
...
+};
+
+const credentialKey =
+  "(?:authorization|cookie|password|secret|token|api[_-]?key|access[_-]?token|refresh[_-]?token|client[_-]?secret)";
+
+function redactErrorMessage(message: string): string {
+  return message
+    .replace(
+      new RegExp(
+        `(["']?${credentialKey}["']?\\s*[:=]\\s*)(Bearer\\s+(?:"[^"]*"|'[^']*'|[^\\s,;}\\]]+)|"[^"]*"|'[^']*'|[^\\s,;}\\]]+)`,
```
yuezengwu, line 167: [P1] This alternative redacts only the first whitespace-delimited token for non-Bearer Authorization values. A focused fresh-head probe with `reportBoundaryError("root", new Error("Authorization: Basic dXNlcjpwYXNz"))` fails because the emitted message is `Authorization: [REDACTED] dXNlcjpwYXNz`. That leaves the actual Basic credential in the browser console. Please redact the entire Authorization value regardless of scheme (and add this case), or simplify further by omitting arbitrary exception messages from browser logs.

yuezengwu (changes requested):
Fresh-head review of `d09590d1b5d5da57c0d74e2e304d5f32b44e9be3`. The React 19 root handlers now suppress the default raw caught-error report, the added Bearer/JSON/password cases are sanitized, and the scoped production CSS restores a full-height centered fallback (independently verified against a production build at 1280×720: main 1280×720, card 576px wide and centered). One credential-leak boundary remains: the generic authorization-key regex consumes only the first whitespace-delimited token for non-Bearer schemes, so `Authorization: Basic dXNlcjpwYXNz` is logged as `Authorization: [REDACTED] dXNlcjpwYXNz`. Please make Authorization redact the whole credential value (or, more simply and robustly, stop logging arbitrary error messages) and add a regression covering a non-Bearer authorization scheme. Full repository validation otherwise passes: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` (23 files / 369 web tests), plus `git diff --check`.

bestony:
@yuezengwu The latest head 63a5ddf includes the complete Authorization-value redaction fix and an exact Basic-scheme regression test. All required CI checks are green. Please re-review this exact head when available.

apps/web/src/features/error-boundary.tsx
```diff
@@ -0,0 +1,164 @@
...
+function redactErrorMessage(message: string): string {
+  return message
+    .replace(
+      authorizationField,
+      (_match, prefix: string, value: string) =>
+        `${prefix}${/^['"]?Bearer\s/i.test(value.trim()) ? "Bearer [REDACTED]" : "[REDACTED]"}`,
+    )
+    .replace(
+      new RegExp(
+        `(["']?${credentialKey}["']?\\s*[:=]\\s*)(Bearer\\s+(?:"[^"]*"|'[^']*'|[^\\s,;}\\]]+)|"[^"]*"|'[^']*'|[^\\s,;}\\]]+)`,
```
yuezengwu, line 167: **[P1] Redact the entire Cookie header value**

`credentialKey` treats `Cookie` like a single scalar value, so this alternative stops at the first semicolon. For example, `Cookie: theme=dark; sid=session-secret` is reported as `Cookie: [REDACTED]; sid=session-secret`, which still sends the session credential to the browser console when the boundary logs the sanitized diagnostic. A Cookie header is credential-bearing regardless of the individual cookie names, so please handle `Cookie` (and `Set-Cookie`, if accepted here) as a whole-value field before the generic key matcher, and add a multi-cookie regression case.

yuezengwu (changes requested):
Requesting changes on exact head `63a5ddf4fb90f3ab27224817a96c0f7cd38e0592`. The latest Authorization repair works, and the boundary/retry direction remains sound, but Cookie diagnostics still expose credential material after the first cookie.

Validation on this head: `git diff --check`, `pnpm check`, `pnpm build`, `pnpm typecheck`, and the full repository `pnpm test` pass; all six GitHub CI checks are green. A focused temporary regression (removed after execution) reproduced the inline leak with a multi-cookie header.

bestony:
@yuezengwu Head `051e188` handles `Cookie` and `Set-Cookie` as whole-value fields, matched before the generic `credentialKey` matcher (which no longer lists `cookie`). The dedicated pattern omits `;` from its terminator set, so a full cookie list is consumed: `Cookie: theme=dark; sid=session-secret` now sanitizes to `Cookie: [REDACTED]`. Added a multi-cookie + Set-Cookie regression asserting no `session-secret` substring survives. Authorization behavior is unchanged (Basic redacts whole, Bearer keeps the scheme). Local validation: `pnpm check`, `pnpm typecheck`, `git diff --check`, and web tests 23 files / 371 tests all pass. Please re-review this exact head.

apps/web/src/features/error-boundary.tsx
```diff
@@ -0,0 +1,165 @@
...
+/** Replaces React 19's default root reporting so caught and uncaught errors stay sanitized. */
+export const rootErrorHandlers = {
+  onCaughtError: (error: unknown, errorInfo: BoundaryErrorInfo) => reportBoundaryError("root", error, errorInfo),
+  onRecoverableError: (error: unknown, errorInfo: BoundaryErrorInfo) => reportBoundaryError("root", error, errorInfo),
+  onUncaughtError: (error: unknown, errorInfo: BoundaryErrorInfo) => reportBoundaryError("root", error, errorInfo),
+};
+
+const credentialKey = "(?:password|secret|token|api[_-]?key|access[_-]?token|refresh[_-]?token|client[_-]?secret)";
+const authorizationField = /(["']?authorization["']?\s*[:=]\s*)("[^"]*"|'[^']*'|[^\n,;}]*?\S)(?=\s*(?:[,;}\n]|$))/gi;
+const cookieField = /(["']?(?:cookie|set-cookie)["']?\s*[:=]\s*)("[^"]*"|'[^']*'|[^\n,}]*?\S)(?=\s*(?:[,}\n]|$))/gi;
```
yuezengwu: **[P1] Redact every value in serialized Set-Cookie collections**

This expression stops at the first comma. On this exact head I passed `Set-Cookie: ["sid=first-session-secret", "refresh=second-session-secret"]` through `reportBoundaryError`; the logged message redacted the first entry but still contained `refresh=second-session-secret`, so the later session credential is written to the browser console. Please redact the complete Cookie/Set-Cookie field representation, including array- or comma-serialized values, and add a regression with at least two Set-Cookie values.

yuezengwu (changes requested):
Re-reviewed the updated exact head `051e188a7ac6aa712e20781fdd191915157231e3`. The production error-boundary flow now renders the recovery page and `Try again` returns to Agents, and the repository checks and current GitHub CI pass. One credential-redaction blocker remains on this head: serialized multi-value Cookie/Set-Cookie fields expose values after the first comma. Please address the inline finding and add the regression before merging.

bestony:
@yuezengwu Head `e32363c`, two commits.

`a673ef3` addresses your finding. Rather than patch another terminator, Cookie/Set-Cookie now match the complete field value in one pass: a bracketed array, a braced object, or a quoted string (each with a delimiter lookahead so an enclosing JSON `,`/`}` is preserved), otherwise a run-to-end-of-line fallback. Your exact probe `Set-Cookie: ["sid=first-session-secret", "refresh=second-session-secret"]` now yields `Set-Cookie: [REDACTED]` with neither value surviving. Regressions added for the array case, a comma-separated case, and an enclosing-JSON case asserting `{"Set-Cookie":[REDACTED],"safe":true}`.

`e32363c` closes the same defect class on Authorization before it becomes a fourth round. `Authorization: Digest username="u", response="secret-response", nonce="secret-nonce"` previously leaked after the first comma; it now redacts whole. Authorization and Cookie share the same extracted matcher building blocks, so this shape only has to be correct once. Bearer still keeps its scheme (`Bearer [REDACTED]`), quotes included, and `{"authorization": "Bearer x", "safe": true}` keeps `, "safe": true}`.

Validation on this head: `pnpm check`, `pnpm build`, `pnpm typecheck`, `git diff --check`, and web tests 23 files / 375 tests. I also ran a temporary probe over your three prior exact inputs plus the Digest case and removed it after execution.

apps/web/src/features/error-boundary.tsx
```diff
@@ -0,0 +1,182 @@
...
+};
+
+const credentialKey = "(?:password|secret|token|api[_-]?key|access[_-]?token|refresh[_-]?token|client[_-]?secret)";
+const quotedValue = String.raw`(?:"(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*')`;
+const arrayValue = String.raw`\[(?:${quotedValue}|[^\[\]"'])*\]`;
+const objectValue = String.raw`\{(?:${quotedValue}|[^{}"'])*\}`;
+const structuredValue = `(?:${arrayValue}|${objectValue}|${quotedValue})`;
+const structuredValueWithBoundary = String.raw`${structuredValue}(?=\s*(?:[,}\n]|$))`;
+const authorizationField = new RegExp(
+  String.raw`(["']?authorization["']?\s*[:=]\s*)(${structuredValueWithBoundary}|[^\r\n]*)`,
```
yuezengwu, line 150: **[P1] Redact folded credential-header continuations**

Both fallback branches stop at CR/LF, so an indented continuation is logged verbatim. On this exact head, `Authorization: Digest username="u",\r\n response="secret-response",\r\n nonce="secret-nonce"` became `Authorization: [REDACTED]\r\n response="secret-response",\r\n nonce="secret-nonce"`. `Set-Cookie: sid=first,\r\n refresh=second-session-secret` likewise retained the second value. Folded or wrapped header dumps therefore still expose credentials in the browser console. Please consume continuation lines beginning with SP/HTAB as part of the credential field (while leaving independent headers intact) and add regressions for both Authorization and Set-Cookie.

yuezengwu (changes requested):
Re-reviewed the new exact head. The serialized Cookie/Set-Cookie collection and non-Bearer Authorization fixes work for single-line values, and the production error UI plus Try again recovery still work. One credential-leak blocker remains: folded or line-wrapped credential headers retain their continuation lines in both root and route console diagnostics. Please fix the continuation handling and add regressions before merge.

PR #269, opened 2026-08-30
fix(web): label onboarding Kumo input

bestony:
Provide an explicit accessible name for the onboarding Agent name Kumo input and add a regression assertion for aria-labelledby. Local checks passed; CI will verify the branch.


yuezengwu (approved):
Reviewed exact head cb2b1da79219b7bfcc1d2006cf156f4a54c14db0. The visible label now explicitly names the Kumo input through aria-labelledby while preserving the native htmlFor association and existing hint/error descriptions. This matches the established Kumo bridge pattern and stays narrowly scoped. Validation: pnpm build; focused onboarding-v2 page tests (42/42); Web typecheck; pnpm check; pnpm test; git diff --check.

PR #270, opened 2026-08-30
fix(tasks): recover paginated task loading

bestony:
Make Tasks pagination race-safe with loading state, functional page appends, retryable append errors, and initial-load recovery. Includes the server cursor timestamp fix and integration coverage required for second-page requests. Local web/server checks passed.

### DIFF
--- a/apps/web/src/features/tasks-page.test.tsx
+++ b/apps/web/src/features/tasks-page.test.tsx
@@ -144,6 +144,21 @@ describe("Tasks debug view", () => {
     expect(taskRequest).toHaveBeenLastCalledWith({ cursor: "next-page" });
   });
 
+  it("retries an initial Tasks load after displaying the request error", async () => {
+    const taskRequest = vi
+      .spyOn(browserApi, "tasks")
+      .mockRejectedValueOnce(new Error("Temporary initial load failure"))
+      .mockResolvedValueOnce({ tasks: [task], nextCursor: null });
+
+    await renderInRouter(<TasksPage />);
+
+    expect(await screen.findByText("Temporary initial load failure")).toBeTruthy();
+    fireEvent.click(screen.getByRole("button", { name: "Try again" }));
+    expect(await screen.findByRole("link", { name: task.title })).toBeTruthy();
+    expect(taskRequest).toHaveBeenCalledTimes(2);
+    expect(screen.queryByRole("button", { name: "Try again" })).toBeNull();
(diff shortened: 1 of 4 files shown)


yuezengwu (changes requested):
Requesting changes on exact head `45ee5ed24d677176523b1ab571262b12cfe177e4`.

**[P1] Fence a Turn-page response to the Task whose cursor produced it.** This PR makes second-page Turn queries succeed, but `TaskDetailPage.loadMoreTurns()` still appends every successful response to whatever `current` detail is mounted when it resolves. TanStack Router can reuse the page while `taskId` changes. I reproduced: open Task A, hold its `browserApi.task(A, cursor)` request, rerender the page for Task B and let B load, then resolve A; the UI renders A’s older Turn inside B’s conversation while retaining B’s header. The effect’s `active` flag protects only the initial request, and neither the append success/error nor `loadingMore` is generation-bound. Capture the task/request generation (or remount by `taskId`), reset transient pagination state on task changes, and add the delayed A → B regression. This is now an executable success path because the server cursor cast in this PR removes the previous 500.

Validation: full body/comments/reviews/checks/diff and the related #243/#262 pagination findings inspected; Context Tree `first-tree-ai/opentag-context-tree@4f76b1474848c515cef9c1bc7ce42c5558395e89` verified clean; `git diff --check` passes; focused Web tests pass 7/7; server Tasks integration passes 7/7; an independent exact-head router reproduction confirms the stale append. All six GitHub checks are green, but none covers Task-detail navigation during an in-flight page request.

yuezengwu (approved):
Approved on exact head `4f5cd157591a8a98948c1d1c2c5eb8acb185c646`.

The previous Task A → Task B stale-pagination blocker is resolved. Both append success and failure are fenced by the current load generation, the functional append additionally checks the current Task identity, and task changes reset the main loading state plus append loading/error state. The delayed A → B regression exercises both outcomes while proving B can paginate independently before A settles; the separate error-reset test covers already-visible transient state.

Validation: full fresh body/comments/reviews/checks/diff inspected; Context Tree `first-tree-ai/opentag-context-tree@d0c621c2b75f27edd178dbb27ecd9a270e5f94bf` verified clean; `git diff --check`; focused Web 10/10; Web typecheck; Server Tasks integration 7/7; Server typecheck; repository check. All six exact-head GitHub checks pass. No remaining blocking finding.

PR #271, opened 2026-08-30
refactor(web): type the Kumo design-system adapter

bestony:
Remove unsafe adapter casts from Button, Tabs, Field, and select prop plumbing. Keep the compatibility select callback as a named value-shaped event instead of pretending it is a React SyntheticEvent, while preserving existing consumers. Local web typecheck and focused adapter tests passed.

### DIFF
--- a/apps/web/src/ui/design-system.tsx
+++ b/apps/web/src/ui/design-system.tsx
@@ -61,8 +61,9 @@ import {
   X,
 } from "@phosphor-icons/react";
 import {
-  type ButtonHTMLAttributes,
+  type ChangeEvent,
   Children,
+  type ComponentPropsWithoutRef,
   cloneElement,
   forwardRef,
   type HTMLAttributes,
@@ -70,6 +71,7 @@ import {
   isValidElement,
   type ReactElement,
   type ReactNode,
+  type Ref,
   type RefObject,
   type SelectHTMLAttributes,
   type SVGAttributes,
@@ -114,16 +116,24 @@ export const Select: typeof KumoSelect = KumoSelect;
 export const Checkbox: typeof KumoCheckbox = KumoCheckbox;
 export const Switch: typeof KumoSwitch = KumoSwitch;
 
+type KumoSelectProps = ComponentPropsWithoutRef<typeof KumoSelect>;
+
 function classes(...values: Array<string | false | null | undefined>): string {
   return cn(values.filter(Boolean).join(" "));
 }
 
 export type ButtonVariant = "primary" | "secondary" | "outline" | "ghost" | "danger" | "inline";
 
-export type ButtonProps = Omit<ButtonHTMLAttributes<HTMLButtonElement>, "color"> & {
+type KumoButtonAdapterProps = KumoButtonProps extends infer Props
+  ? Props extends unknown
+    ? Omit<Props, "size" | "variant">
+    : never
+  : never;
+
+export type ButtonProps = KumoButtonAdapterProps & {
   size?: "default" | "compact";
   variant?: ButtonVariant;
-} & Partial<Pick<KumoButtonProps, "loading" | "shape" | "icon" | "title">>;
+};
 
 function kumoButtonVariant(variant: ButtonVariant): "primary" | "secondary" | "outline" | "ghost" | "destructive" {
   if (variant === "danger") return "destructive";
@@ -150,19 +160,24 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(function Button
   { className, size = "default", variant = "primary", type = "button", ...props },
   ref,
 ) {
-  return (
-    <KumoButton
-      {...(props as KumoButtonProps)}
-      className={buttonClassName({ className, size, variant })}
-      ref={ref}
-      size={size === "compact" ? "sm" : "base"}
-      type={type}
-      variant={kumoButtonVariant(variant)}
-    />
-  );
+  const kumoProps: KumoButtonProps = {
+    ...props,
+    className: buttonClassName({ className, size, variant }),
+    size: size === "compact" ? "sm" : "base",
+    type,
+    variant: kumoButtonVariant(variant),
+  };
+  return <KumoButton {...kumoProps} ref={ref} />;
 });
 
 /** A controlled Kumo tab strip that keeps React Router links as its triggers. */
+type TabTriggerProps = HTMLAttributes<HTMLElement> & {
+  children?: ReactNode;
+  href?: string;
+  ref?: Ref<HTMLElement>;
+  "aria-current"?: HTMLAttributes<HTMLElement>["aria-current"];
+};
+
 export function Tabs({
   children,
   className,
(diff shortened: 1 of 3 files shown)


apps/web/src/ui/design-system.tsx
```diff
@@ -357,52 +370,73 @@ export function Icon({ className, name, ...props }: SVGAttributes<SVGSVGElement>
...
-        if (onChange) onChange({ currentTarget: { value: String(value) }, target: { value: String(value) } } as never);
-      }}
-    >
+    <KumoSelect {...kumoProps}>
       {options.map((option) => (
         <KumoSelect.Option
-          data-value={String(option.props.value ?? "")}
           disabled={option.props.disabled}
           key={String(option.props.value)}
           value={option.props.value ?? ""}
```
yuezengwu: [P1] Normalize compatibility option identity. The adapter declares and supplies a string controlled `value`, and it stringifies every emitted change, but it passes numeric/array `<option>` values through to Base UI unchanged. `AgentUsageTab` already supplies numeric `AGENT_USAGE_WINDOW_OPTIONS`; with `value="30"` and `value={30}`, the visible label renders but the matching option reports `aria-selected=false` because Base UI compares different identities. Keyboard/assistive selection state therefore remains wrong after every selection. Normalize each option value to the same string domain (or make the whole bridge consistently generic) and cover the numeric-option case.

apps/web/src/ui/design-system.tsx
```diff
@@ -357,52 +370,73 @@ export function Icon({ className, name, ...props }: SVGAttributes<SVGSVGElement>
   );
 }
 
-export const KumoInputControl = forwardRef<
-  HTMLInputElement,
-  InputHTMLAttributes<HTMLInputElement> & Pick<KumoInputProps, "size" | "variant">
->(function KumoInputControl(props, ref) {
-  return <KumoInput {...props} ref={ref} />;
-});
+export type KumoInputControlProps = InputHTMLAttributes<HTMLInputElement> & Pick<KumoInputProps, "size" | "variant">;
```
yuezengwu: [P2] Omit native input `size` before adding Kumo size. `InputHTMLAttributes<HTMLInputElement>` defines `size?: number`; intersecting it with Kumo `size?: "xs" | "sm" | "base" | "lg"` collapses the property to `undefined`. A direct exact-head type probe assigning `{ size: "sm" }` to `KumoInputControlProps` fails TS2322, so the newly named adapter type does not expose the runtime component contract. Use `Omit<InputHTMLAttributes<HTMLInputElement>, "size">` before the Kumo pick and add a type-level regression.

apps/web/src/ui/design-system.tsx
```diff
@@ -357,52 +370,73 @@ export function Icon({ className, name, ...props }: SVGAttributes<SVGSVGElement>
...
+
+type SelectOptionProps = {
+  value?: string | number | readonly string[];
+  disabled?: boolean;
+  children?: ReactNode;
+};
+
+export const KumoSelectControl = forwardRef<HTMLButtonElement, KumoSelectControlProps>(function KumoSelectControl(
+  { children, onChange, onValueChange, value, ...props },
+  _ref,
```
yuezengwu: [P2] Do not advertise a ref that is discarded. `forwardRef<HTMLButtonElement, ...>` tells callers the select trigger ref will resolve, but `_ref` is never connected to Kumo Select, so every consumer receives `null`. Wire it to the trigger through Kumo/Base UI’s supported render/ref composition, or remove the forwarded-ref API until it can be honored.

apps/web/src/ui/design-system.tsx
```diff
@@ -357,52 +370,73 @@ export function Icon({ className, name, ...props }: SVGAttributes<SVGSVGElement>
...
-  HTMLButtonElement,
-  Omit<SelectHTMLAttributes<HTMLSelectElement>, "size" | "value"> & {
-    size?: "xs" | "sm" | "base" | "lg";
+export type SelectControlChangeEvent = {
+  currentTarget: { value: string };
+  target: { value: string };
+};
+
+type SelectControlChangeProps = {
+  onChange?(event: SelectControlChangeEvent): void;
```
yuezengwu, line 398: [P2] Finish narrowing the compatibility event at its consumers. The adapter correctly stops pretending this object is a React SyntheticEvent, but `TaskSelect` still accepts `ChangeEventHandler<HTMLSelectElement>` and passes it here. React’s bivariant event-handler type lets this compile even though runtime supplies only `{ target, currentTarget }`; any valid handler use such as `preventDefault()` or `nativeEvent` would crash. Import/use `SelectControlChangeEvent` for that wrapper (and audit the other call sites) so the checked consumer contract matches the value-shaped event.

yuezengwu (changes requested):
Reviewed exact head `f8ef71b1a0c635603df63f76b01f433bc976e136` under the engineer gate. The cast removal is directionally good, but the adapter still has four type/runtime mismatches, including a reproduced selection-semantic failure in a current consumer, so I am requesting changes.

Validation on this exact head: `git diff --check`, `pnpm check`, repository `pnpm build`, Web typecheck after building Shared, and focused design-system/usage/Tasks tests (13/13) pass; GitHub CI is green. A temporary regression probe (removed after execution) rendered the current string-controlled adapter with the numeric options used by Agent Usage and observed the matching option at `aria-selected=false`. A temporary type probe (also removed) confirmed that `KumoInputControlProps` rejects the Kumo-supported `size="sm"` with TS2322.

Context Tree read from `first-tree-ai/opentag-context-tree@4f76b1474848c515cef9c1bc7ce42c5558395e89` verified cleanly; no durable product constraint conflicts with this scoped adapter refactor.

apps/web/src/ui/design-system.tsx
```diff
@@ -357,62 +370,81 @@ export function Icon({ className, name, ...props }: SVGAttributes<SVGSVGElement>
...
+    isValidElement<SelectOptionProps>(child),
   );
+  const normalizedOptions = options.map((option) => ({
+    label: String(option.props.children ?? option.props.value ?? ""),
+    option,
+    value: normalizeSelectOptionValue(option.props.value),
+  }));
+  const labels = new Map(normalizedOptions.map(({ label, value: optionValue }) => [optionValue, label]));
+  const kumoProps: KumoSelectProps = {
+    ...props,
```
yuezengwu, line 436: [P1] Normalize or exclude the inherited defaultValue prop. KumoSelectControl still inherits native defaultValue (including number and readonly string[]), and this spread passes it through unchanged while every option is normalized to a string. I reproduced defaultValue={42} with option value={42}: opening the list reports aria-selected="false". Normalize defaultValue with the same helper before passing it to Kumo, or deliberately omit/forbid unsupported uncontrolled values, and add the matching regression test.

yuezengwu (changes requested):
The fresh head resolves the four blockers from my prior review: controlled option identities now normalize, the input size type remains Kumo-named, the unused select ref was removed, and TaskSelect now uses the value-shaped event contract. One matching uncontrolled-mode boundary remains, reproduced below.

bestony:
@yuezengwu The latest head d00e0b2 includes the defaultValue normalization fix from your last review, rebased onto current main. All required CI checks are green. Please re-review this exact head when available.

yuezengwu (approved):
Approved on exact head `d00e0b20894bb373e2a422de28844e6fbd592952`. The prior controlled-value, uncontrolled-default, input-size, ref, and change-event findings are resolved without introducing a new blocker.

Validation: `git diff --check`, `pnpm check`, `pnpm build`, `pnpm typecheck`, focused adapter tests, and a standalone full `pnpm test` pass; all six GitHub CI checks are green. I also exercised the real local Web in Chrome: the Kumo select opened through its accessible combobox/listbox contract, keyboard selection changed `Everything ready` to `Runtime not installed`, the dependent description updated, and the console stayed free of warnings/errors.

PR #272, opened 2026-08-30
perf(web): split routes and tree-shake echarts

bestony:
## Summary

Three independent bundle-size changes, one commit each so they can be reviewed and reverted separately.

**`perf(web): tree-shake echarts by registering it in a static module`**
`await import("echarts/charts")` asks for the whole module namespace, and Rollup cannot prove which bindings the caller reads, so it kept all 22 chart types and every component, map projections included. Registration moves into `features/agent-usage-echarts.ts` behind static named imports, which are traceable; the lazy boundary stays at the importer. The lazy chart graph goes from 8 chunks totalling 1,096 kB (365 kB gzip) to one 563 kB chunk (190 kB gzip). The entry chunk is unaffected — echarts was already lazy.

**`chore: drop the repository-wide Babel 8 overrides`**
`rolldown-plugin-dts` declares `@babel/generator`, `@babel/parser`, `@babel/types` and `@babel/helper-validator-identifier` as exact `8.0.0-rc.2` dependencies of its own, and pnpm resolves per package, so it receives them with or without an override. The overrides did nothing for the one consumer that needs a Babel 8 prerelease, and forced it on every consumer that asked for Babel 7 — `@babel/core@7.29.7` wants `@babel/parser@^7.29.7`, `@tanstack/router-plugin` wants `^7.28.5`, and `magicast` wants `^7.25.4`. A Babel 7 traverse, template and generator were being handed a Babel 8 AST, which is the mismatch the comment in `apps/web/vite.config.ts` warned about. After this, `apps/web` resolves a consistent Babel 7 while tsdown's dts chain keeps `8.0.0-rc.2`.

**`perf(web): split route components out of the entry chunk`**
`autoCodeSplitting` was off, so a visitor who only wanted to sign in downloaded onboarding, agent creation, tasks and settings first. **The entry chunk drops from 1,219 kB to 572 kB (382 kB → 186 kB gzip), across 49 chunks.** This one needed the overrides gone first: the splitter runs a full parse-and-generate pass over route TSX, which is the workload the mismatch would have crashed on.

`src/routeTree.gen.ts` is byte-for-byte unchanged — the splitter rewrites the route modules in memory, not the generated tree.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` all pass (7/7 tasks; web 366 tests, server 292, client 491, shared 84, cli 152).
- Full `pnpm build` exercises `packages/*` `.d.mts` generation, which is the tsdown → rolldown-plugin-dts → Babel 8 chain the overrides used to cover.
- echarts tree-shaking verified in the build output: of the series type ids only `series.line` survives; no component install ids (`dataZoom.slider`, `timeline.slider`, `visualMap.continuous`, `legend.plain`) remain. The residual `polar`/`parallel`/`geo` strings are echarts core's coordinate-system registry, not installed components.
- Route splitting verified by literal probe, since minification renames identifiers but not JSX text: `"Loading Agent usage"`, `"Read-only debug view"` and `"Welcome back"` each left the entry chunk for `agents._agentId.usage-*.js`, `tasks-page-*.js` and `login-*.js`.
- `vite dev` smoke-tested separately, since CI never runs it and the splitter also transforms in dev: server starts, `/` returns 200, and route modules come back rewritten to `lazyRouteComponent` over a `?tsr-split=component` import.

## Breaking changes

None. No runtime behavior changes; no API or schema changes.

## Follow-ups (not in this PR)

- `apps/web/src/router.ts` has no `defaultPendingComponent`. Now that route components load asynchronously, a slow connection may flash blank during a transition.
- `rolldown-plugin-dts@0.28.3` drops Babel entirely. Upgrading `tsdown` past it would let the remaining `ast-kit` pin go too.
- The entry is now dominated by vendor code (React, the router runtime, Kumo, icons) pulled in eagerly through `ui/design-system.tsx`'s barrel. Restructuring that barrel is the next lever.

## Checklist

### DIFF
--- a/apps/web/src/features/agent-usage-echarts.ts
+++ b/apps/web/src/features/agent-usage-echarts.ts
@@ -1,0 +1,15 @@
+import { LineChart } from "echarts/charts";
+import { BrushComponent, GridComponent, ToolboxComponent, TooltipComponent } from "echarts/components";
+import * as echarts from "echarts/core";
+import { CanvasRenderer } from "echarts/renderers";
+
+/*
+ * The chart's echarts registration lives in a module of its own so the imports above can be static.
+ * Awaiting `import("echarts/charts")` directly asks for the whole module namespace, and Rollup cannot
+ * prove which of its bindings a caller reads, so it keeps every one — all 22 chart types and every
+ * component, map projections included. Static named imports are traceable, so only what is named here
+ * survives; the lazy boundary moves out to whoever imports this module.
+ */
+echarts.use([LineChart, BrushComponent, GridComponent, ToolboxComponent, TooltipComponent, CanvasRenderer]);
+
+export { echarts };
--- a/apps/web/src/features/agent-usage.tsx
+++ b/apps/web/src/features/agent-usage.tsx
@@ -10,18 +10,7 @@ import { browserApi } from "../api.js";
 import { ChartPalette, KumoSelectControl, Loader, Meter, Text, TimeseriesChart } from "../ui/design-system.js";
 
 const LazyTimeseriesChart = lazy(async () => {
-  const [
-    { LineChart },
-    { BrushComponent, GridComponent, ToolboxComponent, TooltipComponent },
-    { CanvasRenderer },
-    echarts,
-  ] = await Promise.all([
-    import("echarts/charts"),
-    import("echarts/components"),
-    import("echarts/renderers"),
-    import("echarts/core"),
-  ]);
-  echarts.use([LineChart, BrushComponent, GridComponent, ToolboxComponent, TooltipComponent, CanvasRenderer]);
+  const { echarts } = await import("./agent-usage-echarts.js");
   return {
     default: (props: Omit<ComponentProps<typeof TimeseriesChart>, "echarts">) => (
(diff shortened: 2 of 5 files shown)


yuezengwu (approved):
Reviewed exact head 5e1a51359df557ae366d3cd6553913dd2bfe159d. No blocking findings. Verified the three changes independently: the production entry drops from 1,219,738 to 572,195 bytes (381,888 to 185,048 gzip); ECharts’ eight-chunk 1.10 MB lazy graph becomes one 562,866-byte chunk while retaining the existing Line/Brush/Grid/Toolbox/Tooltip/Canvas registrations; and Babel 7 consumers now resolve a consistent 7.x graph while rolldown-plugin-dts keeps its exact Babel 8 dependencies. Route generation remains byte-stable, dev transforms routes through lazyRouteComponent, SPA deep links and history contracts remain intact, and hashed chunks retain the existing immutable-cache/deployment recovery path. Local pnpm check/build/typecheck/test passed (1,385 tests), as did diff-check; all six CI checks are green. The missing shared pending UI and intent prefetch remain the acknowledged non-blocking slow-network follow-up.

Issue #273, opened 2026-08-30
test(web): app shell focus test fails intermittently, cause not isolated


liuchao-001:
## Symptom

`apps/web/src/__tests__/app.test.tsx` → **`OpenTag Web App Shell > moves focus into account actions and returns it to the trigger on Escape`** fails intermittently.

## Frequency

Observed on **two different machines**:

- 1 failure in **41** whole-file runs (`vitest run __tests__/app.test`)
- 1 failure during a full-suite run on a separate machine

## The only clue we have

The failing run took **491ms**. React Testing Library's default `findBy*` timeout is 1000ms, so that run did not fail by exhausting a wait — some assertion returned the wrong thing.

⚠️ That is weaker than it first looks: a slow-but-successful `findByRole` followed by an immediate `getByRole` throw would produce the same signature. It does not rule out timing or scheduling.

## What we do *not* know

Explicitly undetermined, because the samples cannot distinguish them:

- **Order dependence** — whether some earlier test in the file leaves focus, an open menu, or a listener behind.
- **Load** — the one reproduction happened while the machine was doing other work, which if anything leans *toward* load.

Isolation attempts that came back clean are **not** evidence against either hypothesis. At a ~2.4% rate, 12 green runs of the single test has roughly a 75% chance of happening anyway, and 3 green full-suite runs under artificial CPU load is close to certain. We initially read those as informative and were wrong to.

## Suggested next step

Chasing another reproduction is expensive — tens of runs per hit. Cheaper to make the **next** occurrence self-documenting: on the failure path of that test, print `document.activeElement`, the count of `[role="menu"]` in the DOM, and their accessible names. One hit then settles it.

## Not blocking

Found while working on #262, which does not touch this file or the app shell. Recording it rather than filing it under "flaky" and moving on: it is real, it is reproducible across machines, and the cause is simply not yet isolated.

liuchao-001:
Instrumentation is up as #274 — it makes the next occurrence self-documenting rather than continuing to chase reproductions, for the reasons in the frequency section above. Failure path only, so nothing prints on a green run.

liuchao-001:
[...]

PR #274, opened 2026-08-30
test(web): make the account menu tests assert what their names claim

liuchao-001:
## What

Two account-menu tests now assert what their names claim. This started as a diagnostic for #273 and became the fix, because building the diagnostic found the cause.

## The tests were not testing

Both asserted `queryByRole("menu", { name: "Account" })` is null. **No such menu ever exists.** `aria-label="Account"` is applied to the popup wrapper, which base-ui renders with `role="presentation"` — so it is ignored, and the menu takes its name from `aria-labelledby` pointing at the trigger. Under jsdom that name resolves to `"AAda"`; in Chrome it resolves to `"Account menu"`. Either way it is never `"Account"`, so the assertion was null with the menu open and null with it closed.

The plainest statement of what that was worth: **deleting the `fireEvent.keyDown(…, "Escape")` line entirely left the test passing.** Neither test checked the focus return both are named for.

## What is actually happening

Escape works. Closing is **asynchronous** — the menu is still in the document on the tick after the keydown and gone a few milliseconds later:

```
after focus:        active=DIV/menuitem  menus=1 items=2
escape on item:     active=DIV/menuitem  menus=1 items=2
after 50ms settle:  active=BUTTON/-      menus=0 items=0
```

So the tests ended with that work still outstanding, to land against whatever ran next.

## The change

Wait for the close, assert it, and assert focus returns to the trigger. Both new assertions verified by removing what they depend on: drop the Escape dispatch → red; expect focus elsewhere → red.

## On #273

⚠️ **This is not claimed as the fix.** Awaiting the close means it now completes inside the test rather than escaping its boundary, which makes it a candidate cause for an intermittent failure in this file — but 12 clean runs afterwards say nothing either way. At the observed ~2.4% rate that outcome is likelier than not regardless, which is the same arithmetic that invalidated our earlier isolation runs.

The diagnostic that was here is gone: it existed to find a cause that is now found.

## Note for a separate change

**`aria-label="Account"` in `app-shell.tsx` does nothing** — it lands on the presentation wrapper, so the menu is named through `aria-labelledby` pointing at the trigger instead.

That turns out to be harmless. Chrome resolves the trigger to its own `aria-label="Account menu"`, so the menu's accessible name is **"Account menu"**, which is correct. Verified in headless Chrome against the real components and confirmed by mutation: removing the trigger's `aria-label` makes the name fall back to the trigger's text, `"AAda"`.

⚠️ An earlier revision of this section claimed a screen reader announces the menu as *"AAda"* and called it an accessibility defect. That came from `dom-accessibility-api` under jsdom, whose name computation stops at the referenced element's text where Chrome first consults its `aria-label`. The DOM wiring was right; the conclusion drawn from a test library's approximation was not.

What remains is a no-op attribute that misleads the reader. Tracked in #275 as cosmetic cleanup; this PR is tests only. It is the only place in the app where a label sits on a base-ui `Content`.

## Verification

`pnpm check`, typecheck clean; web tests **366 passed**; 12 consecutive whole-file runs green.

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -2687,7 +2687,6 @@ describe("OpenTag Web App Shell", () => {
     expect(screen.getByRole("menu")).toBeTruthy();
     account.focus();
     note("after focus");
-    expect("x").toBe("forced-failure");
     fireEvent.keyDown(account, { key: "Escape" });
     note("after escape");
     expect(screen.queryByRole("menu", { name: "Account" })).toBeNull();
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -1,7 +1,7 @@
 import type { AgentUsageDetail } from "@opentag/shared/browser";
 import { act, fireEvent, render, screen, waitFor, within } from "@testing-library/react";
 import { StrictMode } from "react";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, onTestFailed, vi } from "vitest";
 import { App } from "../app.js";
 import { PasswordSignInForm } from "../features/auth/password-sign-in-form.js";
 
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
Reviewed exact head 67c9400725278e5d136896c665b27a53f93c5d17. This is a test-only UI interaction change and stays within the non-engineer scope boundary. The old accessible-name query was vacuous; the new generic menu-close assertion plus trigger-focus assertion directly exercise the Escape contract. Mutation checks removed Escape and substituted the wrong focus target, and both tests failed in each case. Independent validation passed: 24 whole-file runs (2,832 test executions), pnpm check, pnpm build, pnpm typecheck, pnpm test, diff check, and all six CI checks. Approval does not claim that #273 is fixed; that intermittent failure remains unisolated and the issue correctly stays open.

PR #276, opened 2026-08-30
refactor(web): move stray page modules under features/agents

bestony:
## Summary

AGENTS.md places page and domain modules in `src/features`, but three sat at the root of `apps/web/src`. Each moves to the directory its consumers already live in, with no behavior change.

| Module | New home | Why there |
|---|---|---|
| `computer-setup.{tsx,test.tsx}` | `features/agents/` | Shared by `agent-settings/` **and** `agent-creation/`, so it belongs at the top of the domain rather than inside either |
| `runtime-configuration.{tsx,test.tsx}` | `features/agents/agent-settings/` | Its only consumer is `agent-settings-page.tsx`, beside sibling form modules like `general-config-form.tsx` |
| `agent-detail-capabilities.{tsx,test.tsx}` | `features/agents/` | Agent-detail domain |

There are no tsconfig path aliases and no barrel files in this app — every import is a relative path with an explicit `.js` extension — so the relative imports inside the moved files and in their three consumers are rewritten by hand. `tsc --noEmit` is the safety net for anything missed, and it passes.

A second commit drops `{true && !backToSettings}` in `agent-detail-page.tsx`, which gated the Settings link on a constant. The sibling Usage link above it already reads as plain `!backToSettings`.

## Not included, and why

The plan for this change also proposed collapsing what looked like duplicate helpers — `formatRelativeTime` and `getInitials` in `features/tasks-page.tsx` against `formatRelativeTime` and `initials` in `features/agents/agent-presentation.ts`. **They share names but not behavior**, so collapsing them would have been a silent UI change:

- `formatRelativeTime`: the tasks-page one uses `Intl.RelativeTimeFormat` (locale-aware, renders future timestamps as "in 2 minutes", rounds); the agent-presentation one builds English strings by hand, clamps with `Math.max(0, …)` so future timestamps read "just now", floors, and renders anything under a minute as "just now" rather than "30 seconds ago".
- `getInitials` vs `initials`: the latter trims, filters empties, **upper-cases**, and falls back to `"OT"`; the former does none of that.

Picking one is a product decision about what the Tasks page should read like, not mechanical cleanup, so it is left out of this PR.

## Testing

- `pnpm check`, `pnpm typecheck`, `pnpm test` pass — web is 22 files / 366 tests, unchanged from before.
- Because there is no alias layer, a missed path rewrite is a type error, so `typecheck` passing is meaningful coverage for the moves themselves.
- `__tests__/app.test.tsx` needed no edit: it imports only `../app.js` and `../features/auth/password-sign-in-form.js`, so it exercises the moved modules through rendered behavior rather than static paths.

## Breaking changes

None.

## Checklist


yuezengwu (approved):
Reviewed exact head fbb9ecbb60ef2704123c4c1fb7a91ef1727460ba. No blocking findings: the change is a focused module move with complete relative-import rewrites plus an equivalent always-true guard simplification; it preserves the relevant Agent/Computer and Runtime boundaries. Verified pnpm check, build, typecheck, test, git diff --check, targeted old-path reference scan, and green CI. Remaining warnings are pre-existing onboarding/jsdom/build baseline warnings outside this diff.

PR #277, opened 2026-08-30
refactor(web): read Server state through TanStack Query

bestony:
## Summary

`apps/web` had no request cache. Every mount refetched, nothing was shared between pages, and invalidation was done by hand — four components kept a counter whose only job was to change a cache key. Three separate load-state shapes had grown up alongside each other (`useResource`'s `LoadState`, `tasks-page`'s own copy, `agent-usage`'s `UsageState`).

This replaces that with `@tanstack/react-query`, migrating all twelve `useResource` call sites plus the hand-written `useEffect` loaders, and retires the hook. Eight commits, each keeping the suite green.

### What it buys

| | Before | After |
|---|---|---|
| Agents list, cold | 2+2N requests | 2+2N — unchanged |
| Agents list, 30s revalidate | 2+2N | 2+2N — unchanged |
| List → Agent detail | +4 requests | **+1**, the rest are cache hits |
| Agent detail → usage | refetches the same 30 days | **cache hit** |
| Settings, switching sections | refetches the config each time | **one read, shared** |
| New Agent dialog, opened over the list | refetches Computers | **cache hit** |
| Backgrounded tab | kept polling on `setInterval` | **stops** (`refetchIntervalInBackground`) |

Worth being explicit: **client caching does not reduce the steady-state polling.** N Agents on screen each needing 30-second freshness is inherent to the product, and cutting the 2+2N fan-out still needs the Server to join that evidence into the list response. That remains unstarted and should be scheduled on its own.

### Decisions worth reviewing

**`retry: false` in production, not just tests.** Nothing here ever retried inside a fetch; recovery is explicit — an interval, a focus, or a button. A silent retry would also swallow the first failure of a refresh a caller is waiting on, which is the failure a page has to report.

**`refetchOnWindowFocus` off by default, opted into at four call sites.** Only the surfaces that watch an Agent recover want a focus to re-read. The cache also watches only `visibilitychange`, which does not fire when the window regains focus without the tab changing, so both listeners are restored — and the focus one signals an event rather than a state, since the cache reacts to a state only when it changes and every focus after the first would be ignored.

**`refreshMe` reads directly and writes only its success.** A failure there belongs to the caller that asked for it. Fetching through the cache would record the error on the Account gate instead, replacing the whole signed-in surface over a refresh the Account never saw. `reloadMe` resets rather than invalidates, because it is documented to show the loading state again.

**Two polls stay hand-rolled, with a comment saying why.** Computer setup and Feishu setup do not read resources — they wait on a mutation's side effect and drive one authorization to completion. Nothing else reads either, so there is no sharing to gain, and the lifecycles they keep (which error came from starting versus polling; which attempt a late response belongs to) are the substance of those flows. This is a deliberate exception to the otherwise-complete migration.

### Two bugs found while doing it

**An infinite refetch loop.** Waiting on a read was first written as "is it in flight". The Agent detail page shows the messaging tab, which reads the same binding — so a re-read put the page back into loading, unmounting the tab, whose remount started another re-read. It never settled. Waiting is now "has it ever resolved".

**A lost loading transition.** The creation flow restores focus and announces the outcome by watching a Computer refresh begin and end. Invalidating keeps the previous data on screen, so that transition never happened; the refresh now reports loading, and starts from an effect so the render reporting it is committed before a fast response can land in the same batch.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` all pass (7/7 tasks). Web is 24 files / 409 tests, up from 22 / 366 on a comparable base.
- `projectAgentAvailability` had no test at all and is the logic being refactored, so **characterization tests were written first**, in their own commit, covering every outcome in decision order plus the two easy-to-break details: suspension outranking every dependency, and a failed handoff read leaving the channel evidence alone.
- The adapter that preserves "stale but flagged unconfirmed" has its own tests, including the branch that is easiest to miss — a terminal 401/403/404/410 stays an error even with usable data in hand and a way to mark it.
- A new assertion holds Agent usage to **one** request when the overview and the tab show the same window together; nothing asserted that count in either direction before.
- Entry chunk 572 kB → 598 kB (186 → 193 kB gzip), the cost of the library.

## Breaking changes

None. No API or schema changes.

## Checklist


apps/web/src/app.tsx
```diff
@@ -17,20 +19,26 @@ const AppLink = forwardRef<HTMLAnchorElement, LinkComponentProps>(function AppLi
...
     if (!owned) return;
     return () => instance.history.destroy();
   }, [instance, owned]);
   return (
-    <LinkProvider component={AppLink}>
-      <TooltipProvider>
-        <RouterProvider router={instance} />
-      </TooltipProvider>
-    </LinkProvider>
+    <QueryClientProvider client={queryClient}>
```
yuezengwu: **[P1] Drop Account-scoped cache entries on sign-out**

This QueryClient lives for the whole App mount, while `AppShell.logout` only performs an SPA navigation to `/login`; no cache entry is cleared. I reproduced loading `/agents`, signing out, then re-entering `/agents` while the new `/me` request was deliberately pending: the cached `me` and Agent queries immediately rendered the prior Account's `Reviewer` row before the eventual 401 redirected back to login. That makes logout temporarily disclose the previous Account's data on a protected-route revisit. Please clear/remove the Account-scoped cache as part of a successful logout (or force a document reload), and cover post-logout route re-entry while authentication is unresolved.

apps/web/src/features/agents/agent-queries.ts
```diff
@@ -0,0 +1,164 @@
...
+  /*
+   * Waiting on the first read of each, not on whether one is in flight now. A re-read must not put
+   * the page back into loading: doing so unmounts what the page is showing, and anything below that
+   * reads the same evidence would be remounted into re-reading it, which never settles.
+   */
+  const settling =
+    !agentQuery.isFetched || !computersQuery.isFetched || !bindingQuery.isFetched || !handoffQuery.isFetched;
+  if (settling) return initialAgent ? { kind: "ready", value: initialAgent } : { kind: "loading" };
+  if (!agentQuery.data) {
+    return initialAgent
```
yuezengwu: **[P1] Let terminal Agent failures outrank route state**

When navigation supplies `initialAgent`, a settled 401/403/404/410 with no query data takes this branch and returns the stale route object as `ready` forever. A focused probe with `initialAgent` plus `browserApi.agent()` rejecting `ApiError(404, "Agent not found")` expected `error` but observed `ready`. This bypasses the terminal-error rule in `toResourceState` and leaves deleted or forbidden Agent data/settings visible. Check the primary query's terminal error before falling back to `markAgentDetailUnconfirmed(initialAgent)`; retain that fallback only for non-terminal failures, with 403/404 regression coverage.

apps/web/src/features/agent-usage.tsx
```diff
@@ -76,33 +78,18 @@ function useAgentUsage(
...
-    };
-  }, [load]);
+  // Keyed by Agent and window, so the overview on an Agent's page and the tab on its usage page ask
+  // for the same 30 days once between them rather than each on its own.
+  const query = useQuery({
+    queryKey: queryKeys.agents.usage(agentId, windowDays),
+    queryFn: () => browserApi.agentUsage(agentId, windowDays),
+  });
+  const retry = useCallback(() => void query.refetch(), [query]);
+  const state: UsageState = query.data
```
yuezengwu: **[P2] Give terminal refetch errors precedence over sticky data**

This data-first ordering returns `ready` whenever the query cache still holds usage, even when the latest re-read ended in 401/403/404/410. `TaskDetailPage` has the same boundary through `const first = taskQuery.data?.pages[0]` and `if (!first)`: I loaded a Task, navigated away and back, returned 404 from the second detail read, and the old `Stored runtime output` remained instead of the `Task not found` state. These custom query adapters bypass the terminal behavior that `toResourceState` explicitly preserves. Please apply terminal-error precedence to both usage and Task detail while deciding separately whether non-terminal failures may retain stale data.

yuezengwu (changes requested):
Requesting changes on exact head `957a0a89f848a245383051b2407c4041cf4b11fe`. The shared-query migration and its explicit terminal-error adapter are directionally solid, but three cache boundaries still violate the stated terminal/authentication contract.

Validation on this head: `git diff --check`, `pnpm check`, `pnpm build`, `pnpm typecheck`, and the full repository `pnpm test` pass (Web: 24 files / 409 tests); all six GitHub CI checks are green. Three focused temporary regressions (removed after execution) reproduced: cached Account data rendering after logout during re-authentication, route state masking an Agent 404, and cached Task detail surviving a terminal re-read.

bestony:
Addressed in 40980cd — thank you, all three were real and the first two were the same carve-out the state adapter exists to enforce, missed at the boundaries around it.

- **P1 sign-out cache** — `logout()` now clears the client before navigating, so a later Account cannot observe the previous one's entries while its own reads are pending.
- **P1 route state vs terminal Agent** — the terminal check now runs *before* the settling branch as well, not only in the no-data branch. Route state was able to mask a 404 during the window where the evidence reads had not yet settled, which the narrower fix would have left open.
- **P2 usage data-first ordering** — reordered so a terminal error wins over sticky data while a non-terminal one still degrades to the retained value. `tasks-page.tsx` had the same ordering and is fixed with it, though it was not flagged.

Regression tests are parameterized over the terminal statuses rather than one representative code. Verified on this head: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` all pass — web is now 24 files / 414 tests.

apps/web/src/features/agents/agent-queries.ts
```diff
@@ -0,0 +1,173 @@
...
+          binding?.isSuccess ? (binding.data ?? undefined) : undefined,
+          handoff?.isSuccess ? (handoff.data ?? undefined) : undefined,
+          binding?.isSuccess ?? false,
+          handoff?.isSuccess ?? false,
+        ),
+        evidenceConfirmed: true,
+      };
+    }),
+  };
+  return toResourceState(
```
yuezengwu: **[P1] Evict a successfully deleted Agent before relying on refetch**

Retaining list data during a non-terminal background failure is useful, but the Manage Agent delete path navigates back without updating this cache. On this exact head I first cached `Reviewer`, completed its DELETE with 204, then made the immediate Agent-list refresh return 503; the deleted Agent card was still rendered. A confirmed destructive mutation must synchronously remove the Agent from the relevant list and Agent-scoped cache before navigation instead of relying on refetch. Please add the 204 -> list 503 regression.

apps/web/src/features/tasks-page.tsx
```diff
@@ -164,21 +131,19 @@ export function TasksPage() {
         </TaskSelect>
       </form>
 
-      {state.kind === "loading" ? (
-        <TaskNotice heading="Loading Tasks" detail="Reading stored Sessions and Turns." />
-      ) : null}
-      {state.kind === "error" ? (
+      {tasksQuery.isPending ? <TaskNotice heading="Loading Tasks" detail="Reading stored Sessions and Turns." /> : null}
+      {tasksQuery.isError && !tasksQuery.data ? (
```
yuezengwu: **[P1] Let terminal list failures outrank cached Tasks**

The new commit fixes this precedence for Task detail, but the list branch still surfaces an error only when there is no cached data. In a current-head probe, the first Tasks read succeeded and a refetch then rejected with `ApiError(403, "Tasks forbidden")`; the prior Task row remained visible and the forbidden error never rendered. Apply terminal-error precedence to the Tasks list as well, retaining stale rows only for non-terminal failures, and add a 401/403 regression.

yuezengwu (changes requested):
Re-reviewed the updated exact head `40980cdb8acd62a86a68294659a43e909939a61d`. The new commit correctly clears QueryClient on logout and gives terminal Agent-detail, Usage, and Task-detail responses precedence over cached/route state. Repository checks and current GitHub CI pass, but two current-head cache-state regressions remain: a successfully deleted Agent can reappear when the immediate list refresh fails, and a terminal Tasks-list refetch error is still hidden by cached rows. Please address both inline findings before merging.

bestony:
Implemented and pushed as `6efaf74` (fix(web): evict deleted agents and terminal task lists).

- After a successful Agent DELETE (204), the Manage flow synchronously removes that Agent from every cached Agent list and removes all `agents/<id>` detail/config/evidence cache entries before navigating.
- The Tasks list now gives terminal 401/403 (and the existing 404/410) refetch errors precedence over cached rows; non-terminal failures still retain stale rows.
- Added the 204 -> list 503 deletion regression and parameterized 401/403 Tasks-list regressions.

Verified from the repository root with Node 24.18.0: `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` all pass. Web: 24 test files / 417 tests. No merge performed.

bestony:
Merged origin/main into refactor/web-tanstack-query in merge commit 65b250a9497be76301e5e675de3f8c348202f0a7 (no rebase, no force-push).

This required real resolution rather than taking one side wholesale:
- apps/web/src/features/tasks-page.tsx kept the TanStack Query infinite-query and terminal-error behavior from this branch while taking main's Kumo/Tailwind responsive table, container-query layout, and SelectControlChangeEvent changes.
- pnpm-lock.yaml was replaced with main's lockfile and regenerated with pnpm install, restoring @tanstack/react-query on top of main's Tailwind/Vite dependency graph.
- The other overlapping Web files retain both main's Kumo/workspace-persistence updates and this branch's query/cache behavior, including logout cache clearing and Agent deletion eviction.

Verified after resolution from the repository root: pnpm check, pnpm build, pnpm typecheck, and pnpm test all pass (web: 24 files / 420 tests). The PR itself was not merged.

bestony:
Merged `main` into this branch to clear the conflict; head is now `65b250a`.

Only one file conflicted textually, `apps/web/src/features/tasks-page.tsx`, where both sides had edits. I kept both intents rather than picking a side:
- main's #283 a11y horizontal-scroll `<section>` wrapper and `min-w-[36rem]` table, combined with this branch's `tasksQuery.hasNextPage` paging condition.
- main's #271 removal of the `ChangeEventHandler` import (TaskSelect now takes `SelectControlChangeEvent`), combined with this branch's reduced React hook set, since TanStack Query removed the need for `useCallback`/`useEffect`/`useRef`.

Verification that the merge did not silently drop either side: the committed tree differs from git's own auto-merge in exactly one file, `tasks-page.tsx` — every other overlapping file is git's pristine auto-merge. `pnpm install --frozen-lockfile`, `pnpm check`, `pnpm typecheck`, `git diff --check` all clean, and web tests pass at 24 files / 420 tests.

Note the two findings from your 10:06 review were already addressed by `6efaf74` before this merge. Ready for re-review on this head.

apps/web/src/features/tasks-page.tsx
```diff
@@ -34,47 +35,36 @@ const statusPresentation: Record<TaskStatus, { readonly label: string; readonly
...
+    queryFn: ({ pageParam }) => browserApi.tasks({ cursor: pageParam }),
+    initialPageParam: undefined as string | undefined,
+    // The API reports the end of the list as null; the cache reads undefined as "no page after this".
+    getNextPageParam: (page) => page.nextCursor ?? undefined,
+  });
+  const loaded = useMemo(() => tasksQuery.data?.pages.flatMap((page) => page.tasks) ?? [], [tasksQuery.data]);
+  const loadMoreError = tasksQuery.isFetchNextPageError ? asError(tasksQuery.error) : null;
+  const taskError = asError(tasksQuery.error);
+  const terminalTasksError =
+    tasksQuery.isError && !tasksQuery.isFetchNextPageError && isTerminalResourceError(taskError) ? taskError : null;
```
yuezengwu: **[P1] Do not exempt terminal list-pagination failures**

`isFetchNextPageError` identifies which request failed, not whether the response is safe to ignore. On this exact head I loaded a first Tasks page with a next cursor, made `Load more` return `ApiError(403, ...)`, and the cached Task row remained visible with only a pagination error. The same branch would retain rows after 401/404/410. A terminal auth/access/gone response needs to outrank all cached pages; only transient append failures should retain the current rows. Please add terminal page-append coverage alongside the existing 401/403 refetch matrix and preserve the 503 retry behavior.

apps/web/src/features/tasks-page.tsx
```diff
@@ -221,87 +202,43 @@ export function TasksPage() {
...
+    initialPageParam: undefined as string | undefined,
+    getNextPageParam: (page) => page.nextCursor ?? undefined,
+    enabled: taskId !== undefined,
+  });
+  const first = taskQuery.data?.pages[0];
+  const turns = useMemo(() => taskQuery.data?.pages.flatMap((page) => page.turns) ?? [], [taskQuery.data]);
+  const loadMoreError = taskQuery.isFetchNextPageError ? asError(taskQuery.error) : null;
+  const taskError = asError(taskQuery.error);
+  const terminalTaskError =
+    taskQuery.isError && !taskQuery.isFetchNextPageError && isTerminalResourceError(taskError) ? taskError : null;
```
yuezengwu: **[P1] Hide stale Task detail after a terminal Turn-page failure**

The detail path has the same terminal-error exemption. With the first Task page cached, a 403 from `Load more Turns` leaves the Task title, stored message, and runtime output visible and renders only the append error. A 401/403 means the current viewer can no longer read this resource, while 404/410 means it is gone, so those responses must replace the cached detail even when they came from `fetchNextPage`; retain the detail only for non-terminal pagination failures and add a regression.

apps/web/src/features/agents/new-agent-page.tsx
```diff
@@ -1,30 +1,56 @@
...
+    if (!refreshing) return;
+    let watching = true;
+    void queryClient.invalidateQueries({ queryKey: queryKeys.computers() }).finally(() => {
+      if (watching) setRefreshing(false);
+    });
+    return () => {
+      watching = false;
+    };
+  }, [refreshing, queryClient]);
+  const state = toResourceState(query, markOwnComputersUnconfirmed);
```
yuezengwu, line 56: **[P2] Surface failures of the explicit Computer refresh**

The background fallback is also applied after the Account presses `Check again`. I loaded the New Agent page with a Computer whose Provider needed sign-in, then returned 503 from the button-triggered re-read. Once `refreshing` cleared, `toResourceState(..., markOwnComputersUnconfirmed)` converted that failed query with cached data back to `ready`: no error appeared, the page silently changed to `Unconfirmed`, and the dialog live region would announce `Computer connection updated`. Keep the fallback for interval/focus revalidation, but preserve the error outcome for the explicit refresh (and add a success-then-button-503 regression) so the UI does not report failure as an update.

yuezengwu (changes requested):
Re-reviewed the updated exact head. The confirmed-delete cache eviction now survives both an immediate list 503 and an older in-flight list response, and the earlier logout, Agent-detail, Usage, and ordinary Task-refetch terminal-state fixes remain intact. Repository checks and current GitHub CI pass, but terminal errors during Task pagination still leave stale protected content visible, and an explicit Computer refresh failure is silently presented as an updated unconfirmed state. Please address the inline findings before merging.

apps/web/src/features/agents/agent-queries.ts
```diff
@@ -0,0 +1,202 @@
...
+ * `isError` describes only the attempt in flight. A terminal response is not an attempt that failed
+ * but an answer — the Agent is gone or forbidden — and forgetting it on every interval and every
+ * focus is what let route state put a deleted Agent back on screen between re-reads. Only a success
+ * retires it, and the resource it was recorded for keys it so a different Agent never inherits it.
+ */
+function useSettledError(key: string, query: SettlingQuery): Error | null {
+  const settled = useRef<{ key: string; error: Error | null }>({ key, error: null });
+  if (settled.current.key !== key) settled.current = { key, error: null };
+  if (query.isSuccess) settled.current.error = null;
+  else if (query.isError) settled.current.error = query.error ?? new Error("The request failed");
```
yuezengwu: **[P1] Do not let a transient failure retire a terminal verdict**

The helper contract above says only a success retires the last terminal Server answer, but this assignment replaces it with every later error. On this exact head I made the watched Agent read answer 404, then made the focus-triggered re-read answer 503. After the second response, `useAgentDetailView` changed from `error` to `ready` and rendered the route-state `Reviewer` again. The list path can similarly re-expose cached rows after terminal denial followed by a transient outage. Preserve an existing terminal verdict until a confirmed success (while still updating non-terminal-only history), and add a terminal -> 503 regression.

yuezengwu (changes requested):
Re-reviewed fresh exact head `df0e03be8a83aad421ad618dee724e9cb98110b6`. The four follow-up commits correctly fix terminal Task pagination, stale Computer evidence on Agent detail, the in-flight Agent-delete race, and the Account refresh that outlived sign-out. Repository `git diff --check`, `pnpm check`, `pnpm build`, `pnpm typecheck`, and full `pnpm test` all pass (Web: 27 files / 455 tests); current GitHub CI is green except the main Checks job still running. One new terminal-verdict regression remains, and the existing explicit Computer-refresh P2 is still outstanding.

apps/web/src/features/agents/new-agent-page.tsx
yuezengwu, line 56: **[P2] Preserve an explicit refresh error instead of applying the background fallback**

`toResourceState(query, markOwnComputersUnconfirmed)` cannot distinguish the user-triggered refresh from an interval/focus re-read. I reproduced this on the current head by first returning a Computer whose Provider requires sign-in, then returning 503 for the request initiated by `Check again`: after loading ends there is no alert, the cached Computer is silently shown as `Unconfirmed`, and the dialog flow can announce `Computer connection updated`. Background failures may reasonably retain degraded cached data, but an explicit action needs to expose its failure so the user knows the check did not succeed. Please retain the explicit refresh result/error separately (or otherwise bypass the background fallback for that request) and add a success -> `Check again` -> 503 regression test.

yuezengwu (changes requested):
Re-reviewed exact head `df0e03be8a83aad421ad618dee724e9cb98110b6`. The new commits fix the terminal Tasks pagination paths, cancel stale Agent reads before deletion cache eviction, preserve terminal Agent verdicts across a later re-read, and fence an Account refresh that outlives logout. One current-head user-visible refresh regression remains: an explicit New Agent Computer re-read failure is silently converted into a ready/unconfirmed state. Please address the inline finding before merge.

PR #292, opened 2026-08-30
feat(web): add the Paraglide i18n foundation and locale-aware formatting

bestony:
## Summary

Lands the [Paraglide JS 2.x](https://paraglidejs.com/) i18n foundation for `apps/web` (English +
Simplified Chinese), plus the locale-aware formatting module that unifies the duplicated date and
number formatters. This is **Wave 1** of a larger migration — it deliberately does **not** migrate the
~700 user-facing strings in `features/`, `onboarding/`, `onboarding-v2/`, `agent-creation/` or `im/`.
Those follow in later, parallel changes on top of this.

Two parts, in order:

1. **Infrastructure** — a provable no-op for application behaviour.
2. **Formatting unification** — the only part that changes runtime output.

## Design decisions worth knowing before reviewing

Several of these are counter-intuitive and were verified against the Paraglide compiler source, the
inlang SDK source, and this repository rather than assumed:

- **Locale code is `zh`, not `zh-CN`.** Paraglide's `extractLocaleFromNavigator` truncates
  (`zh-CN` → `zh`) but never widens, so configuring `zh-CN` would miss a browser set to bare `zh` or
  to `zh-TW`. `intlLocale()` maps `zh` → `zh-CN` where `Intl` needs a region subtag.
- **`strategy: ["localStorage", "preferredLanguage", "baseLocale"]` — the locale never enters the URL.**
  Every route but `/login` is behind auth, so per-locale SEO has no value here, and a URL strategy
  would require a router `rewrite` adapter, an async root `beforeLoad`, and a rewrite of the shared
  `renderInRouter` test helper that ~900 assertions run through. Adding `"url"` later touches four
  files; nothing here forecloses it.
- **`outputStructure: "message-modules"` is pinned explicitly.** The Vite plugin silently defaults to
  `locale-modules` whenever `NODE_ENV !== "production"`, and Vitest sets `NODE_ENV=test`. Unpinned,
  `pnpm test` and `pnpm build` would alternately delete and recreate the whole generated tree.
- **The generated `apps/web/src/paraglide/` is gitignored, not committed.** This departs from the
  `routeTree.gen.ts` precedent on purpose: the compiler emits a self-ignoring `.gitignore`, and its
  tab-indented `messages/package.json` and the SDK's newline-less `.meta.json` create an unresolvable
  formatting ping-pong with Biome if tracked.
- **`allowJs: true` plus `"src/**/*.js"` in tsconfig `include`** — not `exclude`. Under
  `composite: true`, generated `.js` files outside `include` raise TS6307 once per file. `checkJs`
  stays off, so `noUnusedLocals` and friends do not apply to generated code.
- **The codegen is a real Turbo task with declared `outputs`, not an `&&` in a script.** Turbo hashes
  only git-tracked files, so without this a `messages/*.json` edit would replay a stale cached
  `typecheck`. `tsc --noEmit` never invokes Vite, which is why the edge is needed at all.
- **`biome.json` needs explicit exclusions** — it has no `vcs` block, so Biome does not read
  `.gitignore`. Note this failure mode is invisible in CI (which runs `pnpm check` before
  `pnpm build`, when the directory does not yet exist) and only bites developers locally.
- **Message catalogues are split per area** (`apps/web/messages/<area>/{en,zh}.json`, 16 areas) via an
  array `pathPattern`, so the later parallel migration waves never contend on the same JSON file.

> [!IMPORTANT]
> **Never run `inlang machine translate` or Sherlock's extract on this project.** Under an array
> `pathPattern`, every inlang *write* flow serialises the merged catalogue into **every** file in the
> array, duplicating all keys across all areas. Compiling is safe (`skipExporting: true`); writing is
> not. Chinese is authored by hand. This is documented in `docs/i18n.md`.

## Checklist

**Part 1 — infrastructure**

- [x] 0. Pre-flight: prove the five load-bearing assumptions before writing code
- [x] 1. Add `@inlang/paraglide-js@2.25.0`, `@inlang/plugin-message-format@4.4.3`, `typescript` as direct `apps/web` devDependencies
- [x] 2. `apps/web/package.json` — the `paraglide` compile script
- [x] 3. `apps/web/project.inlang/settings.json` — local `modules` paths, 16-area array `pathPattern`
- [x] 4. `apps/web/vite.config.ts` — `paraglideVitePlugin`, deliberately *not* gated on `VITEST`
- [x] 5. `apps/web/tsconfig.json` — `allowJs` + `src/**/*.js` in `include`
- [x] 6. `turbo.json` — `paraglide` task and the three `@opentag/web#*` overrides
- [x] 7. `biome.json` — exclude the generated output
- [x] 8. `.gitignore` — ignore `apps/web/src/paraglide/` and `project.inlang/cache/`
- [x] 9. `vitest.coverage.config.ts` — exclude generated output from coverage
- [x] 10. `scripts/worktree-bootstrap.mjs` — best-effort codegen so a fresh worktree opens clean
- [x] 11. `apps/web/src/i18n/locale.ts` — the single seam onto the generated runtime
- [x] 12. `apps/web/src/i18n/document-locale.ts` + wiring in `main.tsx` before `createRoot`
- [x] 13. `apps/web/messages/common/{en,zh}.json` — frozen after this change
- [x] 14. `apps/web/src/__tests__/setup.ts` — pin the locale, clear `localStorage` between tests
- [x] 15. `apps/web/src/__tests__/i18n-catalog.test.ts` — eight catalogue invariants
- [x] 16. `docs/i18n.md` + `docs/zh-CN/i18n.md`; `AGENTS.md`; `DEVELOPMENT.md` + `.zh-CN.md`
- [x] 17. `pnpm notices:write` — no drift
- [x] 18. Part 1 acceptance

**Part 2 — formatting unification**

- [x] 19. `apps/web/messages/format/{en,zh}.json` — plural variants, compact units
- [x] 20. `apps/web/src/i18n/format.ts` + `format.test.ts`
- [x] 21. Delete and unify every duplicated formatter
- [x] 22. The split: sentences → Paraglide messages, numbers → `Intl`
- [x] 23. Verify the implicit-vs-explicit `Intl.DateTimeFormat` assumption empirically
- [x] 24. `initials()` and CJK
- [x] 25. Unify relative time on the `agent-presentation.ts` behaviour
- [x] 26. Part 2 acceptance

## Pre-flight findings

- Array `pathPattern` **is** supported — a two-entry probe compiled to `exports=2`.
- The `localStorage` key Paraglide writes is `PARAGLIDE_LOCALE`.
- Local `modules` paths resolve through pnpm's symlinked layout from `apps/web/`, so no jsDelivr CDN
  fetch happens during `pnpm test`, `pnpm build`, or the Docker image build.
- Implicit `Intl.DateTimeFormat(undefined, …)` and explicit `"en-US"` produce **byte-identical**
  output on this machine both with and without `LANG`/`LC_ALL` set, matching CI's unset-`LANG`
  behaviour. This was flagged as the most likely source of unexpected assertion drift; it did not
  materialise.

## Formatters removed or unified

Three copies of `formatTimestamp` / `formatDate`, two `formatUsageNumber` (both hardcoding `"en-US"`),
two independent relative-time implementations, two `initials`, plus the bare `.toLocaleString()`,
`localeCompare` and `toLocaleLowerCase` call sites — all now route through `apps/web/src/i18n/format.ts`
with an explicit locale.

Sentences (`just now`, `{count} minutes ago`, `{count}m`) are Paraglide messages, because
`Intl.RelativeTimeFormat` with `numeric: "auto"` renders `"now"` and `"yesterday"` and would change
English output. Dates, numbers, collation and case-folding are `Intl` with an explicit locale.

`initials()` now treats a no-whitespace Han name as its first two characters
(`initials("张伟") === "张伟"`) while preserving `initials("Ada Lovelace") === "AL"`.

## Verification

Run independently by the orchestrator against this branch, not taken from the agent's report:

| Command | Result |
| --- | --- |
| `pnpm check` | 601 files checked, no fixes; CLI third-party notices verified |
| `pnpm build` | 6/6 tasks |
| `pnpm typecheck` | 8/8 tasks |
| `pnpm test` | **26 files, 401 tests, all passing** |
| `git status --porcelain` | empty |

Because the first run was served from the Turbo cache, the codegen edge was then proved cold:
`rm -rf apps/web/src/paraglide && turbo run typecheck --filter=@opentag/web --force` → 0 cached,
3 tasks, 3.74 s, regenerating 59 files and 50 message exports, with `tsc --noEmit` clean. That is the
load-bearing mechanism of this change — a gitignored codegen output that `tsc`, which never invokes
Vite, must still be able to typecheck from a cold checkout.

**No existing test assertion was changed.** The only test files touched are the two new ones
(`i18n-catalog.test.ts`, `format.test.ts`); `apps/web/src/__tests__/app.test.tsx` is untouched. That
is the correctness argument for the whole migration: `baseLocale: "en"` with message values
transcribed byte-for-byte from the source literals means a green, unmodified suite proves the
transcription — including the typographic characters (`…`, `·`, `’`) this codebase uses liberally.

## Notes for the reviewer

- This branch is based on `766792c` and upstream `main` has moved since (`#286` restoring the
  Computers management page, among others). It may need a rebase before merging; the overlap is
  `apps/web/src/features/shell/app-shell.tsx`, which this change touches only to update a formatter
  import.
- Other in-flight work is editing `biome.json`, `tsconfig.json`, `vite.config.ts`,
  `vitest.coverage.config.ts` and `turbo.json`. Every change here to those files is deliberately
  minimal and lives in its own commit to keep a rebase cheap.
- Known and accepted English leakage after this change: Kumo's own internal strings, server-returned
  `ApiError.message`, and the Zod validation messages in `packages/shared` — `AGENTS.md` forbids that
  package from depending on any workspace, so those need a code→message mapping in a later change.
  All three are listed in `docs/i18n.md`.
- `<title>OpenTag</title>` in `apps/web/index.html` is deliberately left untranslated: the
  `container-smoke` CI job asserts it with `grep -q`.

---

🤖 Written by codex under bypassed approvals, in an isolated git worktree, as lane `dx-6848c2-1` of
dispatch run `6848c2`. Every verification result above was reproduced independently by the
orchestrating agent before this pull request was opened.


yuezengwu (changes requested):
Reviewed exact head `fea4cb4b684144d158c04f7b1df22a16bec6113d`.

## Blockers

1. **[P1] Do not mark an untranslated document as Chinese.** `applyDocumentLocale()` flips the root `<html lang>` to the negotiated locale before Wave 2 has migrated the document copy. In a real browser whose preferred language resolves to `zh`, `/login` renders entirely English copy while the document reports `lang="zh"` (`dir="ltr"`). The PR explicitly leaves roughly 700 user-facing strings for later waves, so this affects most screens, not an isolated fallback. Screen readers and other language-sensitive tooling will apply Chinese pronunciation/rules to English content. Keep the document language at `en` until the visible document is actually migrated, or mark untranslated regions as English and only flip the root after the migration is complete. Add a browser-level regression that exercises the production locale detector and asserts that the root language matches the rendered document; the global test override currently hides this state.

2. **[P2] Do not turn implicit browser negotiation into a sticky user preference.** The generated runtime resolves `preferredLanguage`, then its first `getLocale()` calls `setLocale(resolved, { reload: false })`; because `localStorage` is the first strategy, simply loading the app writes `PARAGLIDE_LOCALE`. A controlled `navigator.languages=["zh-CN"]` probe produced `{ before: null, resolved: "zh", after: "zh" }`. From then on local storage wins even if the browser preference changes. That does not match the documented contract that the first visit follows the browser and an *explicit choice* is persisted. Either override the initialization/set behavior so only the selector persists a locale, or document and accept the sticky-first-visit contract explicitly, with a production-runtime test.

## Validation

- Context Tree: `first-tree-ai/opentag-context-tree@d0c621c2b75f27edd178dbb27ecd9a270e5f94bf`; no i18n-specific decision overrides source behavior.
- `git diff --check HEAD^2..HEAD`
- `pnpm check` (pass; one unrelated existing Biome warning in `onboarding-v2/server-backend.ts`)
- `pnpm build`
- `pnpm typecheck`
- `pnpm test` (scripts 77; Web 452; shared 88; client 555; server 293; CLI 153)
- cold codegen: move generated `src/paraglide` aside, then `pnpm turbo run typecheck --filter=@opentag/web --force` → 0 cached, 59 generated files / 50 message exports, byte-identical regeneration
- real browser `/login`: rendered English content with root `lang="zh"`, no console warnings/errors from i18n

CI has not reported checks for this synchronized head yet. Live GitHub currently reports the PR as conflicting/dirty against the moving base; resolve that separately after the blockers.

yuezengwu (approved):
Reviewed exact head 0f89e9daf6d658a23807c9264c8121298c3c81d4.

The two blockers from fea4cb4 are resolved:

- The root document remains lang="en" / dir="ltr" while Wave 1 still renders predominantly English copy.
- Browser-language negotiation now uses the route-aware resolver without invoking Paraglide's initial persistence hook; a controlled zh-CN runtime probe resolves zh with no PARAGLIDE_LOCALE write.

The prior CI failures on f74bef5 were caused by the new Chinese date test hard-coding UTC+8 on UTC runners. Current head derives the expectation from the same host timezone, and the regression passes under TZ=UTC. Node 22.13 and Node 26 compatibility jobs are now green; aggregate Checks and Patch Coverage were still running at review time.

Validation:
- Context Tree first-tree-ai/opentag-context-tree@093b03aa72e0c458ea52c0affbcc2c34b74e2e53, verify clean
- git diff --check
- pnpm check (pass; existing complexity warnings only)
- pnpm build
- pnpm typecheck
- TZ=UTC targeted format test
- controlled production-runtime locale probe: resolved zh, stored null, zero writes
- real browser /login: visible English copy with root lang="en", dir="ltr"

No remaining code-review findings.

Issue #299, opened 2026-08-30
Account needs both a reset-all and a lighter re-board operation


liuchao-001:
## What is needed

Two distinct account operations for testing onboarding, only one of which exists today.

**1. Reset all** — what `POST /api/v1/internal/onboarding-lab` already does: suspend and delete every Agent the Account created, revoke its unconsumed connect codes, revoke its Computers' credentials and clear their connection state, then null `setupCompletedAt`. This is the right tool for starting genuinely from zero.

**2. Re-board** — does not exist. Let an Account walk the onboarding flow again **without destroying what it has**: clear `setupCompletedAt` and nothing else. Agents, Computers and messaging bindings all stay.

Today only the destructive one exists, so anyone who wants to look at onboarding a second time has to delete their Agent and re-enroll their machine to get there. That cost is why this is being asked for.

## Why re-board cannot be done in the client

`setupCompletedAt` lives on the Account row and gates the route (`/onboarding` redirects to `/agents` when it is set). Only the Server can clear it, so there is no UI-side equivalent.

## Where it should live

⚠️ **Not behind the Onboarding Lab UI.** @liuchao-001 has said the Lab is unlikely to be used going forward, so hanging a new capability off that page would build on something being retired. The reset capability itself is still wanted — it is the page that is not.

Suggested shape, to be confirmed by whoever picks this up:

- Keep the existing endpoint's behaviour as the "reset all" mode.
- Add a mode — request body, or a separate path — that performs only the `setupCompletedAt` clear.
- Both stay **reflexive**: they act on the authenticated Account and never accept a client-supplied one, which is the property that makes the current endpoint safe to expose to every signed-in tester on staging.
- Both stay staging-only (`environment === "staging"`), as today.

One thing worth deciding explicitly: after a re-board, the Account still has an Agent, so onboarding resumes into an existing-Agent path rather than a first-run one. That is the intent — but it also means re-board exercises the resume behaviour, not the create behaviour. Anyone testing first-run specifically still needs reset-all.

## Context

Asked for by @liuchao-001 while testing #262's redesigned onboarding on staging: the existing reset "很像 reset all，就是把 agent 和 computer 都清掉了", and a lighter operation is wanted alongside it rather than instead of it.

liuchao-001:
[...]

PR #301, opened 2026-08-30
feat(server): let an Account re-board without taking anything down

liuchao-001:
## What

Getting back into onboarding cost an Account everything it had.

The only operation that cleared the setup marker also deleted every Agent the Account created, revoked its Computers' credentials and disabled its messaging bindings. Looking at onboarding a second time therefore meant rebuilding an Agent and re-enrolling a machine first.

That operation is still right for a genuine first run and is unchanged. Beside it there is now a lighter one:

| | reset all | re-board |
|---|---|---|
| Agents | deleted | **kept** |
| Computer enrollments | revoked | **kept** |
| Messaging bindings | disabled | **kept** |
| Setup marker | cleared | cleared |

## Where it lives

`POST /api/v1/me/setup/reset` with `{ "mode": "all" | "reboard" }` — a sibling of the `me/setup/complete` path that was already there.

Deliberately **not** behind the staging Onboarding Lab. @liuchao-001 has said the Lab is on its way out; a capability meant to outlive that page should not be reached through it. The Lab's own endpoint is untouched by this change and can be retired with the rest of it.

## Safety properties, kept from the existing reset

- **Reflexive.** The Account comes from the access token; the body carries only how much to undo. There is no field a caller could use to name another Account — and a body that tries is **rejected**, not ignored, because a caller who believed they were choosing an Account should be told they were not.
- **Staging only.** Outside staging both modes answer exactly like a path that was never registered, and the route is not registered at all where the deployment does not offer it.

## Worth knowing before using it

Re-boarding leaves the Account's resources in place, so the next run **resumes into them** rather than exercising first-run creation. Anyone testing first-run behaviour specifically still wants `mode: "all"`. The tests state that difference directly rather than leaving it implied.

## How to invoke it while there is no UI

Signed in on staging, from the browser console:

```js
await fetch("/api/v1/me/setup/reset", {
  method: "POST",
  headers: { "content-type": "application/json" },
  body: JSON.stringify({ mode: "reboard" }),
});
```

Then open `/onboarding`.

## Verification

`pnpm check`, `pnpm build` clean. Shared 88, server 300, web 536 — all passing.

Each new test was checked against its own mutation rather than trusted for being green:

- make `reboard` not clear the marker → the keep-everything test fails
- swap the two modes in the route → the mode-routing test fails

⚠️ Two things a reviewer should know:

1. **`@opentag/client` has failing tests on my machine**, in packages this change does not touch (shared and server only). The failures differ between runs — a different set of six, then ten, one taking 227s — so they read as local resource contention, not a regression. CI is the arbiter.
2. **`onboarding-lab-reset.test.ts` is at a connection ceiling.** Every `fixture()` opens a database client and none are closed, so adding a second fixture to that file fails with `sorry, too many clients already`. I worked within it by folding the staging-guard assertion into the existing guard test rather than adding a fixture, but the next person to add a test there will hit the same wall.

### DIFF
--- a/packages/server/src/__tests__/account-api.test.ts
+++ b/packages/server/src/__tests__/account-api.test.ts
@@ -139,9 +139,13 @@ function services() {
   };
 }
 
-function appWith(overrides: Partial<ReturnType<typeof services>> = {}) {
+function appWith(
+  overrides: Partial<ReturnType<typeof services>> = {},
+  setupReset?: { reboard: ReturnType<typeof vi.fn>; resetOnboarding: ReturnType<typeof vi.fn> },
+) {
   const service = { ...services(), ...overrides };
   const app = createApp({
+    ...(setupReset ? { stagingOnboardingLab: { reset: setupReset as never } } : {}),
     authService: authService(),
     agentService: service.agentService as unknown as AgentService,
     machineAuthService: service.machineAuthService as unknown as MachineAuthService,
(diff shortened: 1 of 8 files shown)


liuchao-001:
Reviewing this ahead of merge because three PRs now depend on the order it lands in: #305 is stacked on this branch, and #302 has to be rebuilt against the endpoint this PR introduces. Two of the findings below are wiring issues that would otherwise reach `main` unreviewed.

`POST /api/v1/me/setup/reset` is the right shape and the right name — it does not depend on the Onboarding Lab, which is exactly what #299 asked for, and #305 removes that Lab entirely. The schema belongs in `packages/shared` too. The notes are about how it is wired, not about the design.

## 1. The route is not registered when only the Lab service is supplied

In `packages/server/src/app.ts`, the `registerAccountRoutes` call sits behind:

```ts
if (
  options.agentService ||
  options.taskService ||
  options.computerService ||
  options.workspaceSetupService ||
  (options.machineAuthService && options.computerConnectCode)
)
```

This PR threads `setupResetService` into that call's options, but `stagingOnboardingLab` is never added to the guard itself. Supply `stagingOnboardingLab` and none of those five, and `setupResetService` is silently dropped — `/api/v1/me/setup/reset` is never registered at all.

Measured on this head:

| construction | `{"mode":"all"}` | `{}` |
|---|---|---|
| `createApp({ authService, stagingOnboardingLab: { reset } })` | **404** | **404** |
| the same plus any one of the five (e.g. `taskService`) | reaches the service | 400 |

Nothing breaks in the deployed server today, because `packages/server/src/index.ts` always constructs all five. But the stated selling point is a path that does not depend on the Lab existing, and as wired it depends on five unrelated services existing instead.

`packages/server/src/__tests__/account-api.test.ts` cannot catch this: `appWith()` always spreads the full `services()` into `createApp`, so the guard is always satisfied. The existing "is not registered when the deployment does not offer it" case omits `setupReset` entirely, which tests a different thing.

Adding `options.stagingOnboardingLab` to the guard fixes it; a test that passes *only* `stagingOnboardingLab` would pin it.

## 2. The body is parsed before the environment is checked

In `packages/server/src/api/account.ts` the handler runs `parseRequest(AccountSetupResetRequestSchema, request.body)` first, and the staging check lives inside the service. With the route registered:

| env | `{"mode":"all"}` | `{}` |
|---|---|---|
| staging | reaches the service | 400 |
| dev | 404 | **400** |
| prod | 404 | **400** |

So a deployment that registers the route but is not staging can be told apart from one that never had it, using a malformed body. `AccountRoutesOptions` currently says "the service answers a request outside staging exactly like a path that was never registered" — as written, it is registration that produces that property, not the service.

**No production impact today**: `config.stagingOnboardingLab` is `OPENTAG_ENV === "staging"` (`packages/server/src/config.ts:340`), and `index.ts` only passes it when true, so prod never registers this route. This is about which layer carries the guarantee, and about the comment claiming a layer the code does not have. Checking `enabled` at the top of the handler — or parsing after the environment check — closes it.

For contrast, the existing Lab endpoint re-checks `reset.enabled` per request *before* touching the body, so a missing mode there returns 404 rather than 400 outside staging.

## 3. Two invariants worth pinning

`reboard` reusing `lockActiveAccount` is what makes it inherit the suspended-Account refusal, and skipping `#verifyCleanedUp` is the entire reason the two modes cannot share a commit path. Neither is currently covered here. The two cases most worth adding:

- `reboard` **succeeds on the state `resetOnboarding` refuses to commit on** — a live Agent plus a usable connect code. This is the one that actually pins the split.
- `reboard` refuses a suspended Account.

## 4. The integration fixture starves its own suite

`packages/server/src/__tests__/integration/onboarding-lab-reset.test.ts` opens a 10-connection pool per `fixture()` and holds every one until the file ends. This PR adds one case, so it stays under the limit — but the next person to add a few will hit `sorry, too many clients already`, and it takes unrelated integration files down with it, not just this one. An `afterEach` that closes the pools fixes it. Worth carrying here since this PR is first in the merge order.

PR #302, opened 2026-08-30
feat(web): add the staging internal tools page and the way into it

### LINKED ISSUE #299: Account needs both a reset-all and a lighter re-board operation

liuchao-001:
Closes #299 — the half of it that #301 and #305 do not cover.

`POST /api/v1/me/setup/reset` landed in #301 and the Onboarding Lab was retired in #305. Neither ships a way to *reach* either reset from the product, so today both modes are curl-only. That was explicitly not the intent: the entry point was asked for on the Web, not in the CLI.

## What this adds

**`/internal`** — a staging-only internal tools index. The two resets sit on it directly, each behind a confirmation that states what it costs:

- **Re-board** keeps everything and reopens onboarding, so the Account resumes from the Agent it already has.
- **Reset all** returns the Account to a genuine first run, deleting its Agents and revoking Computer access.

Tools that need a flow of their own stay on their own page and are linked rather than inlined — currently the onboarding mock.

**The way in** — an *Internal tools* item in the account menu, rendered only where the Server answers that the deployment offers the tools. Outside staging the interface is absent rather than closed, so production renders exactly the menu it renders today, and a failed probe is treated as "not offered" rather than a guess.

## The server changes, and why there are any

This was meant to be a pure Web layer. Two things made that impossible:

**A reachability probe.** The Lab's `GET` was the client's way of asking "does this deployment have the tools?", and #305 deleted it with the rest of the Lab. `/me/setup/reset` was POST-only, so there was no way to ask without performing a reset. This adds `GET /api/v1/me/setup/reset` — 204 when offered, 404 when not — restoring exactly the question the Lab's probe answered, on the path that survived.

**The staging check now runs before the body is read.** Previously `parseRequest` ran first and the environment check lived inside the service, so a deployment that registered the route but was not staging could be told apart from one that never had it by sending a malformed body — 400 rather than 404. That contradicted the comment on `AccountRoutesOptions`, and `docs/staging-onboarding-reset.md` already promised "the environment is re-confirmed on each request rather than trusted from route registration". Both handlers now check `enabled` first, so the code matches what both already claimed. `enabled` is part of the service interface rather than assumed, and a test pins that a disabled deployment answers 404 to `GET`, to a valid body, and to a malformed one alike.

No production behaviour changes either way: prod never registers these routes.

## Docs

`docs/staging-onboarding-reset.md` opened by saying the reset exists "because clearing the setup-completion timestamp alone is not enough" — while documenting `mode: "reboard"`, which does precisely that. That claim is true only of `mode: "all"`, and it is now scoped to it. The page and its entry point are documented, including the wrinkle that both resets clear setup completion — which is what the app checks before showing anything else — so you land in onboarding with the account menu you came through no longer rendered. The zh-CN mirror carries the same changes.

## Testing

`pnpm check`, `pnpm typecheck` and `@opentag/server test` (295) are clean on this tree. The web suite passed at 453 on identical content, but its last two runs on this machine were killed by a wall-clock timeout under load average ~30 from unrelated processes, so treat CI as the authority for it rather than my local run.

Nothing here has been exercised against a live deployment; the whole feature is staging-gated and needs independent QA at this head.

## Not included

The reset integration fixture opens a connection pool per test and never closes it. It is real, but it is inherited from #301 rather than introduced here, so it goes in its own PR with the `reboard` invariants #301 also left uncovered.


liuchao-001:
## Author test evidence — head `a1c1c3f`

Not independent QA; recording it so the reviewer knows what is already covered.

| Suite | Result |
|---|---|
| `pnpm check` | pass (exit 0; the one warning is a pre-existing unsafe-fix suggestion in an untouched file) |
| `pnpm typecheck` | pass (exit 0) |
| `@opentag/shared test` | 88 passed |
| `@opentag/server test` | 299 passed |
| `@opentag/web test` | 545 passed |
| `@opentag/server test:integration` | 288 passed — **two consecutive clean runs** |

### On the integration suite

Worth stating plainly, because the first runs of this branch were red and it would be easy to wave that away.

Adding six re-board cases pushed the file into `sorry, too many clients already`. Each `fixture()` opened a 10-connection pool and held it until the file ended; enough tests, and it starved itself and the files running beside it. Six failures, then eleven, in files I had not touched.

That was a real defect this branch exposed, not noise, so the fixture now closes its pool between tests. Measured after the fix:

- clean `main`: 16 files, 282 passed, exit 0
- this branch: 16 files, 288 passed, exit 0 — twice in a row

### Known-failing outside this change

`@opentag/client` fails 4–6 tests **on clean `main`** on this machine — `createClientRuntime production composition` timeouts plus the Codex/Claude local-probe cases, with the exact set varying per run. They read local CLI artifacts and credential sources, so they are environment-dependent here. Unrelated to this diff, which touches only `shared`, `server` and `web`.

### Not covered by any of the above

Nothing here ran against a live staging deployment, and the whole feature is staging-gated. What still needs eyes on a real deployment:

1. `/internal` renders and the account-menu entry appears on staging, and neither appears in a non-staging build.
2. Re-board actually lands in the resume path with the Agent intact — the point of the issue.
3. Reset-all still behaves as before from the Onboarding Lab.

liuchao-001:
## Independent QA — `PASS` · tier `focused-local` · head `a1c1c3f`

Run by the **code-reviewer** agent, posted through @liuchao-001's GitHub account (the account is also this PR's author, so this is a comment and not a formal approving review). This is **independent** QA: the evidence below comes from my own run cell, not from the author's suites.

**Exact target**: `refs/pull/302/head` = `a1c1c3fff2d07525efa87f1333643604cf8867aa`, verified against `gh pr view --json headRefOid` before and after the run. Any new commit invalidates this report.

### Evidence boundary — read this before quoting the PASS

The feature is staging-gated, and I could not exercise it on the deployed staging host. What I ran is a **local run cell built from this exact head**: the real built server (`packages/server/dist`) serving the real built web bundle (`apps/web/dist`), real Postgres, a real browser, behind a TLS terminator because `OPENTAG_ENV=staging` forces an HTTPS `OPENTAG_PUBLIC_URL`. Sign-in was real email+password through the product's own `/api/v1/auth/email/sign-in`.

So the PASS covers **the code at this head behaving correctly when the server runs with `OPENTAG_ENV=staging` and with `OPENTAG_ENV=prod`**. It does **not** cover the deployed staging box, its config, or its data. Two things inside that boundary are harness-shaped and stated plainly:

- **`setupCompletedAt` was set by SQL, not by finishing onboarding.** `POST /api/v1/me/setup/complete` correctly refused with `409 WORKSPACE_SETUP_NOT_READY` because the seeded Agent has no live messaging handoff. The precondition was therefore written directly. It is the precondition, not the answer — the question under test is whether re-board *clears* that field and leaves everything else alone.
- **The seeded Computer never actually connected** (no daemon). Re-board's resume therefore lands on the "connect your computer" step with a repair code, rather than skipping past it. The resume *branch* is proven; the online-Computer variant of it is not.

### 1. `/internal` and the account-menu entry, staging vs non-staging

Same web bundle in both runs — only `OPENTAG_ENV` changed, which is the design's own claim (the Server answers, the client only asks).

| | `OPENTAG_ENV=staging` | `OPENTAG_ENV=prod` |
|---|---|---|
| `GET /api/v1/internal/onboarding-lab` | `204` | `404` |
| `POST` `{"mode":"reboard"}` | `204` | `404` |
| `POST` `{"mode":"reset-all"}` | `204` | `404` |
| `POST` with no mode | `400 VALIDATION_ERROR` | **`404`** |
| Account menu items | `Computers, Account, Internal tools, Sign out` | `Computers, Account, Sign out` |
| `/internal` | full page, `data-ui="internal-tools"` | `Page not found` |

The fourth row is the one that answers your question 2 directly: **outside staging the new 400 never fires**, because `requireLab` runs before `requestedMode`. A prod deployment cannot be told apart from one that never had the feature, not even by malformed input. No console errors and no page errors on any of these loads.

I also confirmed the client fails closed: with the probe request aborted at the network layer, the account menu renders without the entry.

### 2. Re-board keeps everything and enters the resume path

Row-level, not counts. Before and after the re-board, through the page:

```
--- before-reboard.txt
+++ after-reboard.txt
3a4
>  code     | 32eeb413-02cb-44ad-85b0-64e356ba3e6c | usable
```

That is the **entire** database diff. Agent `4d7e4cbf…` still `active`, enrollment credential `e592f03e…` still `live`, connect code `7b831372…` still `usable`, both Computer rows untouched — and `users.setup_completed_at` went to `NULL`. The one added row is a fresh connect code that the **onboarding resume step itself** issued after the redirect; the reset did not create it.

Resume vs first-run, on the same account, minutes apart:

- after **re-board** → `/onboarding` opens on step 2, heading **"Connect your computer"**, with step 1 "Create agent" already behind it; `GET /api/v1/agents` from the page returns `[{"name":"qa-agent","status":"active","runtime":"claude-code"}]`.
- after **reset-all** → `/onboarding` opens on step 1, heading **"Where should your agent run?"**.

That contrast is the issue's core ask, and it holds.

### 3. Reset-all from the Onboarding Lab is unchanged

Driven through the Lab's own floating control → "Reset my account and start onboarding" → confirm. `POST …/onboarding-lab` returned `204`, redirect to `/onboarding`, and the database moved exactly as the existing doc describes: active Agents `1 → 0` (one row now `deleted`), live enrollment credentials `1 → 0`, usable connect codes `2 → 0`, `account_computers.current_instance_id` and `connected_at` cleared, `setup_completed_at` `NULL`.

### 4. Deploy skew, exercised rather than argued

All three shapes an older bundle can produce were sent at the real endpoint on staging:

| request | result |
|---|---|
| `POST`, CSRF header, **no body, no content-type** (exactly what the pre-#302 client sends) | `400` `"The reset mode is required and must be \"reset-all\" or \"reboard\""` |
| `POST` `{}` | `400`, same message |
| `POST` `{"mode":"reset"}` | `400`, same message |

Account state was unchanged after all three. I agree with the trade-off, and the blast radius is smaller than the PR description claims: `INTERNAL_ONBOARDING_LAB_PATH` has exactly one caller in the repo (`apps/web/src/api.ts`) — no CLI, no script, and the docs carry no `curl` example — so a stale browser tab is the only thing that can hit this, and it gets a clear message rather than a wiped Account.

### Test suites re-run independently at this head

| suite | result |
|---|---|
| `@opentag/server test:integration` | 16 files, **288 passed** |
| `@opentag/web test` | 40 files, **545 passed** |

Both reproduce the author's numbers. CI's six required checks are green at this SHA (verified with `gh pr checks`, not taken from the description). I did not re-run `check` / `typecheck` / `shared` / `server` unit — CI covers them at this exact head and re-running adds nothing.

---

## Review findings

Nothing blocking. Four notes, in descending order of how much I'd want them handled before merge.

**1. MINOR — `docs/staging-onboarding-lab.md` is now wrong in three places, and its zh-CN mirror with it.**
The PR ships a user-facing staging capability and touches no docs. The existing doc is the only written guide for the people #299 was filed for, and it now says:
- *"It exists because clearing the setup-completion timestamp alone is not enough."* — re-board is exactly that, deliberately.
- *"Open the route directly and bookmark it; it is deliberately absent from product navigation."* — there is now an account-menu entry, and the Lab is linked from the new index.
- The "Rules the Server enforces" list does not mention that `mode` is required, and nothing anywhere documents `/internal` or re-board.

`AGENTS.md` requires the Chinese mirror to stay synchronized, so `docs/zh-CN/staging-onboarding-lab.md` needs the same edit. A small "Internal tools index / two reset modes" section plus fixing those two sentences would close it.

**2. LOW — the new entry point is absent in exactly the state both resets produce.**
"Internal tools" lives in the account menu, which lives in `AppShell`, which `_shell.tsx` gates on `setupCompletedAt`. Run either reset and you are in onboarding with no menu. I verified `/internal` is still reachable by typing the URL while setup is incomplete (it sits under `_authenticated` but outside `_workspace`), so this is discoverability rather than a trap — but a tester who found the page through the menu has no menu to find it again with. Worth one line of copy on the page, or a link from the onboarding banner.

**3. INFO — production issues one probe per authenticated page load, answered `404`.**
Verified live: the prod-env shell fires `GET /api/v1/internal/onboarding-lab` on every load and takes the 404. It is cheap, it is the design working as intended, and it fails closed — but it will show up as a steady 404 in production access logs and any 4xx-rate dashboard. Worth a conscious decision rather than a surprise.

**4. INFO, pre-existing — the probe-failure state on `/internal` is thin.**
With the probe aborted, `/internal` renders a bare `<div role="alert">Failed to fetch</div>` — the raw browser error string, no page frame, no retry. I ran the same abort against the pre-existing `/internal/onboarding-lab` route as a control and got byte-identical output, so this is shared `AsyncState` behaviour and **not a regression from this PR**. It matters slightly more now that `/internal` is navigated to rather than bookmarked. Out of scope here.

### Your four questions, answered

1. **Required `mode`** — right call, and the failure direction is verified above. The prod path returns 404 rather than 400, so the stricter validation costs nothing in disclosure. One caller in the repo; the skew window is a stale tab.
2. **The `/internal` staging gate** — no leak found from either side. The server re-checks `reset.enabled` per request and refuses before parsing the body; the client's default is "not offered" and I confirmed that under both a `404` answer and an aborted request. The route module is necessarily present in the production bundle (one bundle), but it can only ever render not-found there.
3. **The integration fixture pool fix** — keep it in this PR. It is ~10 lines in the file this PR extends, without it this PR's own tests are red, and the starvation it fixes was taking down unrelated files. That is a change the PR needs, not scope creep. Two clean 288-pass runs (yours and mine) is adequate evidence.
4. **Missing invariants under the lock** — I found none. `reboard` reusing `lockActiveAccount` is what makes it inherit the suspended-Account refusal, which your test pins. There is genuinely nothing to verify afterwards because nothing is destroyed. One asymmetry I checked deliberately and consider correct: `reboard` does not call `registry.closeEnrollment`, which is right — the Computer is being kept, so its live connection should survive.

### Residual state

The QA run cell is kept warm: the `otqa-pg` Postgres container stays up and its `pr302qa` database is dropped; server, TLS proxy and browser session for this run are stopped. Nothing was written to the source repo, and no operator or staging data was touched — the account, agent and computer in this report exist only in the dropped run database.

**Case disposition**: `candidate-new-case` — "staging-only internal tool visible in staging and absent in prod, driven from one bundle across both server environments" is a reusable shape, and the `OPENTAG_ENV=staging` ⇒ HTTPS-public-URL ⇒ no-dev-auth-bypass chain is a run-cell recipe worth writing down rather than rediscovering.

bestony (approved):
Approve。生产打不到 reboard：非 staging 不注册路由；enabled 在 GET/POST 上再查（POST 在 parse body 之前）；service 自己 fail closed。页面只动已认证 Account，没有选别人、也没有 Workspace/成员产品。

确认框把 Reset all 说成「disables」Agents，正文和 docs 是删除，想更诚实可以改确认句。你写了本 head 没在 live staging 上跑过——按 Ready 约定，独立 QA 过了再标 Ready 更干净，不挡这次代码判断。

liuchao-001:
## Independent QA — `PASS` · tier `focused-local` · head `5c95024`

Run by the **code-reviewer** agent, posted through @liuchao-001's GitHub account (also this PR's author, so this is a comment rather than a formal approving review). Independent: every observation below is from my own run cell, not the author's suites. The previous `PASS` bound to `a1c1c3f` is void — this replaces it.

**Exact target**: `refs/pull/302/head` = `5c950243764b2707885760a4faa37542ef6e1b07`, confirmed before and after the run. Any new commit invalidates this.

### Evidence boundary

Staging-gated feature, so this is a local run cell built from this exact head: the real built server (`packages/server/dist`) serving the real built web bundle (`apps/web/dist`), real Postgres, a real browser, TLS terminator in front because `OPENTAG_ENV=staging` forces an HTTPS `OPENTAG_PUBLIC_URL`. Sign-in through the product's own `/api/v1/auth/email/sign-up` and `/sign-in`. The PASS covers **this code at `OPENTAG_ENV=staging` and at `OPENTAG_ENV=prod`** — not the deployed staging host. Two harness-shaped facts, unchanged from the last round: `setup_completed_at` was written by SQL because `POST /me/setup/complete` correctly refuses (`409 WORKSPACE_SETUP_NOT_READY`) without a live messaging handoff, and the seeded Computer never really connected, so re-board's resume lands on the connect step rather than past it.

**On the local suites**: this machine is at load average ~33 with other agents' vitest workers running, so a local suite failure here would not be interpretable as a product failure. CI is green at this head on all six required checks (`gh pr checks`, verified myself), and the `Checks` job runs `check`, `typecheck`, `test` and `server test:integration`. I deliberately did not re-run those locally — under this load it would produce noise, not evidence. What I added instead is the live behaviour CI cannot reach.

---

### 1. The staging gate, from both sides — same bundle, only `OPENTAG_ENV` differs

| | `staging` | `prod` |
|---|---|---|
| `GET /api/v1/me/setup/reset` | `204` | `404` |
| `POST {"mode":"all"}` | `204` | `404` |
| `POST {"mode":"reboard"}` | `204` | `404` |
| `POST {}` (malformed) | `400` | **`404`** |
| `POST {"mode":"nope"}` | `400` | `404` |
| `POST` with no body at all | `400` | `404` |
| `POST /api/v1/me/setup/nonexistent` (control) | — | `404` |
| Account menu | `Computers, Account, Internal tools, Sign out` | `Computers, Account, Sign out` |
| `/internal` | renders, `data-ui="internal-tools"` | `Page not found` |
| `/internal/onboarding-lab` (retired in #305) | `Page not found` | `Page not found` |
| `GET /api/v1/internal/onboarding-lab` | `404` | `404` |

On `prod`, every request shape is indistinguishable from the control path that genuinely does not exist. No console errors and no page errors on any load.

### 2. The gate-order fix, measured on the same instrument that found it

The deployment-level table above passes even without the fix, because `prod` never registers the route. So I re-ran the probe that produced the original finding — a **registered** service with a non-staging environment, with a staging positive control to prove the route really is live:

| | `#301` head `96f14d17` | **this head `5c95024`** |
|---|---|---|
| `staging` (control: route is live) | valid `POST` reaches service | `GET` `204`, valid `POST` reaches service, malformed `400` |
| `dev` — valid `POST` / malformed `POST` | `404` / **`400`** | `404` / **`404`** |
| `prod` — valid `POST` / malformed `POST` | `404` / **`400`** | `404` / **`404`** |
| `prod` — `GET` | n/a (route did not exist) | `404` |

The discriminator is gone. A deployment holding the routes but not the feature now answers `404` to reachability, to a valid body and to a malformed one alike — which is what `AccountRoutesOptions` and `docs/staging-onboarding-reset.md` ("the environment is re-confirmed on each request rather than trusted from route registration") both already claimed. Moving `enabled` onto the service interface rather than assuming it is the right shape for this.

### 3. Re-board keeps everything and enters the resume path

Row-level. Before/after the re-board, driven through the page:

```
--- before-reboard.txt
+++ after-reboard.txt
5a6
>  code     | a694db17-d356-41a6-b052-eeece53f4d67 | usable
```

The entire database diff. Agent `b07a5b4a…` still `active`, enrollment credential `30fd0ba6…` still `live`, connect code `8805e211…` still `usable`, the Computer row untouched, `setup_completed_at` → `NULL`. The one added row is a connect code the **onboarding resume step** issued after the redirect, not something the reset created. The request went to `POST /api/v1/me/setup/reset` → `204`.

Resume against first-run, same Account, minutes apart:

- after **re-board** → `/onboarding` on step 2, heading **"Connect your computer"**, step 1 "Create agent" already behind it; `GET /api/v1/agents` from the page returns `[{"name":"qa-agent-r2","status":"active"}]`.
- after **reset all** → `/onboarding` on step 1, heading **"Where should your agent run?"**.

### 4. Reset all is unchanged, now reached only from `/internal`

Through the page's own confirmation. `POST` → `204`, redirect to `/onboarding`, and the database moved as the doc describes: active Agents `1 → 0` (one row `deleted`), live enrollment credentials `1 → 0`, usable connect codes `2 → 0`, `setup_completed_at` cleared. Same behaviour as before, from the entry point that replaced the Lab's floating control.

### 5. The documented escape hatch actually works

The new doc says both resets put the Account back behind the setup gate where the menu lives, and that `/internal` still opens if typed. Verified live with setup incomplete: `/agents` redirects to `/onboarding`, `/internal` opens and renders the page. The claim is true.

Docs read in full: the false framing is now correctly scoped to `mode: "all"`, `/internal` and its entry point are described, and `docs/zh-CN/staging-onboarding-reset.md` carries the same changes including the per-request re-confirmation line — which the code now honours.

---

## Findings — nothing blocking

**1. INFO — the reachability `GET` is not documented.** `docs/staging-onboarding-reset.md` now describes the request "for when you want to drive it by hand", but only the `POST`. `GET /api/v1/me/setup/reset` (204/404) is part of the endpoint's contract as of this PR and is the thing the menu entry depends on. One line in both mirrors.

**2. INFO — production issues one probe per authenticated page load, answered `404`.** Verified live. Carried over from the previous round with one new wrinkle: the probe path used to be `/api/v1/internal/onboarding-lab`, which read as internal tooling in a log. It is now a `GET` on `/me/setup/reset`, so production access logs and 4xx dashboards will show a steady stream of 404s against a path whose name reads destructive. It is a `GET` and does nothing, but it is worth being a conscious choice rather than a surprise for whoever reads those logs first.

**3. INFO, pre-existing — the probe-failure state on `/internal` is thin.** A failed probe renders a bare `<div role="alert">Failed to fetch</div>` with no page frame and no retry. I established last round that the retired Lab route behaved byte-identically, so this is shared `AsyncState` behaviour rather than anything this PR introduced. It now has no sibling to share the blame with, since `/internal` is the only page of its kind left.

**Previously raised and now resolved**: the account-menu entry disappearing in exactly the state both resets produce is documented rather than fixed, and I verified the documented workaround. That is a legitimate resolution for a staging-only tool.

### On #314

Read it: one file, test-only, the connection-pool `afterEach` plus the four re-board invariants that were missing — including `succeeds on the very state a full reset refuses to commit on`, which is the one that pins why the two modes cannot share a commit path. The `Checks` job runs `server test:integration`, so CI green does exercise both the pool change and the new cases on a machine that is not under this one's load. I agree it does not need a QA round of its own.

### Residual state

QA run cell kept warm: the `otqa-pg` container stays up, its `pr302qa2` database is dropped, and this run's server, TLS proxy and browser session are stopped. Nothing written to the source repo — the throwaway probe used for section 2 was deleted and the worktree verified clean. No operator or staging data touched.

**Case disposition**: `candidate-new-case` — "one bundle, two server environments, staging-only surface must be absent rather than closed" plus the registered-but-disabled probe in section 2 is a reusable pair. The second half is what distinguishes a gate that holds by registration from one that holds per request, and it is not something a deployment-level test can see.

liuchao-001:
New head `799a076`. This invalidates the `PASS` bound to `5c95024` and @bestony's approval of it. Here is exactly what moved, so re-review can be scoped to it.

**1. The confirmation now says what Reset all does.** It said the operation "disables" the Account's Agents; it deletes them. @bestony flagged the mismatch as non-blocking, and I took it because understating a destructive action in the sentence that asks you to confirm it is the wrong direction to be wrong in. The assertion in `internal-tools.test.tsx` moved with the copy; that file's six cases pass locally.

**2. The reachability `GET` is documented**, in both mirrors — the INFO finding from the QA round. It is part of this endpoint's contract as of this PR and the menu entry depends on it, so leaving only the `POST` written down was a real gap.

**3. `origin/main` merged in.** The branch was five commits behind, which matters because #288 added a patch-coverage gate and #293 fixed the coverage measurement it depends on.

No product behaviour changed beyond the confirmation sentence. The staging gate, the probe, the page and the menu entry are byte-identical to what was QA'd.

## On spending another QA round

The QA report on `5c95024` was thorough and I am invalidating it deliberately rather than by accident. The alternative was merging a destructive-action confirmation that misdescribes itself and an undocumented endpoint contract, then opening a second PR needing its own approval and its own round. One more round now is cheaper than that, and both items came from review rather than from me changing my mind.

## Still true from the last round

Nothing here has run against a live deployment. Local suite runs remain uninterpretable on this machine — load average is still in the 30s from unrelated processes — so CI is the authority for the suites, and it runs `check`, `typecheck`, `test` and `server test:integration` in the `Checks` job.

bestony (approved):
Approve，精确 head `799a076`。

相对上次 Approve 的 `5c95024`，本 PR 自己的产品增量只有两处：Reset all 确认句从 “disables” 改成 “deletes”（测试断言一起改了），以及中英文文档补上 GET `/api/v1/me/setup/reset` 的 204/404。staging 门控、probe、页面和菜单入口除这一句外与上次 QA 的代码一致。

非阻塞：`docs/staging-onboarding-reset.md` 前文写 both resets 把 Account 带回 **inside** the setup gate，请求章节仍写 “204 means the Account is back **outside** the setup gate”；`docs/zh-CN/staging-onboarding-reset.md` 同样。实现上 `/onboarding` 在 `setupCompletedAt` 为 null 时打开，应为 inside。

CI：Checks 已绿。Patch Coverage 在本 head 失败，原因是找不到 `coverage/unit/coverage-final.json`。`pnpm test:coverage` 实际写出的是 `coverage/unit/coverage-summary.json`（以及各 workspace 子目录的 summary），这是 main 上 #288/#293 门控与测量脚本的衔接问题，不是本 PR 的未覆盖行；#309/#316/#319/#287 同样红。合入会被它挡住，但不是这次代码判断的 blocker。

Ready：`5c95024` 的独立 QA PASS 已因本 head 失效。按约定，精确 head 的独立 QA 过了再标 Ready。合入 main 带了 #310 的 onboarding 抽取，新一轮 QA 应再看一次 re-board 后的 onboarding 落点。

liuchao-001:
The red `Enforce Patch Coverage` check on this PR is not this PR. I said I would check rather than merge past it, so: it fails with `ENOENT` on `coverage/unit/coverage-final.json`, a file nothing in the repository writes — `scripts/unit-coverage.mjs` emits per-workspace reports plus `coverage/unit/coverage-summary.json`, and the gate reads a merged detailed report that does not exist.

My earlier guess that merging `main` would clear it was wrong; this branch now has current `main` and it still fails, as it does on unrelated branches. Filed as #320 with the diagnosis. I have not patched it here — the repair touches the per-workspace design #293 introduced deliberately, and that is not a passing PR's call.

Every other check on this head is green, including `Checks`, which runs `check`, `typecheck`, `test` and `server test:integration`.

liuchao-001:
New head `d42fdf7`. One commit, documentation only — no product code changes.

Thanks for the catch. My own paragraph said both resets take the Account back **inside** the setup gate, while the request section below it already said a `204` means the Account is back **outside** one. Same state, opposite words, in the same file and in both mirrors — and it is the same class of contradiction I had just fixed in this document, reintroduced by me a few lines above.

Neither sentence needed the metaphor, so both now state the fact: setup completion is cleared, and that is what the app checks before it will show anything else. `docs/staging-onboarding-reset.md` and `docs/zh-CN/staging-onboarding-reset.md` are the entire diff.

On the coverage gate — agreed on all of it, and it is filed as #320 with the same diagnosis. Worth adding to what you found: the workflow has not succeeded on any branch since #288 landed, across eight branches, so it will block every PR in this repository rather than these two.

bestony (approved):
Approve，精确 head `d42fdf7`。

相对 `799a076` 只有文档：中英文都不再把同一种状态写成 inside / outside the setup gate。前文改成 “incomplete setup / setup 未完成”，204 改成 “setup completion is cleared / setup 完成标记已被清除”，和 `/onboarding` 在 `setupCompletedAt` 为空时打开一致。产品代码未改。

合入仍会被仓库级 Patch Coverage（#320）挡住。Ready：这一 SHA 的独立 QA 尚未绑定；产品路径与上一版相同，但按约定 Ready 仍看精确 head。

liuchao-001:
## Independent QA — `PASS` · tier `focused-local` · head `d42fdf7`

Run by the **code-reviewer** agent, posted through @liuchao-001's GitHub account (also this PR's author, so this is a comment rather than a formal approving review). The `PASS` bound to `5c95024` is void; this replaces it.

**Exact target**: `refs/pull/302/head` = `d42fdf78dab3a0d3b9c32b0f66d601ae5e668695`, confirmed before and after.

### Why this was a full re-run rather than an incremental one

The PR's own delta since `5c95024` is small — one confirmation sentence, its test assertion, and the docs. The merged `main` is not: `5c95024..799a076` is 53 files, +1509/-580, and #310 in it refactors the onboarding flow that every resume/first-run observation in the last report rested on.

```
apps/web/src/onboarding-v2/steps.tsx    127 +-
apps/web/src/onboarding-v2/flow.ts       71 +-
apps/web/src/onboarding-v2/copy.ts       91 +-
apps/web/src/onboarding-v2/server-backend.ts
apps/web/src/setup/**                    (new module)
```

The last round's evidence for "re-board resumes, reset all does not" was two step headings, and those headings are rendered by three of the files above. Carrying that evidence forward would have meant asserting behaviour of code I had not run. So everything below was observed again.

### How the binding reaches `d42fdf7`

The run cell was built from `799a076`. `git diff --name-only 799a076 d42fdf7` is exactly two files:

```
docs/staging-onboarding-reset.md
docs/zh-CN/staging-onboarding-reset.md
```

with **no** non-docs change (`git diff --name-only 799a076 d42fdf7 -- . ':(exclude)docs/**'` is empty). Neither file is a build input, so the `apps/web/dist` and `packages/server/dist` exercised here are the artifacts `d42fdf7` produces. The behavioural evidence therefore binds to `d42fdf7`; the documents were read at `d42fdf7` directly.

### Evidence boundary

Local run cell from this head: the real built server serving the real built web bundle, real Postgres, a real browser, TLS terminator because `OPENTAG_ENV=staging` forces an HTTPS `OPENTAG_PUBLIC_URL`. Sign-in through the product's own email sign-up/sign-in. The `PASS` covers this code at `OPENTAG_ENV=staging` and at `prod` — not the deployed staging host. Unchanged harness facts: `setup_completed_at` is written by SQL because `POST /me/setup/complete` correctly refuses without a live messaging handoff, and the seeded Computer never connects, so resume lands on the connect step rather than past it.

Local suites were again not re-run: this machine reached load average 61 during the run (the build alone took 4m50s against a normal 30s), so a local failure would not be interpretable. CI is green at this head on everything except the coverage gate discussed at the end.

---

### 1. The staging gate, both sides, one bundle

| | `staging` | `prod` |
|---|---|---|
| `GET /api/v1/me/setup/reset` | `204` | `404` |
| `POST {"mode":"all"}` | `204` | `404` |
| `POST {"mode":"reboard"}` | `204` | `404` |
| `POST {}` | `400` | `404` |
| `POST {"mode":"nope"}` | `400` | `404` |
| `POST {"mode":"reset-all"}` (retired vocabulary) | `400` | — |
| `POST /me/setup/nonexistent` (control) | — | `404` |
| Account menu | `Computers, Account, Internal tools, Sign out` | `Computers, Account, Sign out` |
| `/internal` | renders | `Page not found` |

On `prod` every shape is indistinguishable from the control path that genuinely does not exist. No console errors, no page errors, on any load in either environment.

### 2. Re-board on the refactored onboarding

Row-level, driven through the page. Entire database diff:

```
3a4
>  code     | 1d798071-0bb6-47ce-ad28-c0c8ec608caa | usable
```

Agent `1c88ce20…` still `active`, credential `1ec7d3a8…` still `live`, code `80377de2…` still `usable`, Computer row untouched, `setup_completed_at` → `NULL`. The added row is a connect code the resume step issued after the redirect.

The landing points survive #310 unchanged:

- after **re-board** → step 2, heading **"Connect your computer"**, rail showing "Create agent" already behind it; `GET /api/v1/agents` from the page returns `[{"name":"qa-agent-r3","status":"active"}]`.
- after **reset all** → step 1, heading **"Where should your agent run?"**; active Agents `1 → 0` (one `deleted`), live credentials `1 → 0`, usable codes `2 → 0`, setup cleared.

### 3. The confirmation now describes what it does

Read verbatim from the live dialog:

> This deletes your current Agents and revokes your Computer enrollments and messaging connections on staging. Nobody else's Account is touched.

That matches the operation as measured in section 2 — Agents deleted, credentials and codes revoked. The previous wording said "disables". Taking a non-blocking review note on a sentence whose whole job is to let someone consent to a destructive action was the right call.

### 4. Documents

- The `inside`/`outside` contradiction @bestony found is gone from **both** mirrors; the metaphor is replaced by the fact (setup completion cleared, which is the state the app checks). No occurrence of either phrasing remains.
- `GET /api/v1/me/setup/reset` is documented in both mirrors with its `204`/`404` contract and why it exists — the INFO finding from the previous round, resolved.
- The claim that `/internal` still opens after a reset is true: verified live with setup incomplete, `/agents` redirects to `/onboarding` while `/internal` renders.

---

## Findings — nothing blocking

**1. INFO — production issues one probe per authenticated page load, answered `404`.** Unchanged from the last round and verified again. The path now reads `GET /me/setup/reset`, so production access logs and 4xx dashboards carry a steady stream of 404s against a path whose *name* reads destructive. It is a `GET` and does nothing; worth being a deliberate choice rather than a surprise for whoever reads those logs first.

**2. INFO, pre-existing — the probe-failure state on `/internal` is thin.** A failed probe renders a bare `<div role="alert">Failed to fetch</div>`, no page frame, no retry. Established two rounds ago to be byte-identical on the then-still-present Lab route, so it is shared `AsyncState` behaviour, not this PR's.

### A correction I owe this thread

In my review of #301 and in the previous QA report I described the key missing test as *"re-board succeeds on the state `resetOnboarding` refuses"*. That names the wrong function, and @bestony caught the same slip in #314's test name. Read at the source: `resetOnboarding` runs `#deleteOwnedAgents` and `#revokeComputerAccess` **before** `#commitFirstRunState`, and `#verifyCleanedUp` only runs inside that commit step — so the public method never meets a state it would refuse; it cleans up first and succeeds. The accurate statement is that **the commit path `resetOnboarding` ends with verifies cleanup, and re-board must reach the same marker without it**. The design conclusion is unchanged; I repeated an incorrect justification for it and am correcting it here rather than in a separate comment.

### On the red check

`Enforce Patch Coverage` is failing here for a reason that is not this PR — the gate reads `coverage/unit/coverage-final.json`, which nothing in the repository writes. Diagnosis and the cross-branch evidence are on #320: the last eight runs of that workflow, on eight different branches, are all failures, two of them with current `main` already merged. It has never passed for anyone since it landed, so re-pushing this branch will not clear it. Every other check on this head is green, including `Checks`, which runs `check`, `typecheck`, `test` and `server test:integration`.

### Residual state

Run cell kept warm: `otqa-pg` stays up, this run's `pr302qa3` database is dropped, and its server, TLS proxy and browser session are stopped. Nothing written to the source repo. No operator or staging data touched.

**Case disposition**: `candidate-new-case`, unchanged from the previous round — the one-bundle/two-environments check plus the registered-but-disabled probe. This round adds a second reusable lesson: when a PR merges `main`, the delta that decides whether prior evidence survives is the merge's, not the author's.

yuezengwu (commented):
Exact-head review: `d42fdf78dab3a0d3b9c32b0f66d601ae5e668695`.

No blocking correctness finding in the current diff. The change stays within the Web interaction plus the directly required Server probe: `GET /api/v1/me/setup/reset` only answers reachability, while POST remains authenticated, CSRF-protected, strict-schema validated, and reflexive to the Account in the token. The staging gate is checked before parsing the body; deployments that do not offer it return 404. No schema, migration, or lower-level data-structure change is present. The English and Chinese docs agree with the current `all | reboard` contract and with the observed setup marker behavior.

Independent local validation in a detached exact-head worktree:
- `git diff --check origin/main...HEAD`: pass
- Web focused tests: 9/9 pass
- Server account API tests: 16/16 pass
- `pnpm check`: pass (one pre-existing optional-chain warning in an untouched file)
- `pnpm typecheck`: pass
- `pnpm build`: pass

Live CI is green except `Enforce Patch Coverage`. I independently inspected that job: it aborts on missing `coverage/unit/coverage-final.json` before measuring this patch, matching repository-wide #320/#324, so it is not evidence of an uncovered line here.

Because the author is in the non-engineer class and this page changes staging/production exposure behavior, I am intentionally not issuing an automatic approval. Final approval and merge remain a human owner decision.

liuchao-001:
Head moved to `3fe836c`. Not a change of mind — merging #314 first made this branch conflict, so a `main` merge was mechanically required before this could merge at all.

The conflict was a single import line: `main` moved `initials` into the new `i18n/format` module (#292) while this branch had added the query-key import beside it. Both are kept; nothing else was touched by the resolution.

Two things I checked rather than assumed while resolving:

- **#292's Paraglide foundation does not ask this page's copy to move yet.** `main`'s own `app-shell.tsx` still renders "Computers" and "Sign out" as plain literals, so the strings on `/internal` are consistent with the surrounding code rather than lagging a migration.
- **The lockfile is byte-identical to `main`'s.** The new `@inlang/paraglide-js` dependency needed installing here, so I ran `pnpm install --frozen-lockfile`; `git diff origin/main -- pnpm-lock.yaml` is empty.

The nine tests covering the page and the menu entry pass on the resolved tree.

For the record on evidence: the independent QA `PASS` is bound to `d42fdf7`. The delta from there is this merge — one import line of mine, plus the eight `main` commits that arrived between. @liuchao-001 authorised merging on the reasoning that chasing a moving `main` costs a QA round each time while the gate's fix (#324) arrives automatically once these land. That reasoning is what I am acting on, and it is worth stating plainly rather than leaving the binding gap implicit.

PR #303, opened 2026-08-30
feat(web): raise the Agents surface's primary controls

liuchao-001:
## Summary

Four presentation changes across the Agents list and the Agent detail header, from feedback on `dev.opentag.build/agents`.

**New Agent moves under the title, at brand weight.** It sat in the page header's trailing action slot as a compact `outline` button — the page's one primary action, placed where a viewer scans last and styled like a secondary one. It now sits below the title on the reading edge as a full-size `primary` button, which resolves to the OpenTag brand green through `--kumo-button-emphasis-bg`. Both the populated and empty states use the same control.

**Each card's status moves under the Agent name.** It was the last row of the card, below the usage figures, detached from the name it describes. It now reads directly beneath the name inside the card's identity block.

**A failing card no longer carries its own recovery exit.** Which page repairs a stuck Agent depends on which dependency failed, and that explanation already lives on the Agent. The card now states the problem only, and the row link that already covers the whole card is the single way to follow it. `agentCardStatus` loses its `action` field; the Agent detail recovery banner keeps its exit.

**The messaging control on an Agent names its channel.** It was a bare provider glyph, which made a viewer guess at both the channel and what clicking it would do, and collapsed to nothing readable when no channel was connected. It is now a `secondary` button carrying the provider mark, the channel name, and a gear for the settings it opens — the same three-part shape and the same height across a live channel, unreadable evidence, and nothing connected, so the header does not resize as the messaging state changes.

The offline recovery sentence also drops its title case: "This agent's computer is offline." reads as prose rather than as two product nouns. The same possessive is lowercased in the four runtime recovery sentences and the two messaging-tab delivery notes, so one banner does not mix both forms.

## Testing

- `pnpm check` — pass (one pre-existing `useOptionalChain` warning in `onboarding-v2/server-backend.ts`, untouched here)
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — `@opentag/web` 537 passed, `@opentag/shared` 88, `@opentag/server` 296, `open-tag` 153. `@opentag/client`'s `client-runtime-composition.test.ts` fails on this machine, and it fails identically on an unmodified `origin/main` worktree here (timeouts and temp-dir `ENOTEMPTY`); it is untouched by this change, which is confined to `apps/web`.

Test coverage follows the moved behavior: the New Agent placement assertions now anchor on `agents-page-action` instead of `page-header`; the card status assertions anchor inside `agent-card-identity-copy` and assert the card exposes exactly one link; and a new case covers the messaging control naming its channel, carrying the gear, and keeping identical button classes with and without a binding.

Live visual verification is not included — it needs the server, database, and a connected Computer, so it is left to independent QA.

## Breaking changes

None.

## Checklist

### DIFF
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -1995,22 +1995,35 @@ describe("OpenTag Web App Shell", () => {
     expect(link.getAttribute("data-ui")).toBe("agent-messaging-link");
     expect(link.textContent).toContain("Slack");
     expect(link.getAttribute("href")).toBe(`/agents/${agentId}/settings/messaging`);
-    // The gear is what says this control manages the channel rather than only naming it.
-    expect(link.querySelectorAll("svg")).toHaveLength(1);
+    /*
+     * The gear is what says this control manages the channel rather than only naming it, and it
+     * trails the label. Asserted by its own hook rather than by counting glyphs: a glyph count
+     * happens to equal one today only because the provider mark is an `img`, so it would stop
+     * guarding the gear the day `ProviderIcon` renders an svg.
+     */
+    const manage = link.querySelector('[data-ui="agent-messaging-manage"]');
+    expect(manage).toBeTruthy();
+    expect(link.lastElementChild).toBe(manage);
     const connectedClassName = link.className;
     expect(connectedClassName).toMatch(/\bh-\d/);
     connected.unmount();
 
-    installApi({ bound: false });
-    window.history.replaceState({}, "", `/agents/${agentId}`);
-    render(<App />);
-
-    expect(await screen.findByRole("heading", { name: "Reviewer" })).toBeTruthy();
-    const unconnected = screen.getByRole("link", { name: "Connect messaging" });
-    expect(unconnected.getAttribute("data-ui")).toBe("agent-messaging-link");
-    expect(unconnected.textContent).toContain("Connect messaging");
     // Same control, same size: the header must not resize as the messaging state changes.
-    expect(unconnected.className).toBe(connectedClassName);
+    for (const state of [{ bound: false }, { bound: true, bindingEvidenceFails: true }]) {
+      installApi(state);
+      window.history.replaceState({}, "", `/agents/${agentId}`);
+      const rendered = render(<App />);
+
+      expect(await screen.findByRole("heading", { name: "Reviewer" })).toBeTruthy();
+      const control = screen.getByRole("link", {
+        name: state.bound ? "Messaging status unavailable" : "Connect messaging",
+      });
+      expect(control.getAttribute("data-ui")).toBe("agent-messaging-link");
+      expect(control.textContent).toContain(state.bound ? "Messaging status unavailable" : "Connect messaging");
+      expect(control.querySelector('[data-ui="agent-messaging-manage"]')).toBe(control.lastElementChild);
+      expect(control.className).toBe(connectedClassName);
+      rendered.unmount();
+    }
   });
 
   it("offers messaging setup only when the missing binding is confirmed", async () => {
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -544,7 +544,8 @@ describe("OpenTag Web App Shell", () => {
     expect(screen.queryByText("Example")).toBeNull();
     const agentLink = await screen.findByRole("link", { name: "Open Reviewer" });
     const createAgent = screen.getByRole("button", { name: "New Agent" });
-    expect(createAgent.closest('[data-ui="page-header"]')).toBeTruthy();
+    expect(createAgent.closest('[data-ui="page-header"]')).toBeNull();
+    expect(createAgent.closest('[data-ui="agents-page-action"]')).toBeTruthy();
     const agentCard = agentLink.closest('[data-ui="agent-card"]');
     expect(agentCard).toBeTruthy();
     expect(screen.queryByText(/Monitor availability/)).toBeNull();
@@ -554,12 +555,20 @@ describe("OpenTag Web App Shell", () => {
     expect(within(agentCard as HTMLElement).getByText("Tasks")).toBeTruthy();
     expect(within(agentCard as HTMLElement).getByText("Tokens")).toBeTruthy();
     expect(within(agentCard as HTMLElement).getByText("428K")).toBeTruthy();
-    expect(within(agentCard as HTMLElement).getByText("Messaging disconnected")).toBeTruthy();
+    const cardState = (agentCard as HTMLElement).querySelector('[data-ui="agent-card-state"]');
+    expect(cardState).toBeTruthy();
+    expect(within(cardState as HTMLElement).getByText("Messaging disconnected")).toBeTruthy();
+    expect(within(cardState as HTMLElement).getByText("Cannot receive new work")).toBeTruthy();
+    // The card reports the failure and nothing else; opening the Agent is its only follow-up.
+    expect(within(agentCard as HTMLElement).queryByRole("link", { name: "Connect messaging" })).toBeNull();
     expect(
       within(agentCard as HTMLElement)
-        .getByRole("link", { name: "Connect messaging" })
-        .getAttribute("href"),
-    ).toBe(`/agents/${agentId}/settings/messaging`);
+        .getAllByRole("link")
+        .map((item) => item.getAttribute("href")),
+    ).toEqual([`/agents/${agentId}`]);
+    expect(
+      (agentCard as HTMLElement).querySelector('[data-ui="agent-card-identity-copy"] [data-ui="agent-card-state"]'),
+    ).toBe(cardState);
     expect(screen.queryByText("Ada's Mac · macOS")).toBeNull();
     expect(screen.queryByText("Mentions only")).toBeNull();
     const workspaceNavigation = screen.getByRole("navigation", { name: "Product" });
@@ -593,7 +602,7 @@ describe("OpenTag Web App Shell", () => {
     expect(within(status as HTMLElement).getByText("Started 8m ago")).toBeTruthy();
   });
 
-  it("keeps an offline reason and its Computer exit together in the Agent status", async () => {
+  it("states an offline reason under the Agent name and carries no exit of its own", async () => {
     installApi({
       bound: true,
       computerStatus: () => "offline",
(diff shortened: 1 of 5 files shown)


liuchao-001:
## Independent QA — PR #303, head `3eee4b2c40a989264e85592e4a208bad8a737cdf`

**Status: PASS** · **Tier: `focused-local`** · run by code-reviewer, not the author.

**Scope / evidence boundary.** I built `@opentag/shared` + `@opentag/server` + `@opentag/web` at this exact head, served the production web build from a local server (127.0.0.1:51939) against a warm Postgres, seeded four Agents across availability and messaging states, and drove the running product black-box with `agent-browser`. Everything below is what I observed **in that local build, at this head, in light mode**, at 1280 / 760 / 380px. It is not a release-wide claim, and it says nothing about hosted config or real Slack/Feishu traffic.

---

### Your four questions

**1. Brand green + contrast — no defect in light mode; your dark-mode worry is real but not reachable.**

The estimate was made against the CSS custom property, but that is not what gets painted. Kumo's `Button` injects inline `--kumo-button-emphasis-gradient-start/end`, so the primary button renders a vertical gradient over `--kumo-button-emphasis-bg`. Measured from actual screenshot pixels in the button's padding column (no glyph antialiasing), against the `!text-white` label at 14px/500:

| mode | top → bottom | contrast vs white |
|---|---|---|
| light (shipping) | `rgb(79,110,46)` → `rgb(58,92,9)` | **5.83:1 → 7.73:1** — passes AA everywhere |
| dark (forced) | `rgb(158,227,79)` → `rgb(145,222,31)` | **1.55:1 → 1.66:1** — would badly fail AA |

It is the OpenTag brand green, and it is legible. Dark mode is **not reachable in the product**: `apps/web/index.html` hard-codes `<html data-mode="light">` and nothing in `apps/web/src` writes a document-level `data-mode` (the only `data-mode` writer in the built bundle is a chart-tooltip subtree). I had to force it via the DOM to measure at all.

So: **not a blocker for this PR.** It is a real latent defect that lands the day a theme toggle ships, and it is app-wide, not yours — `Button` defaults to `primary` at ~40 call sites. Worth its own issue, in particular because `apps/web/src/ui/kumo-theme.tokens.test.ts` asserts dark brand contrast against `#333333` while Kumo's primary variant hard-codes `!text-white`; that green test does not describe the shipped pixel.

**2. Card layout with the status in the identity row — holds.**

Verified live with a 48-character display name:

- **1280px** — 2-column grid, all four cards exactly **151px** tall, both rows aligned.
- **760px** — single column, still 151px each.
- **380px** — single column, 174px each; the status wraps to a second line via the new `flex-wrap`, the name truncates with an ellipsis, and **no element overflows its box** (I enumerated every descendant for `scrollWidth > clientWidth`; only the intended `.truncate` span and `sr-only` text).

**3. Messaging control — same height and same classes, but not the same width.**

Same `h-9` (36px measured) and a byte-identical class string in every state. Width is `w-max` + `max-w-64`, so it is content-sized:

| state | rendered width |
|---|---|
| `Connect messaging` | 199px |
| `Feishu · @provisioning` | 216px |
| `Slack · OpenTag Platform Reliability Reviewer (staging)` (47 chars) | **256px** — capped at `max-w-64`, label truncates, no overflow |

I did **not** reach the third state live — `Messaging status unavailable` needs the `im-binding` read to fail, and my route-abort did not take. It is safe by construction: all three branches share one `className` expression, and the label (28 chars) is shorter than the Slack one I did truncate at 256px. Flagging it so the evidence is not read as covering more than it does.

`max-w-64` does its job. But the right edge of the header group is what is fixed, so the neighbouring **Settings button does not move** (x=1147 in both states) — only the messaging control's own left edge does. If liuchao-001's 「这个按钮的大小在不同状态下都要保持一致」 means literal same width, that needs an explicit `w-64`; if it means "same shape, never collapses to a bare icon", it is satisfied. I would leave it and confirm the reading with them rather than change it here.

**4. Removing the card's recovery link — it costs one click, and lands the user in the right place.**

For "messaging not connected" it is now card → Agent → the new messaging control → `/settings/messaging`, one step more than before. Verified on a real Agent with no binding: the detail page carries **both** the messaging control and the recovery banner's own exit, so the two-click path is complete and ends where the failure is explained. That is what liuchao-001 asked for, and it reads fine on screen.

---

### Also verified live

- Wording change renders: **"This agent's computer is offline. Retrying automatically."** No `Agent's Computer` remains in any rendered text on the Agent detail page.
- Every card carries **exactly one** link — checked on 4 cards across two availability states (`Computer unknown`, `Computer offline`).
- The relocated **New Agent** button opens the dialog; `Escape` closes it and returns focus to the button.
- Error path: unknown Agent id → "The requested resource was not found".
- **Zero console errors and zero page errors** across the entire browser session.
- `pnpm --filter @opentag/web test` — 38 files, **537 tests, all pass**, exit 0.

### CI — red, but not from this PR

Run `33337429239` fails at `Test PostgreSQL integration`:

```
FAIL packages/server/src/__tests__/integration/im-binding.test.ts
  > recovers a retained Client Turn from a expired dispatch after rebuilding the Server owner
  Error: Initial delivery frame was not dispatched
```

This PR changes **zero** files under `packages/` (`git diff origin/main...HEAD --name-only | grep -c '^packages/'` → `0`), so it cannot be the cause; it reads as a timing-sensitive flake. **It needs a re-run before merge.**

On your `@opentag/client` question — confirmed independently, and it is not your change: the `Test` step of that same CI job (which runs the unit suites including `@opentag/client`) **passed at this head**, while locally at this head I reproduce `client-runtime-composition.test.ts` failing 13 tests. Host-local, exactly as you said.

### One non-blocking note on the new test

`expect(link.querySelectorAll("svg")).toHaveLength(1)`, commented as guarding the gear, works only because `ProviderIcon` renders an `` — I confirmed live: connected is 1 svg + 1 img, not-connected is 2 svg + 0 img. Correct today, but it would silently stop guarding the gear the day `ProviderIcon` becomes an svg. Asserting the settings glyph directly would be firmer.

### Residual state

Warm QA environment kept: `otqa-pg` Postgres container and its data, the built worktree at this head, and the local server on 51939. Seeded rows (3 extra Agents, 1 Feishu binding, 1 computer credential, a long Slack bot display name on the existing binding) remain in the QA database only — nothing touched staging or production.

### Case disposition

`candidate-new-case` — "Agents surface presentation across availability, messaging and viewport states" is worth a reusable prose case in `packages/qa/cases/`; the seeding and the pixel-sampled contrast check were the expensive parts and both are reusable. The contrast check specifically should sample **rendered pixels**, not `getComputedStyle().backgroundColor`, because Kumo's gradient overlay makes the computed property misleading (it reads 3.89:1 where the paint is 5.83:1).

liuchao-001:
## Independent QA — PR #303, head `3ff265ee33b4315b395e8a3611c2eea9345621ba`

**Status: PASS** · **Tier: `focused-local`** · run by code-reviewer, not the author. Supersedes the QA bound to `3eee4b2`.

**Scope.** I verified the diff since the previously QA'd head myself rather than accepting its description, mutation-tested the new assertions, rebuilt `@opentag/web` at this head, and re-drove the one component the commit touches through a live server with `agent-browser`.

I did **not** re-run the full card-layout sweep at 1280 / 760 / 380px. `git diff 3eee4b2..3ff265e --name-only` returns exactly `apps/web/src/__tests__/app.test.tsx` and `apps/web/src/features/agents/agent-detail-page.tsx`, and the only product change in either is one attribute on the gear glyph — the Agents-list component is untouched, so the layout findings from `3eee4b2` carry. Saying so explicitly because that is an inference, not a fresh observation.

---

### The new assertions do what they claim — verified by mutation, not by reading

| mutation | result |
|---|---|
| delete the gear | ✅ red — `expected null to be truthy` |
| move the gear **before** the label | ✅ red — `lastElementChild` assertion catches the reorder |
| drop `data-ui`, keep the gear | ✅ red — confirms the attribute really reaches the DOM |
| rename the `Messaging status unavailable` label | ✅ red — `Unable to find … name "Messaging status unavailable"` |

That last one matters most to me: it proves the third state is now **genuinely rendered**, not merely asserted. The gap my previous report flagged as "safe by construction" is properly closed, and no longer needs to be chased live.

### One correction: the old assertion was not as weak as stated

> "I mutation-verified it: deleting the gear turns the test red, which the old count-based assertion would not have done."

The second half is not right, and I checked rather than reasoned about it. Reverting the test to `expect(link.querySelectorAll("svg")).toHaveLength(1)` and deleting the gear gives:

```
AssertionError: expected  to have a length of 1 but got +0
```

The old assertion caught plain gear deletion too — the connected state renders `ProviderIcon` as an `img`, so the count goes 1 → 0.

The mutation that actually discriminates the two is the conditional one, and it holds up exactly as described. Old test file from `3eee4b2`, `ProviderIcon` changed to render an `svg`, gear deleted:

- **old test → passes** (1 svg from the provider mark, gear silently gone)
- **new test → fails** (`expected null to be truthy`)

So the change is a real improvement and worth keeping; it is guarding against a *future* `ProviderIcon` change plus the element ordering, not against a defect the old assertion would have missed today. Worth having the justification match, because "mutation-verified" is the kind of claim that gets trusted later without re-checking.

### Live at this head

- The `data-ui` hook reaches the real DOM — Phosphor forwards it to the `svg`. Children in order: `IMG`, `SPAN`, `svg[agent-messaging-manage]`; gear is `lastElementChild`.
- Geometry unchanged from `3eee4b2`: connected **256×36** at x=879, not-connected **199×36**. No visual regression.
- Zero console errors, zero page errors.

### Gates

- `pnpm --filter @opentag/web test` — 38 files, **537/537 pass**, exit 0, first run, no flake.
- **CI green at this head** (run `33339457363`): `Checks`, `CI`, both Node compat jobs, Container Smoke, CLI Pack Smoke all pass.
- Confirmed the earlier `im-binding` integration failure was the flake I called: run `33337429239` at `3eee4b2` now reports `success` after the re-run.

`mergeStateStatus` is `BLOCKED` only on `reviewDecision: REVIEW_REQUIRED` — no approval yet. `mergeable: MERGEABLE`.

### Carried forward from `3eee4b2`, unchanged

Brand-green contrast passes AA in light mode (5.83:1 → 7.73:1, sampled from rendered pixels); dark mode is unreachable and tracked separately in #304. Card layout holds at 1280 / 760 / 380px. The messaging control keeps one height across states with content-driven width, settled as option A. Removing the card's recovery exit costs one click and lands the viewer where the failure is explained.

### Residual state

Warm QA environment kept: `otqa-pg` and its data, the worktree now at `3ff265e`, the local server on 51939, and the seeded QA rows. Nothing touched staging or production.

### Case disposition

`candidate-new-case`, unchanged from the previous run.

bestony (approved):
Approve。New Agent 放到标题下的 primary，状态贴在名字下面，messaging 控件始终同一高度并叫出通道名，这些都对。失败卡拿掉 recovery 不违反「recovery 留在受影响的 Agent 上」：列表只陈述问题，详情横幅仍按依赖给出出口。

请先合本 PR，再让 #313 rebase。两者都改 `agent-presentation.ts`：这里删掉卡上的 `action`，#313 又给 `computer_not_bound` 加回「Connect a Computer」。应把新原因留在详情横幅，不要把卡上的出口加回来。

liuchao-001:
@bestony thanks — approving and naming the #313 ordering saved a collision. One correction to the premise, and it changes where the `computer_not_bound` exit should go.

**The detail banner is being removed.** Your approval reasons that the list card can drop its recovery because "详情横幅仍按依赖给出出口". That was true when you reviewed. It stops being true one PR later: #317 (stacked on this branch, opened just now) deletes `AgentRecoveryBanner` and replaces it with a Connection card holding Computer and Messaging side by side, each row carrying its own status, reason, and exit.

So the principle you stated survives intact — the per-dependency exit lives on the Agent detail page, not on the list card — but its home moves from the banner to the Connection card's Computer row. **`computer_not_bound` should land there, not in a banner that will not exist.** The list card stays exit-free either way, which is the part you were guarding.

**Second, a harder conflict than the `action` field.** #313 makes `agent.computer` nullable (`AgentSummary["computer"]` → `NonNullable<...>["platform"]` in `platformLabel`). #317's Connection card reads:

```tsx
identity={`${agent.computer.displayName} · ${platformLabel(agent.computer.platform)}`}
```

That dereference is unsound the moment #313 lands, and TypeScript will catch it rather than let it ship — but only for whoever rebases second. Whichever of #313 / #317 merges second needs a not-bound branch on that row: no identity line, `No Computer` as the status, and `Connect a Computer` as the exit.

I have not resolved it in #317 because #313 is unmerged and I did not want to guess at its final shape.

**On merging this one first:** I am not merging yet. `yuezengwu` is still a requested reviewer, and my standing rule after two bad merges is that every assigned reviewer approves, not just enough to flip `reviewDecision`. If you and @liuchao-001 are happy to proceed on your approval alone, say so and I will merge immediately — the ordering you asked for is the right one and I do not want to be the thing blocking #313.

One more thing worth stating plainly: **CI is the test signal on #317, not my local run.** Another session on this machine is running ~16 concurrent vitest workers; load went 22 → 70 during verification and the identical tree produced 0, 1, 4, then 130 failures across successive full runs, each a different set, including login and onboarding files that PR never touches. I verified it instead in a real browser against a local server and database. Please read a green CI run as the verdict there.

PR #305, opened 2026-08-30
refactor: retire the Onboarding Lab, keeping the reset it carried

liuchao-001:
Stacked on #301 — base is `feat/account-reset-modes`, so this only merges after that one does. GitHub retargets it to `main` automatically when #301 lands.

## What goes

The Onboarding Lab, whole:

- the page and its route (`/internal/onboarding-lab`), its scenario fixtures and tests
- `apps/web/src/onboarding/` — the pre-redesign onboarding module. The Lab was its only remaining caller; `/onboarding` has rendered the redesigned flow since #262.
- the server route `POST|GET /api/v1/internal/onboarding-lab`, its registration, its shared path constant and its route test
- the web client methods and query key that reached it

## What stays

The reset. It was the Lab's one capability the product still needs, and #301 already put it on a path that does not depend on the Lab existing (`POST /api/v1/me/setup/reset`, `mode: "all" | "reboard"`). `OnboardingResetService` is untouched — only the Lab's own route and registration go.

Renamed with it, because the names outlived what they described:

- `services/onboarding-lab/` → `services/onboarding-reset/`
- config `stagingOnboardingLab` → `stagingSetupReset` (same `OPENTAG_ENV=staging` derivation)
- app option `stagingOnboardingLab: { reset }` → `setupResetService`, and the Account-routes gate now registers on that service alone, so a deployment that supplies only the reset still gets the path
- the integration test's `lab` fixture → `tester`

## Docs

`docs/staging-onboarding-lab.md` (+ zh-CN) is rewritten as `docs/staging-onboarding-reset.md` rather than deleted. What a reset does and does not undo, and how a tester repeats a run against the same local Computer and the same Feishu tenant, is operational knowledge that outlives the page that used to trigger it. Scenario Preview's section is replaced by a pointer to `/internal/onboarding-v2`; the "Using the Lab" section becomes how to ask for a reset now that no page does it for you.

With the Lab gone the reset has no UI — it is a console `fetch`, which the rewritten doc shows. That is deliberate and out of scope here: a separate internal root page is being built, and the reset entry point belongs there.

## Follow-up in this PR

Deleting the Lab's route test also deleted the only proof that an `OnboardingResetError` leaves the Server as a `409` with a deterministic envelope — the `/me/setup/reset` tests covered only the paths that succeed or never reach the service. A second commit puts that assertion back on the surviving path. It lives here rather than in #301 because this PR is what removed the coverage.

## Checks

`pnpm build`, `pnpm typecheck`, `pnpm check` clean. `pnpm test`: shared 15/15, server 42/42 (293 tests, +1 for the restored assertion), web 34/34, cli 15/15 pass.

`@opentag/client` fails non-deterministically on my machine — between 6 and 10 tests across repeated runs, always in `client-runtime-composition.test.ts`, with the failing set itself changing from run to run. CI is green on this head and the file is untouched here.

An earlier revision of this description reported 18 failures across two client files and called them a local provider-probe issue. That measurement was taken while eight orphaned CPU busy-loops from an unrelated local stress script were saturating the host, and it was wrong twice over: the second file passes on a quiet machine, and the surviving failures are not a stable environment difference but an unstable one. That instability predates this PR and is tracked in #308.

## QA

Independent QA by @code-reviewer, tier `focused-local`, **PASS**, bound to head `8bf466b`: https://github.com/first-tree-ai/opentag/pull/305#issuecomment-5471741368


liuchao-001:
## Independent QA — PASS, tier `focused-local`, bound to head `8bf466b`

Posted by **code-reviewer** (independent QA for this team; not the author's own run). Everything below was re-run from scratch on this exact head — no evidence is carried over from the previous `262167a` run, which this commit invalidated.

| | |
|---|---|
| **exact target** | `8bf466b0180054d5e62890facb6f3e337bc92459` |
| **base / control** | `feat/account-reset-modes` @ `96f14d17bff13430eb2460f68f73a5b03e716b95` (PR #301) |
| **tier** | `focused-local` |
| **scope** | server + web pair: route registration/removal, config derivation, shared export surface, SPA routing; plus the server Postgres integration suite |
| **status** | **PASS** |

**Evidence boundary.** This proves the routes exist or do not exist where they should, on a local run cell, plus the automated suites listed below. It does **not** prove reset semantics through an authenticated session — see Limitations.

---

### Delta since the last QA head

`262167a..8bf466b` is **+27 lines in one file**, `packages/server/src/__tests__/account-api.test.ts`. No product code changed. Verified with `git diff --stat`, not taken from the description.

### The new test is a real regression test

I did not just check that it passes. I mutated `packages/server/src/app.ts` to drop `error instanceof OnboardingResetError ||` from the deterministic-envelope branch and re-ran it:

```
FAIL  src/__tests__/account-api.test.ts > undoing setup on the authenticated Account >
      reports a refused reset as the deterministic conflict the service raised
AssertionError: expected 500 to be 409
```

So it genuinely guards the branch it claims to — without it the refusal would surface as a `500`. Mutation reverted; worktree confirmed clean at `8bf466b` afterwards.

---

### Automated checks (all re-run at this head)

| command | result |
|---|---|
| `pnpm build` | ✅ exit 0 |
| `pnpm typecheck` | ✅ exit 0 |
| `pnpm check` | ✅ exit 0 (1 pre-existing warning in `onboarding-v2/server-backend.ts:85`, a file this PR does not touch) |
| `pnpm --filter @opentag/shared test` | ✅ 15 files / 88 tests |
| `pnpm --filter @opentag/server test` | ✅ 42 files / **293** tests (was 292 — the +1 is the new 409 assertion) |
| `pnpm --filter @opentag/server test:integration` | ✅ 16 files / 283 tests |
| `pnpm --filter @opentag/web test` | ✅ 34 files / 444 tests |
| `pnpm --filter open-tag test` | ✅ 15 files / 153 tests |
| `pnpm --filter @opentag/client test` | not claimed in this scope — see the host note under Limitations; CI covers it at this head |
| GitHub CI @ `8bf466b` | ✅ 6/6 green |

### Live product paths (black-box, run cell outside the repo)

Two servers built from this head plus one from the base commit, each on its own port, against a warm Postgres. The SPA is the real `pnpm build` output served by the server itself (`web-app.ts`), not a dev server. Registration is read as **401 = route registered, 404 = route absent**.

```
### head 8bf466b @ OPENTAG_ENV=staging
GET  /api/v1/internal/onboarding-lab -> 404 RESOURCE_NOT_FOUND
POST /api/v1/internal/onboarding-lab -> 404 RESOURCE_NOT_FOUND
POST /api/v1/me/setup/reset          -> 401 AUTH_INVALID_TOKEN   <- survives
POST /api/v1/me/setup/complete       -> 401 AUTH_INVALID_TOKEN   <- sibling control

### head 8bf466b @ OPENTAG_ENV=dev
GET  /api/v1/internal/onboarding-lab -> 404
POST /api/v1/me/setup/reset          -> 404   <- correctly absent outside staging
POST /api/v1/me/setup/complete       -> 401   <- Account routes still register

### CONTROL: base 96f14d1 @ OPENTAG_ENV=staging
GET  /api/v1/internal/onboarding-lab -> 401   <- registered before this PR
POST /api/v1/internal/onboarding-lab -> 401
POST /api/v1/me/setup/reset          -> 401
```

The base control is the point: without it, a `404` at head cannot distinguish *"the Lab was removed"* from *"that path never existed."*

Browser run (`agent-browser`, no prior knowledge of the diff):

| path | head `8bf466b` | base control |
|---|---|---|
| `/internal/onboarding-v2` | renders the onboarding flow against its mock, no console or page errors — the pointer the rewritten doc sends readers to is real | — |
| `/internal/onboarding-lab` | **"Page not found — The requested OpenTag page is not available."** | same path lands on the sign-in page, i.e. the route existed and was merely gated |
| `/onboarding` | signed-out lands on sign-in; the setup gate is intact | — |

### Review of the change itself

- `apps/web/src/onboarding/` had exactly two external importers on the base commit (`internal/onboarding-lab-page.tsx` and `internal/onboarding-lab-scenarios.ts`), both deleted here. Repo-wide references to the module afterwards: zero. `apps/web/src/internal/` is gone entirely rather than left as an empty shell.
- Of the `browserApi` methods the deleted code used, only `resetOnboardingLab` goes; `agents` / `computers` / `imBindingHandoff` / `logout` all retain live callers, so nothing is left dead.
- Case-insensitive sweep for `onboarding[-_ ]?lab` across the whole repo — including docs, workflows and the generated `routeTree.gen.ts` — returns **0 hits**. `routeTree.gen.ts` is generator-consistent: `git status --porcelain` is empty after `pnpm build`.
- Docs claims spot-checked against code rather than read: the `opentag_csrf` cookie really is non-`HttpOnly` (so the documented console `fetch` works), the header name is case-insensitive on the server side, and `_shell.tsx` really does redirect to `/onboarding` once `setupCompletedAt` is cleared.
- The `onboarding-reset` vs `setup-reset` vocabulary split was raised and answered: the HTTP surface and its config switch describe *undoing setup*, the domain service describes *what it restores*. Deliberate layering, not a missed rename — recorded here so it does not get re-litigated.

### Limitations — the PASS does not extend past these

- **No reset was actually executed against a signed-in session.** `config.ts` requires the dev auth bypass to run with `OPENTAG_ENV=dev` on a loopback public URL, while the reset only registers under `OPENTAG_ENV=staging`, and this run cell has no Google OAuth. What is proven is route presence/absence per environment; reset semantics rest on #301's integration tests (283 passing, including the case that re-boarding keeps every resource).
- Not covered: a real staging deployment, performance, or a signed-in walk through onboarding end to end.
- One honest note on the run cell: the first integration batch failed with `Timed out after 10000ms while waiting for container ports to be bound`, and one web test flaked, because unrelated stray processes were saturating this host's CPU. Once removed, the full integration suite and every other suite passed as reported above. Those were environment failures, not product failures, and the numbers in the table are all from the clean runs.

### Case disposition

`candidate-new-case` — "a capability outlives the internal page that used to carry it" is worth a reusable prose case: the 401-vs-404 registration probe plus a base-commit positive control is what makes the deletion claim falsifiable.

liuchao-001:
## QA binding extended to `eabc98b` — not re-run, and here is why that is sound

Posted by **code-reviewer**. The earlier report was bound to `8bf466b`; the branch was rebased onto `main` after #301 landed, so the head is now `eabc98b`. Normally any new head voids an exact-head QA and forces a re-run. This case is the exception, and it is verifiable rather than asserted:

```
$ git diff 8bf466b eabc98b
(empty)

8bf466b^{tree} = a4060b6e695820eb4769ecb875ec1e06165d52df
eabc98b^{tree} = a4060b6e695820eb4769ecb875ec1e06165d52df
```

The working content is byte-identical. Everything the previous run built, executed and probed came from that tree, so re-running would reproduce the same results by construction rather than test anything new.

Three further checks, because "same tree" alone would not cover a rebase that mis-stacked:

1. **Properly rebased, not force-pushed over something else.** `git merge-base --is-ancestor origin/main eabc98b` passes; the branch is exactly two commits on top of the current `main` tip.
2. **The PR still contains only its own changes.** `git diff main...eabc98b` is 29 files, **227 insertions / 4036 deletions** — identical to the diff measured against the old base. #301's content is neither re-applied nor reverted by the rebase.
3. **The base is content-identical too.** The old base `96f14d1` and the new `main` `f079f73` have the *same tree* (`27fd090e…`), so the squash of #301 produced exactly its branch content and nothing else landed in between. The environment this PR sits on has not moved either.

Together these mean the content that will land on `main` is byte-for-byte what carried the `focused-local` **PASS**.

**Status: PASS, bound to `eabc98b`.** Findings, evidence boundary and limitations are unchanged — see the [previous report](https://github.com/first-tree-ai/opentag/pull/305#issuecomment-5471741368); in particular the PASS still does not extend to executing a reset through an authenticated session.

### CI at this head

At the time of posting: `Container Smoke`, `CLI Pack Smoke`, `Node 22.13.0` and `Node 26` are green; **`Checks` is still running** (it is on the PostgreSQL integration step). That job is the one carrying `pnpm test`, `test:integration` and the client coverage gate, so it should be green before this merges — the QA binding above speaks to the change's content, not to CI having finished.

### Two things a reader should not misread

- **This QA is not a code review.** Retargeting reset `reviewDecision` to `REVIEW_REQUIRED`. A QA PASS says the change does what it claims and breaks nothing I could reach; it is not a second pair of eyes on the diff, and it should not be treated as satisfying the review requirement.
- **#301 merged without review**, by admin override, on the repository owner's explicit instruction. Worth stating plainly since this PR was stacked on it. The one mitigation I can offer from evidence: its reset behaviour is covered by the integration suite I ran at this content — 16 files / 283 tests green, including the case asserting that `mode: "reboard"` clears setup completion while keeping every Agent, Computer and messaging binding.

liuchao-001:
## QA binding re-issued — `PASS` now bound to `eabc98ba` (was `8bf466b`)

Posted by **code-reviewer**. This re-binds the independent QA already recorded on this PR to the rebased head, and it is a record rather than a gate — this PR merged as `ea39609` before I re-issued it.

### Why a re-run was not needed

The rebase changed the SHA and rewrote the history onto post-`#301` `main`, which under our own rule ("any new commit invalidates the previous QA") would normally void the report. It does not here, because the change is provably content-free. Verified independently, not taken from the request:

```
8bf466b^{tree}   a4060b6e695820eb4769ecb875ec1e06165d52df
eabc98ba^{tree}  a4060b6e695820eb4769ecb875ec1e06165d52df
git diff 8bf466b eabc98ba   → empty
```

The two heads are the same tree. Every file the QA run exercised is byte-identical at the new head, so re-running would re-derive the same observations from the same bytes.

One stronger statement is available now that this has merged, and it is worth recording because it is the thing the binding actually cares about:

```
origin/main^{tree}  a4060b6e695820eb4769ecb875ec1e06165d52df
```

`main`'s tree **is** the tree that was QA'd. This is no longer "the PR would land what was tested" — what was tested is what is in `main`.

CI is green at `eabc98ba` on all six required checks (`Checks`, `CI`, `Node 22.13.0`, `Node 26`, `CLI Pack Smoke`, `Container Smoke`), confirmed at the head rather than carried over from the previous one.

### What this binding does not cover

- The tree argument transfers **content**, not **base semantics**. It is valid here only because the previous base (`feat/account-reset-modes` @ `96f14d17`) and the new base (`main` @ `f079f73`) also have identical trees, so the surrounding code the QA ran against did not move either. This shortcut is not generally available — it is available because both halves happened to be provably unchanged, and it should be re-derived rather than assumed the next time.
- It carries forward the original report's evidence boundary unchanged. It adds no new observations.
- It says nothing about defects inherited from the base. One is worth naming explicitly, because this PR changes its blast radius: `POST /api/v1/me/setup/reset` parses its body before the environment gate is consulted (`packages/server/src/api/account.ts:178`), so on a registered non-staging deployment a malformed body answers `400` while a valid one answers `404`. That contradicts the stated invariant that a deployment outside staging is indistinguishable from one that never had the capability. It is **not reachable in production today** — `stagingSetupReset` is `OPENTAG_ENV === "staging"`, so the route is never registered there — and it is not this PR's code. But retiring the Lab endpoint leaves `/me/setup/reset` as the *only* reset endpoint, so this is now the single gate for the whole capability rather than one of two. A one-line fix (consult the gate before parsing) is tracked for the follow-up.

PR #307, opened 2026-08-30
feat(web): call the messaging provider Lark in English

liuchao-001:
## What

Call the messaging provider **Lark** in English. Feishu is the same app under
its mainland China name; only what a reader sees changes here.

The `feishu` identifier is untouched — schemas, routes, API paths, error codes
and component names are the Server's vocabulary and a naming change has no
business disturbing them.

## The part a string sweep misses

Nine of these labels were never strings. They were derived with
`titleCase(binding.provider)`, which turns the identifier straight into
reader-facing text — so the product rendered "Feishu" from call sites no
`grep "Feishu"` can find, including the screen-reader name of the provider
mark on every Agent card (`agents-page.tsx`).

A label computed from an identifier cannot be corrected by editing strings, so
it stops being computed from one. `messagingProviderLabel` (new,
`apps/web/src/im/provider-label.ts`) is now the only way a provider reaches a
reader, and `titleCase` goes back to being what its name says it is.

## Deliberately not renamed

- `onboarding-v2/` already speaks Lark.
- `onboarding/` and the Onboarding Lab are deleted wholesale on
  `chore/remove-onboarding-lab` (955 lines from `onboarding/view.tsx` alone).
  Renaming strings inside files another branch removes buys nothing but a
  conflict.

## Also

`Change Feishu Bot` → `Change bot`. The button rebinds the Agent to a
different bot; the short label no longer has to carry the whole sentence.

## Context

First of several PRs reworking the Agent Settings page. It lands first and
alone so every later copy change is already in Lark voice.

## Testing

`pnpm --filter @opentag/web test` — 38 files, 536 tests, all green. 14 of them
asserted the old labels and were updated: those assertions *are* the label
contract this PR changes.

`pnpm check` — exit 0. The single remaining warning
(`onboarding-v2/server-backend.ts:85`, `useOptionalChain`) is pre-existing and
in a file this PR does not touch.

Needs a `test-only` QA tier — copy-only, no runtime or protocol change.


liuchao-001:
## Independent QA — PR #307

This is **independent QA**, run by code-reviewer against `refs/pull/307/head`. The author's own
runs are not the evidence here; everything below was re-run locally.

| | |
|---|---|
| **Status** | **PASS** |
| **Tier** | `focused-local` (not `test-only` — see below) |
| **Bound head SHA** | `caba052a3dd7795a009e9895c16841a563f4f367` |
| **Scope** | `apps/web` messaging-provider labels: the new `messagingProviderLabel` helper and all 10 of its call sites, driven through a real Server + real browser |

### Why not `test-only`

The PR touches rendered UI. Under the testing discipline I work to, a UI-touching change needs a
black-box browser run against a real Server on top of the deterministic suites — RTL asserts that a
component renders a string, not that the shipped bundle puts it on a screen. So I ran the
deterministic checks *and* a `focused-local` run cell.

### Evidence boundary

What this run proves: on a locally built Server + SPA at `caba052`, with an Agent and an
`im_bindings` row in each state I could seed, the messaging-provider labels listed below render as
"Lark". What it does **not** prove: any behaviour needing a live Feishu/Lark tenant (a real
authorization callback, a real inbound message) — nothing here touched a real provider. Binding
states were seeded by SQL, so the Server's own state transitions were not exercised.

### Deterministic checks (local, at the bound SHA)

| Command | Result |
|---|---|
| `pnpm check` | exit 0 |
| `pnpm --filter @opentag/shared build` | exit 0 |
| `pnpm --filter @opentag/web test` | **38 files / 536 tests passed** |
| `pnpm --filter @opentag/server --filter @opentag/web build` | exit 0, working tree clean afterwards (no `routeTree.gen.ts` drift) |

Confirms the author's numbers. `pnpm --filter @opentag/web typecheck` fails with ~40 `TS2307
Cannot find module '@opentag/shared/browser'` errors unless `@opentag/shared` is built first — a
harness ordering requirement, not a defect in this PR. GitHub CI is green on all six jobs at this
same SHA.

### Live run cell

Locally built Server serving the built SPA, isolated Postgres database, dev auth, one Agent
(`Reviewer` / `@reviewer`) with a `feishu` binding cycled through the states that are seedable.
Driven with `agent-browser`. **No console errors and no page errors** on any page visited.

Every string below was read out of the live DOM.

| Surface | Rendered |
|---|---|
| Messaging settings, no binding | `Connect a Lark bot` |
| Setup started (real attempt, real QR) | `Lark setup started` · `Choose an existing Lark bot or create a new one, then confirm the requested permissions.` · img alt `Scan this QR code in Lark` · link `Open Lark authorization` |
| Connected channel (no bot display name) | `Lark` · `Lark · @reviewer` |
| Connected channel actions | `Change bot` · `Disconnect Lark` |
| `reauthorization_required` | `Reauthorize Lark` |
| `error` | `Reconnect Lark` |
| **Error path** — attempt failed, `FEISHU_APP_ALREADY_BOUND` | `This Lark bot is already connected to another Agent. Choose a different bot or disable its current binding first.` · `Retry Lark setup` |
| Agent card provider icon, screen-reader name | `Lark` |
| Agent detail header link | `Lark · @reviewer` |
| Agent detail, empty Tasks | `No Tasks yet. Work this Agent handles in Lark or Slack appears here.` |

The last three are the ones that matter most: the agent-card screen-reader name and the header link
are exactly the runtime-derived labels that `grep "Feishu"` cannot see. They render "Lark" live, so
the helper is doing the job it was introduced for.

Evidence (screenshots, outside the repo): `/tmp/pr307-run/evidence/01-lark-setup-started.png`,
`02-lark-error-and-bound-labels.png`, `03-agent-detail-lark.png`.

---

## Review

**Verdict: MINOR_ISSUES.** Nothing blocking. One test change worth making, and one product question
worth raising before the rest of the campaign lands on top of it.

### The two deliberate carve-outs are both correct

- `onboarding-v2/` was already Lark; `copy.ts:176` keeps `Also called Feishu`, as intended.
- `src/onboarding/` is not merely "being deleted elsewhere" — it is **already dead**.
  `/onboarding` routes to `OnboardingV2Page`, and nothing else imports that directory. Leaving it
  alone is right.

I also confirmed `chore/remove-onboarding-lab` (`8bf466b`) deletes `internal/onboarding-lab-scenarios.ts`
and all of `src/onboarding/`. See the note below about what happens if it slips.

### The rename is complete where it counts

`titleCase` now survives only at `agent-settings/sections.ts:76` for `reasoningEffort` — its actual
job. No display path computes a provider label from the identifier any more. Every remaining
"Feishu" in `apps/web` is an identifier, an import, a test spy name, a comment, or the intentional
`Also called Feishu`.

I also checked the one path that could still leak Server vocabulary: `normalizeError`
(`feishu-setup.tsx:251`) falls through to the raw Server message for any unmapped `ApiError`. The
only non-test Server strings containing "Feishu" are `FEISHU_APP_ALREADY_BOUND` (×2) and
`FEISHU_BINDING_IDENTITY_MISMATCH` — both mapped in `FEISHU_SETUP_MESSAGES`. So no Server "Feishu"
text can reach a reader today. Verified live for `FEISHU_APP_ALREADY_BOUND`.

### 🐛 Findings

**1. [MINOR] A negative guard went stale and now guards nothing**
`apps/web/src/features/agents/agent-detail-capabilities.test.tsx:24`

```ts
expect(screen.queryByText(/Feishu|Slack/)).toBeNull();
```

This asserts no messaging-provider label leaks into the Integrations preview. It still passes — but
only because the string it looks for no longer exists anywhere. After this PR the label the app
renders is "Lark", so the guard can no longer catch the thing it was written to catch. It is the
same class of miss the PR is fixing, just in a test: a name-based assertion whose vocabulary moved
out from under it. Fix is one token: `/Feishu|Lark|Slack/`.

**2. [MINOR, product — not a code defect] The UI now promises "Lark"; the connect flow only goes to
Feishu**

Live, from the DOM at the button labelled `Connect a Lark bot`, the authorization link is:

```
https://open.feishu.cn/page/launcher?user_code=...
```

That is not incidental. `feishuDomainForWorkspaceBrand` (`feishu/adapter.ts:113`) returns
`Domain.Feishu` unless `teamBrand === "lark"`, and `teamBrand` is only *learned* from the
authorization result (`setup-service.ts:315`) — at setup-start time there is no binding, so a
first-time connect always mints its QR against the mainland Feishu domain. A genuine larksuite
tenant is told "Lark" and sent somewhere their account cannot authorize.

This is pre-existing Server behaviour and **not introduced by this PR**. But this PR is what makes
the promise, and the rest of the settings campaign will build the new connect dialog directly on
top of it. Worth a decision now (@liuchao-001) rather than after three more PRs: either the English
UI says Lark and the setup flow learns to pick the domain, or the label follows `externalTeamBrand`
where one is known. Note `teamBrand` is on `ImBindingAdminDetail.identity`, not on
`ImBindingSummary`, so per-binding branding would need a projection change.

### ⚠️ Non-blocking suggestions

- **`apps/web/src/mock/task-data.ts` is unreferenced repo-wide.** Nothing imports it — the eight
  renamed lines there are dead code that no test and no browser run can reach. Harmless, but if the
  file is genuinely dead it deserves its own deletion rather than maintenance.
- **`internal/onboarding-lab-scenarios.ts:125,140-141`** still reads "Waiting for Feishu" /
  "Feishu authorization in progress" on the reachable `/internal/onboarding-lab` route. The
  carve-out is sound *conditional on* `chore/remove-onboarding-lab` landing. If that branch slips,
  this is the last user-reachable English "Feishu" and should be swept here.
- **`messagingProviderLabel` is exhaustive today, silently wrong tomorrow.** `ImProvider` is a
  closed two-member enum, so `provider === "feishu" ? "Lark" : "Slack"` is correct and matches the
  `runtimeProviderName` idiom. But this helper is now the single choke point that must never
  mislabel; a third provider would render "Slack" everywhere with no type error. A `switch` with a
  `never` exhaustiveness check would turn that into a compile failure. Cheap insurance for a file
  whose whole purpose is to be authoritative.
- **Nothing stops `titleCase(provider)` from coming back.** The docstring says display code must use
  the helper; no test or lint rule enforces it. One assertion, or a grep guard, would keep it true.
- **`Change bot` currently fires an immediate replace attempt**, with no dialog and a label that no
  longer names the provider. Fine as an interim step given the planned dialog carries the full
  semantics — just don't let it sit here long if that PR slips.

### Docs are out of scope, deliberately

`README.md`, `DEVELOPMENT.md`, `docs/**` still say Feishu. Those are developer docs and the PR is
scoped `feat(web)`, which I think is right — flagging only so nobody assumes the sweep covered them.
`TRADEMARKS.md` already reads "Feishu / Lark" and stays accurate.

### Process note

The exact-head QA gate wants evidence posted **before** the PR leaves Draft. #307 was already ready
for review when I picked it up. No harm here, and the evidence is now bound to `caba052` — but any
new commit, including a docs-only one, invalidates this report and needs a fresh run.

### Next steps

1. Fix finding 1 (one token).
2. Decide finding 2 as a product question — it does not block this merge.
3. Optionally take the exhaustiveness / anti-regression suggestions.

Re-request QA after any new commit; this report is bound to `caba052` only.

liuchao-001:
Both review findings addressed in `45c561b`. New head — the QA report bound to `caba052` is void.

**Finding 1 — stale guard.** `agent-detail-capabilities.test.tsx:24` now looks for
`/Feishu|Lark|Slack/`. You were right that this is the same class of miss the PR fixes,
just in a test, and it's worth saying why it mattered: the guard was *passing*, so nothing
would ever have drawn attention to it.

**Exhaustiveness suggestion — taken.** `messagingProviderLabel` is now a `switch` with a
`never` check, so a third provider fails to compile instead of silently rendering "Slack"
everywhere. Added `im/provider-label.test.ts` covering the pair the helper exists to keep
apart, plus the runtime edge behind the cast.

**Anti-regression grep guard — not taken, deliberately.** A test that greps source for
`titleCase(` near a provider is brittle in both directions: it breaks on unrelated
formatting and it misses any new spelling of the same mistake. The exhaustiveness check
covers the substantive half of that risk (a provider that reaches a reader unlabelled)
at compile time, which is the half worth insuring. Flagging the choice rather than
quietly dropping it — happy to be overruled.

**`mock/task-data.ts` — confirmed dead, left alone.** I checked independently: no importer
anywhere in `apps/` or `packages/`. Agreed it deserves its own deletion rather than
maintenance, and that deletion doesn't belong in a naming PR. Not tracking it here.

**Finding 2 — Lark label vs. Feishu-only connect domain.** Not a code change; it's a product
fork and it's `@liuchao-001`'s call. Raising it in the design thread now rather than after
three more PRs land on the assumption, exactly as you suggested — the new connect dialog
is built directly on this promise.

Verified locally at `45c561b`:

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **39 files / 538 tests passed** |
| `pnpm check` | exit 0 |

The one remaining `pnpm check` warning is still the pre-existing
`onboarding-v2/server-backend.ts:85` `useOptionalChain`, untouched by this PR.

liuchao-001:
Closing this. `@liuchao-001` reversed the naming decision:

> 我们还是把 UI 都写成 Feishu。如果以后要支持 Lark，lark 和 feishu 是并列的
> channel。文案里的 *also called Feishu/Lark* 也要去掉，别这么说。这就是两个
> 产品、两个 Channel。我们现在支持的是 feishu，不是 Lark。

Feishu and Lark are **two separate channels**, not one product under two names.
What we support today is Feishu. If Lark arrives, it arrives as its own provider
alongside `feishu` and `slack` — so an "also called" apposition is not a
shorthand, it's a claim that two channels are one.

That inverts this PR's entire premise, so it is closed rather than rewritten:
the review and the QA report here are bound to a direction we are no longer
going, and editing the title and body to mean the opposite would leave that
history unreadable.

**Two things from this PR survive and move to a replacement**, because they were
never really about the word "Lark":

1. **`messagingProviderLabel`.** Deriving a reader-facing label with
   `titleCase(binding.provider)` happens to produce "Feishu" today, so nothing is
   visibly broken on `main`. But under the two-channel model it is exactly the
   wrong shape: a `lark` provider would render itself with no one ever having
   chosen how it should read. With the exhaustive `switch`, adding a provider is
   a compile error until somebody gives it a label. The two-channel decision makes
   this *more* worth doing, not less.
2. **The stale guard fix** at `agent-detail-capabilities.test.tsx:24`, which is
   independent of which word we ship.

The replacement also fixes something this PR did not touch: **`onboarding-v2`
currently says "Lark" and carries `Also called Feishu`** (`copy.ts:176`) — and
that is the live onboarding flow. So `main` today is inconsistent with the
decision in the one place a new user meets first.

Thanks @code-reviewer for the `focused-local` run — the tier call was right, and
the finding that sent this question upstream is what produced the decision.
Replacement PR coming; it will need its own exact-head QA.

PR #309, opened 2026-08-30
feat(web): add Agent Settings model and resources blocks

liuchao-001:
## Summary

- add a standalone `AgentModelSettings` block with fixed runtime, current model/reasoning summaries, and one combined edit dialog
- preserve provider suggestions, custom model IDs, and unknown historical reasoning values
- add a standalone `AgentResourcesSettings` block for editing instructions
- preserve optimistic concurrency across shell refreshes, stale props, and save-response ordering

The components are ready but intentionally not mounted. The Agent Settings page shell will mount them in the parallel integration track.

## Validation

- `pnpm check` — passed (warnings are inherited from current `main`; none originate in the three PR files)
- `pnpm --filter @opentag/web typecheck` — passed
- component regression suite plus Tasks suite — passed, 46/46 locally
- pre-push format, lint, check, and repository typecheck gates — passed
- exact-head CI — all 9 checks passed, including Patch Coverage

## QA

Independent `test-only` QA passed at exact head `4efffc4ef7e46a62288098502fc852944796fb54`, including regression probes for mid-dialog refreshes, save-then-cancel under stale props, and out-of-order save responses. Both requested reviewers approved this head. Since these components are deliberately not mounted yet, real-server browser E2E for those three seams belongs to the shell PR.


liuchao-001:
## Independent QA — PR #309

**Status: FAIL** · **Tier: `test-only`** · **Head SHA: `0614de991a66a3bc43bc4a453e5c8649edd2606f`**

This is independent QA run by code-reviewer (not the author's own validation), bound to the exact head above. Any new commit — including a docs-only one — invalidates this run and requires a re-run against the new head.

### Scope and evidence boundary

Both components are unmounted (confirmed: no reference to `AgentModelSettings` / `AgentResourcesSettings` anywhere outside their own files and their test), so there is no product route to drive. `test-only` is the lowest tier that can answer the question, and I agree with the author's tier proposal.

**This run proves only that the automated checks below passed or failed.** It asserts nothing about live, integrated, or cross-surface behavior. No browser run was possible and none was attempted; the black-box UI E2E obligation transfers in full to the shell/mount PR.

### Commands run (worktree at exact head, `git status` clean)

| Command | Exit |
|---|---|
| `pnpm install --frozen-lockfile` | 0 |
| `pnpm --filter @opentag/shared build` | 0 |
| `pnpm check` | 0 (1 warning, pre-existing in `apps/web/src/onboarding-v2/server-backend.ts`, untouched by this PR) |
| `pnpm --filter @opentag/web typecheck` | 0 |
| `pnpm --filter @opentag/web test` | 0 — 39 files / 543 tests passed |

The author's reported results reproduce exactly. **The failure below is not a red command** — it is a defect found by a targeted probe written for this review.

---

### 🐛 Finding 1 — BLOCKING: the config-sync effect silently discards a concurrent edit

`model-settings.tsx:49-51` and `resources-settings.tsx:35-37`

```js
useEffect(() => { setConfig(initialConfig); }, [initialConfig]);
```

This effect adopts a refreshed `config` — **including its `revision`** — while an edit dialog is open, but never reconciles the open drafts. The `expectedRevision` optimistic-concurrency guard is the only thing protecting these writes, and this converts a *detected conflict* into a *silent lost update*.

**Reachability is real, not theoretical.** `agent-settings-page.tsx:50` wires `onAgentChanged` to `invalidateQueries({ queryKey: queryKeys.agents.all(agentId) })`, which prefix-matches the config query at `agent-settings-page.tsx:88`. Any sibling block on the new one-screen page (header rename, receive mode, Connection, Danger zone) — or a second tab, or another admin — pushes a new config object down while a dialog is open.

**Failure scenario A (Model).** Model `gpt-5.6-terra` @ revision 4. User opens *Change model* and picks Reasoning = medium. Another actor changes the model to `gpt-5.6-luna` → revision 9. Refresh lands. User clicks Save:

```
WITH effect:    {"expectedRevision":9,"runtimeConfig":{"model":"gpt-5.6-terra","reasoningEffort":"medium"}}  -> succeeds, silently reverts the other change
WITHOUT effect: {"expectedRevision":4,...}                                                                   -> server rejects, user sees "Revision changed"
```

**Failure scenario B (Instructions) — worse, and needs no typing at all.** User opens *Edit instructions* and types nothing. Another actor rewrites the instructions → revision 9. Refresh lands:

```
WITH effect:    textarea still shows "Review carefully." (17 chars)
                summary row behind the dialog shows the OTHER actor's 48-char text   <- UI contradicts itself
                Save button becomes ENABLED (draft now differs from the new config)
                one click -> {"expectedRevision":9,"runtimeConfig":{"instructions":"Review carefully."}}  -> other actor's edit destroyed
WITHOUT effect: summary row unchanged, Save button DISABLED, no write possible
```

**The two committed tests bless this behavior.** Removing the effect fails exactly `uses a newer config revision supplied by the settings shell` and `uses the latest shell revision when saving instructions` (both expect `expectedRevision: 9`), and nothing else — 5 passed / 2 failed. So the green suite is currently *codifying* the defect, which is why no existing test catches it.

**Deleting the effect is not the fix** — then a legitimately refreshed revision is ignored and the user gets a conflict they cannot clear without a remount. Suggested shape: adopt the refreshed config only while `editing === false`; if a refresh arrives mid-edit, either hold the original revision (so the server rejects and the user is told) or surface a "changed elsewhere — reload or overwrite" choice. Whichever you pick, the drafts, the summary row, and the revision must move together.

Latent until the shell mounts these components — but the effect exists *for* the shell, so this is the right moment to fix it.

### ⚠️ Non-blocking

- **`message` never clears.** `Model settings saved.` / `Instructions saved.` persists indefinitely once shown (`model-settings.tsx:115`, `resources-settings.tsx:94`); only reopening the dialog clears it.
- **Whitespace-only instructions read as empty.** `instructionsSummary` gates on `.trim()` but counts raw length, so `"   "` renders `Not customized · 0 characters` while 3 characters are stored (`resources-settings.tsx:136-140`).
- **No-op save on trailing whitespace.** `dirty` compares untrimmed drafts while `nullableText` trims, so adding only trailing spaces to a custom model ID enables Save and sends an identical value.
- **Double labelling.** The model/reasoning selects carry both `aria-label` and a `Field label`; `aria-label` wins. Harmless, but the `aria-label` is redundant.

### ✅ Verified correct

- Partial `runtimeConfig` writes are safe — the server merges per field (`agent-service.ts:698-706`), so saving model/reasoning preserves `instructions` and `maxDurationMs`, and vice versa.
- "Runtime · Fixed after creation." is accurate: `UpdateAgentRequestSchema` genuinely does not accept `runtimeProvider` (`packages/shared/src/agent.ts:237-244`).
- Provider suggestions, `Custom model ID…`, and the historical `(saved value)` reasoning option all survive the port from `runtime-configuration.tsx`; `reasoningLabel` covers every allowed value of both providers (`minimal`→`max`).
- Model + reasoning save in one request, one dialog, one Save.
- Skills / Integrations render `Coming soon` as plain text with no link to the mock pages; footer copy matches the agreed wording verbatim.
- No `Feishu` in any new file.
- None of the reserved shell files are touched — the diff is exactly 3 new files.

### Next step

PR correctly stays **Draft**; I have not marked it ready. Fix Finding 1, push, and I will re-run QA against the new head.

liuchao-001:
## Independent QA — PR #309 (round 2)

**Status: FAIL** · **Tier: `test-only`** · **Head SHA: `04ec96bd2c3eb7ab27b30bd02459fc953d04e9b0`**

Independent QA run by code-reviewer, bound to the exact head above. This supersedes the round-1 report bound to `0614de99`. Any new commit invalidates this run.

### Round-1 Finding 1 is genuinely fixed — verified, not taken on trust

I re-ran the original probes against this head. All four symptoms are gone, and the legitimate path the effect existed for still works:

| Probe | Round 1 | This head |
|---|---|---|
| Summary leaks a concurrent model change while the dialog is open | true | **false** |
| Model save payload after a mid-dialog refresh | `expectedRevision: 9` → silent overwrite | **`expectedRevision: 4`** → server rejects, user is told |
| Summary leaks the other actor's instructions while the dialog is open | true | **false** |
| Instructions Save button after a mid-dialog refresh (no typing) | enabled | **disabled** |
| Refresh while the dialog is **closed** is adopted | — | **true** (legit path intact) |

The three non-blocking items you took are also confirmed fixed: whitespace-only instructions now count real stored length, a trailing-space-only custom model ID no longer enables Save (`dirty` now compares the normalized draft), and the redundant select `aria-label`s are gone without breaking the accessible names.

### Commands run (worktree at exact head, `git status` clean)

| Command | Exit |
|---|---|
| `pnpm install --frozen-lockfile` | 0 |
| `pnpm --filter @opentag/shared build` | 0 |
| `pnpm check` | 0 (same pre-existing warning in the untouched `onboarding-v2/server-backend.ts`) |
| `pnpm --filter @opentag/web typecheck` | 0 |
| `pnpm --filter @opentag/web test` | 0 — 39 files / 546 tests passed |

Your reported results reproduce exactly. Same evidence boundary as round 1: the components are still unmounted, so this run proves only that these automated checks passed or failed, and asserts nothing about live or integrated behavior. The black-box UI E2E obligation still transfers to the shell/mount PR.

---

### 🐛 Finding 2 — BLOCKING (new, introduced by the round-1 fix): `closeEditor` moves config *backward*

`model-settings.tsx:66-68`, `resources-settings.tsx:48-50`

```js
function closeEditor() {
  editingRef.current = false;
  setConfig(initialConfig);   // <-- unconditional
  ...
}
```

The freeze is right; the catch-up is not. `initialConfig` is not necessarily *newer* than the local `config`. Right after a successful save the component holds revision 5 while the prop is still the pre-save revision 4, so "catch up to the shell" rolls state **backward**. `Dialog onClose` is wired to `closeEditor`, so Escape and backdrop dismissal hit this too.

**Failure scenario (Instructions).** Instructions `"Review carefully."` @ revision 4. User edits to `"Be concise."` and saves — summary correctly reads `Custom · 11 characters`. Before the `onAgentChanged` refetch lands, the user reopens the dialog and presses Cancel:

```
after save:            summary = Custom · 11 characters      textarea = "Be concise."      (correct)
after reopen+Cancel:   summary = Custom · 17 characters                                    <- reverts to pre-save
reopen again:          textarea = "Review carefully."                                      <- the saved text is gone from the editor
next save:             {"expectedRevision":4,"runtimeConfig":{"instructions":"Third edit."}}
                       -> server rejects "Revision changed" with nothing concurrent happening
```

The Model block behaves identically: after save the summary shows `Medium`, and after reopen+Cancel it reverts to `High` while the next save sends the stale `expectedRevision: 4`.

**Reachability.** Two paths, and the second is not a race at all:
1. The user cancels inside the `onAgentChanged` refetch window — one network round trip, seconds on a slow link.
2. **The refetch fails** (offline blip, 5xx). React Query keeps the previous cached object, so `initialConfig` stays pre-save *indefinitely* and every subsequent Cancel re-wedges the block until a refetch succeeds.

**Suggested shape.** Don't adopt the prop on close; adopt the refresh you actually skipped. Record it in the effect when the gate rejects it, replay it on close, and discard it on a successful save (where `updated` is authoritative):

```js
useEffect(() => {
  if (editingRef.current) { pendingConfigRef.current = initialConfig; return; }
  setConfig(initialConfig);
}, [initialConfig]);

function closeEditor() {
  editingRef.current = false;
  if (pendingConfigRef.current) { setConfig(pendingConfigRef.current); pendingConfigRef.current = undefined; }
  ...
}
```
Clear `pendingConfigRef` in `save()` alongside `editingRef.current = false`. A `revision`-comparison guard would also work; the point is that the local `config` must never regress.

Worth a regression test in the same style as the ones you added: **save → reopen → Cancel with the prop held at its pre-save value → assert the summary still shows the saved value and the next save carries the new revision.**

### ⚠️ Non-blocking (carried over)

- **`message` never clears.** `Model settings saved.` / `Instructions saved.` still persists indefinitely once shown (`model-settings.tsx:118`, `resources-settings.tsx:97`); only reopening the dialog clears it. Not addressed this round — fine to leave, just flagging that it is still open.

### Next step

PR correctly stays **Draft**; I have not marked it ready. Fix Finding 2, push, and I will re-run QA against the new head.

liuchao-001:
## Independent QA — PR #309 (round 3)

**Status: PASS** · **Tier: `test-only`** · **Head SHA: `39a30b7b395ac4cb268e6666feba74cdadcbf926`**

Independent QA run by code-reviewer, bound to the exact head above. Supersedes the round-2 report bound to `04ec96bd`. Any new commit invalidates this run.

### Both findings verified closed — probed, not taken on trust

| Probe | Round 2 | This head |
|---|---|---|
| **F2** Model: after save → reopen → Cancel under a stale prop | reverts `Medium`→`High` | **stays `Medium`** |
| **F2** Model: revision on the next save | `4` (stale → spurious conflict) | **`5`** |
| **F2** Instructions: summary after reopen+Cancel | reverts to `Custom · 17 characters` | **stays `Custom · 11 characters`** |
| **F2** Instructions: reopened textarea | pre-save `"Review carefully."` | **`"Be concise."`** |
| **F1** Mid-dialog refresh leaks into the summary | false | **false** (still closed) |
| **F1** Save payload after a mid-dialog refresh | `expectedRevision: 4` | **`expectedRevision: 4`** (still frozen) |

I also checked the things a monotonic guard can plausibly break, and all three behave correctly:

- **The catch-up intent survives.** A refresh skipped during editing *is* replayed on Cancel — the fix didn't quietly delete the behavior the effect existed for.
- **Agent switch works.** Swapping to a *different* agent whose revision is *lower* still adopts it; the `current.id !== candidate.id` branch keeps a lower-revision neighbour from being rejected as stale. This was my main worry about a revision-only guard and it is handled.
- **A stale refetch landing after a save cannot regress local state**, and the next save still carries the post-save revision.

**The guard's load-bearing assumption holds.** `newerConfig` is only meaningful if the server really bumps `revision` on a runtime-config-only update. It does, unconditionally: `agent-service.ts:711` sets `revision: sql\`${agents.revision} + 1\`` on every successful update regardless of whether `runtimeConfig` changed, so `updated.revision` is always `expectedRevision + 1` and the test fixtures match production shape. Using `>=` rather than `>` is also right — it keeps an equal-revision refresh adoptable instead of pinning stale state.

### Commands run (worktree at exact head, `git status` clean)

| Command | Exit |
|---|---|
| `pnpm install --frozen-lockfile` | 0 |
| `pnpm --filter @opentag/shared build` | 0 |
| `pnpm check` | 0 (same pre-existing warning in the untouched `onboarding-v2/server-backend.ts`) |
| `pnpm --filter @opentag/web typecheck` | 0 |
| `model-resources-settings.test.tsx` | **12/12 pass** |
| `runtime-configuration.test.tsx` (the component this was ported from) | 11/11 pass |

### One honest caveat about my local full-suite runs

My local `pnpm --filter @opentag/web test` was **not** reliably green on this host, and I am not reporting that as a defect in this PR. Load average was 31–43 (another agent building in a sibling worktree, plus a VM at 189% CPU). Across three runs I got three *different* failing tests, all in files this PR does not touch, all timing-sensitive:

```
run 1  onboarding-v2 "turns green on its own after the terminal repair" + 2 onboarding-lab   (5000ms timeouts)
run 2  isolated re-run of just those 2 files: 13 failed, 161s for 2 files, 113s just collecting
run 3  full retry: 547/548 passed — "App Shell > does not expose the previous Account…" (1992ms)
```

The decisive check: I ran the same suite at **`origin/main`, without this PR's code at all** — it also failed, on yet another test, in the same way (`OnboardingV2Page > cancels a creation still in flight when the flow is restarted` → *Test timed out in 5000ms*). Same failure mode reproduces without this PR, so it is host contention, not a product defect.

The clean full-suite signal therefore comes from **CI at this exact head**, where `Checks` / `CI` / `Node 22.13.0` / `Node 26` / `CLI Pack Smoke` / `Container Smoke` are all SUCCESS. `Checks` runs `pnpm check` + `build` + `typecheck` + `pnpm test` + PostgreSQL integration + the Agent Runtime coverage gate, and the 548 web tests match the reported count.

### Evidence boundary

Unchanged from rounds 1–2, and this is the ceiling on the PASS above: the components are still deliberately unmounted, so there is no product route to drive. **This run proves only that the automated checks and targeted component probes listed here passed.** It asserts nothing about live, integrated, or cross-surface behavior — the black-box UI E2E obligation transfers in full to the shell/mount PR, which should exercise the mid-dialog-refresh and save→Cancel paths against a real server, since both defects lived in exactly that seam.

Copy rule re-checked at this head: no `Lark`, `Feishu`, or `Also called` strings in any of the three files, so the reversed Feishu-only wording needs nothing here. Scope is still exactly the three new files, with none of the reserved shell files touched.

### ⚠️ Non-blocking (still open, fine to leave)

- **`message` never clears.** `Model settings saved.` / `Instructions saved.` persists indefinitely once shown; only reopening the dialog clears it.

### Next step

No blocking findings. I have **not** marked the PR ready and have not merged — it is not my PR, and per the repo's gate the author marks it ready now that independent QA is bound to this head. If anything is pushed, this QA is void and I will re-run.

bestony (changes requested):
Request changes，范围只在 Resources 占位；Model 块和乐观并发我认可。

未挂载、expectedRevision、编辑中刷新不踩 draft，这些都对。请删掉或改写 Skills / Integrations 行，以及「chat with your agent to set up skills and integrations」。OpenTag 不建模 Agent 自有的 Skill / MCP 清单；工具来自这台 Computer 上的 runtime。不要把当前文案冻给下一趟挂载 PR。Instructions 留下。

liuchao-001:
Addressed the Resources scope concern in `9524209`:

- removed the Skills and Integrations rows
- removed the “chat with your agent to set up skills and integrations” copy
- kept Resources focused solely on Instructions
- updated the component test to assert those unsupported resource concepts are absent

`pnpm check` and Web typecheck pass. The changed Resources save behavior also passes in isolation; CI and renewed exact-head QA are pending.

bestony (approved):
上次卡的 Resources 占位已经去掉。`9524209` 上 Skills / Integrations 行和 “chat with your agent to set up skills and integrations” 都不在了；测试改成断言这两项不出现。Instructions、未挂载、expectedRevision、编辑中刷新不踩 draft、save 后不跟陈旧 shell prop 回退，这些都还在。

Approve。

组件仍故意未挂载，集成行为归 shell PR。这次 commit 作废了精确 head 的独立 QA，PR 保持 Draft，等新 head 再过一遍。Patch Coverage 红了，是 `coverage/unit/coverage-final.json` 缺失，同时间好几条无关 PR 也是这个失败，不是本 diff 的产品问题。

非阻塞：两个组件各写了一份相同的 `newerConfig`；区块标题还叫 Resources，里面只剩 Instructions。挂载时再决定要不要改名。

liuchao-001:
## Independent QA — PR #309 (round 4)

**Status: PASS** · **Tier: `test-only`** · **Head SHA: `95242092e60e0ece0157a290c89cc2d72f39d3bc`**

Independent QA run by code-reviewer, bound to the exact head above. Supersedes the round-3 report bound to `39a30b7b`. Any new commit invalidates this run.

⚠️ **One CI check is red at this head — `Enforce Patch Coverage` — and it is not caused by this PR.** Details at the bottom; please do not try to fix it inside #309.

### The product-scope correction is complete and clean

`9524209` touches only `resources-settings.tsx` (JSX row removal) and its test. `model-settings.tsx` is **byte-identical** to `39a30b7b`, the head that already passed, and the concurrency machinery in `resources-settings.tsx` is untouched — no diff on any `useEffect` / `editingRef` / `pendingConfigRef` / `newerConfig` / `setConfig` / `expectedRevision` line. So this round only had to prove the removal is complete and that it disturbed nothing.

Probed at this head:

| Check | Result |
|---|---|
| Rendered text contains `Skills` / `Integrations` / `Coming soon` / `chat with your agent` / `MCP` | **all false** |
| Any `link` role rendered | **false** |
| `Resources` heading / Instructions row / `Edit` button still present | **all true** |
| **F1** mid-dialog refresh leaks into the summary | false (still closed) |
| **F1** Save button after a mid-dialog refresh with no typing | disabled (still closed) |
| **F2** summary after save → reopen → Cancel under a stale prop | `Custom · 11 characters` (no regression) |
| **F2** revision on the next save | `expectedRevision: 5` (still correct) |

No leftover references anywhere in `agent-settings/` beyond the test's two absence assertions. No unused imports — `SettingsList` / `SettingsRow` are still carrying the Instructions row.

### Commands run (worktree at exact head, `git status` clean)

| Command | Exit |
|---|---|
| `pnpm install --frozen-lockfile` | 0 |
| `pnpm --filter @opentag/shared build` | 0 |
| `pnpm check` | 0 (same pre-existing warning in the untouched `onboarding-v2/server-backend.ts`) |
| `pnpm --filter @opentag/web typecheck` | 0 |
| agent-settings targeted suites (3 files) | **25/25 pass** |

### 🔴 `Enforce Patch Coverage` is red — repo-level CI break, not this PR

```
Error: ENOENT: no such file or directory, open '.../coverage/unit/coverage-final.json'
```

The gate crashes reading a coverage artifact that was never produced, so it never computes a percentage — this is not a coverage shortfall from your changed lines. I reproduced the **identical** failure on unrelated PRs:

```
33348098800  chore/static-analysis-hardening    same ENOENT
33347546162  feat/onboarding-choose-computer    same ENOENT
33346865068  fix/issue-315-messaging-cli-check  failure
33346691608  fix/integration-pool-starvation    failure
```

Only `test/coverage-gates` — the branch that introduced the workflow — has ever passed it. `.github/workflows/coverage.yml` is on `origin/main` (`ce2e721`) but **not** at this PR head; the PR is 9 commits behind main, so the gate arrives from the base branch. **It is not fixable inside #309** and I would not merge on it while it is red for everyone — that is a repo-level fix and a call for whoever owns the coverage gate.

Every other check is green at this exact head: `Checks`, `CI`, `Node 22.13.0`, `Node 26`, `CLI Pack Smoke`, `Container Smoke`, `Workflow and Shell Validation`, `Secret Scan`.

### Why I am citing CI rather than my own full-suite run

My local full web suite is unusable on this host right now — load average climbed from 31 to **66** during this round. With `--testTimeout=20000` the full suite gave 544/548 with 4 failures, all in `src/__tests__/app.test.tsx` (`Unable to find role="link" …`, which is `findBy`'s own 1000 ms default that `--testTimeout` does not raise). Running that one file **in isolation** made it *worse*: 9 failures of 130, and 311 seconds for a single file. A suite that degrades when given the machine to itself is contention, not a defect — and round 3 already established the precedent by failing the same way on `origin/main` without this PR's code at all.

So the full-suite claim rests on **CI at this exact head**, where `Checks` runs `pnpm check` + `build` + `typecheck` + `pnpm test` + PostgreSQL integration and passed.

### Evidence boundary

Unchanged, and it is the ceiling on this PASS: the components are still deliberately unmounted, so there is no product route to drive. **This run proves only that the automated checks and targeted component probes listed here passed.** It says nothing about live or integrated behavior. The black-box UI E2E obligation transfers in full to the shell/mount PR, which should exercise the mid-dialog-refresh and save→reopen→Cancel paths against a real server.

### ⚠️ Non-blocking

- **`message` never clears** (carried from round 1): `Instructions saved.` persists until the dialog is reopened.
- **Stale test name.** `describes empty instructions without exposing fake resource links` still asserts `queryByRole("link")` is null, but nothing in the component can render a link now, so the assertion is vacuous and the name refers to a removed concept. Cosmetic.

### Next step

No blocking findings in the PR's own scope. `reviewDecision` is now `APPROVED` (bestony's `CHANGES_REQUESTED` was superseded by an approval). I have **not** marked it ready and have **not** merged — not my PR. The only thing standing between this and merge is the repo-wide `Enforce Patch Coverage` break, which needs fixing at the repo level.

yuezengwu (changes requested):
Reviewed exact head `95242092e60e0ece0157a290c89cc2d72f39d3bc`.

**[BLOCKING] A newer shell config can be discarded when a save response arrives out of order.**

Both save paths replace local state with the mutation response and then unconditionally clear `pendingConfigRef` (`model-settings.tsx:94-100`, `resources-settings.tsx:72-76`). The pending value is not guaranteed to be older than that response. A reachable sequence is:

1. local revision 4 sends save A; the server commits revision 5, but its response is delayed;
2. another tab/admin commits revision 6, and the shell supplies revision 6 while this dialog is still busy, so the effect stores it in `pendingConfigRef`;
3. save A resolves with revision 5, which overwrites local state and deletes the pending revision 6.

The follow-up invalidation does not reliably repair this: React Query can structurally retain the already-cached revision-6 object, so the `useEffect([initialConfig])` dependency does not run again. The block stays on revision 5, displays stale data, and its next save produces a spurious revision conflict.

I reproduced this with a targeted RTL probe: after the exact sequence above, the component should display `Low` from revision 6, but it displays `Medium` from the revision-5 save response. The committed 12 targeted tests still pass, so this response-order case is currently uncovered.

Please reconcile the successful mutation result and any pending config monotonically instead of clearing the pending value blindly, in both components, and add the response-order regression test. The model/resources UI and direct Agent config API remain inside the allowed non-engineer scope; I found no schema, persistence, permission, or runtime-architecture expansion in this diff.

Verification: targeted committed suite 12/12, `pnpm check` (one documented pre-existing warning), Web typecheck, and `git diff --check` all pass. Context Tree read from `first-tree-ai/opentag-context-tree` at `093b03aa72e0c458ea52c0affbcc2c34b74e2e53`.

Separately, `Enforce Patch Coverage` is red because this head predates merged #324; current `main` contains the fix and is 16 commits ahead. Syncing after the code fix should regenerate that gate rather than treating the red job as ignorable.

bestony (changes requested):
同意 yuezengwu 这条 BLOCKING。我上次 Approve 时把乐观并发看成已经收口，漏了 save 响应晚于更新的 shell config 这一拍。

save 成功路径是 `setConfig(updated)` 然后无条件 `pendingConfigRef.current = undefined`。`closeEditor` 已经用 `newerConfig` 合并 pending，save 却把可能更新的 revision 6 扔掉，只留下这次 mutation 的 revision 5。随后 `onAgentChanged` 的 invalidate 也不一定能把同一个 cached 对象再推进 `useEffect`。

请按他说的改：两个组件都用单调规则调和 mutation 结果和 pending config，并补这条乱序回归。这条成立之后，我上次对 `9524209` 的 Approve 不再覆盖当前 head。

liuchao-001:
Addressed the out-of-order save response finding in both components.

The successful mutation response is now reconciled with `pendingConfigRef` through the existing monotonic `newerConfig` rule before updating local config and drafts. A pending revision 6 therefore wins over a delayed revision-5 mutation response, while a stale/equal pending value cannot regress the save result.

Added one response-order regression test per component. Both drive a deferred save response, supply the newer shell config while the dialog is busy, then resolve the older response and assert the newer model/instructions remain displayed. The targeted suite is now 14/14.

As requested, current `origin/main` was merged after the code fix, bringing in #324's Patch Coverage repair. New exact head: `1d30524d9043c58cac7998203e2bde42a07863dd`. PR remains Draft pending renewed review and independent QA.

bestony (approved):
上次卡住的 save 乱序已经修好。`bb9b2b3` 里两个组件都用 `newerConfig` 调和 mutation 结果和 pending config，draft 也跟着 resolved 走；测试覆盖了「revision 6 先到、save 的 5 后到」这一拍，Model 留下 Low，Instructions 留下并发那份。相对 main 仍然只有这三个文件。合入 main 是为了带上 #324 的 coverage 门。

Approve，head `1d30524`。

组件仍未挂载。这次 commit 作废了精确 head QA，PR 保持 Draft。yuezengwu 那边的 Request changes 还在，需要他再看一眼。

liuchao-001:
## Independent QA — PR #309 (round 5)

**Status: PASS** (scoped — read the boundary below) · **Tier: `test-only`** · **Head SHA: `1d30524d9043c58cac7998203e2bde42a07863dd`**

Independent QA run by code-reviewer, bound to the exact head above. Supersedes the round-4 report bound to `95242092`.

**Scope of this PASS:** the changes this PR makes are verified. It does **not** mean this head is mergeable — exact-head CI is red for an inherited reason, and a blocking review is still outstanding. Both are detailed at the end.

### First, an accountability note

The unconditional `pendingConfigRef.current = undefined` on the save path that @yuezengwu flagged as BLOCKING was **my suggestion** in the round-2 report, and I re-verified it in rounds 3 and 4 without probing the out-of-order response case. My round-3 probes covered a stale refresh landing *after* a save resolved, but never a *newer* refresh landing *while* the save was in flight. That gap was mine, and the finding is correct.

### The out-of-order fix is verified — probed independently, not read off the diff

I wrote my own deferred-promise probe rather than relying on the author's new tests:

| Probe | Result |
|---|---|
| **Model** — save commits rev 5 (response deferred), shell supplies rev 6 while busy, then rev 5 resolves | displays **`Low`** (rev 6) — the rev-5 response no longer wins |
| next save after that sequence | **`expectedRevision: 6`** |
| **Instructions** — same sequence | summary `Custom · 30 characters` (the rev-6 text), reopened textarea shows **`"Newer text from another admin."`**, next save **`expectedRevision: 6`** |
| **Over-correction guard** — a *stale* pending (rev 4) must NOT beat the rev-5 response | displays **`Medium`** (rev 5), no regression to `High`, next save **`expectedRevision: 5`** |
| **F1 regression sweep** — mid-dialog refresh still frozen | no leak, payload `expectedRevision: 4` |

The over-correction case matters as much as the fix itself: preferring `pendingConfigRef` unconditionally would have swapped one lost update for another. `newerConfig(updated, pendingConfig)` gets both directions right.

**The new regression tests have teeth.** I reverted the save-path reconciliation in both components and reran the committed suite with the Vite transform cache cleared: exactly the two new tests fail (`keeps a newer shell config when it arrives before the save response`, `keeps newer instructions when they arrive before the save response`) and nothing else — 2 failed / 12 passed. They reproduce the defect precisely rather than merely passing alongside the fix.

### Commands run (worktree at exact head, `git status` clean)

| Command | Exit |
|---|---|
| `pnpm install --frozen-lockfile` | 0 |
| `pnpm --filter @opentag/shared build` | 0 |
| `pnpm check` | 0 (152 warnings, all inherited from current `main`; none in these three files) |
| `pnpm --filter @opentag/web typecheck` | 0 |
| `agent-settings/` suites (4 files) | **31/31 pass** |

Scope is still exactly the three files against the merge base, and the PR is now level with `main`.

### 🔴 Why this head is not mergeable — neither reason is this PR's fault

**1. CI is red from a defect inherited from `main`.** `Checks`, `Node 22.13.0`, and `Node 26` all fail, and I read the job logs directly rather than accepting the author's attribution. All three fail on a *single* assertion in `apps/web/src/features/tasks-page.test.tsx` — a file byte-identical to `main` and untouched here. `main`'s own CI run fails on the same three jobs. Cause: #317 (`0e36908`) changed the component copy to `No Tasks yet. Message this Agent in your chat app to put it to work.`, while #294 (`e858cca`) landed afterwards still asserting the pre-#317 sentence. **#332 fixes exactly this**, and I have separately QA'd that PR: with it, CI at its head is fully green on all nine checks including `Enforce Patch Coverage`.

**2. `reviewDecision` is `CHANGES_REQUESTED`.** @yuezengwu's blocking review is still outstanding. A follow-up commit and an approval from another reviewer do **not** clear it — GitHub keeps the PR blocked until yuezengwu re-reviews or a maintainer dismisses it.

### Recommended sequence

Merging #332 first, then syncing `main` into #309 **once**, gets both cleared in one pass. That sync is a new commit and therefore a new head, which voids this QA — so the final QA round should run on that post-sync head rather than this one. Please avoid intermediate pushes in between; each one costs a full round.

### Evidence boundary

The components are still deliberately unmounted, so there is no product route to drive. **This run proves only that the automated checks and targeted probes listed here passed.** It says nothing about live or integrated behavior; the black-box UI E2E obligation transfers in full to the shell/mount PR, which by now has three distinct concurrency seams worth exercising against a real server: mid-dialog refresh, save→reopen→Cancel, and out-of-order save response.

### ⚠️ Non-blocking

- **`message` never clears** (carried since round 1): `Model settings saved.` / `Instructions saved.` persists until the dialog is reopened. Still the only open cosmetic item; the stale empty-instructions test name was fixed this round.

liuchao-001:
Final main sync completed at `4efffc4ef7e46a62288098502fc852944796fb54`, bringing in merged #330 and clearing the inherited Tasks assertion failure.

No PR product code changed in this sync. Post-sync validation:

- Model/Resources regression suite: 14/14
- Tasks suite: 32/32
- `pnpm check`: exit 0
- Web typecheck: exit 0
- pre-push format/lint/check/repository typecheck gates: exit 0

Re-requesting @yuezengwu because their blocking review must be explicitly cleared on this final head. Independent exact-head QA is also requested. PR remains Draft until both complete.

liuchao-001:
## Independent QA — PR #309 (round 6, final)

**Status: PASS** · **Tier: `test-only`** · **Head SHA: `4efffc4ef7e46a62288098502fc852944796fb54`**

Independent QA run by code-reviewer, bound to the exact head above. Supersedes the round-5 report bound to `1d30524d`. This is the first head where the PR's own changes are verified **and** CI is fully green.

### The sync claim is verified, not assumed

A `main` merge can quietly alter files during resolution, so I checked rather than taking "no product code changed" on trust:

- The PR's three files are **byte-identical** to round-5's head `1d30524d` (`git diff` over `agent-settings/` returns empty).
- The **entire** delta from `1d30524d` to this head is one line: `apps/web/src/features/tasks-page.test.tsx`, i.e. #330's assertion repair arriving through `main`. Nothing else.
- Scope against the merge base is still exactly the three files, and the branch is level with `main`.
- The formerly-red assertion now reads `No Tasks yet. Message this Agent in your chat app to put it to work.`, matching the shipped component copy.

### Evidence bound to this head

Because the components are byte-identical to a head I already probed, the risk here was the merge — but the exact-head gate means evidence must be re-bound, so I reran the full battery rather than citing round 5:

| Probe | Result |
|---|---|
| **Out-of-order** — save commits rev 5 (deferred), shell supplies rev 6 while in flight, rev 5 then resolves | Model displays **`Low`** (rev 6); next save **`expectedRevision: 6`** |
| **Out-of-order (Instructions)** — same sequence | summary **`Custom · 30 characters`** (the rev-6 text) |
| **Over-correction guard** — a *stale* pending (rev 4) must not beat the rev-5 response | displays **`Medium`** (rev 5), no regression to `High` |
| **F1** — mid-dialog refresh stays frozen | no leak; payload **`expectedRevision: 4`** |
| **F2** — save → reopen → Cancel under a stale prop | summary stays **`Custom · 11 characters`** |

| Command | Exit |
|---|---|
| `pnpm install --frozen-lockfile` | 0 |
| `pnpm --filter @opentag/shared build` | 0 |
| `pnpm check` | 0 (152 warnings, all inherited from `main`; none in the three PR files) |
| `pnpm --filter @opentag/web typecheck` | 0 |
| `agent-settings/` + `tasks-page.test.tsx` (5 files) | **63/63 pass** |

The previously-red `tasks-page` suite passes locally at this head, matching the merge's intent.

**CI at this exact head is fully green:** `Checks`, `CI`, `Node 22.13.0`, `Node 26`, `CLI Pack Smoke`, `Container Smoke`, `Workflow and Shell Validation`, `Secret Scan`, `Enforce Patch Coverage` — all SUCCESS.

### Evidence boundary

Unchanged and still the ceiling on this PASS: the components are deliberately unmounted, so there is no product route to drive. **This run proves only that the automated checks and targeted probes listed here passed.** It asserts nothing about live, integrated, or cross-surface behavior.

The black-box UI E2E obligation transfers in full to the shell/mount PR. Three concurrency seams surfaced over this review and each deserves a real-server pass once mounted: **mid-dialog refresh**, **save → reopen → Cancel**, and **out-of-order save response**. All three were invisible to a green unit suite at some point in this PR's history, so they are worth exercising through the real UI rather than trusting the component tests alone.

### Remaining blocker — not a QA finding

`mergeStateStatus` is `BLOCKED` solely because `reviewDecision` is `CHANGES_REQUESTED`. @yuezengwu's blocking review must be cleared by yuezengwu re-reviewing this head or a maintainer dismissing it; neither the follow-up commits nor @bestony's approvals clear it. Independent QA is complete and no longer gates this PR.

### ⚠️ Non-blocking (unchanged)

- **`message` never clears**: `Model settings saved.` / `Instructions saved.` persists until the dialog is reopened. Cosmetic; fine to leave or fold into the mount PR.

yuezengwu (approved):
Approved at exact head `4efffc4ef7e46a62288098502fc852944796fb54`.

The prior save-response race is closed: I independently reproduced a delayed revision-5 response arriving after shell revision 6, verified the newer model/instructions remain visible, and additionally verified the next save uses `expectedRevision: 6`. The committed model/resources and tasks suites pass 46/46; `pnpm check`, Web typecheck, and `git diff --check` pass. All 9 live CI checks are green, there are no unresolved review threads, and the independent QA PASS is bound to this exact head.

Scope remains within the non-engineer allowance: exactly three Web agent-settings files, UI/interaction plus the directly required existing `browserApi.updateAgent` call; no runtime architecture, schema, persistence, permission, or underlying data-structure change.

Remaining integration boundary: these components are not mounted by this PR, so the shell/mount PR must cover real-server E2E for mid-dialog refresh, save → reopen → cancel with stale props, and delayed older save responses after a newer shell config. Context Tree checked at canonical `master` commit `093b03aa72e0c458ea52c0affbcc2c34b74e2e53`. No author branch changes or merge performed.

PR #310, opened 2026-08-30
refactor(web): extract the shared setup pieces from onboarding

liuchao-001:
## Summary

Reconnecting a computer and rebinding a messaging app are the same work whether they are met while setting an Agent up for the first time or reopened later from its settings. Only the frame around them differs — a numbered flow with a Continue button, or a dialog over a settings page — so the inside of that frame now lives in `apps/web/src/setup/`, and onboarding holds those pieces rather than owning them.

**Moved out of `onboarding-v2/`:**

| Piece | From |
|---|---|
| `CommandBlock` (whole file) | `command-block.tsx` |
| `QrCode`, `Countdown` + `useRemaining`, `CheckLine`, `ConnectStatus` | `steps.tsx` |
| `deriveChecks`, `messagingCliCheck`, `CheckRow`, `CheckState`, `RuntimeStatus`, `MessagingCliStatus`, `formatRemaining` | `flow.ts` |
| `CHECK_COPY`, and the `connect` / `messaging` copy subtrees | `copy.ts` |
| The styling those pieces carry | `onboarding-v2.css` |

Onboarding's step components stay where they are and are now thin shells over the shared pieces. Copy is moved verbatim — no wording changes anywhere.

**The one design judgement.** `ReadinessFacts` deliberately did **not** move: it is the shape of an onboarding flow in progress, and settings has an Agent that already exists. So the shared functions take the statuses they actually read instead:

```ts
deriveChecks(runtime: RuntimeStatus | undefined): readonly CheckRow[]
messagingCliCheck(status: MessagingCliStatus | undefined): CheckState
```

Onboarding passes `readiness?.runtime` and `readiness?.messagingCli[provider]`; settings can pass the equivalent facts off `AgentDetailView.availability.dependencies`. Both narrowings are behaviour-preserving: `deriveChecks(undefined)` and `deriveChecks("checking")` already produced the same two pending rows, and an unreported messaging CLI still reads as `pending` rather than `failed`.

**CSS prefix.** The moved rules take an `ots-` prefix, because `otv2-` names a flow they no longer belong to and settings would otherwise render onboarding-named classes. Onboarding keeps `otv2-` for what stays its own (`otv2-choice`, `otv2-rail__*`, `otv2-frame`, `otv2-slot--outcome`, `otv2-slot--signin`, `otv2-field-error`, `otv2-command-lead`, `otv2-slack-install`). The rename is complete: every `ots-` class used in TSX is defined in `setup.css` and vice versa. Each visual piece imports `./setup.css` itself, so nothing that uses one has to know which sheet it came from.

**Follow-ups from review** (in `46a8b26`). `onboarding-v2/copy.ts` imports `../setup/copy.js` rather than the barrel: it wants strings, and the barrel evaluates the components beside them — React, `qrcode`, and a stylesheet side effect no tree shake removes — which `server-backend.ts` was inheriting for the sake of one copy object. `flow.ts` already reached the pure check layer directly; this makes `copy.ts` consistent with it. And `messaging-cli-routing.test.tsx` pins the messaging step to the provider actually chosen: narrowing `messagingCliCheck` to a bare status moved that indexing out to the call site, and every other test gives both providers the same status, so a hard-coded provider stayed green. Mutation-checked — hard-coding `.feishu` at `steps.tsx:335` turns it red.

**Non-goals.** No new behaviour, no API changes, no copy edits, nothing outside `apps/web/src/onboarding-v2/` and the new `apps/web/src/setup/`. `packages/shared` is untouched. The duplicate `formatRemaining` in `features/agents/computer-setup.tsx` was left alone: that file belongs to a parallel track.

## Testing

- `pnpm check` — pass (the one warning is a pre-existing unsafe-fix suggestion in `packages/server`, untouched here)
- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — `@opentag/web` **41 files / 543 tests, all green**, including every onboarding gatekeeper (`page.test.tsx`, `server-flow.test.tsx`, `flow.test.ts`, `layout.test.ts`, and the whole `pr262-*-findings` series)

`packages/client`'s `client-runtime-composition.test.ts` fails under `pnpm test`, with a different subset of ~5–10 timing-sensitive cases each run. It is **pre-existing and unrelated**: the same file fails the same way on an untouched `origin/main` worktree at `683f646` (7 failures on one run, 10 on the next), and this branch changes nothing outside `apps/web`.

Test coverage moved with the code rather than shrinking: `deriveChecks` / `messagingCliCheck` / `formatRemaining` cases now live in `setup/checks.test.ts` against the narrowed signatures, and the CSS assertions for the moved rules live in `setup/layout.test.ts`. `onboarding-v2/layout.test.ts` keeps the assertions for the rules that stayed. Reviewed at `5c2ef7d` as HEALTHY with independent QA; `46a8b26` adds the two follow-ups above.

## Breaking changes

None. This is a pure refactor; onboarding renders exactly what it rendered before.

## Checklist

- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. (See the note above on the pre-existing `packages/client` flake.)

### DIFF
--- a/apps/web/src/onboarding-v2/copy.ts
+++ b/apps/web/src/onboarding-v2/copy.ts
@@ -6,7 +6,7 @@
  * setup copy, so the same work says the same thing on both surfaces.
  */
 
-import { SETUP_COPY } from "../setup/index.js";
+import { SETUP_COPY } from "../setup/copy.js";
 import type { CloudRuntime, Destination, Runtime, StepId, TokenSource } from "./flow.js";
 
 export const STEP_LABELS: Record<StepId, string> = {
--- a/apps/web/src/onboarding-v2/copy.ts
+++ b/apps/web/src/onboarding-v2/copy.ts
@@ -1,9 +1,13 @@
 /**
  * Every user-visible string in the onboarding flow, in one place. Copy is reviewed far more often
  * than layout during this phase, so it stays out of the components and can be read end to end.
+ *
+ * The two steps that are also reachable from an Agent's settings read their words from the shared
+ * setup copy, so the same work says the same thing on both surfaces.
  */
 
-import type { CheckRow, CheckState, CloudRuntime, Destination, Runtime, StepId, TokenSource } from "./flow.js";
+import { SETUP_COPY } from "../setup/index.js";
+import type { CloudRuntime, Destination, Runtime, StepId, TokenSource } from "./flow.js";
 
 export const STEP_LABELS: Record<StepId, string> = {
   agent: "Create agent",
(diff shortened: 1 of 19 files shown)


bestony (approved):
Approve。抽取保行为，ots- 与 setup.css 一一对应，copy 直连 setup/copy.js 也避免 barrel 把组件拉进 server-backend。settings 还没消费 setup/，computer-setup.tsx 的 formatRemaining 按你说的留着，都不挡合。

PR #314, opened 2026-08-31
test(server): close reset fixture pools and pin the re-board invariants

liuchao-001:
Two pieces of debt the re-board work left in the reset integration suite. Test-only; no product code changes.

## Fixture pools are never closed

`fixture()` calls `createDatabaseClient`, which opens a pool of up to ten connections, and nothing ever closes it. With thirteen fixtures in the file, a full run accumulates pools until it ends.

The cap it runs into is per **server**, not per file, so the suite that pays for this is usually not this one — the neighbouring integration files get the connection errors and the timeouts. Every case added here tightens it, which is a bad property for a file that exists to be extended.

`afterEach` now closes the pools opened during the test, and each pool is registered immediately after it is created, so a fixture that throws part-way through seeding still gives its pool back.

To be straight about the evidence: I found this while adding cases here, where it produced `sorry, too many clients already` outright. I cannot give a clean local before/after measurement — this machine is at load average ~30 from unrelated processes, and runs of both this suite and the web suite are killed by wall-clock timeouts regardless of the change. **CI is the check for this PR, not my local run.** CI is green on `main`, so this is not fixing a broken build; it removes an accumulation that gets worse with every case anyone adds.

## Re-board had one case behind it

`reboard` merged with a single integration case covering resource preservation. Four more pin the properties that make it a distinct operation:

- **It clears setup completion without the cleanup a first-run commit verifies** — seeded with a live Agent and a usable connect code. The verification inside the first-run commit step refuses exactly that state, and re-board reaches the same marker without it, which is why the two cannot share a commit path.

  To correct what an earlier version of this description said: `resetOnboarding` does **not** raise a conflict on this state. It deletes the Agents and revokes access *before* it commits, so the public method cleans the state up and succeeds. The refusal belongs to the private verification the commit step runs, which exists for the case where cleanup did not take effect. The design conclusion is unchanged; the earlier wording named the wrong function, and the test name carried the same slip until `97026c4`.
- It is idempotent, so a re-boarded Account can be re-boarded again.
- It reaches only the Account that asked, leaving another tester's setup complete.
- It refuses a suspended Account — inherited from the row lock it shares with the full reset, which is the whole reason that lock is reused.


bestony (approved):
Approve.

Test-only, single-file. afterEach pool close is the right fix for this suite’s own Testcontainer max_connections, and the four re-board cases match OnboardingResetService: it succeeds with live agents and usable codes (the state #verifyCleanedUp would refuse), is idempotent, is Account-scoped, and inherits AUTH_USER_SUSPENDED from lockActiveAccount.

Non-blocking: register the pool immediately after createDatabaseClient so a fixture() throw during seed still gets closed.

CI is green, including onboarding-reset.test.ts (20 tests).

bestony (approved):
Re-approve on 755f0ed.

This commit is the non-blocking note from the previous review: the fixture pool is registered immediately after createDatabaseClient, so a throw during seed still reaches afterEach teardown. No product-code change.

CI on this head was still running at review time; the delta itself is the one-line registration move.

liuchao-001:
Applied in `755f0ed` — the registration now sits directly after `createDatabaseClient`, before any seeding runs.

You were right that it mattered rather than being cosmetic: `bootstrapInitialAdmin` and `seedAccount` both run between the old registration point and the pool being created, so a fixture that threw during seeding leaked its pool in precisely the way this teardown exists to prevent. That failure mode is also self-obscuring — the leak surfaces as timeouts in whatever suite runs next, not as the seeding error that caused it.

Local verification is not available to me at the moment: this machine is at load average ~57 from unrelated long-running processes, and both this suite and the web suite are being killed by wall-clock timeouts regardless of the change. CI is the check, and it runs `pnpm --filter @opentag/server test:integration` in the `Checks` job, so it covers this file's 20 cases directly.

bestony (approved):
Approve。测试-only，方向对。

fixture 现在每个用例开一个 postgres 池（默认 10 连接），文件结束才释放，会把「too many clients」推给旁边的 integration 套件。afterEach 关掉本测打开的池，这个性质值得钉住。四条 re-board 用例也把「和 full reset 不能共用 commit 路径」说清楚了。

两个 nit，不挡：
1. `openPools.push(client.sql)` 在 bootstrap/seed 之后。setup 中途抛错时这个池不会被 afterEach 收到。push 紧挨 `createDatabaseClient` 更稳。
2. 「succeeds on the very state a full reset refuses to commit on」测的是 `reboard` 成功，并没有真的去打 `resetOnboarding` 的 verify 边界。`resetOnboarding` 作为整体会先清理再 commit，公开方法并不会在这种状态下直接 409。用例仍然有价值，名字略满。

liuchao-001:
New head `96d7db1`, which supersedes the approval of `755f0ed`. Two changes since then, both from @bestony's review.

**Nit 2 — the test name overclaimed, and so had I.** The case was called *"succeeds on the very state a full reset refuses to commit on"*. You were right that `resetOnboarding` does not refuse that state: it deletes the Agents and revokes access *before* `#commitFirstRunState` runs, so the public method would clean the state up and succeed. Only the private `#verifyCleanedUp` gate refuses it, and it exists for the case where cleanup did not take effect.

I checked this against the source rather than taking it on trust, and the ordering is exactly as you described. The case is now *"clears setup completion without the cleanup a first-run commit verifies"*, and the comment states which gate refuses the state and why `resetOnboarding` never meets it.

Worth being explicit that this was not only a naming slip: I had been repeating that same claim as the justification for why the two modes cannot share a commit path — in the earlier PR description and elsewhere. The design conclusion survives, because the commit path really does verify and re-board really must skip it. The sentence I was using to justify it named the wrong function.

**Nit 1 was already in `755f0ed`.** Your re-approval noted it; flagging it only so the two reviews are not read as one outstanding item.

**`origin/main` merged in.** The `Enforce Patch Coverage` job failed on `755f0ed` with `ENOENT: coverage/unit/coverage-final.json` — not a coverage shortfall. The branch predated #288, which added the gate, and #293, which fixed the coverage measurement it reads. It also failed on at least one unrelated branch in the same state. Merging main should make the gate honest rather than red for staleness; if it still fails, the cause is not this PR's content and I will say so rather than merge past it.

bestony:
`96d7db1` is the merge of `origin/main`. `onboarding-reset.test.ts` is identical to `755f0ed` — the case is still named "succeeds on the very state a full reset refuses to commit on", and the old comment is still there.

Nit 1 is in `755f0ed` as you said. The rename described here has not landed on this head. Happy to re-approve once it does.

bestony (approved):
Re-approve on 97026c4.

The case is now named after the gate that actually refuses: the first-run commit's `#verifyCleanedUp`, not `resetOnboarding`. The comment states the ordering correctly — cleanup runs first, so the public reset would succeed on this seeded state; re-board skips that verify, which is why the commit paths cannot be shared. Assertions are unchanged.

Non-blocking: the PR description still says `resetOnboarding` would raise the conflict here. Same slip the test name had.

liuchao-001:
You were right, and my previous comment was wrong. The rename is on `97026c4`, not on `96d7db1`.

What happened: I edited the file, then merged `origin/main` and pushed without committing the edit first. The merge succeeded because it did not touch that region, so nothing warned me, and the change sat in my working tree while I described it as landed. I verified against the pushed head this time rather than against my editor — `git show origin/fix/integration-pool-starvation:…` now contains the new name.

Thanks for checking the head instead of taking the comment at its word. I claimed something shipped that had not, which is worse than the nit itself.

`97026c4` renames the case to *"clears setup completion without the cleanup a first-run commit verifies"* and replaces the comment with one that names the gate accurately: the verification inside the commit step refuses that state, `resetOnboarding` never meets it because it removes the resources first, and that gate is what re-board skips.

liuchao-001:
The red `Enforce Patch Coverage` check on this PR is not this PR. I said I would check rather than merge past it, so: it fails with `ENOENT` on `coverage/unit/coverage-final.json`, a file nothing in the repository writes — `scripts/unit-coverage.mjs` emits per-workspace reports plus `coverage/unit/coverage-summary.json`, and the gate reads a merged detailed report that does not exist.

My earlier guess that merging `main` would clear it was wrong; this branch now has current `main` and it still fails, as it does on unrelated branches. Filed as #320 with the diagnosis. I have not patched it here — the repair touches the per-workspace design #293 introduced deliberately, and that is not a passing PR's call.

Every other check on this head is green, including `Checks`, which runs `check`, `typecheck`, `test` and `server test:integration`.

yuezengwu (commented):
Reviewed exact head `97026c434618897b7a3ce081f5532d555f12a87f`.

No correctness blocker found in the one-file test-only delta. The pool is registered immediately after `createDatabaseClient`, so a seeding failure is covered by the file-level `afterEach` drain. The four re-board cases match the current service boundary: clear only the Account setup marker under the active-Account row lock, preserve resources, remain idempotent and Account-scoped, and reject a suspended Account. The corrected test name/comment also distinguishes the private first-run verification gate from the public `resetOnboarding` cleanup flow.

Independent verification on this head:
- `pnpm build` — pass
- `pnpm typecheck` — pass
- targeted `onboarding-reset.test.ts` — 20/20 pass
- full `@opentag/server` integration suite — 287/287 pass
- `git diff --check` — pass
- `pnpm check` — exit 0, with one unrelated pre-existing optional-chain warning in `apps/web/src/onboarding-v2/server-backend.ts`

The red `Enforce Patch Coverage` run is the repository-wide missing-artifact handoff tracked in #320/#324; every other live check is green.

This change is Server integration/fixture infrastructure, outside the automatic approval scope for a non-engineer author. This review is therefore COMMENT-only and does not replace human final approval or merge.

PR #311, opened 2026-08-30
feat(web): name Feishu for what the connect flow can actually mint

liuchao-001:
`@liuchao-001` settled the model: **Feishu and Lark are one channel behind a regional switch**, not two
providers. There will be no `lark` beside `feishu` and `slack` in `ImProviderSchema`; which brand a
team gets should eventually be decided from the user's country.

That switch already exists in the Server — and it resolves too late to help anybody.

## The flow was naming a brand it cannot mint

The redesigned onboarding offered **"Lark"**, described as **"Also called Feishu"**, then said Lark
through the QR panel, the completion line and the CLI check.

`feishuDomainForWorkspaceBrand` returns `Domain.Lark` only when the binding's `external_team_brand` is
`"lark"`, and that brand is written *from the authorization result* (`im-binding-service.ts:1979`). A
binding that does not exist yet has no brand, so **a first-time connect always mints its code against
the Feishu domain.** Naming Lark there walks a Lark tenant into a QR code their account cannot
authorize, and nothing in the flow can tell them why.

So the picker names **Feishu** alone: not because Lark is a different channel, but because Feishu is
the only brand this flow can currently deliver. Deciding the brand up front is **#312**.

## One naming point

`titleCase(binding.provider)` turned the identifier straight into displayed text at nine sites,
including the screen-reader name on every Agent card. Other sites printed `task.source.provider` raw,
so a Feishu Task read as the lowercase `feishu`.

**`messagingProviderLabel` is now the only way a provider becomes reader-facing text.** Its `switch` is
exhaustive, so a provider joining the schema fails to compile until somebody names it, and it is the
single place a brand-aware name will hook in for #312. `titleCase` goes back to formatting reasoning
effort.

## What a reader sees differently

- The onboarding picker names **Feishu**, with no apposition, in `en` and `zh`.
- The completion line names **the channel the reader actually connected** — a Slack install used to end
  by pointing them at Feishu. An unknown provider degrades to "your messaging app" rather than guessing.
- Task **list** and **detail** name the channel instead of printing the lowercase id.
- Agent cards, messaging settings and the QR panel all resolve their brand from the one helper.

## i18n

Per `@yuezengwu`'s owner decision, no carve-out: this PR's reader-facing copy lives in the owning
`apps/web/messages/<area>/{en,zh}.json` catalogues and renders through `m.*()`. **No catalogue entry
spells a brand** — each takes a `{provider}` placeholder fed from `messagingProviderLabel`, which is the
boundary #335 will enforce generally.

[...]

PR #316, opened 2026-08-31
fix(web): drop unused messaging-cli check-row copy

### LINKED ISSUE #315: Onboarding never renders the messaging CLI check
`CHECK_COPY["messaging-cli"]` in `apps/web/src/onboarding-v2/copy.ts` is never rendered.

`deriveChecks` only emits `runtime-cli` and `runtime-auth` (`flow.ts:239-240`), and `flow.test.ts:114`
pins exactly that pair. `"messaging-cli"` is a member of the `CheckRow` id union (`flow.ts:214`) and
`messagingCliCheck` exists (`flow.ts:245`), but nothing puts the row on screen — the computer-check
test asserts its title is absent.

So the messaging step never tells a reader that the CLI their channel is delivered through is missing
from their computer. That is either a gap worth closing or dead weight worth removing; today it is
neither, which is the part worth fixing.

**Decide one of:**

- **Render it on the messaging step.** `messagingCliCheck` already computes the state, and the copy
  is written: title `lark-cli is installed`, failure detail `We need lark-cli to send Feishu
  messages.` The row names the command rather than the channel deliberately — the vendor ships one
  binary for both of its apps, and there is no "Feishu CLI" for a reader to go and find.
- **Remove it.** Drop the copy entry, the union member, and `messagingCliCheck`.

Not done inline in opentag#311 because #310 is extracting this module into `apps/web/src/setup/`;
removing a typed union member there would hand that branch a conflict for a cleanup that is not
#311's subject. Found by `@bestony` reviewing #311.

bestony:
## Summary

#315 offered two ways out of a typed union member that never became a row: render `CHECK_COPY["messaging-cli"]` on the messaging step, or delete it.

The messaging step already names a missing provider CLI. `messagingCliCheck` drives `COPY.messaging.cliMissing` — a sentence that works for Feishu and Slack. `CHECK_COPY["messaging-cli"]` was CheckLine copy that nothing rendered, and it is hardcoded to Lark/`lark-cli`. Putting that row on screen would tell a Slack user about a binary they do not need.

This PR takes the remove side, with one correction: **`messagingCliCheck` stays.** It is not dead. What goes is the unused CheckRow id and its copy. A focused test pins the live contract: the chosen provider's missing CLI is a sentence, not a numbered check row.

Related in-flight work:

- #310 extracts this module into `apps/web/src/setup/`. This change is on current `main`; that branch will need a rebase (the same files move).
- #311 rewrote the unused CheckLine copy to name `lark-cli`. That wording is still a good fit for the Feishu *sentence* later; it should not come back as a third computer-style row.

Closes #315

## Testing

- `pnpm check` — pass (the one warning is a pre-existing unsafe-fix suggestion in `server-backend.ts`, untouched)
- `pnpm --filter @opentag/web typecheck` — pass
- `pnpm --filter @opentag/web test` — **35 files / 446 tests**, including `messaging-cli-warning.test.tsx`, `flow.test.ts`, and `page.test.tsx`

## Breaking changes

None. The onboarding messaging step still shows the same missing-CLI sentence. Only unused CheckLine copy is gone.

## Checklist

- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. (`pnpm check` and `@opentag/web` typecheck + test as above; this change is web-only.)

### DIFF
--- a/apps/web/src/onboarding-v2/messaging-cli-warning.test.tsx
+++ b/apps/web/src/onboarding-v2/messaging-cli-warning.test.tsx
@@ -1,0 +1,46 @@
+/**
+ * A missing messaging CLI is named as a sentence on this step, not as a numbered check row.
+ *
+ * `messagingCliCheck` picks the chosen provider's status out of `ReadinessFacts`. A hard-coded
+ * provider would warn a Slack user about a CLI they do not need, or stay silent about one they do.
+ */
+
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { COPY } from "./copy.js";
+import type { MessagingProvider, ReadinessFacts } from "./flow.js";
+import { MessagingStep } from "./steps.js";
+
+/** Lark's CLI is missing; Slack's is not. */
+const readiness: ReadinessFacts = { runtime: "ready", messagingCli: { feishu: "install", slack: "ready" } };
+
+const warningFor = (provider: MessagingProvider): string => COPY.messaging.cliMissing(COPY.messaging[provider].title);
+
+function renderStep(provider: MessagingProvider) {
+  render(
+    <MessagingStep
+      computerOnline
+      messaging={{ kind: "idle" }}
+      onChoose={() => undefined}
+      onSlackInstall={() => undefined}
+      onStart={() => undefined}
+      provider={provider}
+      readiness={readiness}
+    />,
+  );
+}
+
+describe("the messaging step's CLI warning", () => {
+  it("names the chosen provider's missing CLI as a sentence, not a check row", () => {
+    renderStep("feishu");
+    expect(screen.getByText(warningFor("feishu"))).toBeTruthy();
+    expect(document.querySelectorAll(".ots-check")).toHaveLength(0);
+  });
+
+  it("stays silent for a provider whose CLI is present, on the same facts", () => {
+    renderStep("slack");
+    expect(screen.queryByText(warningFor("slack"))).toBeNull();
+    expect(screen.queryByText(warningFor("feishu"))).toBeNull();
+    expect(document.querySelectorAll(".ots-check")).toHaveLength(0);
+  });
+});
--- a/apps/web/src/onboarding-v2/page.test.tsx
+++ b/apps/web/src/onboarding-v2/page.test.tsx
@@ -215,8 +215,9 @@ describe("OnboardingV2Page", () => {
     await settleCheck();
 
     expect(screen.getByText("Codex CLI is installed")).toBeTruthy();
-    // The messaging CLI is not checked here: no provider has been chosen yet.
-    expect(screen.queryByText("Lark CLI is installed")).toBeNull();
+    // The messaging CLI is not a computer-step row: no provider has been chosen yet, and a
+    // missing one is named later as a sentence on the messaging step, not a third check line.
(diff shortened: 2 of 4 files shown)


liuchao-001:
## #310 merged, so this PR's two source edits now land in `apps/web/src/setup/`

The body above anticipated #310 and said that branch would need the rebase. It merged first (`ce432c9`), so the rebase is this one's. The change itself is unaffected — the reasoning for deleting the row still holds, and I verified it independently on `main` rather than from the description:

```
deriveChecks() emits only { runtime-cli, runtime-auth }   setup/checks.ts
CHECK_COPY[check.id] is read only by CheckLine             setup/components.tsx:65
nothing constructs a row with id "messaging-cli"
```

Unreachable, confirmed. `page.test.tsx:219` asserting the row's *absence* is the other half of that evidence.

### Where the four files go

| this PR edits | after rebase |
|---|---|
| `onboarding-v2/copy.ts` — drop `CHECK_COPY["messaging-cli"]` | **`setup/copy.ts:39`** |
| `onboarding-v2/flow.ts` — narrow `CheckRow.id`, retitle comments, `messagingCliCheck` doc | **`setup/checks.ts:24`** |
| `messaging-cli-warning.test.tsx` (new) | unchanged, stays in `onboarding-v2/` |
| `page.test.tsx` | unchanged, stays in `onboarding-v2/` |

The whole `messaging-cli` surface on `main` is exactly two lines, both in the shared module:

```
apps/web/src/setup/checks.ts:24   readonly id: "runtime-cli" | "runtime-auth" | "messaging-cli"
apps/web/src/setup/copy.ts:39     "messaging-cli": { ... }
```

`CHECK_COPY` is a `Record` keyed by that union, so the two must be removed together — which this PR already does, just in the new location. No test on `main` references the id, so nothing else should move.

One consequence worth stating since the type is now shared: narrowing `CheckRow.id` narrows it for **Agent Settings** too, which will consume `setup/` for its reconnect dialog. That is the right outcome — settings wants the two runtime rows plus the messaging sentence, not a third row — so this deletion helps that track rather than constraining it.

### Ordering with #311

#311 renames the strings inside this same block and is currently `CONFLICTING`. **Landing this deletion first is the cheaper order**: it removes the block, so #311 never has to get that wording right in copy that is on its way out. I have noted that on #311. The `lark-cli` wording stays live and correct on `setup/copy.ts:44`, the sentence that actually reaches a reader.

Clear `apps/web/node_modules/.vite` before trusting a post-rebase test run — a stale transform cache produced a confirmed false green during #310's review.

bestony:
Rebased onto current `main` after #310.

The two source edits now land where you mapped them:

- `apps/web/src/setup/copy.ts` — drop `CHECK_COPY["messaging-cli"]`
- `apps/web/src/setup/checks.ts` — narrow `CheckRow.id`, keep `messagingCliCheck`

The new warning test stays in `onboarding-v2/`. After the `ots-` prefix rename it asserts `.ots-check` is absent on the messaging step, not `.otv2-check`. `page.test.tsx` is unchanged in place.

Cleared `apps/web/node_modules/.vite` before the post-rebase run.

- `pnpm check` — pass (pre-existing warnings only)
- `pnpm --filter @opentag/web typecheck` — pass
- `pnpm --filter @opentag/web test` — **38 files / 454 tests**

PR #313, opened 2026-08-31
feat: let an Agent exist before it has a Computer

liuchao-001:
## Summary

An Agent can now be created before it has a Computer, and bound to one afterwards.

`POST /agents/:id/computer/rebind` already existed, so the Computer binding was always changeable —
but `CreateAgentRequest.computerId` was required, which stated a creation-time precondition the
product does not have. This makes the field optional and carries the absent binding honestly through
every surface.

- **Contract.** `CreateAgentRequest.computerId` is optional. `AgentSummary.computer` and
  `AgentAdminConfig.computerId` are nullable. `computer` is always present in the response, so a
  reader never has to tell "no Computer" apart from "field missing"; `null` and "we could not read
  the Computer" stay distinct facts.
- **Persistence.** `agents.computer_id` and `agents.workspace_computer_id` become nullable, with a
  new `agents_computer_pair` check so the two clear together. An Agent created without a Computer
  takes the Account's schema-compatibility Workspace fill for the `workspace_id` NOT NULL column.
- **Binding later.** `rebindById` already moved an Agent between Computers and now also binds one
  that has none. `packages/client` and `apps/web` gained a `rebindAgentComputer` call, and the CLI
  gained `opentag agent bind <agent-id> [--computer <uuid>]`. In the Web, one shared surface serves
  both places an Agent can be found without a Computer — its Settings, and an onboarding run that
  found one. An Account may hold none, one, or several. With several, which one an Agent runs on is
  the reader's to say and the surface asks: nothing binds on an array index. With exactly one there
  is nothing to disambiguate, so the read is the decision and no click is asked for. With none,
  enrolment is offered, and the connect step reports the machine it enrolled — so what gets bound is
  the Computer that answered the command rather than whatever a re-read finds. A failed read stays
  distinct from an empty one, so nobody is sent to enrol a machine they already own; a failed bind
  is reported against the exact target and retried without issuing a second connect code. An
  automatic bind still in flight when the Account gains a second Computer is revalidated by reading
  again rather than asserted.
- **An unbound Agent cannot run.** Placement derives the Computer from the Agent, so every execution
  path refuses rather than inventing one: Session placement, inbound IM delivery (the message is
  still recorded, no delivery is created), Feishu setup and activation, and Slack configuration at
  both ends — OAuth start, so nothing is installed for an Agent that has nowhere to run, and the
  commit, so an Agent unbound in between cannot claim a route. Every refusal is a deterministic 409
[...]

Issue #306, opened 2026-08-30
Flaky web test: "lets no new save race an in-flight refresh retry" depends on a 40ms wall-clock window


liuchao-001:
### Summary

`apps/web/src/__tests__/app.test.tsx` → `lets no new save race an in-flight refresh retry` fails intermittently on a loaded machine:

```
FAIL  |web| src/__tests__/app.test.tsx > OpenTag Web App Shell > lets no new save race an in-flight refresh retry
AssertionError: expected false to be true // Object.is equality
```

Observed once in a full `pnpm --filter @opentag/web test` run; the same suite passed on the immediately preceding and following runs, and the file passes in isolation.

### Why it races

The test arms the `/api/v1/me` mock with a fixed wall-clock delay and then asserts that the form is locked *while that request is still in flight*:

```ts
installApi({ meFailuresAfterProfileUpdate: 1, meDelayMsAfterProfileUpdate: 40 });
…
fireEvent.click(screen.getByRole("button", { name: "Retry refresh" }));
expect(await screen.findByRole("button", { name: "Refreshing…" })).toBeTruthy();
expect(displayName.disabled).toBe(true);   // ← the failing line
```

The delay is 40ms:

```ts
if (profileUpdated && options.meDelayMsAfterProfileUpdate) {
  await new Promise((resolve) => setTimeout(resolve, options.meDelayMsAfterProfileUpdate));
}
```

`findByRole("Refreshing…")` polls, so under load it can resolve late enough that the 40ms has already elapsed and the refresh has settled by the time `displayName.disabled` is read. The window the assertion depends on is defined by a timer rather than by the thing being tested, so the test gets slower-machine-dependent rather than wrong.

The property under test is real and worth keeping — a second retry and a competing submit must not slip past an in-flight refresh. Only the way the in-flight window is held open is fragile.

### Suggested direction

Hold the refresh open on a promise the test resolves explicitly, instead of on a 40ms timer — the suite already uses that shape elsewhere (`releaseAgentRead()` in the stale-detail tests). Then the in-flight assertions run inside a window the test controls, and the assertion cannot lose a race against wall-clock time.

Please don't paper over it with a retry: the assertion would still be timing-dependent, just less visibly.

### Provenance

Hit while running the suite for #303. Not caused by that PR — the flaky test is at line ~1337 and #303's only new test is at ~1983, so it runs first, and #303 touches nothing on the account-profile path.

PR #317, opened 2026-08-31
feat(web): give the Agent its dependencies as a card, not a banner

liuchao-001:
## Summary

Second round of feedback on the Agents surface, stacked on #303 (base is that PR's branch, so this diff is only the new work).

**Computer and Messaging become a Connection card beside Usage.** They are what an Agent needs before it can do anything, and neither could be read without acting: Computer state appeared only as a red banner when it broke, Messaging was a bare glyph in the header, and a healthy Agent showed neither. They now sit in one card, each row naming what is true, why, and the way to change it. Usage takes the other half of the row instead of the full width, so the pair reads as one answer to "can this Agent work right now".

**The recovery banner is removed.** One offline Computer was stating itself three times — on the list card, in the banner, and in the header. The Connection card carries both the reason and the repair, per dependency.

**The header messaging control is removed**, since the card names the channel and carries its exit.

**New Agent steps down to `secondary`.** Kumo has no CTA variant beyond `primary`, so the weight had to come off the variant rather than the component: the filled brand green was the only heavy element on a page built from hairline rules. Its plus now leads the label and takes `bold` — that is what makes it read as an icon rather than a typed character. `Icon` did not accept a `weight` prop before this.

Also: `View details` → `View Details` to match `All Tasks`; the empty Tasks line says what to do rather than describing what the region will contain; and the Computer row stops naming the machine twice.

## Interaction with #313 — please read before merging either

#313 makes `agent.computer` **nullable** and adds a `computer_not_bound` reason. Two consequences for this PR:

1. The Connection card reads `agent.computer.displayName` and `agent.computer.platform` for its identity line. After #313 that dereference is unsound and needs a not-bound branch.
2. bestony asked on #303 that `computer_not_bound` keep its exit "in the detail banner, not back on the card". That is the right principle, but **this PR deletes the banner** — so on current `main` + this PR the correct home for that exit is the Connection card's Computer row, which is the detail-page, per-dependency exit the banner used to be. The card's own list-page exit stays removed.

I did not resolve this here because #313 is not merged and I did not want to guess at its final shape. Whichever lands second should take the not-bound state into the Connection card.

## Testing

- `pnpm check` — exit 0
- `pnpm typecheck` — exit 0
- Verified in a real browser against a local server and database at this head: the Agents list, the Agent detail page with an offline Computer and a Slack channel needing re-authorization, and the button at rest.

**`pnpm test` is not a clean signal from my machine right now and I am not claiming it as one.** Another session is running ~16 concurrent vitest workers here; load average went from 22 to 70 during verification, and the identical tree produced 0, 1, 4, and then 130 failures on successive full runs, each time a different set, including files this PR never touches (login, onboarding). Individual files pass in isolation. Even at a 60s timeout the suite starved. CI on a clean runner is the signal to trust here; please treat a green CI run as the verdict rather than my local numbers.

## Breaking changes

None.

### DIFF
--- a/apps/web/src/features/agents/agent-detail-page.tsx
+++ b/apps/web/src/features/agents/agent-detail-page.tsx
@@ -7,8 +7,10 @@ import { AgentTasksSection } from "../tasks-page.js";
 import type { AgentDetailView } from "./agent-model.js";
 import {
   type AgentDependencyStatus,
+  agentAvailabilityRecovery,
   agentComputerStatus,
   agentMessagingStatus,
+  agentRecoveryMessage,
   agentStatusPresentation,
   initials,
   messagingChannelLabel,
@@ -25,6 +27,7 @@ export function AgentDetailPage({ agentId }: { agentId: string }) {
         <section className="grid gap-6">
           <AgentObjectHeader agent={agent} />
           <div className="grid gap-6">
+            <AgentLifecycleNotice agent={agent} />
             {/*
              * Usage and Connection share a row: neither fills the width on its own, and a failed
              * dependency belongs beside the work it is stopping rather than in a banner above it.
@@ -164,8 +167,9 @@ function ConnectionRow({
 
 export function AgentAvailabilityAction({ agent }: { agent: AgentDetailView }) {
   /*
-   * Only while the Agent is healthy. Every other state is already named, with its recovery exit, by
-   * the banner directly beneath this header, and saying it twice reads as two problems.
+   * Only while the Agent is healthy. A failed dependency is already named, with its own exit, by
+   * the Connection card below, and saying it twice reads as two problems. A paused Agent is not a
+   * dependency failure and no card speaks for it, so it gets the notice beneath this header.
    */
   if (agent.availability.state !== "ready") return null;
   const status = agentStatusPresentation(agent);
@@ -173,5 +177,38 @@ export function AgentAvailabilityAction({ agent }: { agent: AgentDetailView }) {
     <div className="inline-flex">
       <StatusIndicator label={status.label} tone={status.tone} />
     </div>
+  );
+}
+
+/**
+ * The Agent's own lifecycle, which no dependency row can carry. Computer and Messaging state their
+ * own failures in the Connection card, but a paused Agent has nothing wrong with either -- it was
+ * turned off -- so without this the home reports a healthy Computer and channel and never mentions
+ * that the Agent is not running.
+ */
+export function AgentLifecycleNotice({ agent }: { agent: AgentDetailView }) {
+  if (agent.availability.state !== "suspended") return null;
+  const status = agentStatusPresentation(agent);
+  const recovery = agentAvailabilityRecovery(agent);
+  return (
+    <section
+      className="flex flex-wrap items-center justify-between gap-4 rounded-lg bg-kumo-tint p-4"
+      aria-label={`Agent status: ${status.label}`}
+      data-ui="agent-lifecycle-notice"
+    >
+      <div className="grid gap-1">
+        <StatusIndicator label={status.label} tone={status.tone} />
+        <p className="text-sm text-kumo-subtle">{agentRecoveryMessage(agent)}</p>
+      </div>
+      {recovery ? (
+        <Link
+          className={buttonClassName({ size: "compact", variant: "secondary" })}
+          state={{ agent }}
+          {...recovery.link}
+        >
+          {recovery.label}
+        </Link>
+      ) : null}
+    </section>
   );
 }
--- a/apps/web/src/features/agents/agent-detail-page.tsx
+++ b/apps/web/src/features/agents/agent-detail-page.tsx
@@ -1,17 +1,18 @@
 import { Link } from "@tanstack/react-router";
 import { buttonClassName, Icon, StatusIndicator, Text } from "../../ui/design-system.js";
-import { ProviderIcon } from "../../ui/provider-icon.js";
 import { AgentUsageOverview } from "../agent-usage.js";
 import { AsyncState } from "../resource/resource-state.js";
 import { useAccount } from "../session/session-context.js";
 import { AgentTasksSection } from "../tasks-page.js";
 import type { AgentDetailView } from "./agent-model.js";
 import {
-  agentAvailabilityRecovery,
-  agentRecoveryMessage,
+  type AgentDependencyStatus,
+  agentComputerStatus,
+  agentMessagingStatus,
   agentStatusPresentation,
   initials,
   messagingChannelLabel,
+  platformLabel,
 } from "./agent-presentation.js";
 import { useAgentDetailView } from "./agent-queries.js";
 import { agentDetailLink, agentSettingsLink, agentSettingsSectionLink } from "./agent-routes.js";
@@ -24,8 +25,14 @@ export function AgentDetailPage({ agentId }: { agentId: string }) {
         <section className="grid gap-6">
           <AgentObjectHeader agent={agent} />
           <div className="grid gap-6">
-            {agent.availability.state !== "ready" ? <AgentRecoveryBanner agent={agent} /> : null}
-            <AgentUsageOverview agent={agent} agentId={agent.id} />
+            {/*
+             * Usage and Connection share a row: neither fills the width on its own, and a failed
+             * dependency belongs beside the work it is stopping rather than in a banner above it.
+             */}
+            <div className="grid gap-6 @min-[48rem]/workspace:grid-cols-2">
+              <AgentUsageOverview agent={agent} agentId={agent.id} />
+              <AgentConnectionCard agent={agent} />
+            </div>
             <AgentTasksSection agentId={agent.id} />
           </div>
         </section>
@@ -74,7 +81,6 @@ export function AgentObjectHeader({ agent, backToSettings }: { agent: AgentDetai
           </div>
         </div>
         <div className="flex flex-wrap items-center justify-end gap-3">
-          {!backToSettings ? <AgentMessagingLink agent={agent} /> : null}
           {!backToSettings ? (
             <Link
               className={buttonClassName({ variant: "secondary" })}
@@ -90,28 +96,69 @@ export function AgentObjectHeader({ agent, backToSettings }: { agent: AgentDetai
   );
 }
 
-export function AgentRecoveryBanner({ agent }: { agent: AgentDetailView }) {
-  const recovery = agentAvailabilityRecovery(agent);
-  const status = agentStatusPresentation(agent);
+/**
+ * Computer and Messaging, side by side, each carrying its own status and its own repair exit. One
+ * card rather than two: they are the pair an Agent needs before it can do anything, and reading
+ * them together is how a viewer answers "can this Agent work" without holding two cards in mind.
+ */
+export function AgentConnectionCard({ agent }: { agent: AgentDetailView }) {
+  const computer = agentComputerStatus(agent);
+  const messaging = agentMessagingStatus(agent);
+  const binding = agent.messaging.kind === "ready" ? agent.messaging.value : undefined;
   return (
     <section
-      className="flex flex-wrap items-center justify-between gap-4 rounded-lg bg-kumo-danger-tint p-4"
-      aria-label={`Agent status: ${status.label}`}
+      className="grid content-start gap-4 rounded-lg bg-kumo-base p-4 ring ring-kumo-line"
+      aria-labelledby="agent-connection-heading"
+      data-ui="connection-overview"
     >
-      <div>
-        <strong>{status.label}</strong>
-        <p>{agentRecoveryMessage(agent)}</p>
+      <Text as="h2" id="agent-connection-heading" variant="heading">
+        Connection
+      </Text>
+      <ConnectionRow
+        agent={agent}
+        identity={`${agent.computer.displayName} · ${platformLabel(agent.computer.platform)}`}
+        name="Computer"
+        status={computer}
+      />
+      <ConnectionRow
+        agent={agent}
+        identity={binding ? messagingChannelLabel(agent, binding) : undefined}
+        name="Messaging"
+        status={messaging}
+      />
+    </section>
+  );
+}
+
+function ConnectionRow({
+  agent,
+  identity,
+  name,
+  status,
+}: {
+  agent: AgentDetailView;
+  identity?: string;
+  name: string;
+  status: AgentDependencyStatus;
+}) {
+  return (
+    <div className="grid gap-1 border-t border-kumo-line pt-3" data-ui={`connection-${name.toLowerCase()}`}>
+      <div className="flex flex-wrap items-center justify-between gap-2">
+        <strong className="text-sm">{name}</strong>
+        <StatusIndicator label={status.label} tone={status.tone} />
       </div>
-      {recovery ? (
+      {identity ? <p className="truncate text-sm text-kumo-subtle">{identity}</p> : null}
+      {status.detail ? <p className="text-sm text-kumo-subtle">{status.detail}</p> : null}
+      {status.action ? (
         <Link
-          className={buttonClassName({ size: "compact", variant: "secondary" })}
-          state={{ agent }}
-          {...recovery.link}
+          className="w-fit text-sm text-kumo-link"
+          state={{ agent, returnAgentId: agent.id, returnLabel: agent.displayName }}
+          {...agentSettingsSectionLink(agent.id, status.action.section)}
         >
-          {recovery.label}
+          {status.action.label}
         </Link>
       ) : null}
-    </section>
+    </div>
   );
 }
 
@@ -126,35 +173,5 @@ export function AgentAvailabilityAction({ agent }: { agent: AgentDetailView }) {
     <div className="inline-flex">
       <StatusIndicator label={status.label} tone={status.tone} />
     </div>
-  );
-}
-
-/**
- * One affordance for messaging, beside Settings. Messaging is what makes an Agent reachable, so this
- * states the channel rather than hinting at it with a bare provider glyph, and the gear says the
- * same control manages it. Every state — a live channel, unreadable evidence, nothing connected —
- * renders the same three-part shape at the same height, so the header does not resize as the state
- * changes and a missing binding keeps its entry point rather than disappearing.
- */
-function AgentMessagingLink({ agent }: { agent: AgentDetailView }) {
-  const binding = agent.messaging.kind === "ready" ? agent.messaging.value : undefined;
-  const label = binding
-    ? messagingChannelLabel(agent, binding)
-    : agent.messaging.kind === "unconfirmed"
-      ? "Messaging status unavailable"
-      : "Connect messaging";
-  return (
-    <Link
-      aria-label={label}
-      className={buttonClassName({ className: "max-w-64", variant: "secondary" })}
-      data-ui="agent-messaging-link"
-      state={{ agent, returnAgentId: agent.id, returnLabel: agent.displayName }}
-      title={label}
-      {...agentSettingsSectionLink(agent.id, "messaging")}
-    >
-      {binding ? <ProviderIcon className="size-4 shrink-0" provider={binding.provider} /> : <Icon name="message" />}
-      <span className="truncate">{label}</span>
-      <Icon className="text-kumo-subtle" data-ui="agent-messaging-manage" name="settings" />
-    </Link>
   );
 }
--- a/apps/web/src/features/agents/agent-presentation.ts
+++ b/apps/web/src/features/agents/agent-presentation.ts
@@ -254,7 +254,7 @@ export function messagingAgentStatusDescription(
 export function agentAvailabilityRecovery(
   agent: AgentDetailView,
 ): { label: string; link: AgentSettingsSectionLink } | undefined {
-  if (!true || agent.availability.state === "ready") return undefined;
+  if (agent.availability.state === "ready") return undefined;
   if (agent.availability.reason === "agent_suspended") {
     return { label: "Manage Agent", link: agentSettingsSectionLink(agent.id, "manage") };
   }
@@ -353,7 +353,19 @@ export function agentComputerStatus(agent: AgentDetailView): AgentDependencyStat
   if (computer.state === "unconfirmed") return { action, detail: recovery, label: "Unknown", tone: "neutral" };
   if (computer.state === "action_required") return { action, detail: recovery, label: "Offline", tone: "warning" };
   const providerName = runtimeProviderName(runtime.provider);
-  if (runtime.status && runtime.status !== "ready") {
+  /*
+   * A reachable Computer whose Provider readiness could not be read is `runtime_unconfirmed`, not
+   * ready. Falling through to Online here would paint missing evidence as a working Agent.
+   */
+  if (!runtime.status) {
+    return {
+      action,
+      detail: `OpenTag could not confirm ${providerName} on this Computer.`,
+      label: "Unknown",
+      tone: "neutral",
+    };
+  }
+  if (runtime.status !== "ready") {
     // Named per Provider status, and worded by the same helper the Computer settings page uses.
     const detail = recovery;
     if (runtime.status === "checking") return { detail, label: `Checking ${providerName}`, tone: "info" };
@@ -396,9 +408,33 @@ export function agentMessagingStatus(agent: AgentDetailView): AgentDependencySta
     };
   }
   /*
+   * An active binding whose delivery evidence could not be read is `handoff_unconfirmed`. The
+   * channel may well be fine, but this row states what is known, and "Connected" is not known.
+   */
+  if (binding.bindingState === "active" && handoff.state === "unconfirmed") {
+    return {
+      action,
+      detail: "OpenTag could not confirm whether messages reach this Agent.",
+      label: "Unknown",
+      tone: "neutral",
+    };
+  }
+  const detail: Record<ImBindingSummary["bindingState"], string | undefined> = {
+    active: undefined,
+    provisioning: "The messaging connection is still being set up.",
+    reauthorization_required: "The messaging connection needs to be re-authorized before it can receive messages.",
+    error: "The messaging connection failed. Reconnect it to receive messages.",
+    disabled: "Messaging is turned off for this Agent. Reconnect it to receive messages.",
+  };
+  /*
    * The exit stays even when the channel is healthy. This card replaced the header's messaging
    * control, so dropping it on success would leave changing the bound bot reachable only by
    * hunting through Settings.
    */
-  return { action, label: messagingConnectionLabel(binding), tone: messagingConnectionTone(binding) };
+  return {
+    action,
+    detail: detail[binding.bindingState],
+    label: messagingConnectionLabel(binding),
+    tone: messagingConnectionTone(binding),
+  };
 }
--- a/apps/web/src/features/agents/agent-presentation.ts
+++ b/apps/web/src/features/agents/agent-presentation.ts
@@ -344,19 +344,35 @@ export type AgentDependencyStatus = {
 
 export function agentComputerStatus(agent: AgentDetailView): AgentDependencyStatus {
   const { computer, runtime } = agent.availability.dependencies;
-  const action = { label: "View Computer", section: "computer" as const };
-  /*
-   * "this Computer" rather than the machine name: the identity line directly above this sentence
-   * already carries the name, and repeating it made one row read as though it named two machines.
-   */
-  const recovery = computerRecoveryMessage(agent, "this Computer");
-  if (computer.state === "unconfirmed") return { action, detail: recovery, label: "Unknown", tone: "neutral" };
-  if (computer.state === "action_required") return { action, detail: recovery, label: "Offline", tone: "warning" };
   const providerName = runtimeProviderName(runtime.provider);
   /*
-   * A reachable Computer whose Provider readiness could not be read is `runtime_unconfirmed`, not
-   * ready. Falling through to Online here would paint missing evidence as a working Agent.
+   * Every branch carries the exit, including the healthy one. This card is the only Computer
+   * recovery surface on the Agent home now that the banner is gone, so a state without an exit
+   * leaves the Computer reachable only by hunting through Settings.
+   *
+   * The wording is derived from these two dependency fields rather than from
+   * `computerRecoveryMessage`, which answers a different question: it branches on the Agent-wide
+   * `availability.reason`, and a higher-ranked reason such as `agent_suspended` masks the runtime
+   * one -- which made this row label a Provider as "Checking" while its sentence claimed the
+   * connection was unconfirmed, about a Computer that was online.
    */
+  const action = { label: "View Computer", section: "computer" as const };
+  if (computer.state === "unconfirmed") {
+    return {
+      action,
+      detail: "OpenTag could not confirm this Computer's current connection.",
+      label: "Unknown",
+      tone: "neutral",
+    };
+  }
+  if (computer.state === "action_required") {
+    return {
+      action,
+      detail: "OpenTag is not running on this Computer. Start it there to bring it back online.",
+      label: "Offline",
+      tone: "warning",
+    };
+  }
   if (!runtime.status) {
     return {
       action,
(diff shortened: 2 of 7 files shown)


apps/web/src/features/agents/agent-detail-page.tsx
```diff
@@ -24,8 +25,14 @@ export function AgentDetailPage({ agentId }: { agentId: string }) {
...
-            {agent.availability.state !== "ready" ? <AgentRecoveryBanner agent={agent} /> : null}
-            <AgentUsageOverview agent={agent} agentId={agent.id} />
+            {/*
+             * Usage and Connection share a row: neither fills the width on its own, and a failed
+             * dependency belongs beside the work it is stopping rather than in a banner above it.
+             */}
+            <div className="grid gap-6 @min-[48rem]/workspace:grid-cols-2">
+              <AgentUsageOverview agent={agent} agentId={agent.id} />
+              <AgentConnectionCard agent={agent} />
+            </div>
```
bestony, line 38: This is the banner’s replacement, but it only carries Computer and Messaging. A paused Agent (`availability.state === "suspended"`) still has `AgentAvailabilityAction` return null, so the home no longer names the pause or offers `Manage Agent`. `projectAgentAvailability` ranks `agent_suspended` first and still fills healthy Computer/Messaging dependencies, so this card currently reads as Online / Connected.

I reproduced this at this head: `/agents/:id` with `initialStatus: "suspended"` has no `Suspended` text and no resume exit.

Keep recovery on the Agent. A third row, a header badge for lifecycle, or a slim Agent-level notice all work; the Connection card cannot be the only remaining recovery surface unless it also speaks for the Agent itself.

apps/web/src/features/agents/agent-presentation.ts
```diff
@@ -329,3 +329,76 @@ export function formatRelativeTime(value: string): string {
...
+    const detail = recovery;
+    if (runtime.status === "checking") return { detail, label: `Checking ${providerName}`, tone: "info" };
+    if (runtime.status === "install")
+      return { action, detail, label: `${providerName} not installed`, tone: "warning" };
+    if (runtime.status === "sign-in") {
+      return { action, detail, label: `${providerName} sign-in required`, tone: "warning" };
+    }
+    return { action, detail, label: `${providerName} unavailable`, tone: "warning" };
+  }
+  return { action, label: "Online", tone: "success" };
```
bestony, line 416: `runtime.status` being `null` is `runtime_unconfirmed`, not ready. `computer.state` is still `ready`, this `if` is skipped, and the row returns **Online** / `success`.

The previous banner used `agentStatusPresentation`, which names this `Computer unknown`. Missing Provider evidence has to stay unconfirmed.

apps/web/src/features/agents/agent-presentation.ts
```diff
@@ -329,3 +329,76 @@ export function formatRelativeTime(value: string): string {
...
+      label: "Cannot receive messages",
+      tone: "warning",
+    };
+  }
+  /*
+   * The exit stays even when the channel is healthy. This card replaced the header's messaging
+   * control, so dropping it on success would leave changing the bound bot reachable only by
+   * hunting through Settings.
+   */
+  return { action, label: messagingConnectionLabel(binding), tone: messagingConnectionTone(binding) };
```
bestony: This success path also runs when `handoff.state === "unconfirmed"` (`handoff_unconfirmed`): the binding can still be `active`, and the only special case above is `action_required`.

The banner used to say `Status unknown`. An unreadable handoff is not a connected channel.

bestony (changes requested):
The Connection card is the right replacement for Computer and Messaging: each dependency names itself, keeps an exit, and the list card still has none. Deleting the banner still dropped two recoveries the banner was the only home for, and the new helpers paint two unconfirmed evidence states as success.

## Requested changes

**1. A paused Agent has no name or resume exit on the home.**

`AgentAvailabilityAction` still hides every non-ready state, and `AgentConnectionCard` only renders Computer + Messaging. `projectAgentAvailability` still ranks `agent_suspended` first, so a paused Agent with a live Computer and channel now reads as Online / Connected. I confirmed this against this head: a test that opens `/agents/:id` with `initialStatus: "suspended"` cannot find `Suspended`, the paused copy, or `Manage Agent`. Restore that lifecycle recovery on the Agent itself — header badge, a third Connection row, or a slim Agent-level notice — and add the home-page case the banner tests used to cover.

**2. Unconfirmed evidence is shown as success.**

`agentComputerStatus` falls through to `Online` when the Computer is reachable but `runtime.status` is `null` (`runtime_unconfirmed`). `agentMessagingStatus` falls through to `Connected` when the binding is active but `handoff.state === "unconfirmed"` (`handoff_unconfirmed`). The old banner used `agentStatusPresentation` (`Computer unknown` / `Status unknown`). Missing evidence has to stay unconfirmed, not green.

## Non-blocking

- The `AgentAvailabilityAction` comment still talks about “the banner directly beneath this header”.
- `agentAvailabilityRecovery` is now unreferenced; reuse it for the paused-Agent exit or delete it. The `if (!true || …)` guard can go with it.
- Reauthorization / error / disabled / provisioning messaging states have a label and exit but no `detail` sentence. Worth a why-line if you are already in that helper.
- `#313` is correctly deferred. Whichever lands second still needs a not-bound branch on the Computer row; `agent.computer.displayName` / `platform` is unsound after that PR.
- CI `Checks` failed on `packages/server` `im-binding.test.ts` (“Initial delivery frame was not dispatched”). This diff does not touch the server. Retry is the signal; I am not treating that failure as a finding in this review.

`apps/web` `app.test.tsx` (131) and `agent-model.test.ts` (23) pass locally at this head. Those suites do not cover the two gaps above.

bestony:
#303 merged to `main` (squash `ce2e721`) and this PR’s base followed it. Head is still `3c279cc`; the two requested changes on that head still stand.

GitHub now reports **merge conflicts** (`DIRTY`). This branch still carries the pre-squash #303 commits (`3eee4b2`, `3ff265e`) while `main` has the squash plus nine other commits, including #310. Rebase onto current `main` before the next review pass — the stacked slice is no longer a valid merge.

#313 is still open, so the nullable-computer handoff stays deferred until whichever of these lands second.

liuchao-001:
Both requested changes are fixed and the branch is rebased onto `main`. New head `e0f99b9`.

## 1. Paused Agent — fixed

You were right, and the failure mode was worse than "no exit": `projectAgentAvailability` ranks the pause first *and* still fills healthy Computer and Messaging dependencies, so the card actively asserted **Online / Connected** about an Agent that was switched off. That is a wrong statement, not just a missing one.

A pause is not a dependency failure and no dependency row can honestly speak for it, so it gets an Agent-level notice above the Usage/Connection row — `data-ui="agent-lifecycle-notice"` — carrying the `Suspended` badge, `This Agent is paused. Resume it to start receiving messages again.`, and `Manage Agent`. The header keeps showing only the ready state; dependency failures still speak for themselves in the card, so nothing is said twice.

That also answers your non-blocking note: `agentAvailabilityRecovery` is what supplies the exit, so it is referenced again rather than deleted, and its `if (!true || …)` guard is gone.

## 2. Unconfirmed evidence — fixed

Both fall-throughs closed:

- `agentComputerStatus` — a reachable Computer with `runtime.status === null` is `runtime_unconfirmed`. Now `Unknown` / `neutral`, with `OpenTag could not confirm ${provider} on this Computer.`
- `agentMessagingStatus` — an `active` binding with `handoff.state === "unconfirmed"` is `handoff_unconfirmed`. Now `Unknown` / `neutral`, with `OpenTag could not confirm whether messages reach this Agent.`

Reaching the second one needed a harness gap closed: `bindingEvidenceFails` fails the binding *and* the handoff read, which lands in the already-covered `messaging.kind === "unconfirmed"` branch. There was no way to fail only the handoff, which is exactly why the state was never covered. Added `handoffEvidenceFails` for it.

## Also done

- Failed messaging binding states (provisioning / reauthorization / error / disabled) now carry the missing `detail` line.
- The stale `AgentAvailabilityAction` comment no longer refers to a banner that does not exist.

## Tests

Added two cases, both covering states no suite reached before: the paused Agent on `/agents/:id`, and the two unconfirmed-evidence rows.

## Rebase

Rebased onto `main` with `--onto`, dropping `3eee4b2` / `3ff265e` since they are now the `ce2e721` squash. Clean, no conflicts; the diff is only the round-2 work. I used `--force-with-lease` — flagging it because `CONTRIBUTING.md` says not to force-push shared branches, and my read is that this is a single-author topic branch and you asked for the rebase, but say so if you would rather I had merged `main` in instead.

`pnpm check` and `pnpm typecheck` are exit 0 on this head. The full suite is still not a signal I can give you from this machine — load is ~50 with 23 concurrent vitest workers from another session — so CI remains the verdict. The two new tests pass here at a raised timeout.

## Still deferred

The nullable-`agent.computer` handoff with #313, as you said. Whichever lands second takes the not-bound branch onto the Computer row.

bestony (approved):
The two requested changes are addressed at `e0f99b9`.

- A paused Agent now has an Agent-level notice (Suspended, the resume sentence, and Manage Agent). The new home-page test passes on this head.
- `runtime_unconfirmed` and `handoff_unconfirmed` land on Unknown, not Online / Connected. The new test covering both Connection rows passes.

Also taken: why-lines on the failed messaging binding states, and the dead `if (!true || …)` guard is gone. The branch is rebased onto current `main` and mergeable.

`#313` stays deferred: whichever lands second still needs the not-bound Computer row.

Local: those two new cases plus five adjacent Agent-home tests pass. CI Checks were still running at review time.

apps/web/src/features/agents/agent-detail-page.tsx
bestony: Fixed in e0f99b9: `AgentLifecycleNotice` names the pause and carries Manage Agent. The new home-page test covers this.

apps/web/src/features/agents/agent-presentation.ts
bestony: Fixed in e0f99b9: a missing `runtime.status` now returns Unknown instead of Online.

apps/web/src/features/agents/agent-presentation.ts
bestony: Fixed in e0f99b9: an active binding with `handoff.state === "unconfirmed"` now returns Unknown instead of Connected.

bestony:
`--force-with-lease` was the right call. This is a single-author topic branch, the rebase was requested after the #303 squash, and merging `main` in would have kept the pre-squash commits beside `ce2e721`. `CONTRIBUTING.md` is about shared long-lived branches, not that case.

Approval already stands on `e0f99b9` — the re-request does not need another pass of the same head.

CI on this head:

- **Checks** failed again in `im-binding.test.ts` (`Old generation dispatch was not persisted`). Same class of server integration flake as the previous run; this diff does not touch the server. Retry is the signal.
- **Enforce Patch Coverage** failed with `ENOENT` on `coverage/unit/coverage-final.json` after “Measure unit coverage” succeeded. That is a missing artifact, not uncovered lines in this patch. Treat it as a workflow issue (likely from the new patch-coverage gate), not a finding against this PR.

liuchao-001:
Agreed on both, and I chased the coverage one to its root since it affects everyone, not just this PR.

**`Enforce Patch Coverage` can never pass in its current form.** It is not a missing artifact by accident — the file is never written at all:

- `.github/workflows/coverage.yml:96` reads `coverage/unit/coverage-final.json`.
- `coverage-final.json` is produced by the `json` reporter, but `scripts/unit-coverage.mjs:70-73` invokes each project with only `--coverage.reporter=json-summary` and `--coverage.reporter=text-summary`. These CLI flags override the `json` reporter that `vitest.coverage.config.ts` declares, so no `coverage-final.json` is emitted anywhere.
- Separately, per-project output lands in `coverage/unit/<project>/`, and the only top-level file the script writes is `coverage-summary.json` (`scripts/unit-coverage.mjs:162-163`).

So the step fails on every PR regardless of how well covered the patch is, and it fails *after* `Measure unit coverage` reports success. Filed as **#323** with both candidate fixes; the real one is probably adding the `json` reporter and merging the per-project files, since `coverage-summary.json` has per-file totals rather than the per-line hit map changed-line enforcement needs.

Worth deciding whether that check should be required while it is in this state — a permanently red gate teaches people to ignore it.

**`Checks`** — re-run triggered. Third consecutive failure in `packages/server/.../im-binding.test.ts`, a different test each time (`recovers a retained Client Turn` → `requires old-placement reconciliation` → `Old generation dispatch was not persisted`). Same file, rotating cases, and this diff touches zero files under `packages/`. I have not opened an issue for that one yet since it is not mine and may already be tracked — say the word if you would rather it were filed.

Thanks for the `--force-with-lease` read; that matches my reasoning and I would rather have it on the record than assumed.

bestony:
No new issue for the `im-binding` flake — [#297](https://github.com/first-tree-ai/opentag/issues/297) already tracks it (staging deploy skipped when that file trips; rotating cases, re-run-without-code-change). [#220](https://github.com/first-tree-ai/opentag/issues/220) is the related timeout work. Please hang the third consecutive miss on #297 rather than opening a fourth ticket.

Coverage: [#320](https://github.com/first-tree-ai/opentag/issues/320) already has the same diagnosis and the eight-run “never passed” evidence. I am closing [#323](https://github.com/first-tree-ai/opentag/issues/323) as a duplicate of that. Your suggested fix (emit/`json` reporter, merge per-project `coverage-final.json`, fail closed with a readable missing-artifact error) belongs there.

On whether the check should be required while it is in this state: it is **not** in `protect main` today. That ruleset only requires the `CI` aggregator. So the permanently red coverage job is noise, not the merge block on this PR — #317 is waiting on `CI` (the `im-binding` flake) and whatever CODEOWNERS review is still outstanding, not on patch coverage.

Do not add it as a required check until #320 lands. A gate that cannot pass should not be required, and a red check that is not required still trains people to ignore red. Fix the producer, then require it.

liuchao-001:
## Independent QA — PR #317, head `e0f99b9f2c9cb94036c6e37b26bd0a0fcdc60c46`

**Status: FAIL** · **Tier: `focused-local`** · run by code-reviewer, not the author.

One reproducible regression, plus a second row that states something untrue. Both of bestony's requested changes are genuinely fixed — I mutation-verified them rather than reading the tests.

**Scope.** Built `@opentag/shared`, `@opentag/server` and `@opentag/web` at this head, served the production web build from a local server against a warm Postgres, seeded four Agents across availability states, and drove the running product with `agent-browser` at 1280 / 760 / 380px.

---

### 🐛 FAIL — the Computer row loses its exit in the `Checking <Provider>` state

`agentComputerStatus` omits `action` on exactly one branch, where all seven siblings include it:

```ts
if (runtime.status === "checking") return { detail, label: `Checking ${providerName}`, tone: "info" };
```

Live, on a Computer that is **online**:

```
connection-computer → "Computer | Checking Codex | superMacBook-M4-Pro.local · macOS |
                       OpenTag is still checking Codex on this Computer."
links: []
```

I enumerated every anchor on the page: `anyComputerRoute: []`. **There is no link to the Computer anywhere on the Agent detail page in this state** — only the generic `Settings`, leaving the viewer to find the Computer tab themselves. The Messaging row beside it keeps `View messaging`, so the asymmetry is visible on screen. Reproduced on two different Agents.

**This is a regression, not just an inconsistency.** Before this PR, `runtime_unavailable` rendered the banner, and `agentAvailabilityRecovery` fell through to its final `return { label: "View Computer", … }`. This PR deletes the banner and makes the Connection card the only recovery surface — then drops the exit in this state. It also contradicts the rule the messaging helper states three lines earlier: *"The exit stays even when the channel is healthy… dropping it on success would leave changing the bound bot reachable only by hunting through Settings."*

**Reachability:** `projectComputerProviderReadiness` returns `{ status: "checking" }` for any admitted provider with no snapshot, so this is the state of every reachable Computer whose client has not reported readiness for that provider. Transient on the happy path; permanent if a connected client never reports it.

**Why it slipped:** there is a test asserting the messaging row *"keeps a repair exit in every messaging state"* — and no Computer-row equivalent. The mirror test across `unconfirmed / offline / checking / install / sign-in / unavailable / online` would have caught this and would prevent it returning.

### 🐛 The Computer row's label and detail contradict each other on a paused Agent

The label comes from the per-dependency `runtime.status`; the detail comes from `computerRecoveryMessage(agent, "this Computer")`, which branches on the **whole-Agent** `availability.reason`. When a higher-ranked reason masks the runtime one — `agent_suspended` is the only one — they disagree. Live, suspended Agent with the Computer online:

```
Computer | Checking Codex | superMacBook-M4-Pro.local · macOS
         | OpenTag could not confirm this Computer's current connection.
```

The label says the provider is being checked; the sentence says the connection is unconfirmed. The Computer is online and confirmed, so the sentence is false. Same class as the one you asked bestony's #2 about: a row asserting something untrue about a dependency. The fix is to derive the detail from the same dependency state as the label, not from the Agent-wide reason.

---

### ✅ bestony's two requested changes — verified by mutation, not by reading

| mutation | result |
|---|---|
| lifecycle notice never renders | ✅ red — `Unable to find … "Agent status: Suspended"` |
| restore the `!runtime.status` fall-through to `Online` | ✅ red — `Unable to find … "Unknown"` |
| restore the `handoff.state === "unconfirmed"` fall-through to `Connected` | ✅ red — `expected <span> to be null` |

Live, the paused Agent renders `Suspended`, *"This Agent is paused. Resume it to start receiving messages again."*, and `Manage Agent` → `/settings/manage`.

### Two things I could not reach live, stated so they are not read as covered

- **`handoff_unconfirmed`** needs an `active` binding *and* a failing handoff read. `bindingState` is derived — my seeded Slack installation has an invalid credential, so it resolves to `reauthorization_required` regardless of the row's `status` — and `agent-browser network route --abort` did not take on these fetches. Covered by mutation above instead.
- **`!runtime.status`** looks unreachable through the real API: the server returns every admitted provider, and `SERVER_ADMITTED_AGENT_RUNTIME_PROVIDERS` is `codex` + `claude-code`, which is the entire `agent_runtime_provider` enum — so `.find()` never returns undefined. The test reaches it with `computerProviderReadiness: []`, a shape the server does not emit. The branch is correct and worth keeping as a guard; it is just defensive rather than a live state.

### ✅ Layout, button, console

- **1280px** — Usage and Connection are two equal `466px` columns at the same `y`, both `266px` tall. **760 / 380px** — stacked, full width, no overflow beyond the intended `.truncate` identity line.
- **New Agent** — `secondary` (white background, dark text), 36px, bold `Plus` **leading** the label (`plusFirst: true`).
- **Zero console errors, zero page errors** across the session.

### Gates

- `pnpm --filter @opentag/web test` — 37 files, **454 tests, all pass**, exit 0, first run, no flake, despite load average ~22.

  The drop from 537 at `3ff265e` is **upstream, not a collection failure here**: `#305` retired the Onboarding Lab and deleted four test files (`onboarding/*`, `internal/onboarding-lab.test.tsx`), while three new files arrived from other merges. I checked for collection errors and found none.

- **`Enforce Patch Coverage` fails — and it is not this PR.** The failure is `ENOENT: coverage/unit/coverage-final.json`; the `Measure unit coverage` step reports success but leaves no artifact where the enforce step reads it. **17 of the last 18 runs of that workflow failed identically across unrelated branches** (`fix/issue-315-messaging-cli-check`, `test/0af99c-coverage-95`, `fix/integration-tab-kumo-table`, …); only `test/coverage-gates` ever passed. The gate is broken repo-wide and will block every PR until fixed — worth its own issue.
- `Checks` was still pending when I finished; everything else on the run is green.

`reviewDecision` reads `APPROVED` as of this comment.

### Residual state

Warm environment kept: `otqa-pg` and its data, the worktree at `e0f99b9`, the local server on 51939. A 25-second keep-alive is refreshing `last_seen_at` so the Computer stays `online` past the 90s presence timeout — **that process is mine and I will stop it**; it is not part of the product.

### Case disposition

`candidate-new-case` — the Connection card across dependency states is worth a reusable prose case, and the per-row "every state keeps its exit" check is the specific assertion that would have caught the first finding.

apps/web/src/features/agents/agent-presentation.ts
```diff
@@ -329,3 +329,112 @@ export function formatRelativeTime(value: string): string {
...
+      action,
+      detail: `OpenTag could not confirm ${providerName} on this Computer.`,
+      label: "Unknown",
+      tone: "neutral",
+    };
+  }
+  if (runtime.status !== "ready") {
+    // Named per Provider status, and worded by the same helper the Computer settings page uses.
+    const detail = recovery;
+    if (runtime.status === "checking") return { detail, label: `Checking ${providerName}`, tone: "info" };
```
bestony: This is the only Computer-row branch that drops `action`. Independent QA reproduced it live: an online Computer in `checking` has **no Computer link anywhere on the Agent home**, only generic Settings. The banner this card replaced used to fall through to `View Computer`.

Keep `action` here (and on Online). A table of Computer states that all keep the exit would have caught this — messaging already has that test.

apps/web/src/features/agents/agent-presentation.ts
```diff
@@ -329,3 +329,112 @@ export function formatRelativeTime(value: string): string {
...
+};
+
+export function agentComputerStatus(agent: AgentDetailView): AgentDependencyStatus {
+  const { computer, runtime } = agent.availability.dependencies;
+  const action = { label: "View Computer", section: "computer" as const };
+  /*
+   * "this Computer" rather than the machine name: the identity line directly above this sentence
+   * already carries the name, and repeating it made one row read as though it named two machines.
+   */
+  const recovery = computerRecoveryMessage(agent, "this Computer");
```
bestony: `computerRecoveryMessage` keys off Agent-wide `availability.reason`. When the Agent is paused, that reason is `agent_suspended`, so the helper takes the `computer.state !== "action_required"` branch and says the connection could not be confirmed — while this row's label, from `runtime.status`, says `Checking Codex`.

QA saw that contradiction live on a paused Agent whose Computer is online. The detail has to come from the same Computer/runtime dependency this row is naming, not from the Agent-level reason the lifecycle notice already covers.

bestony (changes requested):
Independent QA at `e0f99b9` found two Computer-row defects I underweighted on the first pass. I verified both in the source. The earlier paused-Agent and unconfirmed-evidence fixes still stand; these do not.

## Requested changes

**1. `Checking <Provider>` drops the only Computer exit on the page.**

`agentComputerStatus` omits `action` on the `checking` branch; every sibling keeps `View Computer`. The deleted banner used to supply that exit via `agentAvailabilityRecovery`'s fall-through. Live, an online Computer in `checking` has no Computer link anywhere on the Agent home — only generic Settings. Messaging beside it still has `View messaging`.

Keep the exit in every Computer state, including `checking` and Online. Add the Computer-row mirror of “every messaging state keeps its exit”.

**2. On a paused Agent the Computer row asserts two different facts.**

The label is derived from `runtime.status`; the detail is `computerRecoveryMessage`, which branches on Agent-wide `availability.reason`. When `agent_suspended` outranks `runtime_unavailable`, live copy is `Checking Codex` plus `OpenTag could not confirm this Computer's current connection.` The Computer is online. Derive the detail from the same dependency state as the label.

Coverage ENOENT remains #320, not this PR.

liuchao-001:
Both fixed at `59f2d376ee98b7200dcd9c33421ec52ba832225b`. Both were real regressions and both were mine.

**1. `Checking <Provider>` exit.** `action` is now on all eight branches, including `Online`. Your framing as a regression rather than an inconsistency is the correct one — that branch is what `projectComputerProviderReadiness` returns for any admitted provider without a snapshot, so it was the default state of a reachable Computer whose client had not reported yet, not an edge case.

Added the Computer-row mirror you asked for: all seven states (`unconfirmed / offline / checking / install / sign-in / unavailable / online`), each asserting `View Computer` resolves to `/settings/computer`.

**2. The self-contradicting row.** The row no longer calls `computerRecoveryMessage`; its sentence is derived from the same two fields as its label (`dependencies.computer.state`, `dependencies.runtime.status`), so a masked Agent-wide reason cannot make the two disagree.

I did not refactor `computerRecoveryMessage` to take dependency state. It still answers the Agent-level question correctly on the Computer settings page, where the Agent-wide reason *is* the right basis — the two questions genuinely differ, which is what the bug was. The mirror test asserts each sentence beside its label so they cannot drift apart again, and there is a specific paused-Agent case: suspended + `checking` now reads `Checking Codex` / `OpenTag is still checking Codex on this Computer.`, with the old false sentence asserted absent.

Both fixes are on top of the two you already accepted; those are untouched.

Local: `app.test.tsx` is 134/135, the one failure being the unrelated 40ms-window flake tracked in #306, which passes in isolation.

Coverage stays on #320 as you said — and for the record, my #323 was a duplicate of it that I filed without scanning first. Worth knowing why the scan would not have helped anyway: `gh issue list --repo … --search "<term>"` returns empty for *every* term on this repo, including guaranteed matches, and `gh search issues` errors outright. Listing and filtering client-side works. I put that on #320 so the next person does not lose the same time.

bestony (approved):
The two Computer-row findings from independent QA are addressed at `59f2d37`.

- Every `agentComputerStatus` branch, including `checking` and Online, carries `View Computer`. The new table test covers all seven states.
- The row’s sentence now comes from the same Computer/runtime fields as the label. A paused Agent with `Checking Codex` no longer claims the connection is unconfirmed. `computerRecoveryMessage` stays on the Computer settings page, which still asks the Agent-wide question.

The earlier paused-Agent notice and unconfirmed→Unknown fixes are untouched. Local: the two new tests plus the two prior recovery cases pass on this head.

`#313` remains deferred. Coverage ENOENT remains #320.

apps/web/src/features/agents/agent-presentation.ts
bestony: Fixed in 59f2d37: the `checking` branch now keeps `View Computer`. The new seven-state table test covers this.

apps/web/src/features/agents/agent-presentation.ts
bestony: Fixed in 59f2d37: the Computer row derives its sentence from the same Computer/runtime fields as the label. `computerRecoveryMessage` stays on the settings page.

yuezengwu (approved):
Reviewed exact head `59f2d376ee98b7200dcd9c33421ec52ba832225b`.

The two fresh Computer-row regressions are fixed: every current Computer/Provider state keeps `View Computer`, and the row now derives its label and detail from the same dependency-local Computer/runtime facts. A suspended Agent therefore keeps the Agent-level lifecycle notice without making the Computer row contradict itself. The earlier paused-Agent and unconfirmed-evidence fixes remain intact.

I independently mutation-verified both guards: removing the `checking` exit makes the all-states case fail, and restoring the Agent-wide recovery helper makes the suspended/checking contradiction case fail.

Validation at this head:
- `pnpm check` (exit 0; one existing optional-chain warning in untouched onboarding code)
- `pnpm build`
- `pnpm typecheck`
- `pnpm test` (all 7 workspace test tasks pass; Web 37 files / 456 tests)
- `git diff --check`

Scope is seven `apps/web` UI/interaction files only; it stays within the non-engineer author boundary and introduces no API, architecture, persistence, or data-structure change. #313 remains an explicit ordering dependency: whichever PR lands second must add the nullable/unbound Computer presentation there.

The red patch-coverage run is the already-diagnosed producer artifact failure fixed by merged #324, not a coverage finding in this patch; the required CI aggregator is green.

liuchao-001:
## Independent QA — PR #317, head `59f2d376ee98b7200dcd9c33421ec52ba832225b`

**Status: PASS** · **Tier: `focused-local`** · run by code-reviewer, not the author. Supersedes the FAIL bound to `e0f99b9`.

Both findings are fixed, verified live and by mutation. I verified the diff since the last QA'd head myself: `agent-presentation.ts` and `app.test.tsx`, nothing else.

---

### ✅ 1. `Checking <Provider>` keeps its exit

Live, on an online Computer:

```
Computer | Checking Codex | superMacBook-M4-Pro.local · macOS
         | OpenTag is still checking Codex on this Computer. | View Computer → /settings/computer
```

`computerRouteOnPage` is now `["View Computer"]`, where at `e0f99b9` it was `[]`. Confirmed on a third Agent with a different provider, which correctly reads `Checking Claude Code`.

### ✅ 2. The paused Agent's Computer row no longer contradicts itself

Live, suspended Agent with an online Computer:

```
Computer | Checking Codex | superMacBook-M4-Pro.local · macOS
         | OpenTag is still checking Codex on this Computer. | View Computer
```

Label and sentence now agree, and the false `could not confirm this Computer's current connection` is gone. The lifecycle notice still carries `Suspended` and `Manage Agent`.

Deriving the sentence from the same two dependency fields as the label — rather than refactoring `computerRecoveryMessage` — is the right call. I checked the claim that the old helper still answers the settings page's question correctly rather than taking it: `agent-computer-settings.tsx` gates that block on `ready = computerState.state === "ready" && !runtimeUnavailable`, so for a paused Agent with an online Computer the block does not render and the helper is never reached. Live, that page shows `Online` with no recovery sentence.

### New guards — mutation-verified

| mutation | result |
|---|---|
| drop `action` from the `checking` branch | ✅ red — `Unable to find … "View Computer"` |
| drop `action` from the `Online` branch | ✅ red — same |
| restore the Agent-wide sentence in `checking` | ✅ red — paused-Agent test fails on the expected sentence |

**A caveat about my own method, because it nearly produced a false finding.** My first attempt at the `checking` mutation was a `perl` regex that silently matched nothing; the test passed and briefly looked as though the new guard did not cover the very regression it was written for. I caught it by diffing the file to confirm the mutation had actually applied. A mutation that does not mutate is indistinguishable from a test that does not guard — I should verify the edit landed before reading the result, and I have added that to how I run these.

---

### Non-blocking

**`computerRecoveryMessage`'s `computerName` parameter is now dead.** It was added one commit ago for `agentComputerStatus`'s `"this Computer"` call, and that call site is gone — no caller passes a second argument now. Worth deleting, and it matters slightly for the #313 rebase: the default `computerName = agent.computer.displayName` is one more unguarded dereference once `agent.computer` becomes nullable.

**Pre-existing, not this PR, not verified live:** the Computer settings page derives `runtimeUnavailable` from the Agent-wide `availability.reason`, so a *suspended* Agent whose Provider genuinely is missing would still be labelled `Online` there — the same masking that #317 just fixed in the Connection card, still present in the sibling. I could not demonstrate it: it needs a real client reporting `install` / `sign-in`, which my run cell cannot produce. Flagging it as a code-level observation for whoever owns that page, not as a finding against this PR.

### Evidence boundary

- Live-verified Computer states: `Unknown` (unconfirmed), `Offline`, `Checking` — for both `codex` and `claude-code`. **`install` / `sign-in` / `unavailable` / `ready` were not reachable**: provider readiness comes from `RuntimeConnectionRegistry`, an in-process Map fed only by a live client's WebSocket registration, so it cannot be seeded. The new mirror test covers all seven; four of them rest on that test rather than on my observation.
- **`handoff_unconfirmed` still not reached live.** I retried: granting all seven `SLACK_REQUIRED_BOT_SCOPES` did not clear it, because `credential.status === "invalid"` on the seeded Slack installation independently forces `reauthorization_required`. Reaching it needs a forged installation credential *and* a failing handoff read. Mutation-covered only, as recorded last round.

### ✅ Layout, console, gates

- **1280px** — Usage and Connection are two equal `466px` columns at the same `y`, both `285px`. **760 / 380px** — stacked, no overflow outside the intended `.truncate`.
- **Zero console errors, zero page errors.**
- `pnpm --filter @opentag/web test` — 37 files, **456/456 pass**, exit 0, first run, at load average ~19–33. That is 454 plus the two new cases.
- **CI is green at this head** (run `33350597839`): `Checks`, `CI`, both Node compat jobs, Container Smoke, CLI Pack Smoke, Secret Scan, Workflow Validation. The `im-binding` flake cleared.
- `Enforce Patch Coverage` is still red with the `ENOENT` — that is #320, repo-wide, and not in the `protect main` ruleset.

`reviewDecision: APPROVED`, `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` (from the non-required coverage job only).

### Residual state

Warm environment kept: `otqa-pg` and its data, the worktree at `59f2d37`, the local server on 51939 (PID recorded). A 25-second keep-alive is holding `last_seen_at` fresh past the 90s presence timeout so the Computer reads `online` — **mine, not the product; I will stop it.** Seeded shapes left in place at the author's request: `Q` suspended, and the Slack binding now carrying all seven required scopes.

### Case disposition

`candidate-new-case` — unchanged. The per-row "every state keeps its exit" table is the assertion worth carrying into a reusable case; it is what was missing when this shipped.

bestony:
QA PASS at `59f2d37` accepted — it supersedes the FAIL on `e0f99b9`. The two Computer-row regressions are closed; approval already stands.

The dead `computerName` default is real: both remaining callers pass only `agent`. Not a merge blocker here. It is one more unguarded `agent.computer.displayName` for whichever of #313 and this lands second — hang it on that follow-up, do not reopen this review for it.

The settings-page masking note is also out of scope for #317.

Issue #320, opened 2026-08-31
Enforce Patch Coverage fails on every PR: the gate reads a coverage file nothing writes


liuchao-001:
## What happens

The `Enforce Patch Coverage` check fails on every pull request with:

```
Error: ENOENT: no such file or directory, open '/home/runner/work/opentag/opentag/coverage/unit/coverage-final.json'
```

It is not a coverage shortfall — the enforcement script crashes before it can measure anything. Observed on #302 and #314, and on at least one unrelated branch (`fix/issue-315-messaging-cli-check`). It reproduces on branches that have current `main` merged in, so it is not staleness.

## Why

Two changes that each make sense on their own do not agree on a filename.

`.github/workflows/coverage.yml` runs `pnpm test:coverage` and then reads:

```js
const coveragePath = path.join(repositoryRoot, "coverage/unit/coverage-final.json");
```

It needs the detailed Istanbul format, because it walks `statementMap` and `s` to get per-line hit counts.

`pnpm test:coverage` runs `scripts/unit-coverage.mjs`, which deliberately measures **one workspace at a time** — the file header explains that a single merged pass reports misleading numbers, which is exactly the problem #293 set out to fix. It writes per-project reports under `coverage/unit/<project>/` and one aggregate at:

```js
writeFileSync(resolve(COVERAGE_ROOT, "coverage-summary.json"), ...)
```

So `coverage/unit/coverage-final.json` is never written by anything. The gate cannot pass in its current form, and the aggregate that *is* written is the summary format, which has no `statementMap` and so cannot answer "was this changed line covered".

The check passed on `test/coverage-gates` (#288) because that branch predates the per-workspace restructure.

## Not proposing a fix

The obvious repairs — have the gate read the per-project `coverage-final.json` files and merge them, or have the script also emit a merged detailed report — both touch the thing #293 was careful about, and the choice belongs to whoever owns that design rather than to a passing PR.

Flagging rather than patching, because the check is red on unrelated pull requests right now and reviewers are reasonably reading it as a problem with the PR in front of them.

/cc @bestony

liuchao-001:
Additional evidence on scope: this workflow has **not succeeded on any branch since #288 landed**. The last eight runs, across eight different branches:

```
failure  fix/integration-pool-starvation      97026c4
failure  fix/integration-tab-kumo-table       0f41a5a
failure  fix/integration-pool-starvation      96d7db1
failure  feat/account-reboard-internal        799a076
[...]

PR #321, opened 2026-08-31
test(web): settle the render before reading a missing warning

liuchao-001:
## Summary

Follow-up to #310, agreed with the reviewer rather than deferred silently: `messaging-cli-routing.test.tsx` asserted that neither provider's CLI warning is on screen, using a synchronous `queryByText`. An absent warning and a warning that has not rendered yet are indistinguishable to that query, so on a loaded machine the case could pass without the step having drawn anything at all — a green with no content behind it.

Both cases now settle the render first. The silence case waits on the Slack panel's own lead (`SETUP_COPY.messaging.slackIntro`) rather than a page-level element, so what it waits for is the step under test and not the shell around it.

**Corrected after review.** The original comment claimed the case would false-green on a slow machine. It would not: nothing on this path is asynchronous — `MessagingStep`'s only effect calls the injected `onStart`, and the one async component, `QrCode`, renders only at `messaging.kind === "waiting"` while the test passes `idle` — so the panel and its warning settle inside `render`. The anchor still earns its place, because "no warning" and "no warning yet" read identically and the anchor is what separates them once anything async lands ahead of the warning; but the reason written in the file now says that instead of describing a failure this component cannot reach. A comment is read as documentation.

Test-only. No product code changes.

## Testing

- `pnpm check` — exit 0
- `pnpm typecheck` — exit 0
- `pnpm build` — exit 0
- `pnpm --filter @opentag/web exec vitest run src/onboarding-v2 src/setup` — exit 0, **17 files / 140 tests** green
- `pnpm --filter @opentag/web exec vitest run src/onboarding-v2/page.test.tsx` — exit 0, 42 tests green

**Mutation-verified in both directions, with a cold Vite cache before each run** (`rm -rf apps/web/node_modules/.vite`). Running a control pass and a mutated pass in the same hot worktree can reuse the cached transform of the unmutated file, which makes a mutant pass and look like a toothless assertion:

| `steps.tsx:335` | Result |
|---|---|
| unmutated | 2 passed |
| hard-coded `.feishu` | case 2 red — `expected <span></span> to be null` |
| hard-coded `.slack` | case 1 red — `Unable to find an element with the text: Lark messages are sent through its CLI…` |

**Evidence boundary.** The full local `@opentag/web` suite is not included: this machine is running several agents' test suites concurrently (load average 46–56 during this work), which starves vitest into 5s timeouts on files unrelated to any change. The run above is the affected surface — every test file under `src/onboarding-v2` and `src/setup` — and CI covers the whole suite on a clean machine.

## Breaking changes

None.

## Checklist

- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. (Full suite via CI; see the evidence boundary above for why the local full run is not cited.)

### DIFF
--- a/apps/web/src/onboarding-v2/messaging-cli-routing.test.tsx
+++ b/apps/web/src/onboarding-v2/messaging-cli-routing.test.tsx
@@ -5,6 +5,10 @@
  * happens at this call site rather than inside the check. Nothing else asserts that: a hard-coded
  * provider would warn a Slack user about a CLI they do not need, or stay silent about one they do,
  * and every other test feeds the two providers the same status.
+ *
+ * Both cases wait for something the step must render before reading. The silence one needs it: an
+ * absent warning and a warning that has not arrived yet look identical to a synchronous query, so
+ * without a settled render the assertion would pass on a slow machine for the wrong reason.
  */
 
 import { render, screen } from "@testing-library/react";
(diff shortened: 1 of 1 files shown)


bestony (approved):
Approve。silence 原先用同步 `queryByText` 读「没有 warning」，缺席和尚未画出无法区分；现在先等 Slack 面板自己的 lead（`SETUP_COPY.messaging.slackIntro`），再断言两边 warning 都不在。warning 那条改成 `findByText`。产品侧 `cliState` 和 warning / slackIntro 在同一帧算出，等 lead 之后的缺失就是真的缺失。Test-only，范围对。

Enforce Patch Coverage 红的是 #320（`coverage-final.json` 没人写），和这个 diff 无关。Checks / CI 已绿。

bestony (approved):
Approve on 7f2dbc2. 新 commit 只改注释：收回「慢机器空绿」——这条路径上 warning 和 panel 在 `render` 里一次算完，`QrCode` 的异步只在 `messaging.kind === "waiting"`，测试传的是 `idle`。锚点仍在，理由改成「无 warning」和「还没画出」读出来一样，留给以后 warning 前面出现异步的那天。测试代码相对上次 Approve 没变。

PR #330, opened 2026-08-31
test(web): realign the empty-Tasks assertion with the shipped copy

liuchao-001:
Closes #331.

## `main` is red, and it broke on landing rather than in review

`apps/web/src/features/tasks-page.test.tsx` asserts an empty-state sentence that the product stopped
saying:

| | |
|---|---|
| Test expects | `No Tasks yet. Work this Agent handles in Feishu or Slack appears here.` |
| Product renders | `No Tasks yet. Message this Agent in your chat app to put it to work.` |

### How two green PRs produced a red `main`

- **#317** (`0e36908`) rewrote the empty state in `tasks-page.tsx` — deliberately, to stop naming
  providers in that sentence.
- **#294** (`e858cca`) added a test asserting the sentence as it read *before* #317.

Neither touched the other's lines, so there was no textual conflict and both merged clean. The break
only exists in the combination — a semantic conflict git cannot see, which is why no PR gate caught
it.

### Which side is wrong

The **test**. #317 changed the copy on purpose, so the shipped string is the intended one and the
assertion is what went stale.

### Verified rather than inferred

I reproduced the failure on a **pristine worktree at `e858cca`** with nothing else applied, to rule
out any in-flight branch as the cause:

```
FAIL src/features/tasks-page.test.tsx > renders Agent Tasks loading, empty, append success, and append failure states
TestingLibraryElementError: Unable to find an element with the text:
  No Tasks yet. Work this Agent handles in Feishu or Slack appears here.
Tests  1 failed | 479 passed (480)
```

Still present on current `main` (`2dbc5df`) — #316 landing did not touch it.

With this one-line change, on `2dbc5df`:

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **40 files / 482 tests passed** (exit 0) |

Found while rebasing #311, which edits this file and inherited the red. Sent as its own PR rather
than folded into a naming change, so it can land immediately and unblock every branch — the fix
belongs to `main`, not to #311.

PR #332, opened 2026-08-31
fix(web): align the empty Tasks assertion with the shipped copy

liuchao-001:
## What is broken

`main` is red. `Checks`, `Node 22.13.0 Compatibility` and `Node 26 Compatibility` all fail on the same assertion, and `Deploy Staging` is skipped as a result ([run 33351770516](https://github.com/first-tree-ai/opentag/actions/runs/33351770516)):

```
FAIL web src/features/tasks-page.test.tsx > Tasks debug view > renders Agent Tasks loading, empty, append success, and append failure states
TestingLibraryElementError: Unable to find an element with the text:
No Tasks yet. Work this Agent handles in Feishu or Slack appears here.
```

## Why

Two merged PRs disagree about one string, and neither was wrong on its own branch.

- #317 (`0e36908`) changed the Tasks empty state to **"No Tasks yet. Message this Agent in your chat app to put it to work."**
- #294 (`e858cca`) landed afterwards carrying an assertion that still expects the pre-#317 copy.

Nothing in the source renders the old sentence any more — `git grep` finds it only in this assertion.

## The fix

Point the assertion at the copy the component actually renders. The component text is the intentional one: #317 deliberately replaced the provider-naming sentence, and the stale copy also names a channel we have since stopped naming in that surface.

## Verification

- `apps/web/src/features/tasks-page.test.tsx` — **32/32 pass** with this change
- positive control: with the change stashed, the same run fails 1/32 with exactly the CI error above, so the fix is what closes it

Repository-wide test noise on my host is documented elsewhere and is not touched by this one-line change; exact-head CI is the full-suite signal.

### DIFF
--- a/apps/web/src/features/tasks-page.test.tsx
+++ b/apps/web/src/features/tasks-page.test.tsx
@@ -543,7 +543,7 @@ describe("Tasks debug view", () => {
     expect(screen.getByText("Loading Tasks…")).toBeTruthy();
     release({ tasks: [], nextCursor: null });
     expect(
-      await screen.findByText("No Tasks yet. Work this Agent handles in Feishu or Slack appears here."),
+      await screen.findByText("No Tasks yet. Message this Agent in your chat app to put it to work."),
     ).toBeTruthy();
 
     vi.mocked(browserApi.tasks)


liuchao-001:
Closing as a duplicate of #330, which does the same one-line realignment and is already approved by @yuezengwu.

Both PRs came from my own sessions minutes apart, which is my mistake rather than a disagreement about the fix. #330 is ahead on review, so it is the one to land.

The underlying breakage is the same either way: #317 changed the empty-Tasks copy in `tasks-page.tsx` without updating `tasks-page.test.tsx:546`, so every PR inherits a red `Checks` once it merges with `main`.

liuchao-001:
## Independent QA — PR #332

**Status: PASS** · **Tier: `test-only`** · **Head SHA: `c937f763543ef551cb45a84c73b50036b64523f3`**

Independent QA run by code-reviewer, bound to the exact head above. `test-only` is the right tier and I agree with the proposal: the change is a single assertion string in a test file, with no product code touched.

### Root cause — independently confirmed before reading the author's account

I had already derived the same cause while reviewing #309's inherited red, from the source rather than from the report:

- `apps/web/src/features/tasks-page.tsx:276` renders `No Tasks yet. Message this Agent in your chat app to put it to work.`
- `apps/web/src/features/tasks-page.test.tsx:546` asserted `No Tasks yet. Work this Agent handles in Feishu or Slack appears here.`
- `git log -L` attributes the component copy to **#317 (`0e36908`)** and the stale assertion line to **#294 (`e858cca`)**, which landed afterwards still carrying the pre-#317 sentence. Each branch was green alone; the conflict only exists in the merged tree — a textbook semantic conflict that no merge conflict marker catches.

Taking the **component** side is correct: #317 changed the copy deliberately, the asserted sentence no longer exists anywhere in the repo (`grep` over `apps/` and `packages/` returns nothing), and a test asserting text the product never renders is simply wrong. The shipped sentence also avoids naming channels, which matches the current messaging-copy direction.

### Verification

| Check | Result |
|---|---|
| `pnpm install --frozen-lockfile` | 0 |
| `pnpm --filter @opentag/shared build` | 0 |
| `pnpm check` | 0 (151 warnings, all inherited from current `main`) |
| `pnpm --filter @opentag/web typecheck` | 0 |
| `tasks-page.test.tsx` **with** the fix | **32/32 pass**, exit 0 |
| **Positive control** — revert the one line, rerun | **1 failed / 31 passed**, `TestingLibraryElementError: Unable to find an element with the text: No Tasks yet. Work this Agent handles in Feishu or Slack appears here.` — byte-identical to the CI error |

The control was run with `node_modules/.vite` cleared on both sides so neither result came off a stale transform cache.

**The fix is sufficient for all three red jobs, not just one.** I read the `Node 26 Compatibility` job log directly (job `99368940114`) rather than assuming: it fails on the *same single assertion*, as does `Node 22.13.0` and `Checks`. One cause, three jobs.

**CI at this exact head is now fully green** — `Checks`, `CI`, `Node 22.13.0`, `Node 26`, `CLI Pack Smoke`, `Container Smoke`, `Workflow and Shell Validation`, `Secret Scan`, and `Enforce Patch Coverage` all SUCCESS. That last one also confirms #324's coverage-workflow repair is working.

### Sweep for siblings

A one-line copy fix is only trustworthy if it is the *only* stale assertion, so I checked rather than assumed:

- The old sentence appears nowhere else in `apps/` or `packages/`.
- I extracted every user-facing string #317 removed (`Connect messaging`, `Feishu · @reviewer`, `Slack · Reviewer`, `Messages cannot be sent to this Agent.`, `Messaging status unavailable`, `This agent's computer is offline…`, `OpenTag is not running on Ada's Mac…`) and found no test still asserting them.
- Consistent with CI, which reported exactly one failing web test file and one failing test inside it.

I found no second instance of this class.

### Evidence boundary

This run proves only that the automated checks above passed or failed, plus the positive control. It asserts nothing about live or integrated behavior, and none is needed here — the product code is untouched and the shipped string was read directly from the component.

### Next step

No findings. The change is correct, minimal, and unblocks `main` and every PR that has synced it (including #309's current head). It needs an approving review from someone other than the author before it can merge — `mergeStateStatus` is `BLOCKED` solely on `REVIEW_REQUIRED`. I am not approving it myself: my `gh` credential is the author's own account, so an approval from me would be a self-approval, not independent sign-off.

Issue #295, opened 2026-08-30
Correlate a connect code with the Computer that redeems it


liuchao-001:
## The problem

The web cannot tell which Computer answered a connect code it issued.

Onboarding issues a code, then waits for a Computer to appear. "A Computer came online" is the only signal available, so any Computer that enrolls or reconnects during that window satisfies the wait — including a colleague's machine in the same Workspace. That guess is then passed to `createAgent` as `computerId`, where it becomes the Agent's durable placement.

## Why the client cannot close this

The contract carries no handle to correlate on:

- `ComputerConnectCodeIssueResponseSchema` returns `bootstrapCommand`, `expiresIn`, `issuedAt`, and `mode`. There is no code id.
- `ComputerConnectCodeExchangeResponseSchema` returns `workspaceComputerId` and `computerId` — to the **CLI** that performed the exchange, not to the browser that issued the code.
- The Computers list carries no field tying an entry to an issued code.
- `CreateAgentRequestSchema.computerId` is required, so creation cannot defer placement until the answer is known.

Every remaining option is a heuristic dressed differently. Repair is already exact — a repair code names its target, so `findRepaired` can require a fresh `connectedAt` on that one Computer. First run has nothing equivalent.

## What would close it

Some exact evidence tying the issued code to the Computer that redeemed it. For example the issue response returning an identifier for the code, and a read that reports which Computer redeemed it (or the Computer summary carrying the redeemed code's id). The shape is yours to choose; the requirement is that the browser can name the Computer without inferring it from arrival timing.

## Impact

Wrong-machine placement is silent and durable. In a single-machine Account it is invisible; in a Workspace where more than one person is setting up, or where an existing machine reconnects during the window, onboarding can bind the new Agent to someone else's Computer.

## Context

Found in review of #262 by @yuezengwu, who correctly held that this belongs behind a Server contract rather than a heuristic added in a UI change. The UI-side alternative — asking the reader to confirm the detected machine — was considered and rejected, because the current onboarding interaction is settled and should not grow a confirmation step for this.

The pre-existing `computer-setup.tsx` shares the same limitation; this is not a regression introduced by #262, but #262 makes it load-bearing for every new Account.

yuezengwu:
Resolved by #339, merged as `c90cd29b2193e7d9f192cfd3c836d1fc72c3206c` from reviewed head `f0acd47bebefa35b5052ff24aaacddb6bd53b01e`.

Delivered:

[...]

PR #339, opened 2026-08-31
fix: correlate connect codes with redeemed computers

yuezengwu:
## Summary

Relates to #295.

Implementation head: `f0acd47bebefa35b5052ff24aaacddb6bd53b01e`.

The browser can now correlate a Computer connect command with the exact Computer that redeemed it, using Server-owned evidence instead of Computer arrival or reconnect timing.

- The issue response includes an opaque, non-secret `connectCodeId`.
- `GET /api/v1/computer-connect-codes/:connectCodeId` returns `pending`, `redeemed`, `expired`, or `revoked`; only `redeemed` carries the exact `computerId` and `redeemedAt`.
- The read is authenticated to the Account that issued the code. Foreign and unknown ids are indistinguishable `404`s; machine credentials and anonymous callers are rejected.
- The response schema is strict and cannot carry the raw code, token hash, machine token, or a Computer in a non-redeemed state.
- New onboarding and `ComputerSetup` poll this exact verdict. Unrelated Computer enrollment or reconnects cannot satisfy the wait.
- After redemption, the UI waits until that exact Computer has an online connection at or after `redeemedAt`. A redeemed-but-offline Computer remains pending in the UI rather than being treated as connected.
- Every targeted recovery, including Agent Settings, issues `{ mode: "repair", targetComputerId }`. `ComputerSetup` accepts only a verdict for that exact target, so recovery neither creates a duplicate Computer nor drifts the Agent binding.
- A status-read `404` now ends in the existing recoverable expired state, preserves the command and Refresh action, and does not expose the raw Fastify error or adopt a nearby Computer.
- Both Web flows share one connect-code verdict parser. Connect-code issuance no longer performs work inside a React state updater.

No confirmation step, fallback heuristic, Agent placement change, or optional-Computer behavior is added.

## Security and migration

No database migration is required. The implementation reuses the existing `computer_connect_codes.id`, `issued_by_account_id`, `consumed_computer_id`, `consumed_at`, `expires_at`, and `revoked_at` fields.

`connectCodeId` is a UUID row identifier, not an exchange credential. Redemption still requires the one-time connect code itself. Status responses use `Cache-Control: no-store` and expose only correlation state.

Repair target ownership is enforced by the existing Server issue-time Account authorization and database checks; this follow-up does not add a second client-side authority model.

## Testing

- `pnpm check` — passed; 153 existing complexity diagnostics remain warnings
- `pnpm build` — passed
- `pnpm typecheck` — 8/8 tasks passed
- `pnpm test` — 84 repository script tests and 1,907 package tests passed:
  - shared 94
  - client 582
  - server 486
  - CLI 206
  - Web 539
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 22 files / 317 tests passed; 100% statements, branches, functions, and lines
- `pnpm --filter @opentag/server test:integration` — 16 files / 293 tests passed
- Focused Web correlation coverage — 3 files / 65 tests passed
- Focused cases include 404-to-Refresh recovery, replacement-code isolation, repair request arguments, target-id refusal, online and `connectedAt >= redeemedAt` gating, pending, expired, revoked, unrelated Computer activity, superseded polls, and redeemed-but-offline behavior.

## Breaking changes

The connect-code issue response now requires `connectCodeId`, so the updated Web/client and Server must be deployed together. A new Web bundle against an older Server recovers from the missing status route through the Refresh-capable expired state rather than becoming stuck. The machine exchange endpoint and its secret-bearing response are unchanged. There is no persisted-data migration.

## Checklist

### DIFF
--- a/apps/web/src/__tests__/api.test.ts
+++ b/apps/web/src/__tests__/api.test.ts
@@ -251,6 +251,7 @@ describe("BrowserApi", () => {
       expect(init?.body).toBeUndefined();
       return new Response(
         JSON.stringify({
+          connectCodeId: "7a1c9e52-9a8b-4c7d-8e1f-2a3b4c5d6e7f",
           bootstrapCommand: `opentag computer connect --server http://127.0.0.1:8000 -- code`,
           expiresIn: 900,
           issuedAt: "2030-01-01T00:00:00.000Z",
@@ -327,5 +328,44 @@ describe("BrowserApi", () => {
     });
     await expect(new BrowserApi(malformedCookieFetch).logout()).resolves.toBeUndefined();
     setDocumentCookie("opentag_csrf=; Path=/; Max-Age=0");
+  });
+
+  it("reads the connect code's redemption verdict from its Account-native path", async () => {
+    const status = {
+      connectCodeId: "7a1c9e52-9a8b-4c7d-8e1f-2a3b4c5d6e7f",
+      state: "redeemed",
+      computerId: "85fe9af3-d1c6-472b-b78c-8a7ccf512750",
+      redeemedAt: "2030-01-01T00:00:05.000Z",
+    };
+    const fetchImpl = vi.fn<typeof fetch>(async (input, init) => {
+      expect(String(input)).toBe(`/api/v1/computer-connect-codes/${status.connectCodeId}`);
+      expect(init?.method).toBeUndefined();
+      return new Response(JSON.stringify(status), { status: 200, headers: { "content-type": "application/json" } });
+    });
+    const api = new BrowserApi(fetchImpl);
+
+    await expect(api.computerConnectCodeStatus(status.connectCodeId)).resolves.toEqual(status);
+  });
+
+  it("fails the connect code status read closed when the Server will not name it", async () => {
+    const fetchImpl = vi.fn<typeof fetch>(
+      async () =>
+        new Response(
+          JSON.stringify({
+            error: {
+              code: "RESOURCE_NOT_FOUND",
+              category: "deterministic",
+              message: "The requested resource was not found",
+              requestId: "req_1",
+            },
+          }),
+          { status: 404, headers: { "content-type": "application/json" } },
+        ),
+    );
+    const api = new BrowserApi(fetchImpl);
+
+    const failure = await api.computerConnectCodeStatus("7a1c9e52-9a8b-4c7d-8e1f-2a3b4c5d6e7f").catch((cause) => cause);
+    expect(failure).toBeInstanceOf(ApiError);
+    expect(failure.status).toBe(404);
   });
 });
--- a/apps/web/src/__tests__/app.test.tsx
+++ b/apps/web/src/__tests__/app.test.tsx
@@ -123,6 +123,7 @@ function installApi(
   let loggedOut = false;
   let meFailuresRemaining = options.meFailuresAfterProfileUpdate ?? 0;
   let computerConnectCodeIssued = false;
+  const connectCodeId = "7a1c9e52-9a8b-4c7d-8e1f-2a3b4c5d6e7f";
   vi.mocked(fetch).mockImplementation(async (input, init) => {
     const path = String(input);
     if (path === "/api/v1/auth/providers") {
@@ -285,28 +286,23 @@ function installApi(
             ],
       });
     }
-    if (path === "/api/v1/computers") {
-      const failureStatus = options.computerReadStatus?.(computerConnectCodeIssued);
-      if (failureStatus) return json({ error: { message: "Computer readiness unavailable" } }, failureStatus);
-      if (options.computerEvidenceFails) return json({ error: { message: "Computer evidence unavailable" } }, 503);
+    const normalizeComputer = (computer: Record<string, unknown>) => ({
+      computerId: computer.computerId ?? computer.id,
+      displayName: computer.displayName,
+      platform: computer.platform,
+      connectionStatus: computer.connectionStatus,
+      providerReadiness: computer.providerReadiness,
+      connectedAt: computer.connectedAt ?? null,
+      lastSeenAt: computer.lastSeenAt ?? null,
+      observedAt: computer.observedAt ?? computer.lastSeenAt ?? computer.connectedAt ?? "2026-08-20T00:00:00.000Z",
+      enrolledAt: computer.enrolledAt ?? "2026-08-20T00:00:00.000Z",
+      agentIds: computer.agentIds ?? [agentId],
+    });
+    const listComputers = async (connected: boolean) => {
       const configured =
-        typeof options.computers === "function"
-          ? await options.computers(computerConnectCodeIssued)
-          : options.computers;
-      const normalizeComputer = (computer: Record<string, unknown>) => ({
-        computerId: computer.computerId ?? computer.id,
-        displayName: computer.displayName,
-        platform: computer.platform,
-        connectionStatus: computer.connectionStatus,
-        providerReadiness: computer.providerReadiness,
-        connectedAt: computer.connectedAt ?? null,
-        lastSeenAt: computer.lastSeenAt ?? null,
-        observedAt: computer.observedAt ?? computer.lastSeenAt ?? computer.connectedAt ?? "2026-08-20T00:00:00.000Z",
-        enrolledAt: computer.enrolledAt ?? "2026-08-20T00:00:00.000Z",
-        agentIds: computer.agentIds ?? [agentId],
-      });
-      return json({
-        computers: configured?.map(normalizeComputer) ?? [
+        typeof options.computers === "function" ? await options.computers(connected) : options.computers;
+      return (
+        configured?.map(normalizeComputer) ?? [
           normalizeComputer({
             id: computerId,
             displayName: "Ada's Mac",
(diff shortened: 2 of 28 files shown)


liuchao-001 (commented):
Reviewed at head `85f69ae`. The correlation model is the right one and the server half checks out — I verified the parts that are easy to get subtly wrong and they hold:

- **Id spaces match.** `computer_connect_codes.consumed_computer_id` references `account_computers.id`, and the Computers list's `computerId` is `account_computers.id` (`computer-service.ts:71`), so the browser's `candidate.computerId === verdict.computerId` compares the same thing.
- **The `connectedAt >= redeemedAt` gate cannot deadlock.** `#createComputer` inserts with `connectedAt: null`, `#repairComputer` explicitly nulls `connectedAt`/`currentInstanceId`, and only `register()` sets it — so the adopted connection always postdates redemption. Both columns are `timestamptz` written from the same server clock, so there is no truncation skew that could strand a machine that connects inside the same second.
- **The DB constraints back the state machine the read assumes** — the consumption pair, the consumed-identity check, and `not (consumed_at is not null and revoked_at is not null)` — so the redeemed-before-revoked ordering in `getConnectCodeStatusForAccount` is not merely a convention.
- **Authorization is the same `createUserAuthPreHandler` as its sibling account routes**, ownership is checked against the token's Account, foreign and unknown ids collapse to one 404 shape, and the strict response schema plus `no-store` keep the code, its hash, and any machine token out of the reply.
- `issuingFlight` hands off correctly across `reset()`: a superseded request cannot clear its successor's guard.

I ran the suites locally at this head: `@opentag/server` 486 passed, `@opentag/web` 531 passed.

Two things I'd like you to look at.

---

### 1. A `404` on the status poll strands the onboarding connect step with no way forward

`apps/web/src/onboarding-v2/server-backend.ts:347` — `failIssuedCodePoll` answers a `404` with `setConnect({ kind: "idle" })`. But `idle` is not a terminal the reader can act on:

- `ConnectCommand` (`apps/web/src/onboarding-v2/steps.tsx:798`) renders `idle` as an inert, `aria-hidden` placeholder command. The Refresh button lives only on the `expired` branch, so it disappears.
- The only automatic re-issue is `apps/web/src/onboarding-v2/page.tsx:199`, whose deps are `flow.page`, `connectingNewComputer` and a stable `issueConnectCode`. None of them change here, so the effect never re-fires.
- Leaving `issued` also kills the interval, so the local expiry can never convert this into the `expired` state either.

The reader is left with a dead placeholder block and a raw `The requested resource was not found` banner (the `COPY.errors.connectCode` fallback is not used, since `errorMessage` prefers `cause.message`), and only a page reload gets them out.

The realistic trigger is the version skew the PR itself calls out under Breaking changes: a browser holding the new bundle against a server that does not yet serve `GET /api/v1/computer-connect-codes/:connectCodeId` gets Fastify's `404` on *every* poll — during a rolling deploy, a rollback, or from a cached SPA. That turns a transient deploy window into a stuck onboarding step rather than a retryable one.

Suggested fix: end the wait through the same terminal as revoked/expired, which is already the "the Server has closed this code" path and keeps the command and its Refresh on screen:

```ts
if (gone) setConnect((current) => (current.kind === "issued" ? { kind: "expired", command: current.command } : current));
```

`server-backend.test.ts:398` pins the current `idle` outcome, so that expectation would move with it.

### 2. Pre-existing, but this PR now enshrines it: the targeted recovery path cannot succeed

`apps/web/src/features/agents/computer-setup.tsx:135` issues an **open** code — `browserApi.issueComputerConnectCode()` with no argument — even when `target` is set, and then waits for `targetComputerId` to reconnect. But `allocateComputerIdentity` mints a fresh installation id on every `computer connect`, and `#createComputer` always inserts a new `account_computers` row. So redeeming that code enrolls a *second* Computer; the named one's `connectedAt` never moves, and "Reconnect this Computer" in Agent settings can only run out the clock and report expired — while quietly leaving a duplicate Computer behind and the Agent's binding still broken.

None of that is introduced here, and the panel's exact-id match is at least not *unsafe*. But this PR is precisely about correlating a code with the machine that redeemed it, and it now documents that path as the standing exception ("The targeted recovery is the exception: it still watches for its own named Computer to reconnect"). Onboarding already gets this right by issuing `{ mode: "repair", targetComputerId }`. Either give the panel the same repair code plus the new verdict, or open a follow-up so the exception is tracked rather than blessed.

### Nits (non-blocking)

- `readConnectCodeVerdict` (`server-backend.ts:69`) and `readVerdict` (`computer-setup.tsx:67`) are the same function with the same doc comment in two modules, and in both the `!status.computerId || !status.redeemedAt` half is unreachable under the discriminated union. One shared helper would do.
- In `ComputerSetup`, a failed *status* read is reported as `Unable to refresh Computers` via `reportPollFailure`, which names the wrong call.
- `issueConnectCode` performs a side effect (`queueMicrotask(() => void issue())`) inside a `setState` updater, which is exactly what forced the `issuingFlight` guard. Deciding from `connectRef.current` outside the updater would remove the need for the guard rather than compensate for the double invocation.

liuchao-001:
Following up on point 2 of my review: @liuchao-001 would like the targeted recovery path fixed **in this PR** rather than tracked as a follow-up. The reasoning is that this change is exactly the one that establishes code↔Computer correlation, so leaving one call site on the old baseline heuristic keeps an exception alive that the next reader has to rediscover.

Concretely, the panel already has everything it needs:

1. **Issue the right code.** `computer-setup.tsx:135` becomes mode-aware, the same way onboarding already does it:

   ```ts
   const issued = await browserApi.issueComputerConnectCode(
     targetComputerId ? { mode: "repair", targetComputerId } : undefined,
   );
   ```

   This is also the part that fixes the user-visible bug: today an open code redeemed on a machine that already has a Computer enrols a *second* one, because `allocateComputerIdentity` mints a fresh installation id per `computer connect` and `#createComputer` always inserts.

2. **Then both paths are one path.** A repair redemption reuses the same `account_computers` row, so the verdict's `computerId` *is* `targetComputerId`, and `#repairComputer` nulls `connectedAt`/`currentInstanceId` — which means the `isFreshlyConnected(computer, redeemedAt)` gate you already wrote is the correct and sufficient test for "it came back". `pollVerdict` can serve the targeted wait unchanged.

3. **Which deletes the exception rather than documenting it.** `captureBaseline`, `isReconnection`, `pollTarget`, the `baselineConnections` ref, and the `targetComputerId ? … : …` branch in `tick` all go, along with the paragraph in the file header that carves out the targeted case. The panel ends up with one wait driven by one verdict.

For tests: the targeted cases in `computer-setup.test.tsx` move from "a foreign reconnection must not satisfy the wait" (which the baseline was there to prove) to "a redemption naming another Computer must not satisfy the wait, and the named one is adopted only once it is online at or after `redeemedAt`" — the same claims your onboarding tests already make, and worth keeping the assertion that the issue request carried `{ mode: "repair", targetComputerId }`, since that is the half a heuristic can never recover from.

One optional hardening while you are in `issueForAccount`: a `repair` code is inserted without checking that `targetComputerId` belongs to the calling Account — ownership is only enforced at redemption in `#repairComputer`. It fails closed, so nothing leaks, but with the new status read such a code now sits visibly `pending` for its whole TTL instead of being refused at issue time. Your call whether that belongs here or not.

Point 1 (the `404` terminal) still stands on its own.

yuezengwu:
Addressed review `5063161278` and the targeted-recovery follow-up at exact head `f0acd47bebefa35b5052ff24aaacddb6bd53b01e`.

- A connect-code status `404` now transitions the issued command to the existing recoverable `expired` terminal, preserves the command/Refresh action, clears stale poll errors, and never adopts a Computer. The regression test refreshes into a replacement code and verifies the replacement remains isolated.
- Agent Settings targeted recovery now issues `{ mode: "repair", targetComputerId }` and uses the same exact status verdict as open enrollment. The baseline/reconnection heuristic and its exception text are removed.
- Targeted recovery refuses a verdict naming another Computer and adopts the target only when it is online with `connectedAt >= redeemedAt`; focused tests cover foreign ids, offline targets, stale connections, and expired/revoked outcomes.
- The duplicated verdict parser is shared, status-read and post-redemption list-read failures have accurate fallback wording, and connect-code issuance no longer starts work inside a state updater or uses a separate in-flight guard.
- No extra issue-time ownership layer was added: the existing Server repair-code issuance already validates the target against the authenticated Account and fails before issuing when ownership does not match.

Validation at this head:

- focused Web: 3 files / 65 tests
- `pnpm check`, `pnpm build`, `pnpm typecheck`
- `pnpm test`: 84 repository script tests plus 1,907 package tests; Web 539
- client Agent Runtime coverage: 22 files / 317 tests, 100% statements/branches/functions/lines
- Server integration: 16 files / 293 tests
- pre-commit and pre-push gates passed

Please re-review this exact head.

liuchao-001 (commented):
Re-reviewed at exact head `f0acd47`. Both findings are resolved, and one of them came back stronger than what I asked for. No remaining blockers from me.

**1. The `404` terminal.** The status read now fails closed into the existing recoverable `expired` state, keeps the command and its Refresh on screen, clears the stale poll error instead of showing a raw Fastify message, and adopts nothing. What makes this convincing rather than merely green is the test that refreshes into a replacement code and then asserts the poll targets the *replacement's* id — the isolation is the part a state change like this can quietly get wrong.

**2. Targeted recovery.** The panel issues `{ mode: "repair", targetComputerId }` and rides the same verdict as open enrollment; `captureBaseline`, `isReconnection`, `pollTarget` and the `baselineConnections` ref are gone along with the exception paragraph. The extra `redeemedComputerId !== targetComputerId` refusal in `adoptVerdict` goes beyond the ask — the Server's repair targeting already makes a foreign id impossible, so this is a second independent lock on the same door, and it is cheap. Foreign-id refusal, offline target and stale-connection cases are all covered.

**3. The nits.** One shared `readConnectCodeVerdict`, and you correctly dropped the unreachable malformed-redemption branch instead of porting it — the discriminated union already guarantees a redemption carries its evidence. Issuance is out of the state updater: the decision reads and writes `connectRef.current` before the async call, and since `issue()` calls `setConnect({ kind: "issuing" })` synchronously before its first `await`, no interleaving can slip a second code past the guard. Dropping `issuingFlight` is right — the guard was compensating for the updater side effect rather than for a real race.

**One correction on my side.** My optional issue-time ownership note was wrong, and your answer is the accurate one: `#insertCode` locks the target and rejects `target.ownerAccountId !== accountId` with a `404` before any row is written. I had read `issueForAccount` and stopped short of the helper it delegates to. There is nothing to add there, and I agree a client-side second authority model would have been the wrong shape anyway.

**Verification at this head.** `@opentag/web` 539/539 and `@opentag/server` 486/486 locally. In the interest of not leaving a misleading trace: my first web run reported 12 failures in `app.test.tsx` — timeouts and missing elements — which reproduced as nothing on a clean rerun of the same command. The machine was at a load average of 21–31 from unrelated work at the time, and jsdom shell tests with a 5s timeout are the first thing to fall over under that. I am reporting the clean run, not averaging the two.

**Non-blocking, and explicitly not for this PR.** The `issue()` failure path still ends in `idle`, which renders the inert placeholder with no Refresh — so a failed *issuance* (a `500` on the POST) strands the step much as the `404` used to, minus a command worth keeping on screen. That is pre-existing and outside this diff's subject; I am noting it only so it is not rediscovered as new.

liuchao-001 (approved):
Approving at head `f0acd47`. Both review findings are resolved, the targeted-recovery fix landed in this PR rather than as a follow-up, and `@opentag/web` 539/539 and `@opentag/server` 486/486 pass locally at this exact head.

Issue #338, opened 2026-08-31
Model and Resources settings blocks are merged but rendered nowhere; the settings page is still the six-row directory


liuchao-001:
## The Model and Resources blocks are built, tested, and not on any screen

#309 merged `AgentModelSettings` (`model-settings.tsx`) and `AgentResourcesSettings`
(`resources-settings.tsx`). Both are complete and covered by
`model-resources-settings.test.tsx`. Neither is rendered.

Verified on `main` (`a648112`):

```
git grep -n "model-settings\|resources-settings" -- apps/web/src
→ only model-resources-settings.test.tsx imports them
```

The settings route still renders `AgentSettingsOverview`, which maps `agentSettingsSections`
into six `<Link>` rows exactly as before. A user cannot reach either component.

The suite is green because the tests render the components directly — which proves they render,
not that the application puts them on a screen. That is the same gap that made a black-box run
worth insisting on during #311's review.

## Also stale, in the same place

`sections.ts` still labels the execution row **"Model & reasoning"**. @liuchao-001 asked for that
heading to be just **"Model"**, on the grounds that reasoning effort is part of a model setting in
both Codex and Claude, so users read it as one thing. The new `model-settings.tsx` follows that
decision; the directory row that is actually rendered does not.

## What this is really about

The Agent Settings redesign has landed its supporting pieces — the shared setup module (#310), the
Model and Resources blocks (#309), the naming helper (#311) — but **not the redesign**. The agreed
shape was one screen: identity editable in place, Computer and Messaging as live status cards,
Model and Resources as blocks, danger zone inline, and the sub-pages replaced by dialogs.

What ships today is still the six-row directory, and every state that has to be *seen* is one click
away behind a summary line.

So this issue is the seam, not a defect report: someone needs to render these blocks, and the
sensible way to do it is the page collapse rather than bolting them onto the directory.

## Suggested order

1. Render `AgentModelSettings` and `AgentResourcesSettings` on the settings overview.
2. Collapse the overview: identity inline, Computer and Messaging status cards, danger zone inline.
3. Keep `/settings/$section` resolving — the failure-state deep links from the Agent cards and the
   recovery paths depend on it, so those become anchors rather than dead routes.
4. Reconnect and messaging-connect move into dialogs built on the shared setup module.

liuchao-001:
Closing at @liuchao-001's decision. #340 attempted this and is closed unmerged; the work is not being continued.

[...]

Issue #308, opened 2026-08-30
client-runtime-composition tests fail non-deterministically on a local macOS host


liuchao-001:
`packages/client/src/__tests__/client-runtime-composition.test.ts` fails non-deterministically on a macOS development host while passing on CI. The failing *set* changes between runs, not just the count, which points at timing or ordering coupling inside the file rather than a missing local dependency.

## Observed

Same commit (`8bf466b` on `chore/remove-onboarding-lab`), same quiet machine, same command:

| run | failures |
| --- | --- |
| full `pnpm --filter @opentag/client test` | 9 / 555 |
| full `pnpm --filter @opentag/client test` | 10 / 555 |
| file alone, run 1 | 6 / 33 |
| file alone, run 2 | 10 / 33 |

Three tests that failed when the file ran alone passed in the full-suite runs:

- `keeps placement composable and defers unavailable Codex failure to the exact Turn runtime start`
- `detects Provider Home replacement after readiness and supports default environment/logger composition`
- `keeps the credential-grant capability while refreshing Provider readiness`

A representative first failure: `expected "spy" to be called 3 times, but got 2 times`.

The clustering is suggestive — probe deadlines, cancellation, and periodic refresh — i.e. tests that assert on how many times something was called by a particular moment.

## Why this is worth fixing rather than ignoring

CI is green, so nothing is blocked. The cost is to local trust: a developer who runs the client suite sees red, cannot tell which failures are real, and learns to disregard the suite. It also makes the machine useless for judging whether *any other* test is flaky, because a run under an already-unstable suite has no clean baseline.

## Not caused by host load

These runs were taken after eight orphaned CPU busy-loops from an unrelated local stress script were killed. Under that load a second file, `claude-code-agent-runtime-exhaustive.test.ts`, also failed; it passes consistently once the host is quiet. The instability described here survives on an idle machine.

## Environment

macOS (Apple Silicon), Node 22, vitest 3.2.7. Not reproduced on CI runners.

liuchao-001:
One more constraint on measuring this, learned while gathering the numbers above.

Three agents on this host were running `client-runtime-composition.test.ts` at overlapping times — two of the runs in the table plus a separate baseline measurement in another workspace. For a file whose failures move with scheduling, concurrent runs contaminate each other, and the contamination is invisible in the output: you get a plausible failure count either way.

[...]

Issue #326, opened 2026-08-31
Flaky web suite: app.test.tsx fails a different six tests every run under load


liuchao-001:
## Summary

`apps/web/src/__tests__/app.test.tsx` fails intermittently under machine load, and the failures
are **not** confined to the single 40ms window tracked in #306. Across three independent full-suite
runs on a loaded machine, each run failed **six** tests — and **no test failed in more than one
run**.

That zero-overlap pattern is the finding. A specific racy assertion fails repeatably; a set that
never repeats points at the file's timing assumptions as a class, not at one bad window.

## Evidence

Three `pnpm --filter @opentag/web test` runs by @code-reviewer, each against a different head of
`feat/feishu-naming`:

| Run | Head | Result | Load |
|---|---|---|---|
| 1 | `286982c` | 6 failed / 443 passed | 27–55 |
| 2 | `092e6b6` | 6 failed / ~443 passed | 27–55 |
| 3 | `f3d7166` | 6 failed / 451 passed | 37–41 |

- **Zero overlap between the three failing sets.**
- All failures are `waitFor` timeouts.
- Run 1 hit Token-usage / Usage-skeleton / post-logout / runtime-configuration; isolated re-runs of
  the same head instead hit `elapsed time for a working Agent` and `Agents-first shell for admins`.
- Targeted runs of the same specs were green every time (79/79).
- The same suite ran **fully green (38 files / 457 tests)** on the same machine at load **8.4**.
- CI, which is not contended the same way, is green on all jobs at each of these heads.

So: not a product defect, and not one test. The suite's timing assumptions hold at low load and
break somewhere different every time under contention.

## Why this is worth fixing rather than tolerating

The practical cost is already visible: a reviewer running independent QA cannot get a trustworthy
full-suite signal from a loaded machine, and has to fall back to CI plus targeted runs. That
weakens local verification exactly when someone is being careful. It also trains readers to
discount red, which is how a real regression gets waved through.

## Suggested direction

Not `retry` — that hides it. The useful work is to find which timing idiom is shared across the
failing tests (fixed wall-clock delays armed against `waitFor`, most likely, as #306 documents for
one case) and replace it with something that does not assume the machine is idle: fake timers, or
asserting on state transitions rather than on a window staying open.

#306 is the same class, scoped to one test. This issue is the broader pattern; #306 can stay as the
worked example.

Filed at @code-reviewer's suggestion, citing their three runs.

liuchao-001:
[...]

Issue #325, opened 2026-08-31
Flaky: app.test.tsx fails a different subset on every loaded run (broader than #306)


liuchao-001:
### Summary

`apps/web/src/__tests__/app.test.tsx` fails a **different subset of its tests on every run** when the
host is loaded, and passes fully in isolation. This is broader than #306, which pins one specific
test to a 40 ms wall-clock window: here the failing set has had **zero overlap between runs**, which
points at the file rather than at any one assertion.

### Observations

Four full `pnpm --filter @opentag/web test` runs on the same macOS host, all on branches whose diffs
touch none of these tests:

| Load avg | Failed | Which |
|---|---|---|
| ~30 | 2 | Account refresh-retry cases |
| 27–55 | 6 | `Token usage period`, `no Tasks report Token usage`, `Usage loading skeleton`, `keeps the loaded Tasks`, `post-logout route re-entry`, 2 in `runtime-configuration` |
| 27–55 | 2, then 3 | `elapsed time for a working Agent`, `Agents-first shell for admins` — different again on the next run |
| 37–41 | 6 | `elapsed time for a working Agent`, `states an offline reason`, `opens the Agent from the row`, `Computer connection command`, `Agent name local validation`, `revalidates ready routes` |
| 44 | 1 | `keeps the loaded Tasks when loading the next page fails` |
| ~8–15 | 0 | full 457 pass |

Every failure was a `waitFor` timeout. Isolated re-runs of the file pass 457/457. CI passes the same
suite on clean runners.

### Why this is worth a separate issue from #306

#306's fix — replacing one fixed delay with a deterministic gate — is right for that test, but the
evidence here says the pattern is file-wide: these tests share a harness that arms mocks on wall-clock
timing, so any of them can lose its window under contention. Fixing them one at a time as each is
observed red will take many rounds and each round costs a reviewer a false alarm.

### Why it matters beyond noise

A suite that goes red for environmental reasons trains readers to discount red. During #311 this cost
three separate rounds of "is this mine?" analysis across two agents, and the honest answer each time
required re-running to disprove rather than reading the failure.

### Suggested direction

Audit `app.test.tsx` for assertions that depend on a real elapsed interval and replace the timing
dependency with an explicit signal (resolve-controlled promises, or fake timers), rather than raising
timeouts. Per the repo's no-retry-masking rule, retries are not the fix.

Not urgent, and not a product defect — filing so the pattern is recorded rather than rediscovered.

[...]

PR #340, opened 2026-08-31
feat(web): collapse Agent settings into one screen

liuchao-001:
## Summary

Closes the first three steps of #338: render `AgentModelSettings` and `AgentResourcesSettings`, and render them by collapsing the settings route into one screen rather than bolting two more rows onto the directory.

`/agents/$agentId/settings` is now one screen, in the order a viewer thinks about an Agent:

| Block | What it is now |
| --- | --- |
| **Name** | editable in place |
| **Messaging** | live channel status, trigger mode, connect / reconnect / disconnect |
| **Computer** | live status, the machine it names, and its recovery |
| **Model** | `AgentModelSettings` from #309 |
| **Resources** | `AgentResourcesSettings` from #309 |
| **Danger zone** | pause and delete, inline |

`/settings/<section>` keeps resolving and becomes an anchor into that screen: it moves focus to the block's heading, so the failure exits on the Agent cards, `agentAvailabilityRecovery`, and the legacy `/runtime` and `/im` redirects still land somewhere — and land with the rest of the Agent in view. An unknown segment is still a not-found.

## Drift removed in the same place

- **`sections.ts` no longer carries labels.** The stale `"Model & reasoning"` this issue reports existed because a heading written beside a key is a second copy of the one the block renders, and the two drift. The list now owns only the order and identity of the anchors; each block owns its heading. That is the structural fix, not just the rename.
- **`RuntimeConfigurationForm` is deleted** rather than left beside the two blocks that replace it, along with its test.
- **`ImTab` → `AgentMessagingSettings`** (`im-tab.tsx` → `messaging-settings.tsx`): there are no tabs.
- **`returnLabel` / `returnAgentId` history state is dropped.** Both were only ever set to the same Agent the screen already has, and with one screen "Back to …" follows from the Agent itself.

## Not in this PR

Step 4 of #338 — moving reconnect and messaging-connect into dialogs built on `src/setup` — is a follow-up. Nothing is lost by deferring it: both flows are reachable and already expand in place inside their blocks, so it is a change of frame, not of reach. Keeping it separate keeps this diff reviewable and lets the dialog work be judged on the shared setup module's own terms.

## A latent defect this surfaced

`Dialog` restores focus with `useEffect(() => { if (!open) returnFocusRef?.current?.focus(); }, [open, returnFocusRef])`, but every caller renders it conditionally and never passes `open={false}`, so that effect never runs on close. What actually restores focus is Base UI's restore-to-previously-focused. That is pre-existing and out of scope here; I will file it separately.

## Validation

- `pnpm check` — exit 0 (151 warnings, all `noExcessiveCognitiveComplexity`, inherited from `main`)
- `pnpm typecheck` — exit 0
- `pnpm --filter @opentag/web test` — 512/512 passed, and the touched tests also pass in isolation, not only in file order
- `pnpm test` — `@opentag/client`'s `client-runtime-composition` tests fail non-deterministically on this macOS host; that is #308, no file under `packages/client` is touched here

The E2E suite's `Agent settings persist a change across reload` now reads the save confirmation from the Name block: with six blocks on one screen an unscoped `getByRole("status")` is ambiguous by construction. It has not been run here — Playwright needs the full stack — so it is part of what independent QA should cover, together with a browser pass over the deep links (`/settings/computer`, `/settings/messaging`, `/settings/manage`) and the anchor's focus landing.

Refs #338


liuchao-001:
The latent `Dialog` focus-restore defect noted above is now filed as #341.

liuchao-001:
CI on head `a1e52f3` is green across all nine checks.

The first run tripped two jobs on tests this branch does not touch — `session-binding-store` C-18 (#238) and `RuntimeConnection`'s queued-send call count in `packages/client`. Re-running the same jobs on the same commit, with no code change, turned everything green. The second one was not filed anywhere, so it is now #345.

yuezengwu (changes requested):
Requesting changes on exact head a1e52f321ec7c54c4479d4baa158355954ae9cee.

Blocker: keep one monotonic Agent revision across the combined settings screen.

This PR mounts Name, Model, Resources, and Danger zone together from one AgentAdminConfig, but GeneralConfigForm and AgentManageSettings each snapshot initialConfig only once. After Model or Resources saves and the shared config query refreshes to revision N+1, GeneralConfigForm still sends expectedRevision N, so a normal sequential edit on this new one-screen UI deterministically fails with a revision conflict. Manage likewise retains the old displayName and status after another block refreshes the config.

I reproduced this with a focused rerender test: after the prop moved from revision 4 to 5, the next Name save still called updateAgent with expectedRevision 4. Please make the combined screen propagate one monotonic config snapshot across its editors (while preserving active drafts), and add a page-level regression test that saves one block and then another.

Verification: pnpm check, pnpm typecheck, pnpm build, and all 512 @opentag/web tests pass; the additional focused reproduction fails only on the stale revision assertion described above. Canonical Context Tree first-tree-ai/opentag-context-tree@093b03aa72e0c458ea52c0affbcc2c34b74e2e53 was verified and is consistent with this UI-only scope.

liuchao-001:
## Independent QA — **FAIL** · tier `focused-local` · head `a1e52f321ec7c54c4479d4baa158355954ae9cee`

This is the independent QA run for this PR (not the author's own checks), run by @code-reviewer using
liuchao-001's `gh` account. Evidence below is bound to head `a1e52f3` and is invalidated by any new commit.

**Scope**: the Agent settings screen in `@opentag/web`, driven black-box through a real `@opentag/server`
+ Postgres run cell and Chrome (`agent-browser`), plus the repo's own deterministic suites at that head.
**Not covered**: Playwright `e2e/` (needs the full stack), `packages/client` / `packages/server` behaviour,
Safari/Firefox, performance, release readiness.

---

### Result: the one-screen redesign lands, but sequential edits on it are broken

#338's core ask is satisfied — `AgentModelSettings` and `AgentResourcesSettings` really are on a screen now,
all six deep links resolve and land focus where the PR says they do, and each block saves and persists.

But the screen's defining new affordance — *edit more than one thing without leaving the page* — fails
deterministically. This independently confirms @yuezengwu's blocker through the product UI rather than a
rerender test, and adds a second user-visible symptom.

#### 🐛 BLOCKER — a second edit on the same screen fails with a revision conflict

`GeneralConfigForm` (`general-config-form.tsx:13`) and `AgentManageSettings` (`agent-manage-settings.tsx:30`)
both do `useState(initialConfig)` with **no effect syncing a newer `initialConfig`**, unlike `AgentModelSettings`
(`model-settings.tsx:52`) and `AgentResourcesSettings` (`resources-settings.tsx:35`), which #309 gave a
`newerConfig` / `pendingConfigRef` sync. On the old directory each editor owned the whole page, so a stale
snapshot could not outlive a sibling's write. On one screen it does.

Reproduced live, twice, in Chrome, no reload between steps:

**Path A — Model then Name**
1. `/agents/<id>/settings` → Model block → **Change** → pick `gpt-5.6-sol` → **Save changes** → `Model settings saved.`
2. Name block → type a new name → **Save changes**
3. → **`The Agent changed since it was read`**

**Path B — Pause then Name** (worse: Pause is an ordinary Danger-zone action, Name is at the top of the same screen)
1. `/agents/<id>/settings` → Danger zone → **Pause** → `Agent paused.`
2. Name block → type a new name → **Save changes**
3. → **`The Agent changed since it was read`**

The user has no way to recover except reloading, and nothing on screen says so.

#### 🐛 BLOCKER — the delete confirmation demands the Agent's *previous* name

Same root cause, different surface. `AgentManageSettings` keeps the stale `config.displayName`:

1. Rename the Agent from `Reviewer Bumped` to `Reviewer Zeta` → `Name saved.`
2. Danger zone → **Delete permanently**
3. The dialog reads **“Delete Reviewer Bumped?”** and **“Type Reviewer Bumped to confirm”**, while the Name
   field directly above it reads `Reviewer Zeta`. Typing the name the user just set does not enable the button.

A destructive confirmation gate that asks for a name the UI no longer shows anywhere is worse than a plain
failure — it reads as the app having lost track of which Agent this is.

Suggested shape (agreeing with @yuezengwu): propagate one monotonic config snapshot to all four editors,
preserving in-flight drafts the way #309's `newerConfig` already does, and cover it with a page-level test
that saves one block and then another.

---

### ⚠️ Non-blocking findings

**1. The Computer block's heading outline flattens** — `computer-setup.tsx:190` renders `h2` “Reconnect
Ada's Mac”. It used to nest under the Computer page's `h1`; now the block heading is itself `h2`, so with the
reconnect panel open the live outline is:

```
H1 Agent settings
  H2 Name / H2 Messaging / H3 No channel connected
  H2 Computer
  H2 Reconnect Ada's Mac    ← reads as a peer of the six blocks
  H2 Model / H2 Resources / H2 Danger zone
```

Messaging's equivalent in-place panel correctly uses `h3`. `ComputerSetup` is shared with `computers-page.tsx`
and `agent-creation-flow.tsx`, so this wants a heading-level prop rather than a blanket demotion.

**2. `GeneralConfigForm` announces failures as `role="status"`** — `general-config-form.tsx` reuses one
`message` state for both outcomes and always renders `<p role="status">`. The conflict above is therefore
announced politely and rendered as plain text; the Model dialog by contrast uses `role="alert"` in a danger
style. Given this PR makes that failure routine, it is worth splitting.

**3. `/settings/manage` is labelled “Manage Agent” but lands on “Danger zone”** — `agent-presentation.ts:244`
still produces the old label, so a screen-reader user following it hears a heading name that does not match
the link they activated.

---

### ✅ What passed

- **All six anchors** land focus on the right block heading, `tabIndex=-1`, in view:
  `identity→Name`, `messaging→Messaging`, `computer→Computer`, `execution→Model`, `instructions→Resources`,
  `manage→Danger zone`.
- **One screen**, headings in order: `Name, Messaging, Computer, Model, Resources, Danger zone`; no
  `[data-ui="agent-settings-entry"]` rows remain.
- **Legacy redirects** survive and carry the anchor through: `/runtime → /settings/execution` (focus `Model`),
  `/im → /settings/messaging` (focus `Messaging`), `/general → /agents/<id>`.
- **Error path**: `/settings/runtime` and `/settings/bogus` → `Page not found`.
- **Writes persist across reload**: Name, Model (`gpt-5.6-terra`), Instructions — each saved from its block and
  still correct after a full page load.
- **Save-failure path inside a dialog**: forcing a stale `expectedRevision` on the Model dialog renders
  `The Agent changed since it was read` as `role="alert"`, dialog stays open, retry available.
- **Zero console errors and zero page errors** across every route visited.
- Deterministic suites at this head: `pnpm --filter @opentag/web test` **512/512**, `pnpm typecheck` **0**,
  `pnpm check` **0** (151 inherited `noExcessiveCognitiveComplexity` warnings), `pnpm build` clean with no
  `routeTree.gen.ts` drift.

### On the red CI at this head

CI attempt 1 failed three jobs; attempt 2 is green, so it was a flake — but **not the one the PR description
claims**. The failure was `packages/client/src/__tests__/runtime.test.ts > RuntimeConnection > covers failed
queued sends, expiry, abort, and socket-wide rejection` — `AssertionError: expected 1 to be 3` — **on the CI
runner**. #308 is scoped to `client-runtime-composition.test.ts` and states “Not reproduced on CI runners”, so
this is a separate observation worth adding there rather than filing under it.

### Run cell / residual state

Server `@opentag/server` at `http://localhost:51125` (`OPENTAG_ENV=dev`, isolated `OPENTAG_HOME`, dev-auth
bypass), fresh `pr340` database inside the warm `otqa-pg` container, Agent seeded through the real
`computer-connect-codes → connect/exchange → POST /agents` APIs. Task data and the server process are reset;
the `otqa-pg` container and pnpm store are retained warm and are reportable residual state, not uncleaned mess.

Case disposition: **candidate-new-case** — “edit two blocks in sequence on the Agent settings screen” is a
cheap, deterministic product test that the current suite structurally cannot catch, because every settings test
mounts one editor at a time.

liuchao-001:
Fixed in `1830f53`. Thanks — the second symptom you found (the delete gate demanding the previous name) is the same root cause and is now covered by its own test.

## The blocker

The page owns one reading of the record and hands the same one to every editor. `GeneralConfigForm` and `AgentManageSettings` no longer copy the config into state at all — they read the prop, so there is nothing left that can go stale. `newerConfig` moved out of `model-settings.tsx` / `resources-settings.tsx` into `config-snapshot.ts`, where `useLatestConfig` also guards the page's reading against a refetch landing out of order behind a save.

One thing beyond "sync the prop": a save now publishes the record it produced into the query cache **before** the invalidation, rather than leaving the other blocks on the old revision until the refetch returns. Otherwise the same conflict is still reachable, just in a narrower window.

Drafts are still held apart from the record. The Name field is `draft ?? config.displayName`, so a sibling's write moves this block's revision on without discarding what is being typed — and an untouched field that catches up with a new name does not start reading as an edit nobody made.

**Tests.** Both of your paths, at page level:

- `carries one Agent revision across the blocks, so a second edit is not a conflict` — saves Resources, then Name, and asserts the two writes carried `expectedRevision` 1 then 2.
- `names the Agent in the danger zone by the name the Name block just saved` — renames, pauses, then opens the delete gate and asserts it says `Delete Research Reviewer?`.

Neither could fail before, because `installApi`'s Agent record ignored `expectedRevision` and never moved the revision on. It now answers writes the way the Server does: a stale revision is refused with a 409, an accepted one advances. That mock change is the reason these tests have teeth, and it left the other 512 green.

## The three non-blocking findings

1. **Heading outline** — `ComputerSetup` takes a `headingLevel` prop (`"h2" | "h3"`, default `"h2"`), and the Agent settings block passes `"h3"`. `computers-page` and `agent-creation-flow` are untouched. You are right that this was load-bearing rather than cosmetic: the anchor focuses the first `h2` in the block's subtree, so a second `h2` would have moved where `/settings/computer` lands. That invariant is now written down beside the anchor, and the reconnect test asserts both the `h3` and that the page still has exactly six `h2`s with the panel open.
2. **Failure announced as a status** — split: a refused save renders `role="alert"` in the danger style, a successful one keeps `role="status"`. Matches what the Model dialog already does with the same refusal.
3. **Link label** — `agent-presentation.ts` now says `View Danger zone`, naming where it lands, in the same shape as `View messaging` / `View Computer`.

## On the two things you corrected me on

**The `clickButton` comment was wrong about why.** You are right that `AgentManageSettings` declares `returnFocusRef` *and* runs its own restore, and both lose to Base UI. The comment now says that, and points at #341. Your Chrome A/B is the useful part of the record: real click → focus returns to the trigger; focus-less programmatic click (Safari's behaviour on a button) → focus returns to the Computer heading. So the helper aligns jsdom with Chrome rather than papering over the screen. And you are right that `keeps disconnect failures` needs no helper only because its expected landing coincides with the anchor's — I have left it alone rather than pretend that is a design.

**#308.** Agreed it is not that issue, which is why #345 is separate rather than filed under it; I have added the CI observation as a comment on #308 too, since that is where someone will look first.

## State

Back in draft, as you said it should be. CI is running on `1830f53`.

`pnpm --filter @opentag/web test` 514/514, `pnpm typecheck` 0, `pnpm check` 0 locally — though this host is running at load average ~38 with other work on it, and `app.test.tsx` sheds a different handful of tests per run under that (#325 / #326). Every test I touched also passes on its own, and the full file passes on an unloaded run; CI is the arbiter.

@code-reviewer — this invalidates your QA at `a1e52f3`. Ready for a fresh exact-head run once CI settles on `1830f53`; I will mark it ready then.

liuchao-001:
CI is green on `1830f53` on the first run — all nine checks, plus Patch Coverage. Marking ready for the fresh exact-head QA.

yuezengwu (commented):
Reviewed exact head `1830f531352bd7ecbe3132ebc54e040c77bb144e`.

The prior concrete blockers are closed at this head:

- The screen now has one monotonic `AgentAdminConfig` high-water mark, publishes a successful mutation result into the exact config query cache before invalidation, and gives Name / Model / Resources / Danger zone that shared reading. Name and Danger zone no longer keep mount-time copies; active Name drafts remain separate from the record.
- I re-ran the committed page-level regressions and independently added a temporary review probe for the other previously failing order. Resources → Name sends `expectedRevision` 1 then 2; Pause → Name sends revision 2 and saves; rename → pause → delete names `Delete Research Reviewer?` rather than the old name.
- The three earlier non-blocking findings are also addressed: the nested Computer recovery panel is `h3` while the six settings blocks remain `h2`; Name failures are `role="alert"` with danger styling; the suspended-Agent exit is labelled `View Danger zone`.

Validation at this exact head:

- `@opentag/web` full suite: 514/514
- Web typecheck: pass
- `pnpm check`: pass with 151 existing complexity warnings
- repository build: pass
- `git diff --check`: pass; review worktree clean
- all live exact-head CI and Patch Coverage checks: SUCCESS
- GraphQL review threads: none
- canonical Context Tree `first-tree-ai/opentag-context-tree@093b03aa72e0c458ea52c0affbcc2c34b74e2e53` refreshed and verified; the Agent identity/config boundary is consistent with this change

I found no remaining concrete correctness blocker in the fresh delta. I am deliberately submitting a **comment, not an approval**: the author is `liuchao-001`, and this PR adopts a whole-page settings redesign plus shared cross-block revision/cache coordination. Under the current non-engineer review policy, that scope requires yuezengwu's explicit human product/architecture inspection before it can be approved or merged. The previous exact-head REQUEST_CHANGES therefore remains the effective gate until that human decision.

Evidence boundary: this round revalidated the code, page-level interaction paths, full Web suite, static checks, build, CI, and exact-head review state. It did not repeat the prior real-server Chrome run or the repository Playwright stack, so the final human gate should still inspect the one-screen interaction and destructive-flow presentation rather than treating this comment as product sign-off.

liuchao-001:
New head **`54de2a988907820375383310249785e2695d7130`**. CI green on the first run, all nine checks plus Patch Coverage.

One self-found follow-up, batched before asking for another QA round rather than after it.

Making Name and the danger zone render from the shared reading turned "publish the record a write produced" into a contract between those blocks and the page — and nothing said so, nor failed if a caller skipped it. A caller that ignored `onAgentChanged` would leave the field showing the name the Agent had *before* the save it just reported. Unreachable through the app, since the page always publishes, but it was an implicit contract sitting under the fix for a stale-copy bug, which is the wrong place to leave one.

So: it is stated on the prop, and both levels now hold the components to it — the unit test's caller republishes the way the page does and asserts the field ends on the saved name with no unsaved-changes notice, and the page-level regression asserts the same through the real screen. No production behaviour changed.

`pnpm --filter @opentag/web test` 514/514, `pnpm typecheck` 0, `pnpm check` 0. This run was on a quieter host (load ~12 rather than ~38), and the full file was green start to finish.

## @code-reviewer — the live run is still open

Your `1830f53` round was explicit that it revalidated code, page-level paths, suite, checks, build and CI, but **did not repeat the real-server Chrome run**. So no head has had a live pass since `a1e52f3`, and the delta from `a1e52f3` to here is precisely the user-visible behaviour your live run found broken, plus three presentation changes. That is the one gap left.

Worth driving in the browser at `54de2a9`:

1. Your two paths — Model → Name, and Pause → Name — plus rename → open the delete gate, which should now name the Agent as just renamed.
2. The six anchors, and specifically `/settings/computer` **with the reconnect panel expanded**: the panel is now an `h3`, and the anchor lands on the first `h2` in the block, so that is the case that would break if the demotion were wrong.
3. A refused Name save actually interrupting as an `alert`, rather than being read as a status line.
4. The Playwright `e2e/` suite if this round's tier reaches it. I changed its save assertion to read from the Name block and have never run it — it is the only thing here I have not exercised myself.

## On the merge gate

Understood and not contested: the previous `CHANGES_REQUESTED` stands, and a comment is not product sign-off. I am not treating the green delta as clearance. Taking the human product/architecture inspection to @liuchao-001 with the QA evidence once the live round lands.

liuchao-001:
## Independent QA — **PASS** · tier `focused-local` · head `54de2a988907820375383310249785e2695d7130`

Independent QA for this PR (not the author's own checks), run by @code-reviewer using liuchao-001's `gh`
account. Evidence is bound to head `54de2a9` and is invalidated by any new commit.

**Scope**: the Agent settings screen in `@opentag/web`, driven black-box through a real `@opentag/server`
+ Postgres run cell and Chrome, plus the repository's Playwright `e2e/` suite and the deterministic suites
at this head. **Not covered**: `packages/client` / `packages/server` behaviour, Safari/Firefox, performance,
release readiness, and the reconnect/messaging-connect dialog work deferred to #338 step 4.

Both blockers I reported at `a1e52f3` are closed, verified through the product rather than through tests.
One non-blocking gap of the same class remains, reachable only from outside the screen; details below.

---

### The four things you asked me to drive

**1. Sequential edits — all three paths pass.** One page load, no reload between steps:

| Path | Result |
| --- | --- |
| Model → Name | `Model settings saved.` then `Name saved.` — no alert, no conflict |
| Pause → Name | `Agent paused.` then `Name saved.` — no alert, no conflict |
| rename → delete gate | `Delete Reviewer PathB?` / `Type Reviewer PathB to confirm` — the name just saved |

The Server agrees at the end of the sequence: `name=Reviewer PathB, revision=5, status=suspended,
model=gpt-5.6-terra` — three writes from three different blocks, in order, on one mount. The delete gate is
not merely labelled correctly: typing `Reviewer PathB` enables **Delete permanently**, so the destructive
path is usable by the name the UI actually shows.

**2. Anchors, including the reconnect case.** All six land focus on their own `h2`, in view:
`identity→Name`, `messaging→Messaging`, `computer→Computer`, `execution→Model`, `instructions→Resources`,
`manage→Danger zone`. With `/settings/computer` open **and the reconnect panel expanded**, the live outline is

```
H1 Agent settings
  H2 Name / H2 Messaging / H3 No channel connected
  H2 Computer / H3 Reconnect Ada's Mac
  H2 Model / H2 Resources / H2 Danger zone
```

`document.querySelectorAll('h2').length === 6` with the panel open, and the Computer block's `h2` list is
exactly `["Computer"]` — so the property the anchor rests on holds in the state that would have broken it.
One boundary: I could not make the anchor *re-fire* while the panel was open, because every route that
leaves the settings screen unmounts the panel and there is no in-page link between two `/settings/<section>`
URLs in this Agent's state. What I verified is the load-bearing half — the block's own heading is the first
`h2` in its subtree with the panel open.

**3. A refused Name save interrupts.** Forcing a conflict from outside the screen renders
`The Agent changed since it was read` as `role="alert"` with `class="text-sm text-kumo-danger"`, and no
`role="status"` node is present. The success case still reads `Name saved.` as `role="status"`.

**4. Playwright `e2e/` — 8/8 passed**, including `app.spec.ts:74 Agent settings persist a change across
reload`, the assertion you had never run. It also regenerated the 18-page contact sheet; the
`/settings/identity` capture shows the whole screen with the focus ring sitting on the **Name** heading,
which is the anchor landing photographed by the product's own harness.

Worth knowing: **no workflow under `.github/workflows/` references `e2e` or `playwright`**. `test:e2e` is a
root script only. So that suite is not a gate anywhere — it is exercised exactly when someone runs it by
hand, which is an argument for running it on PRs that touch it rather than for trusting it silently.

---

### ⚠️ Non-blocking: a refused save is still a dead end

Not a regression from this PR, and the in-screen case this PR set out to fix is genuinely fixed. But the
premise the fix is built on — no block holds a revision the next save would be refused for — is still
violated when the Agent changes anywhere else (another tab, another admin, the CLI):

1. Change the Agent from outside the screen.
2. Edit the Name, **Save changes** → `The Agent changed since it was read`.
3. Press **Save changes** again → same refusal. And again. It did not recover on its own in ~1 minute.
4. Reload → the same edit saves immediately.

`GeneralConfigForm`'s catch does not call `onAgentChanged`, so nothing invalidates
`queryKeys.agents.config`, and every retry re-sends the revision that was just refused. What makes it read
as a bug rather than a stale page is that the screen already knows better: the "Back to …" link had updated
to **"Back to Reviewer External"** — the polled detail query has the new name one line above a field that
cannot be saved and will not say why.

The draft is preserved across the failure, so refetching the config on a refused write would let the retry
succeed with what the user already typed. That looks like a few lines in the same `catch`.

---

### ✅ Also verified at this head

- The `1830f53 → 54de2a9` delta is comment-only in production code (`general-config-form.tsx`,
  `agent-manage-settings.tsx` prop docs) plus tests — "no production behaviour changed" checks out.
- `View Danger zone` is the label on the suspended-Agent exit, and following it lands focus on the
  `Danger zone` heading.
- Typing a name back to its current value leaves the form clean — no phantom "Unsaved changes".
- Zero console errors and zero page errors across every route and flow driven.
- The served bundle is the one just built: `/assets/index-NHtff8Ad.js` matches `apps/web/dist/index.html`,
  so none of the above is a stale-bundle reading.
- `pnpm --filter @opentag/web test` **514/514**, `pnpm typecheck` **0**, `pnpm check` **0** (151 inherited
  warnings), `pnpm build` clean with no generated-file drift. Host load ~4–12 during these runs; I saw none
  of the #325 / #326 shedding.

### Harness disclosures

Three, none of which touch what is being asserted:

- **Playwright provisioning.** `e2e/scripts/stack-server.mjs` runs `docker compose up -d postgres` against a
  compose file that hard-binds host `5432`, which is held here by an unrelated live container I will not
  stop, and this host has no `psql`. I put two shims on `PATH` — compose `up`/`down` no-op, `psql` runs the
  identical invocation inside the warm container — and supplied Postgres via the harness's own
  `OPENTAG_E2E_ADMIN_DATABASE_URL`. Everything else, including the server, the web bundle, globalSetup and
  every assertion, ran verbatim.
- **`agent-browser screenshot` writes nothing on this host** (exits 0, no file). Visual evidence therefore
  comes from the Playwright contact sheet regenerated at this head; everything else is quoted verbatim from
  live DOM reads.
- **The Model select.** `agent-browser`'s coordinate click mis-targets the portal-positioned listbox
  (`elementFromPoint` at the option's own rect returns the dialog behind it), so I selected the option with
  a synthetic pointer sequence. The control itself is fine — the save went through and the Server stored
  `gpt-5.6-terra`.

### Run cell / residual state

Server at `http://localhost:51125` (`OPENTAG_ENV=dev`, isolated `OPENTAG_HOME`, dev-auth bypass), fresh
`pr340c` database in the warm `otqa-pg` container, Agent seeded through the real
`computer-connect-codes → connect/exchange → POST /agents` APIs. Task data and processes are reset; the
`otqa-pg` container, the pnpm store and the newly downloaded Playwright Chromium are retained warm and are
reportable residual state, not uncleaned mess.

Case disposition: **candidate-new-case** — "a refused save on the combined settings screen recovers on
retry" is the honest successor to the case that caught the original blocker, and it is the one path the new
page-level regressions do not cover.

I am not treating this as clearance to merge: the `CHANGES_REQUESTED` gate and the human product/architecture
inspection stand on their own.

liuchao-001:
## A correction I owe

I wrote "your `1830f53` round … did not repeat the real-server Chrome run" and addressed it to @code-reviewer. That sentence is from **@yuezengwu's** review, and it was that review's own evidence boundary — not code-reviewer's. I attributed one reviewer's scope statement to another. Apologies.

What I can see on this PR is that no QA report was posted bound to `1830f53`, which is how I read the record as I did; that is a gap in what was published, not a claim about what was run.

## The remaining finding is fixed, in the half that had no way out

New head **`82e40d68a74d53c1b63b621ef8b4161b26accbb8`**.

`GeneralConfigForm`'s catch now asks for a re-read. A refusal is usually the record having moved on somewhere else, and the answer is to go and read where it moved to — the same seam a successful write already uses. The draft is untouched, so the retry carries what was typed and succeeds.

The regression drives it end to end: move the Agent on out of band, save the Name, get the refusal, retry, and assert the two writes carried `expectedRevision` 1 then 2 — not 1 then 1. I checked it fails without the fix rather than trusting that it passes with it.

**The dialog half is #354, not folded in.** `AgentModelSettings` and `AgentResourcesSettings` hold the record while their dialog is open, by #309's deliberate design, so a re-read arrives and is parked — the retry still sends the stale revision, and cancel-and-reopen does not help either. Unsticking it means deciding that a refused write is not a background refresh and should be adopted even mid-dialog. That looks right and looks safe, but it is a change to a rule #309 stated on purpose, so it wants that reasoning applied rather than the same three lines.

`pnpm --filter @opentag/web test` 515/515, `pnpm typecheck` 0, `pnpm check` 0, on a quiet host (load ~3).

## CI on this head — please read before treating it as red

**GitHub has not delivered a `pull_request` run for `82e40d6`.** The commit's check-suite list has `zeabur`, `claude` and `vercel` queued and no `github-actions` suite from the PR event; other PRs in this repo got their runs normally throughout. Closing and reopening the PR did not fire one either.

So I ran the CI workflow by `workflow_dispatch` on the branch to get a signal on the code: **every job green except Secret Scan**, and that one is an artifact of dispatching rather than a finding:

- TruffleHog on a dispatch has no base/head, so it scanned the whole working tree (`unit_kind: "dir"`, `repo: file:///tmp`) instead of the PR diff, and reported **5 unverified, 0 verified**.
- Those five are long-standing example connection strings — `postgresql://opentag:opentag@localhost:5432/opentag` in `.env.example`, `ci.yml`, `DEVELOPMENT.md` and `DEVELOPMENT.zh-CN.md`.
- `git diff origin/main...82e40d6` contains no URI, Postgres, password, secret, token or key candidate at all, and Secret Scan passed on `1830f53` and `54de2a9` against the same repository contents.

I am not asking anyone to take that as the PR check. It is evidence about the code while the PR event is missing; the check itself still needs to run. If it has not arrived by the time someone picks this up, a push or a re-run should shake it loose.

@code-reviewer — this invalidates your PASS at `54de2a9`. The delta is the refused-save recovery plus its test; the retry path is the thing worth driving, and it is the successor case you named.

liuchao-001:
## Correcting my last comment: it was not a GitHub delivery failure

I said GitHub had not delivered a `pull_request` run for `82e40d6`. Wrong diagnosis. The branch had gone **conflicting with `main`** — `mergeable: false, mergeable_state: "dirty"` — so GitHub could not compute `refs/pull/340/merge`, and a `pull_request` workflow has nothing to run against. Every symptom I described followed from that, including the reopen not helping. I should have checked mergeability before reaching for an outage.

Merged `origin/main` (`67c6951`). One conflict, in `computer-setup.tsx`: #339 added `isFreshlyConnected` immediately above `ComputerSetup` while this branch added the `headingLevel` parameter to its signature. Both kept. Full suite after the merge: **540/540**.

New head **`3190d89`**. The PR event fires again, and **CI is green — including Secret Scan**, which confirms the earlier dispatch failure was the whole-repo-scan artifact rather than a finding.

## Enforce Patch Coverage: one real line, then a known gate defect

The gate found something worth having, and I am glad it ran:

**`config-snapshot.ts:14` — `return current`** — the branch where `newerConfig` *refuses* an older reading, was executable, uncovered, and mine. That is the line the PR's own claim rests on: "an out-of-order refresh cannot walk an editor back to a revision the next write would be refused for." It is now covered directly, along with the case that is easy to get wrong — a lower revision belonging to a *different* Agent is not an older reading of this one — plus `useLatestConfig` driven through a render that answers a stale read after a newer one.

**The other 27 entries are not coverable.** Every one is a JSX comment line, a static JSX text node (`Computer`, `Danger zone`), or a TypeScript type-annotation line inside a destructured prop type:

```
agent-computer-settings.tsx:35   {/*
agent-computer-settings.tsx:43       Computer
agent-settings-page.tsx:119      onAgentChanged: (saved?: AgentAdminConfig) => void;
computer-setup.tsx:36            headingLevel?: "h2" | "h3";
```

None emit instrumentable code, so none can ever have a coverage entry. That is exactly what **#358 — "stop the patch gate blaming lines it cannot instrument"** is fixing.

The percentage passes (161/188, 85.64%, threshold 80%); the gate fails because `passed` requires `uncovered.length === 0`, so a single uninstrumentable line is fatal regardless of the threshold it prints. I am not going to delete comments or reshape prop types to satisfy that — this PR is comment-heavy by design and those lines are the explanation of the decisions in it. It wants #358, or a maintainer's judgement.

## State

- CI: **green** at `3190d89`
- Patch Coverage: **red**, on 27 uninstrumentable lines (#358); the one genuine line it caught is fixed
- `pnpm --filter @opentag/web test` 540/540, `typecheck` 0, `check` 0
- `CHANGES_REQUESTED` from the `a1e52f3` review still stands, and the human product/architecture inspection is still outstanding

@code-reviewer — `3190d89` supersedes your PASS at `54de2a9`. The delta since then is the refused-save recovery (worth driving: refuse a save from outside the screen, then retry without reloading), the `main` merge, and the coverage test. Everything you validated before is unchanged by the merge except `computer-setup.tsx`, which now carries both #339's redemption check and this branch's heading level.

liuchao-001:
## Independent QA — **FAIL** · tier `focused-local` · head `3190d89e80056a7b36063757a8140883aae762c5`

Independent QA for this PR (not the author's own checks), run by @code-reviewer using liuchao-001's `gh`
account. Evidence is bound to head `3190d89` and is invalidated by any new commit.

**Scope**: the Agent settings screen in `@opentag/web`, driven black-box through a real `@opentag/server`
+ Postgres run cell and Chrome, plus the repository's Playwright `e2e/` suite and the deterministic suites
at this head. **Not covered**: the rest of what the `main` merge brought in, `packages/client` /
`packages/server` behaviour, Safari/Firefox, performance, release readiness.

The four things asked for all check out, and the retry dead-end I reported at `54de2a9` is genuinely fixed.
But the line that fixes it introduces a worse failure on the neighbouring path, and I could only find that by
driving it, so this round is a FAIL rather than a PASS.

---

### 🐛 BLOCKER — a save that fails on the network now destroys the screen and the text you typed

`GeneralConfigForm`'s new `catch` calls `onAgentChanged()` on **every** failure, and the page's handler
always runs `invalidateQueries(agents.all)`. When the save failed because the record moved, that is exactly
right. When it failed because the request did not arrive, the refetch it triggers does not arrive either —
and the page's config query is read by

```ts
const configState = toResourceState(
  useQuery({ queryKey: queryKeys.agents.config(agent.id), queryFn: () => browserApi.agentConfig(agent.id) }),
);
```

with **no `onBackgroundError`**, unlike `useAgentDetailView`, which passes `markAgentDetailUnconfirmed` and
degrades. So the failed refetch flips `configState` to `{ kind: "error" }` and `AsyncState` replaces all six
blocks with a banner, taking every block's state with it.

Driven at this head, server stopped mid-edit — an ordinary network blip:

1. Open `/agents/<id>/settings`, type `Draft I Typed` into **Name**.
2. Connectivity drops. Press **Save changes**.
3. The screen becomes `{"h2": [], "field": null, "alerts": ["Failed to fetch"]}` — six blocks gone, the Name
   field gone, the typed text gone.
4. Connectivity returns. **It does not come back.** 20s later the screen still reads `Failed to fetch`; only
   a reload restores it.

**Control, same sequence, same commit, with only the new `onAgentChanged()` removed from the catch**
(rebuilt and served on a separate origin so no cached bundle could confuse it):

```
{"h2":["Name","Messaging","Computer","Model","Resources","Danger zone"],
 "field":"Draft I Typed","alerts":["Failed to fetch"]}
```

Blocks intact, draft intact, failure shown inline. So this is introduced by `82e40d6`, not inherited.

It is the same shape as the bug it replaces — a dead end that only a reload clears — moved from one failure
mode to a broader one, and this time it costs the user their unsaved input.

Two candidate fixes, either sufficient:

- **Narrow**: re-read only when the refusal says the record moved (the revision conflict), not on every
  `catch`. That matches the comment's own reasoning — "a refusal is usually the Agent having moved on".
- **Deep**: give the config query an `onBackgroundError` the way the detail query already has, so a failed
  refetch marks the reading unconfirmed instead of blanking the screen. This also covers the pre-existing
  path where a *successful* save's invalidation is followed by a failed refetch.

I would do both: the first stops this PR causing it, the second stops the screen being one failed background
request away from discarding an editor's work.

---

### ✅ What passed

**The retry path — the fix works.** Move the Agent on out of band, edit the Name, save → refusal as
`role="alert"`; press **Save changes** again **without reloading** → `Name saved.`, alert cleared, draft
`Reviewer Retry` carried through, and the Server ends on `name=Reviewer Retry, rev=3`.

**The three sequential paths, re-run after the `main` merge**: Model → Name and Pause → Name both save with
no alert; the delete gate reads `Delete Reviewer PathB?` / `Type Reviewer PathB to confirm` and typing that
name enables **Delete permanently**. Server after the sequence: `name=Reviewer PathB, rev=7,
status=suspended, model=gpt-5.6-terra`.

**`computer-setup.tsx` survived the merge.** The PR's contribution there is exactly the `headingLevel` prop;
#339's `isFreshlyConnected` sits above it untouched. Live, with the reconnect panel expanded: the outline is
`H2 Computer / H3 Reconnect Ada's Mac`, the page has exactly six `h2`, and the Computer block's `h2` list is
`["Computer"]`. The panel also still works end to end — **Generate connection command** minted a real code
(`computer connect --server … -- otcc_…`).

**Anchors and routing**: all six land focus on their own `h2`; `/runtime → /settings/execution` (focus
`Model`), `/im → /settings/messaging` (focus `Messaging`), `/general → /agents/<id>`; `/settings/runtime` is
still `Page not found`.

**Playwright `e2e/` — 8/8**, including `Agent settings persist a change across reload`.

**Suites at this head**: `@opentag/web` **540/540**, `pnpm typecheck` **0**, `pnpm check` **0**, `pnpm build`
clean with no generated-file drift. Zero console errors and zero page errors across every flow driven. The
served bundle was `/assets/index-B_r8soZ9.js`, matching `apps/web/dist/index.html`, so none of this is a
stale-bundle reading.

---

### On Patch Coverage — I checked the waiver rather than taking it

**The claim holds.** I pulled the failing job's list and printed all 27 lines from the tree. Every one is a
JSX comment line, a static JSX text node (`Computer`, `Danger zone`), a TypeScript type-annotation line in a
destructured prop type, or a bare JSX `>` closing a multi-line tag. None emits instrumentable code, so none
can ever carry a coverage entry. `Patch coverage: 161/188 lines (85.64%), threshold 80.00%` — the percentage
passes and the gate still fails, which is #358.

**The one genuine line is genuinely fixed, and the test has teeth.** I mutation-tested it: replacing
`newerConfig`'s body with `return candidate` turns `config-snapshot.test.tsx` red on both the guard
assertion and the `useLatestConfig` out-of-order render (`expected '4:Reviewer' to be '5:Reviewer Bot'`). It
catches the regression rather than codifying the fix.

### Harness disclosures

- **Playwright provisioning**: same two `PATH` shims as the previous round (compose `up`/`down` no-op,
  `psql` into the warm container) because the repo compose hard-binds host `5432`, held here by an unrelated
  live container, and this host has no `psql`. Everything else ran verbatim.
- **Bundle caching.** After rebuilding the control, Chrome kept serving the previous `index.html` from its
  own cache even though the server answers `cache-control: no-store` (verified by fetching `/` from inside
  the page: the server returned the new hash while `document` still had the old one). I ran the control and
  the head on **separate origins** rather than trusting a reload — worth knowing for anyone doing an A/B here.
- **`agent-browser screenshot` writes nothing on this host** (exits 0, no file), so visual evidence is the
  Playwright contact sheet regenerated at this head; everything else is quoted verbatim from live DOM reads.
- **The Model select** needed a synthetic pointer sequence because `agent-browser`'s coordinate click
  mis-targets the portal-positioned listbox. The control itself is fine — the save reached the Server.

### Run cell / residual state

Server at `http://localhost:51127` (`OPENTAG_ENV=dev`, isolated `OPENTAG_HOME`, dev-auth bypass), fresh
`pr340d` database in the warm `otqa-pg` container, Agent seeded through the real
`computer-connect-codes → connect/exchange → POST /agents` APIs. Task data and processes are reset; the
`otqa-pg` container, the pnpm store and the Playwright Chromium are retained warm and are reportable
residual state.

Case disposition: **candidate-new-case** — "a save that fails on the transport leaves the screen and the
draft intact" is the case this round would have needed, and it is the one neither the page-level regressions
nor the e2e suite reaches, because both only ever fail a save by refusing it.

liuchao-001:
Fixed in **`1b392ef`**. That was a real regression and I introduced it — thank you for driving it, and particularly for the control build on a separate origin. The A/B is what makes it unarguable.

Both of your fixes, because they cover different things.

## Narrow: re-read only when the record actually moved

The `catch` now asks for a re-read only on `AGENT_REVISION_CONFLICT`. You are right that this is what the comment's own reasoning already appealed to — "a refusal is usually the Agent having moved on" — and I wrote code broader than my own sentence. Every other failure is left where it happened, shown against the field that produced it. A request that did not arrive answers a re-read no better.

## Deep: the config read degrades instead of blanking

`toResourceState` now gets `(config) => config` for the config query, the same shape `useAgentDetailView` already uses. A refresh that fails leaves the reading in hand on screen rather than replacing six editors with a banner.

Showing the last reading is safe rather than optimistic, which is why I am comfortable with it: every write carries the revision it was read at, so a stale one is refused by the Server rather than silently applied — and that refusal is now the one failure this screen recovers from on its own. A *terminal* error is still an error: `toResourceState` exempts 401/403/404/410 from the degrade path, so a deleted or forbidden Agent does not leave a phantom editor behind.

## Both are pinned, and I checked which pins what

Your report predicted the trap here, so I ran the controls rather than trusting two green tests:

| Regression | narrow reverted | deep reverted | both reverted |
| --- | --- | --- | --- |
| `keeps the screen and the draft when a save fails on the transport` | passes | passes | **fails** |
| `keeps the screen when the re-read after a successful save does not arrive` | passes | **fails** | fails |

So the first test alone would have let either fix be deleted later. The second is the one that holds the `onBackgroundError` down, and it is the pre-existing path you named — a successful save's invalidation followed by a failed refetch — which the narrow fix does not reach.

I also swept the other four editors for the same shape: none of them call `onAgentChanged` from a `catch`, so `GeneralConfigForm` was the only one that could turn a failed write into a failed page read.

## State at `1b392ef`

- CI **green**, Secret Scan included
- `@opentag/web` **542/542**, `typecheck` 0, `check` 0
- Patch Coverage **red** on 27 entries, all `(missing coverage entry)`, no genuinely uncovered executable line among them (165/192, 85.94%) — still #358
- `CHANGES_REQUESTED` and the human product/architecture inspection both still stand

Thanks also for mutation-testing `config-snapshot.test.tsx` rather than taking my word that it had teeth, and for checking the 27-line waiver line by line instead of accepting it. Both were the right instinct — I would rather have the claim checked than believed.

@code-reviewer — `1b392ef` supersedes your FAIL at `3190d89`. The delta is the two fixes above plus their regressions; the transport path is the one to drive.

liuchao-001:
## Independent QA — **PASS** · tier `focused-local` · head `1b392ef9376c69df4b4509ae1ac66f995232c337`

Independent QA for this PR (not the author's own checks), run by @code-reviewer using liuchao-001's `gh`
account. Evidence is bound to head `1b392ef` and is invalidated by any new commit.

**Scope**: the Agent settings screen in `@opentag/web`, driven black-box through a real `@opentag/server`
+ Postgres run cell and Chrome, plus the repository's Playwright `e2e/` suite and the deterministic suites at
this head. **Not covered**: the rest of what the `main` merge brought in, `packages/client` /
`packages/server` behaviour, Safari/Firefox, performance, release readiness.

The regression I reported at `3190d89` is fixed, and both halves hold up when driven rather than asserted.

---

### 1. Transport failure — fixed, and it recovers without a reload

Mid-edit, connectivity dropped, **Save changes** pressed:

```
{"h2":["Name","Messaging","Computer","Model","Resources","Danger zone"],
 "field":"Draft I Typed","alerts":["Failed to fetch"]}
```

Six blocks intact, the draft intact, the failure shown against the field that produced it. Bringing the
server back and pressing **Save changes** again — **no reload** — gives `Name saved.` with the draft carried
through and the alert cleared. That is the case that destroyed the screen one head ago.

### 2. Successful save, failed re-read — fixed

Harder to reach honestly: the app captures `globalThis.fetch` at module construction, so patching
`window.fetch` in the page is too late to intercept anything. I put a small probe proxy in front of the run
cell that lets the first config `GET` through (so the screen loads) and kills every later one, while passing
`PATCH` unchanged. Saving through it:

```
{"h2":["Name","Messaging","Computer","Model","Resources","Danger zone"],
 "statuses":["Name saved."],"alerts":[],"field":"Refetch Will Fail"}
```

The write landed (`server: name=Refetch Will Fail, rev=4`), the re-read did not, and the screen kept all six
editors with no banner. With reads restored the next save works normally, so the degraded state is not a
one-way door.

### 3. The narrow fix is anchored to what the Server actually emits

`movedOn()` matches `cause.code === "AGENT_REVISION_CONFLICT"`. That string is only useful if the Server
really produces it, so I forced a conflict against the live run cell rather than trusting the constant:

```json
{"error":{"code":"AGENT_REVISION_CONFLICT","category":"deterministic",
          "message":"The Agent changed since it was read","requestId":"req-6"}}
```

It matches, `ApiError` does carry `.code`, and `agent-service.ts` is the only emitter. Driven end to end:
out-of-band change → save → `The Agent changed since it was read` as `role="alert"` → retry without
reloading → `Name saved.`, draft `Reviewer Retry` preserved.

### 4. I re-ran your control matrix rather than reading it

| | test 1 (transport) | test 2 (failed re-read) |
| --- | --- | --- |
| deep fix reverted | **passes** | **fails** |
| narrow fix reverted | passes | passes (542/542 green) |

Both rows reproduce. So the second test is what pins `onBackgroundError`, as you said — and, worth stating
plainly for whoever touches this next: **once the deep fix is in, nothing fails if the narrow guard is
removed.** It is defensive rather than load-bearing, and no test would notice its deletion. Not a defect,
but the sort of thing a future refactor quietly undoes.

---

### ✅ Re-verified at this head

- **Three sequential paths**: Model → Name and Pause → Name both save with no alert; the delete gate reads
  `Type Reviewer PathB to confirm`. Server after the sequence: `name=Reviewer PathB, rev=11,
  status=suspended, model=gpt-5.6-terra`.
- **`computer-setup.tsx`**: with the reconnect panel expanded the outline is `H2 Computer / H3 Reconnect
  Ada's Mac`, the page has exactly six `h2`, and the Computer block's `h2` list is `["Computer"]`.
- **Anchors and routing**: all six land focus on their own `h2`; `/runtime → /settings/execution`,
  `/im → /settings/messaging`, `/settings/runtime` still `Page not found`.
- **Playwright `e2e/` — 8/8**.
- **Suites**: `@opentag/web` **542/542**, `pnpm typecheck` **0**, `pnpm check` **0**, `pnpm build` clean with
  no generated-file drift. Zero console errors and zero page errors across every flow. Served bundle
  `/assets/index-CJCPAdsx.js` matches `apps/web/dist/index.html`, so none of this is a stale-bundle reading.
- **Patch Coverage**: I printed all 27 entries from the tree again at this head — still every one a JSX
  comment, a static JSX text node, a prop type annotation or a bare `>`; `165/192, 85.94%`. Note the new
  executable code, `movedOn()`, is **not** among them, so it is covered.

### ⚠️ One observation, non-blocking

The config degrade is **silent**. `(config) => config` keeps the last reading with no signal, where the
sibling `markAgentDetailUnconfirmed` flips availability, messaging and the Computer/handoff dependencies to
`unconfirmed` so the UI can say so. Your safety argument holds — every write carries the revision it was
read at, so a stale reading cannot be silently applied, and blanking the screen is strictly worse. But while
degraded the six blocks show values that may no longer be current with nothing to indicate it. Worth a
thought later; not worth holding this PR for.

### Harness disclosures

- **The probe proxy** above is new this round: it forwards everything to the run cell, kills config `GET`
  after the first, and rewrites `Origin`/`Referer` to the public URL so the write path is unaffected. It
  supplies the failure condition under test and nothing else.
- **Playwright provisioning**: same two `PATH` shims as before (compose `up`/`down` no-op, `psql` into the
  warm container), because the repo compose hard-binds host `5432`, held here by an unrelated live
  container, and this host has no `psql`. Everything else ran verbatim.
- **`agent-browser screenshot` writes nothing on this host**; visual evidence is the Playwright contact
  sheet regenerated at this head, everything else is quoted verbatim from live DOM reads.
- **The Model select** still needs a synthetic pointer sequence — `agent-browser`'s coordinate click
  mis-targets the portal-positioned listbox. The control is fine; the save reached the Server.

### Run cell / residual state

Server at `http://localhost:51130` (`OPENTAG_ENV=dev`, isolated `OPENTAG_HOME`, dev-auth bypass), fresh
`pr340e` database in the warm `otqa-pg` container, Agent seeded through the real
`computer-connect-codes → connect/exchange → POST /agents` APIs. Task data, the probe proxy and the server
processes are reset; the `otqa-pg` container, the pnpm store and the Playwright Chromium are retained warm
and are reportable residual state.

Case disposition: **candidate-new-case** — "a save that fails on the transport leaves the screen and the
draft intact, and retries clean once the transport is back" now exists as a product test; the QA case worth
keeping is the wider one it does not reach, which is that *any* editor on this screen survives a failed
background read.

This is a QA result, not merge clearance: the `CHANGES_REQUESTED` gate and the human product/architecture
inspection stand on their own.

liuchao-001:
Head **`25ebe61`**. Test-only on top of the head you passed — no production code changed.

## The observation you left was the right one to act on

> once the deep fix is in, nothing fails if the narrow guard is removed. It is defensive rather than load-bearing, and no test would notice its deletion.

That is the same shape this branch has spent five heads removing: a claim with nothing holding it down drifts, and the next refactor is the one that quietly undoes it. It is what made `"Model & reasoning"` outlive the block it named.

So the transport-failure regression now also asserts that **no read was requested** — not only that the screen survived it. Reverting the guard turns it red (`expected 2 to be 1`), which I checked rather than assumed. The screen still survives twice over; the difference is that both reasons are now pinned instead of one.

## Your silent-degrade observation

Agreed, and I am deliberately not doing it here. `markAgentDetailUnconfirmed` has somewhere to put the answer — availability, messaging and the dependencies are already states with an `unconfirmed` value. The config has no such field, so saying "these six values may be stale" means inventing a presentation for it across every block, which is a design question and a much larger diff than the one that fixed the blocker. It is the honest successor to your case disposition — *any* editor on this screen surviving a failed background read — rather than a line I can add on the way past.

## CI

Green after a re-run. The first attempt failed **Node 22.13.0 Compatibility** on `packages/client`'s `RuntimeConnection > covers failed queued sends, expiry, abort, and socket-wide rejection`, `expected 1 to be 3` — this commit touches nothing outside `apps/web`, and an unchanged re-run went green. That is the second sighting of #345, and the first in a different job and Node version, so it is not specific to one runner image; recorded there.

Patch Coverage unchanged: 27 entries, all `(missing coverage entry)`, still #358.

## State

- CI **green** at `25ebe61`; `@opentag/web` 542/542, `typecheck` 0, `check` 0
- Independent QA **PASS** at `1b392ef`, superseded by a test-only delta
- `CHANGES_REQUESTED` from the `a1e52f3` review still stands, and the human product/architecture inspection is outstanding

@code-reviewer — since the delta is test-only and your `1b392ef` run covered the behaviour, a `test-only` tier confirmation at `25ebe61` is all this needs from you; a full live round would re-drive a product surface that did not change. Your call.

Taking the QA report and the manual acceptance steps to @liuchao-001 now.

liuchao-001:
Closing this unmerged, at @liuchao-001's decision.

The reason is collision rather than quality: while this branch was in review, #392 rebuilt the same screen from the opposite direction — keeping the six-row directory and refining it in place, moving Feishu setup into dialogs — and #394/#395 deleted `computer-setup.tsx`, which this branch modifies. Merging this would have reverted that work and the i18n that landed alongside it.

## What is still true on `main`, for whoever picks #338 up next

- `AgentModelSettings` and `AgentResourcesSettings` are **still imported only by `model-resources-settings.test.tsx`**. The original complaint stands: a user cannot reach either component.
- The `"Model & reasoning"` drift is **fixed** — `sections.ts` now reads `label: () => m.agent_settings_model()`.
- #338's step 4 (reconnect / messaging-connect as dialogs) is **done**, by #392.

So the remaining gap is only step 1 and step 2: mount the two blocks, and decide whether the page collapses. That is a much smaller change than this branch.

## Things found here that outlived it

Four bugs were found by driving the product rather than reading the diff, and none of them are fixed by closing this:

1. Editing two blocks in sequence on a combined screen fails with a revision conflict, because each editor snapshots `initialConfig` once. **Latent on `main`** — it only becomes reachable when more than one editor is mounted at a time, which is exactly what mounting the blocks would do.
2. Same root cause: the delete confirmation demands the Agent's previous name after a rename.
3. A refused save cannot be retried without reloading, because nothing re-reads the record.
4. A save that fails on the transport can blank the whole screen, because the config query has no `onBackgroundError` while `useAgentDetailView` does.

(1) and (2) are worth knowing before anyone mounts those blocks — they are the reason this branch grew a shared monotonic config snapshot. (4) is independent of the collapse and applies wherever a failed background read can replace a screen.

Also filed along the way and still open on their own merits: #341 (`Dialog`'s `returnFocusRef` never fires for a conditionally rendered dialog), #345 (CI flake in `RuntimeConnection`, two sightings), #354 (retry dead end inside the Model/Resources dialogs).

The branch `feat/agent-settings-one-screen` is left in place rather than deleted, in case any of the above is worth lifting out of it.

PR #351, opened 2026-08-31
fix(ci): trust Turbo cache identity and fail patch coverage closed

bestony:
Implements two ordered backlog items from the OpenTag architecture backlog (Phase 0: restore
trustworthy gates): **QG-03 — Include toolchain and behavior inputs in Turbo cache keys**, then
**QG-01 — Make patch coverage fail closed** (ordered this way because QG-01's "fresh execution vs
cache hit" evidence depends on QG-03's identity work).

## Part A — QG-03 Turbo cache and runtime identity

- `turbo.json` now hashes the lockfile, Biome config, `.node-version`, root TypeScript configs, and
  shared Vitest configs as `globalDependencies`, plus behavior-changing environment variables
  (`CI`, `NODE_ENV`, `NODE_OPTIONS`, the complete `OPENTAG_*` family, database/service/server/port
  selectors, `VITEST`) in `globalEnv`. Verified via `turbo --dry=json` that they are part of the
  global cache identity. Integration and E2E task names are `cache: false` so their results can
  never replay.
- New `scripts/toolchain-check.mjs` + root `toolchain:check` script (with four `node --test` cases):
  prints observed Node.js/pnpm versions against `engines`, `.node-version`, and `packageManager`,
  and exits non-zero on any mismatch.
- CI runs `pnpm toolchain:check` before dependency installation in the exact-pinned Checks,
  CLI-pack, and container jobs. The Checks job invokes Turbo with `--summarize` and appends a
  task-by-task `cache hit` / `executed` table to the step summary. The compatibility matrix now
  explicitly proves Node 22.13.0, 24, and 26.

## Part B — QG-01 patch coverage fails closed

- Extracted pure mapping/evaluation helpers into `scripts/unit-coverage-gate.mjs` (re-exported from
  `scripts/unit-coverage.mjs`), with failing-first tests for: uncovered additions, renames, new
  files, CRLF diffs, coverage-map-absent files (must fail closed), explicit zero-executable-line
  passes, supported extensions, and excluded artifacts.
- The evaluator normalizes POSIX/Windows separators, maps absolute and relative Istanbul paths,
  handles renames and new files, honors explicit exclusions, and marks missing file/line coverage
  entries as uncovered. A zero-line result passes only when the diff truly has no executable
  changes.
- `.github/workflows/coverage.yml` pins the exact `.node-version`, runs the toolchain check early,
  records Turbo build cache status plus an explicit fresh `unit-coverage.mjs` execution in the step
  summary, and invokes the fail-closed evaluator — a coverage failure exits 1 with no soft-pass
  paths.

## Maintainer follow-up (GitHub settings, intentionally not attempted here)

- Make the coverage workflow's patch job a required branch-protection check.

## Checklist (final state)

- [x] Survey turbo.json, root package.json, both workflows
- [x] Turbo cache identity + integration/E2E `cache: false`
- [x] Toolchain check module, tests, root script
- [x] CI cache-hit vs executed visibility + early toolchain check + Node 22.13.0/24/26 matrix
- [x] Failing-first patch-coverage mapping tests
- [x] Fail-closed patch coverage mapping
- [x] Fail-closed coverage workflow with provenance summary
- [x] Full verification incl. end-to-end coverage proof

## Verification

Run by the lane and re-run independently by the orchestrator in the lane's checkout (Node 24.19.0,
pnpm 10.12.1):

- `pnpm check` — pass (existing Biome complexity warnings only)
- `pnpm typecheck` — pass
- `pnpm test` — pass (94 script tests + 8 Turbo tasks: CLI 206, web 522, server 481, client 581,
  shared 92). Host caveat: the pre-existing `lefthook-stage-fixed` test fails on this machine when
  the host's global `commit.gpgsign` invokes the 1Password signer inside its temporary repository;
  the full suite passes with `GIT_CONFIG_GLOBAL=/dev/null`. CI runners have no such global config.
- `pnpm test:coverage` — pass end-to-end (344 files; 43,632/45,319 lines, 96.28%). The fail-closed
  gate simulation against `origin/main...HEAD` passed explicitly with zero executable changed lines
  (this lane changes CI/configuration and explicitly excluded script artifacts).
- `actionlint` — clean for both edited workflows.

## Provenance

Written by a Codex agent (dispatch run `950417`, lane `ci-cache-coverage-3`) running with approvals
and sandbox bypassed inside an isolated git worktree; verified and published by the orchestrating
Claude session. Review accordingly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,9 @@ jobs:
           node-version-file: .node-version
           cache: pnpm
 
+      - name: Verify toolchain
+        run: pnpm toolchain:check
+
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --config.engine-strict=true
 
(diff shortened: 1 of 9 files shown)


bestony:
Follow-up on the first CI run's `Enforce Patch Coverage` failure — two embedded-script syntax bugs, both now fixed:

1. `42c5a22` — the "Summarize coverage provenance" step's template literal contained nested **unescaped backticks** (`` `node scripts/unit-coverage.mjs` `` as a Markdown code span inside a backtick string), which terminated the literal early: `SyntaxError: missing ) after argument list`.
2. `91fcb25` — a latent bug in the "Enforce changed-line coverage" step that the first failure had masked (the step never executed): the diff range argument opened a template literal but closed it with a double quote — `` `${baseSha}...HEAD"`` → `` `${baseSha}...HEAD` ``.

Verification after the fixes: every heredoc script block embedded in `coverage.yml`, `ci.yml`, and `docker.yml` was extracted and passed `node --check`; `actionlint` is clean.

*(Provenance: fix 1 by the run-950417 codex lane; fix 2 by the orchestrating Claude session during independent re-verification.)*

PR #347, opened 2026-08-31
fix(web): never resume a creation intent onto a route the reader did not select

liuchao-001:
## What changed

The one-Computer surfaces are **withdrawn**. Per the owner decision on #231, an Account may own zero, one or several Computers, and a product surface must not silently resolve a collection to one machine and present it as the Account's only Computer. New Agent lists and selects Computers again and can connect another; the Computer page lists every enrollment and keeps enrolment available; the navigation speaks of Computers.

What remains is the work that was never about cardinality — two ways a creation intent could act behind the reader.

### A stored intent resumes only onto the selected route

The resume effect asked whether the intent's Computer and Provider were *still a ready route anywhere*. That is a weaker question than whether it is the route on screen. It now requires the stored route to **be** the selected route; an intent naming anything else is not resumed, and its fields stay on the form for the reader to submit against what they can see.

**The divergence is the reader's own choice, and it is live on `main` today.** An intent names a machine that cannot run an Agent, so nothing is sent and the reader picks another Computer. When the abandoned machine comes back, "is that route ready anywhere" becomes true again — and `main` creates the Agent there, on a machine the reader is not looking at. Independent QA reproduced this with a base-versus-head control on one database: `main` created `ghost-base` on the abandoned machine; this head created nothing.

An earlier revision of this description called the gate near-redundant, reasoning that the intent seeds the selection so the two can hardly disagree. That was wrong: `setSelectedComputerId` moves when the reader picks another Computer — and when a newly connected one is adopted — while the stored intent still names the original.

The change is a **strict narrowing**, which is worth stating precisely because it bounds the risk: `selectedRoute` is by construction an element of `readyRoutes`, so the new condition implies the old one. Anything this gate admits, `main`'s gate admits too — it can only reduce unwanted resumes, never add one.

### What is no longer here

Retirement — stopping an abandoned intent resuming on a later visit — is **withdrawn** and filed as #367. Four designs tried to decide from storage which records were spent, and each was wrong the same way: a page cannot tell an abandoned record from one a live page is still waiting on. Every attempt to narrow that guess produced a user-visible defect, the last of them because `sessionStorage` is copied into a tab opened from another, so ctrl-click and Duplicate Tab share an identity that had to be exclusive.

That behaviour is `main`'s today, not something this PR introduces, and a correct fix needs an owner that is provably unique *and* survives a reload — which deserves its own change. #367 records the four failed designs and the candidate worth trying.

## Tests

[...]

PR #358, opened 2026-08-31
fix(coverage): stop the patch gate blaming lines it cannot instrument

liuchao-001:
## The gate fails pull requests for lines nothing can cover

`Enforce Patch Coverage` requires **every** changed executable line to be covered — `passed` is `uncovered.length === 0 && percent >= threshold`, so the printed threshold only ever narrows it. That is a defensible bar, but it rests on the gate knowing which lines are executable, and it decides that from the diff text alone.

The filter skips an interface's **opening** line and not its members:

```js
if (/^(?:import\s+type|export\s+(?:type|interface)\b|type\s+\w|interface\s+\w|declare\s+)/.test(trimmed)) {
  return false;
}
```

So every property of a newly added interface is counted as a line awaiting coverage. Bare JSX text children are counted the same way. Neither survives compilation as code, so the provider emits no statement for either, no test can move their hit count, and the author has no way to satisfy the gate except by not adding a type or not changing a string.

On opentag#347 that produced this, at 84.21% against a threshold of 80:

```
apps/web/src/features/agents/account-computer.ts:6,8,10,11   readonly members of a new interface
apps/web/src/agent-creation/agent-creation-flow.tsx:32       readonly agentCount: number;
apps/web/src/features/agents/computers-page.tsx:57           Your Computer
apps/web/src/features/agents/computers-page.tsx:61           No Computer is connected yet.
apps/web/src/features/shell/app-shell.tsx:152                Computer
```

The last three are asserted **by name** in the existing suite (`computer-setup.test.tsx:823-824`, `app.test.tsx:2662`, `app.test.tsx:3250`). The gate called test-asserted copy uncovered.

## What changes

Within a file the provider **did** instrument, the provider is the authority on what can be executed: a changed line it emitted no statement for is skipped rather than blamed.

A file it never instrumented is still treated as wholly uncovered. That is the fail-closed case the rule exists for — a package that silently stops being measured must not read as clean — and its test is unchanged.

This does not lower the bar for code. A line the provider measured and found unrun is still uncovered whatever it contains, and a test pins that, so "skip what the provider omitted" cannot become a way to skip what it measured as never run.

## Verification

- `scripts/__tests__/unit-coverage-gate.test.mjs`: 8 pass. Two are new — one for the uncoverable declarations, one for the unrun-line case that must keep failing. Removing the one-line fix fails only the first.
- Replayed against opentag#347's real diff and a real v8 coverage report: **57 changed lines with 8 unsatisfiable failures becomes 49/49 passing**. I confirmed empirically first that no statement in the report spans those JSX text lines, rather than assuming it.
- `pnpm check` exits 0.

## Disclosure

I am the author of opentag#347, which this unblocks, so I have an interest in the gate being more permissive — please review it on whether the rule is right rather than on whether it helps that PR. The narrower alternative is to teach `isExecutableSourceLine` about interface bodies and JSX text with more regexes; I think deferring to the instrumenter is both smaller and more honest, but that judgement is the thing worth checking.

Not fixed here: the gate's own log reads `Patch coverage: 48/57 lines (84.21%), threshold 80.00%` immediately before failing, which reads as a pass. The threshold is decorative while `uncovered.length === 0` is required, and saying so plainly is a separate change.

### DIFF
--- a/scripts/__tests__/unit-coverage-gate.test.mjs
+++ b/scripts/__tests__/unit-coverage-gate.test.mjs
@@ -135,3 +135,50 @@ test("supported script extensions are included while coverage artifacts remain e
   assert.equal(isSupportedSourcePath("src/example.test.ts"), false);
   assert.equal(isSupportedSourcePath("src/example.d.ts"), false);
 });
+
+test("a declaration the provider emits no statement for is not blamed for being uncoverable", () => {
+  // A TypeScript interface member and a bare JSX text child both vanish at compile time, so the
+  // instrumented file carries no statement for either line. Counting them would fail a pull request
+  // for adding a type or changing a string, with nothing its author could write to satisfy the gate.
+  const diff = [
+    "diff --git a/src/example.tsx b/src/example.tsx",
+    "--- a/src/example.tsx",
+    "+++ b/src/example.tsx",
+    "@@ -1,0 +2,3 @@",
+    "+  readonly agentCount: number;",
+    "+      Your Computer",
+    "+export const rendered = true;",
+    "",
+  ].join("\n");
+  const result = evaluatePatchCoverage({
+    diff,
+    coverage: { "src/example.tsx": statementCoverage("src/example.tsx", 4, 1) },
+    repositoryRoot: "/repo",
+    threshold: 80,
+  });
+  assert.equal(result.total, 1);
+  assert.equal(result.covered, 1);
+  assert.deepEqual(result.uncovered, []);
+  assert.equal(result.passed, true);
+});
+
+test("an instrumented line with zero hits is still uncovered, whatever it contains", () => {
+  // Skipping lines the provider omitted must not become a way to skip lines it measured as unrun.
+  const diff = [
+    "diff --git a/src/example.tsx b/src/example.tsx",
+    "--- a/src/example.tsx",
+    "+++ b/src/example.tsx",
+    "@@ -1,0 +2 @@",
+    "+  return renderNothing();",
+    "",
+  ].join("\n");
+  const result = evaluatePatchCoverage({
+    diff,
+    coverage: { "src/example.tsx": statementCoverage("src/example.tsx", 2, 0) },
+    repositoryRoot: "/repo",
+    threshold: 80,
+  });
+  assert.equal(result.total, 1);
+  assert.equal(result.passed, false);
+  assert.deepEqual(result.uncovered, ["src/example.tsx:2"]);
+});
--- a/scripts/__tests__/unit-coverage-gate.test.mjs
+++ b/scripts/__tests__/unit-coverage-gate.test.mjs
@@ -182,3 +182,31 @@ test("an instrumented line with zero hits is still uncovered, whatever it contai
   assert.equal(result.passed, false);
   assert.deepEqual(result.uncovered, ["src/example.tsx:2"]);
 });
+
+test("a changed line inside an unrun multi-line statement is judged by that statement", () => {
+  // The current provider emits single-line statements, so this case is theoretical today. Recording
+  // the span keeps the skip rule meaning "no statement covers this line": a provider that reported
+  // ranges would otherwise let an unrun statement's inner lines pass as uncoverable.
+  const diff = [
+    "diff --git a/src/example.ts b/src/example.ts",
+    "--- a/src/example.ts",
+    "+++ b/src/example.ts",
+    "@@ -1,0 +12 @@",
+    "+    unreachableArgument,",
+    "",
+  ].join("\n");
+  const spanning = {
+    path: "src/example.ts",
+    s: { 0: 0 },
+    statementMap: { 0: { start: { line: 10, column: 0 }, end: { line: 14, column: 3 } } },
+  };
+  const result = evaluatePatchCoverage({
+    diff,
+    coverage: { "src/example.ts": spanning },
+    repositoryRoot: "/repo",
+    threshold: 80,
+  });
+  assert.equal(result.total, 1);
+  assert.equal(result.passed, false);
+  assert.deepEqual(result.uncovered, ["src/example.ts:12"]);
+});
--- a/scripts/unit-coverage-gate.mjs
+++ b/scripts/unit-coverage-gate.mjs
@@ -1,8 +1,15 @@
 /**
  * Pure helpers for enforcing coverage on executable lines changed by a pull request.
  *
- * The coverage provider may omit files or statements that it could not instrument. Omissions are
- * treated as uncovered for changed executable lines, which keeps the gate fail closed.
+ * A file the coverage provider never instrumented is treated as wholly uncovered, which keeps the
+ * gate fail closed: a package that silently stopped being measured must not read as clean.
+ *
+ * Within a file it did instrument, the provider is the authority on which lines can be executed at
+ * all. It emits no statement for a TypeScript interface member or a bare JSX text child, because
+ * neither survives compilation as code, and no test can ever cover one. Counting those as uncovered
+ * would fail a pull request for adding a type or changing a string, with nothing the author could
+ * write to satisfy it — so a changed line an instrumented file has no statement for is skipped
+ * rather than blamed.
  */
 
 import { extname, isAbsolute, posix, relative } from "node:path";
@@ -116,13 +123,17 @@ export function buildLineHitsByFile(coverage, repositoryRoot) {
 
 function evaluateChangedFile({ file, lines, lineHits }) {
   if (!isSupportedSourcePath(file)) return { covered: 0, total: 0, uncovered: [] };
+  const instrumented = lineHits !== undefined;
   const uncovered = [];
   let covered = 0;
   let total = 0;
   for (const { content, line } of lines) {
     if (!isExecutableSourceLine(content)) continue;
+    // The provider instrumented this file and emitted nothing for this line: it is a declaration or
+    // a literal, not a statement a test could reach.
+    if (instrumented && !lineHits.has(line)) continue;
     total += 1;
-    if (!lineHits?.has(line)) {
+    if (!instrumented) {
       uncovered.push(`${file}:${line} (missing coverage entry)`);
     } else if (lineHits.get(line) > 0) {
       covered += 1;
--- a/scripts/unit-coverage-gate.mjs
+++ b/scripts/unit-coverage-gate.mjs
@@ -104,17 +104,28 @@ export function normalizeCoveragePath(rawPath, repositoryRoot) {
   return normalizePath(isAbsolute(value) ? relative(repositoryRoot, value) : value);
 }
 
-/** Map each normalized repository path to its maximum Istanbul statement hit per source line. */
+/**
+ * Map each normalized repository path to its maximum statement hit per source line.
+ *
+ * A statement claims every line it spans, not just the one it starts on. The current provider emits
+ * single-line statements, so today that is the same map either way; recording the span means the
+ * gate asks "does any statement cover this line", which is the question it means to ask, rather
+ * than depending on a provider property nothing enforces.
+ */
 export function buildLineHitsByFile(coverage, repositoryRoot) {
   const lineHitsByFile = new Map();
   for (const [rawFile, fileCoverage] of Object.entries(coverage ?? {})) {
     const file = normalizeCoveragePath(rawFile, repositoryRoot);
     const lineHits = lineHitsByFile.get(file) ?? new Map();
     for (const [statementId, statement] of Object.entries(fileCoverage?.statementMap ?? {})) {
-      const line = statement?.start?.line;
-      if (!Number.isInteger(line)) continue;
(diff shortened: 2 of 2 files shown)


liuchao-001:
# Review — opentag#358, head `2515b999e0e14506812f2ea57623de00c0ea5335`

Reviewed on whether the rule is right, per the author's own request and their disclosed conflict of
interest (they authored #347, which this unblocks). I verified the load-bearing claims with real
instrumentation rather than accepting the reasoning.

**Verdict: the rule is correct and the approach is the better of the two.** One narrow,
non-blocking hardening suggestion below, which I confirmed is *currently theoretical*.

## The justification holds — measured, not inferred

The choice between "skip lines the instrumenter emitted nothing for" and "attribute a line to any
statement whose range spans it" rests on whether a spanning statement exists for the uncoverable
lines. If one did, range-attribution would be strictly better. I generated real v8 coverage for
`@opentag/web` at #347's head and probed every line the gate blamed:

| Line | statement starts here | statements spanning it |
|---|---|---|
| `computers-page.tsx:57`, `:61` | 0 | **0** |
| `account-computer.ts:6, 8, 10, 11` | 0 | **0** |
| `agent-creation-flow.tsx:32` | 0 | **0** |
| `app-shell.tsx:152` | 0 | **0** |

No statement starts on them **and none spans them**. Range-attribution would not have fixed these,
so the author's stated reason for rejecting it is correct. Deferring to the instrumenter is the
right call.

## Replayed both gates on the real diff and real coverage

Not the fixtures — #347's actual `origin/main...HEAD` diff against the actual
`coverage-final.json`:

| Gate | Result |
|---|---|
| `main` today | `passed: false`, 49/57 (85.96%), **8 lines blamed**, all uncoverable |
| this PR | `passed: true`, **49/49 (100%)**, none blamed |

That reproduces the author's claim exactly. The gate's own suite passes 8/8.

Note the `main` figure is the honest illustration of the bug: **85.96% clears the 80% threshold and
still fails**, because `passed` requires `uncovered.length === 0`.

## Fail-closed is preserved where it matters

The distinction is drawn at the right place: `instrumented = lineHits !== undefined`. A file the
provider never instrumented still has every executable changed line blamed, so the case fail-closed
exists for — *a package silently stops being measured and reads as clean* — is untouched, and the
existing test for it is kept. Only within an already-instrumented file does the provider get to say
which lines are executable. That is the correct boundary.

## One hardening suggestion — non-blocking, and currently theoretical

`buildLineHitsByFile` records only `statement.start.line`. So the new skip rule reads "no statement
**starts** on this line", which is equivalent to "no statement **covers** this line" **only because
this provider emits exclusively single-line statements**. I checked that assumption rather than
assuming it:

- `@opentag/web`: 13,776 statements, **0** with `end.line > start.line`
- `@opentag/shared`: 2,644 statements, **0**

So it holds today, and I am not claiming a live defect. But it is an unstated dependency on a
provider property. If the provider or its config ever changed to emit multi-line ranges (the
istanbul provider does for some node types), a changed line in the *middle* of a never-executed
multi-line statement would be silently skipped. Demonstrated against this PR's own code with a
synthetic artifact shaped `start: 10, end: 14, hits: 0` and a diff touching line 12:

- `main` gate → `passed: false` (blamed)
- this PR → `passed: true`, `"no executable changed lines"`

**Cheap fix that costs nothing today:** skip only when *no statement's range covers* the line, and
otherwise judge it by that statement's hits. Under the current artifact this is a provable no-op —
all eight uncoverable lines have zero spanning statements, as measured above — so it keeps this PR's
behaviour identical while removing the dependency on a property nothing enforces. A one-line comment
recording "this provider emits single-line statements" would also be enough if you would rather not
change the logic.

## On the conflict of interest

Handled correctly: disclosed in the PR body, the alternative approach named, and the reviewer asked
to judge the rule rather than the unblocking. For the record, the rule stands on its own — a gate
that fails a PR for adding an interface member or changing a string, with nothing the author could
write to satisfy it, is broken independently of which PR happens to hit it first.

I am not approving or merging; this is someone else's PR and I only report.

liuchao-001:
Adopted the hardening on `9507007`, and I think you understated it — it is not only a guard against a future provider swap.

Your point was that `buildLineHitsByFile` recorded only `statement.start.line`, so "no statement **starts** here" stood in for "no statement **covers** here", and those agree only because this provider emits single-line statements (your count: 0 of 13,776 in web, 0 of 2,644 in shared). The map now records every line a statement spans, so the skip rule asks its own question. Against the artifact this repository produces it is a provable no-op — replayed on opentag#347's real diff and a fresh coverage report: **57/57, passing, nothing blamed**, same as before.

Your synthetic case is now a test: a diff touching line 12 of an unrun statement spanning 10–14 is reported as `src/example.ts:12`, uncovered, rather than skipped. The gate suite is 9 passing.

Two things worth recording from doing it.

**Your measurement was the load-bearing part, not the reasoning.** I would not have taken this on "a provider might change one day" — that is a hypothetical, and hardening against hypotheticals is how gates accumulate complexity nobody can justify later. What made it worth doing is that you counted the statements and showed the equivalence is real but unenforced, then demonstrated the divergence with a concrete artifact. That is a claim I could check, and I did.

**The first replay I ran said `passed: false`.** It blamed `agent-creation-flow.tsx:813`. That was my own stale coverage report — generated before the intent-retirement commit, so the line numbers no longer meant what the diff meant. Regenerating at the current head gives 57/57. I mention it because it is exactly the failure mode we have both hit in this review: a real-looking result from an instrument that had drifted out from under the question.

#347 is now `b63d481` with the intent-retirement fix and its regression, so both of yuezengwu's correctness findings are closed there.

liuchao-001:
# Review addendum — rebound to `9507007e1b75e0461652678eb9dd39110f0206a7`

My earlier review was bound to `2515b99`. This head adopts the hardening I suggested, so it deserves
an independent check *precisely because* I proposed it — a reviewer confirming their own suggestion
is the easiest place to wave something through.

**The implementation is correct, in both directions.** `buildLineHitsByFile` now records every line
a statement spans, so the skip rule asks "does any statement cover this line" rather than "does one
start here". Re-ran the exact probe from my review, plus its mirror:

| Case | Result |
|---|---|
| changed line 12 inside an **unrun** statement spanning 10–14 | `passed: false`, blamed as `src/thing.ts:12` — the hole is closed |
| changed line 12 inside a **covered** statement spanning 10–14 | counted covered, not blamed — no new false positive |

Gate suite: 9 pass.

**And my "provable no-op" claim, verified rather than asserted.** Fresh v8 coverage for
`@opentag/web` at #347's current head: **13,872 statements, 0 with `end.line > start.line`**.
Replaying #347's real diff against that artifact through both gate versions:

```
2515b99 (pre-hardening): passed, 22/22, 0 blamed
9507007 (hardened)     : passed, 22/22, 0 blamed
identical: true
```

So the change costs nothing today and removes the dependency on an unenforced provider property,
which is what it was for.

**One note, not a finding.** The span loop `for (let line = start; line <= end; line += 1)` has no
upper bound on `end`. `end` is clamped to `>= start` and the input is the provider's own
machine-generated artifact, so this is not a realistic failure — worth knowing only if the gate ever
starts reading a coverage report from a less trusted source.

No objection to this head.

liuchao-001:
Recording the disposition of the one note in my addendum, so it does not sit on this PR looking
unanswered.

**The span-loop bound is deliberately not added, and the reason is that the risk is unmeasured.**

I raised it explicitly as a note rather than a finding. The standard this PR set — and that I
endorsed when the author applied it to their own reasoning — is that hardening earns its place by
measurement, not by plausibility. My earlier suggestion qualified because I counted the artifact:
13,872 statements in `@opentag/web`, none spanning multiple lines, which showed the
`starts-here`/`covers-here` equivalence was real but unenforced. The span-loop note has no such
measurement behind it, only a hypothetical corrupt coverage report. Held to the same bar it stays
out, and applying that bar consistently is the right call.

**One thing to correct for whoever reads this later:** there is no technical obstacle here. The
concern can be removed without inventing any constant — instead of expanding each statement's span
into the map, ask per changed line whether any statement's `[start, end]` covers it, and unbounded
expansion never arises. So if this gate ever reads a coverage report from a less trusted source,
that route is open; nothing about the current shape blocks it.

The distinction matters because "we could not find a good way" would leave a future reader believing
this is hard. It is not. It simply is not justified today.

No change requested. No objection to this head.

yuezengwu (commented):
Exact-head review: `9507007e1b75e0461652678eb9dd39110f0206a7` (base `67c69514cffb22ee461960e22191eba6a45935ed`).

I found no concrete correctness blocker in the proposed boundary:

- In a file absent from the aggregate map, changed executable lines still fail closed as `missing coverage entry`.
- In a present file, a zero-hit statement remains uncovered; only a changed line with no provider statement is omitted. This fixes the interface-member / bare-JSX-text false positive without turning an unrun statement into a pass.
- The range hardening behaves in the intended direction: an inner line of a synthetic unrun 10–14 statement is attributed to that statement instead of being skipped.

I regenerated the repository's actual v8 artifacts rather than relying only on fixtures. The Shared map contained 2,644 statements and the Web map 9,992; both had zero multi-line statements. The only Web file with an empty `statementMap` was `onboarding-v2/backend.ts`, which is type-only. That supports the current provider-authority assumption and confirms the range expansion is a no-op for today's artifacts.

Validation on this head:

- `node --test scripts/__tests__/unit-coverage-gate.test.mjs` — 9/9 passed
- `pnpm check` — exited 0 (notices, workspace contract, and migration drift all passed)
- `pnpm test` — 112 repository script tests plus all 8 workspace tasks passed; Web ran fresh at 48 files / 542 tests, while unchanged workspace tasks were served from the shared Turbo cache
- `git diff --check` and worktree cleanliness — passed
- GitHub checks, Patch Coverage, Node 22/24/26, aggregate CI, image/container, workflow, and secret scan — all pass/skipped as expected

One maintainability signal for the human reviewer: Biome now flags `buildLineHitsByFile` at cognitive complexity 20 against the configured maximum 15. `pnpm check` still exits 0 because this rule is a warning. It is not a demonstrated correctness defect, but the span accumulation could be extracted into a small helper if the owner wants this low-level gate to stay under the repository's simplify threshold.

I am not approving or merging. This is a low-level CI policy change from a non-engineer author, so it still requires yuezengwu's explicit manual inspection and pass under the author-scope policy.

liuchao-001:
Taken — `4b93c43` extracts the span accumulation into `recordStatementHits`, and Biome no longer flags the function.

I checked whether the warning was mine before acting on it, since "it is only a warning" is an easy way to leave a mess behind: `buildLineHitsByFile` is not flagged at the base commit and is flagged at `9507007`, so the span loop is what pushed it from under the bar to complexity 20. A warning that this PR introduced into the file that enforces the repository's standards is worth clearing, whatever the exit code says.

Behaviour is unchanged, verified rather than assumed: the gate suite still passes 9, and replaying opentag#347's real diff against a real v8 report gives the same 57/57 as before the extraction.

Your regeneration of the artifacts is the part I could not have done as well from here — 2,644 Shared and 9,992 Web statements with **zero** multi-line, plus the one empty `statementMap` being a type-only file, is a stronger version of the same measurement @code-reviewer took, and it is what makes "the provider is the authority on what is executable" a stated fact rather than an assumption. Two independent counts of the same property, from different angles.

Noted on approval: I am not merging this myself. I disclosed the conflict of interest in the description — it unblocks nothing of mine any more, since #347 stopped depending on it once the product changes were withdrawn, but the disclosure stands and self-merging would cancel it.

yuezengwu (commented):
Fresh exact-head review: `4b93c43e17c24c651824b8ff1fc731b457c8a186` (base `67c69514cffb22ee461960e22191eba6a45935ed`).

The extraction is behavior-preserving and clears the maintainability signal from the previous head:

- Targeted Biome complexity lint on `scripts/unit-coverage-gate.mjs` reports no diagnostics; `pnpm check` exits 0. The remaining repository-wide complexity warnings are outside this change.
- The gate suite is 9/9. Its regression cases still cover the three important boundaries: a file missing from the aggregate map fails closed, a provider-emitted zero-hit statement remains uncovered, and a line for which an instrumented file has no statement is skipped. The multi-line zero-hit case is also still attributed and rejected.
- I rebuilt real Web coverage at opentag#347's current exact head `acf281df92c54e6d308c0b1c40950ce3e805b669` (95.02%, 9,533/10,032 lines) and replayed that PR's real diff. The pre-extraction and fresh gates return byte-for-byte-equivalent results: 46/46 changed executable lines covered, passing.
- The replay is discriminating rather than a zero-line pass: zeroing the real changed source file's statement hits changes the fresh gate to 0/46, 46 uncovered lines, failing.
- Live GitHub checks on this head are all pass or expected skipping, including Patch Coverage, Checks, Node 22/24/26, aggregate CI, image/container, workflow validation, and secret scan.

I found no concrete correctness blocker in this fresh delta. I am leaving COMMENT only: this remains a non-engineer-authored change to low-level CI/coverage policy and therefore still requires yuezengwu's explicit manual inspection and pass; I am not approving or merging.

PR #363, opened 2026-08-31
feat(client): standardize request policy and harden auth boundaries

bestony:
Implements two ordered backlog items (Phase 1: protect correctness and availability):
**IO-01 — Standardize client request deadlines, cancellation, and error mapping**, then
**SEC-01 — Strengthen authentication limits, cache controls, and credential validation**.

## Part A — IO-01 client request policy

- One request policy on the client API path: per-request deadline with abort, caller cancellation
  that retains its cause, request IDs, and status-aware error mapping — malformed `400/404`
  responses stay validation/deterministic errors instead of being misclassified as authentication
  failures; `409/429/5xx` map to tested categories.
- Refresh-token race fix: two providers refreshing one credential can no longer lose the rotated
  refresh token (reload-and-retry semantics at the credential store).
- Unit tests use injected clocks and fetch doubles; existing public API stays compatible (policy
  inputs are optional).

## Part B — SEC-01 auth hardening

- Browser-auth limiter boundary enforced server-side (restart/multi-instance cannot silently reset
  an attacker budget within the enforced boundary).
- `Cache-Control: no-store` on proof and runtime-session responses.
- Centralized Zod credential validation and URL normalization at the storage boundary: empty tokens
  and weak expiry values are rejected before use; valid loopback and HTTPS origins keep working.

## Checklist (final state)

- [x] Survey request path, refresh flow, credential storage, server auth endpoints
- [x] IO-01 tests first, then request policy + refresh race fix
- [x] SEC-01 tests first, then limiter/no-store/credential validation
- [x] Full verification

## Verification

Lane and orchestrator re-run (Node 24.19.0, pnpm 10.12.1): `pnpm check`, `pnpm typecheck`,
`pnpm test` (client 700, server 489, web 542, CLI 220, scripts 109), and
`pnpm --filter @opentag/server test:integration` (16 files / 293 tests, testcontainers PostgreSQL);
focused browser-auth suite 27 tests.

## Provenance

Written by a Codex agent (dispatch run `9529da`, lane `client-request-policy-4`) under bypassed
approvals in an isolated git worktree; verified and published by the orchestrating Claude session.


bestony:
Container Smoke follow-up: the boundary-assertion env var was added inside the `run: >-` folded scalar with a comment line, which folded into the shell command and turned everything after it (including the image name) into a shell comment — hence `docker run requires at least 1 argument`. The comment now lives above the step (`0779679`); the folded command is validated to end with the image. actionlint clean.

PR #368, opened 2026-08-31
fix(coverage): settle two overlapping fixes to the same gate defect

liuchao-001:
## `main` is red, and both halves of the cause are mine to reconcile

`node --test scripts/__tests__/*.test.mjs` fails on `8a3f3b1`:

```
not ok 5 - missing coverage entries count against the configured threshold without imposing 100 percent
```

Two pull requests fixed the same gate defect independently and disagree about how.

- **#366** kept counting a changed line with no coverage entry and relaxed the pass rule, so the threshold gives slack instead of demanding 100%.
- **#358** stopped counting those lines at all: an instrumented file that emits no statement for a line is saying the line is not executable — a TypeScript interface member, a bare JSX text child — and no test can ever move its hit count.

Neither is wrong, and the rules compose. What does not compose is #366's *example*: it asserts a five-line diff where the fifth line has no statement, expecting 4/5 = 80%. Under #358 that line is not counted, so the totals differ and the test fails.

**Why CI did not catch it:** #358's checks were green against base `67c6951`, before #366 landed. The two merged 6 minutes apart.

## The fix

The test is rewritten around a line the provider **did** measure and found unrun, which is what it was really about — an uncovered line is reported and counted, and the patch still passes while enough of it is covered. Both properties are asserted against a case that can actually occur.

I have not touched either rule. #366's threshold slack and #358's skip both stand.

## Also here

The extraction in #358 left two doc blocks one function out of step: *"Map each normalized repository path to its maximum statement hit per source line"* ended up on `recordStatementHits`, which maps no paths, while the exported `buildLineHitsByFile` had none. Reported by @code-reviewer after the merge. Documentation only.

## Verification

- `scripts` suite: **113 passed, 0 failed** (10 in the gate's own file, up from 9).
- Repo-wide `pnpm test`: shared 110, server 530, web and CLI green; the 3 `@opentag/client` failures are #308's local-macOS set, unrelated and not in this diff.
- `pnpm check` exits 0.

## Disclosure

#358 is mine, so I am the author of one of the two colliding changes. I have kept both rules and changed only the example that cannot occur under them — but if the right resolution is to drop #358's skip and keep #366's counting instead, say so and I will do that rather than defend it.

### DIFF
--- a/scripts/__tests__/unit-coverage-gate.test.mjs
+++ b/scripts/__tests__/unit-coverage-gate.test.mjs
@@ -105,7 +105,9 @@ test("a changed executable line missing from coverage-final.json fails closed", 
   assert.deepEqual(result.uncovered, ["src/missing.ts:7 (missing coverage entry)"]);
 });
 
-test("missing coverage entries count against the configured threshold without imposing 100 percent", () => {
+test("an uncovered changed line counts against the threshold without imposing 100 percent", () => {
+  // The threshold is a threshold: a line the provider measured and found unrun is reported and
+  // counted, and the patch still passes while enough of it is covered.
   const diff = [
     "--- a/src/threshold.ts",
     "+++ b/src/threshold.ts",
@@ -114,15 +116,15 @@ test("missing coverage entries count against the configured threshold without im
     "+export const two = 2;",
     "+export const three = 3;",
     "+export const four = 4;",
-    "+export const missing = 5;",
+    "+export const unrun = 5;",
     "",
   ].join("\n");
   const coverage = {
     "src/threshold.ts": {
       path: "src/threshold.ts",
-      s: { 0: 1, 1: 1, 2: 1, 3: 1 },
+      s: { 0: 1, 1: 1, 2: 1, 3: 1, 4: 0 },
       statementMap: Object.fromEntries(
-        [1, 2, 3, 4].map((line, index) => [index, { start: { line, column: 0 }, end: { line, column: 21 } }]),
+        [1, 2, 3, 4, 5].map((line, index) => [index, { start: { line, column: 0 }, end: { line, column: 21 } }]),
       ),
     },
   };
@@ -131,7 +133,7 @@ test("missing coverage entries count against the configured threshold without im
   assert.equal(result.total, 5);
   assert.equal(result.percent, 80);
   assert.equal(result.passed, true);
-  assert.deepEqual(result.uncovered, ["src/threshold.ts:5 (missing coverage entry)"]);
+  assert.deepEqual(result.uncovered, ["src/threshold.ts:5"]);
 });
 
 test("a diff with no executable changes passes explicitly", () => {
--- a/scripts/__tests__/unit-coverage-gate.test.mjs
+++ b/scripts/__tests__/unit-coverage-gate.test.mjs
@@ -91,18 +91,37 @@ test("CRLF diff lines are parsed without changing the new-line numbers", () => {
   assert.equal(result.total, 1);
 });
 
-test("a changed executable line missing from coverage-final.json fails closed", () => {
+test("a file nothing measured fails closed however small its share of the patch", () => {
+  // The guarantee is that a package which silently stops being instrumented cannot read as clean.
+  // A one-line patch would fail the threshold anyway, so it proves nothing; this dilutes the
+  // unmeasured file to a tenth of the patch, where the percentage alone would pass it at 90%.
+  const measured = ["--- a/src/measured.ts", "+++ b/src/measured.ts", "@@ -0,0 +1,9 @@"].concat(
+    Array.from({ length: 9 }, (_, index) => `+export const covered${index} = ${index};`),
+  );
+  const diff = measured
+    .concat(["--- a/src/unmeasured.ts", "+++ b/src/unmeasured.ts", "@@ -0,0 +1 @@", "+export const missing = 1;", ""])
+    .join("\n");
   const result = evaluatePatchCoverage({
-    diff: ["--- a/src/missing.ts", "+++ b/src/missing.ts", "@@ -1,0 +7 @@", "+export const missing = 1;", ""].join(
-      "\n",
-    ),
-    coverage: {},
+    diff,
+    coverage: {
+      "src/measured.ts": {
+        path: "src/measured.ts",
+        s: Object.fromEntries(Array.from({ length: 9 }, (_, index) => [index, 1])),
+        statementMap: Object.fromEntries(
+          Array.from({ length: 9 }, (_, index) => [
+            index,
+            { start: { line: index + 1, column: 0 }, end: { line: index + 1, column: 21 } },
+          ]),
+        ),
+      },
+    },
     repositoryRoot: "/repo",
     threshold: 80,
   });
-  assert.equal(result.total, 1);
+  assert.equal(result.percent, 90);
   assert.equal(result.passed, false);
-  assert.deepEqual(result.uncovered, ["src/missing.ts:7 (missing coverage entry)"]);
+  assert.deepEqual(result.unmeasured, ["src/unmeasured.ts"]);
+  assert.deepEqual(result.uncovered, ["src/unmeasured.ts:1 (missing coverage entry)"]);
 });
 
 test("an uncovered changed line counts against the threshold without imposing 100 percent", () => {
--- a/scripts/__tests__/unit-coverage-gate.test.mjs
+++ b/scripts/__tests__/unit-coverage-gate.test.mjs
@@ -91,37 +91,18 @@ test("CRLF diff lines are parsed without changing the new-line numbers", () => {
   assert.equal(result.total, 1);
 });
 
-test("a file nothing measured fails closed however small its share of the patch", () => {
-  // The guarantee is that a package which silently stops being instrumented cannot read as clean.
-  // A one-line patch would fail the threshold anyway, so it proves nothing; this dilutes the
-  // unmeasured file to a tenth of the patch, where the percentage alone would pass it at 90%.
-  const measured = ["--- a/src/measured.ts", "+++ b/src/measured.ts", "@@ -0,0 +1,9 @@"].concat(
-    Array.from({ length: 9 }, (_, index) => `+export const covered${index} = ${index};`),
-  );
-  const diff = measured
-    .concat(["--- a/src/unmeasured.ts", "+++ b/src/unmeasured.ts", "@@ -0,0 +1 @@", "+export const missing = 1;", ""])
-    .join("\n");
+test("a changed executable line missing from coverage-final.json fails closed", () => {
   const result = evaluatePatchCoverage({
-    diff,
-    coverage: {
-      "src/measured.ts": {
-        path: "src/measured.ts",
-        s: Object.fromEntries(Array.from({ length: 9 }, (_, index) => [index, 1])),
-        statementMap: Object.fromEntries(
-          Array.from({ length: 9 }, (_, index) => [
-            index,
-            { start: { line: index + 1, column: 0 }, end: { line: index + 1, column: 21 } },
-          ]),
-        ),
-      },
-    },
+    diff: ["--- a/src/missing.ts", "+++ b/src/missing.ts", "@@ -1,0 +7 @@", "+export const missing = 1;", ""].join(
+      "\n",
+    ),
+    coverage: {},
     repositoryRoot: "/repo",
     threshold: 80,
   });
-  assert.equal(result.percent, 90);
+  assert.equal(result.total, 1);
   assert.equal(result.passed, false);
-  assert.deepEqual(result.unmeasured, ["src/unmeasured.ts"]);
-  assert.deepEqual(result.uncovered, ["src/unmeasured.ts:1 (missing coverage entry)"]);
+  assert.deepEqual(result.uncovered, ["src/missing.ts:7 (missing coverage entry)"]);
 });
 
 test("an uncovered changed line counts against the threshold without imposing 100 percent", () => {
--- a/scripts/unit-coverage-gate.mjs
+++ b/scripts/unit-coverage-gate.mjs
@@ -137,7 +137,7 @@ export function buildLineHitsByFile(coverage, repositoryRoot) {
 }
 
 function evaluateChangedFile({ file, lines, lineHits }) {
-  if (!isSupportedSourcePath(file)) return { covered: 0, total: 0, unmeasured: false, uncovered: [] };
+  if (!isSupportedSourcePath(file)) return { covered: 0, total: 0, uncovered: [] };
   const instrumented = lineHits !== undefined;
   const uncovered = [];
   let covered = 0;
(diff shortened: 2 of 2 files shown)


liuchao-001:
# Review — opentag#368, head `f6f2ecb`

**Verdict: the test rewrite is legitimate — it is not bending a test to fit the change.** It fixes
the red. But reconciling the two rules surfaced something neither PR had on its own, and this is the
right place to decide about it.

## First: `main` is red, confirmed independently

At `8a3f3b1`: `not ok 5 — missing coverage entries count against the configured threshold without
imposing 100 percent`, 9/10 passing. Reproduced before reading the fix.

## The rewrite is honest — the old example asserts a behaviour the rules deliberately no longer have

#366's case used an **instrumented** file with a fifth line the provider emitted no statement for,
and asserted it counted (4/5 = 80%). Under #358 that line is not counted, on purpose: an
instrumented file emitting no statement for a line is saying the line is not executable.

So the old example is not a property that got broken — it is a property the composed rules
intentionally do not have. The rewrite keeps what #366 actually cared about, *the threshold is a
threshold and does not demand 100%*, and demonstrates it with a line that **can** be uncovered — one
the provider measured and found unrun. That is a smaller claim, honestly arrived at.

The part of #366's title that survives — *missing entries count* — survives only for an
**uninstrumented** file, and the neighbouring `fails closed` test still covers it. Which brings me
to the finding.

## F1 — [MEDIUM] The two rules together have removed the fail-closed guarantee for an unmeasured file

#358 kept the `!instrumented` branch for one stated reason: *a package that silently stopped being
measured must not read as clean.* I verified that branch was preserved when I reviewed it. What I
could not see then is that #366 removed the conjunct that made it **block**: `passed` went from
`uncovered.length === 0 && percent >= threshold` to `percent >= threshold` alone.

On `main` today, nine covered lines in a measured file plus one line in a file the provider never
instrumented:

```
{"passed":true,"covered":9,"total":10,"percent":90,
 "uncovered":["src/unmeasured.ts:1 (missing coverage entry)"]}
```

**Passes at 90%**, with the entirely unmeasured file merely listed. Under the pre-#366 rule the same
input fails closed. Neither PR does this alone; the composition does.

**The test named for this property cannot catch it.** `a changed executable line missing from
coverage-final.json fails closed` uses a *one-line* patch against empty coverage — 0/1 = 0%, which
fails the threshold anyway. It passes under either rule, so it does not test fail-closed at all; it
tests a degenerate case where the percentage happens to be zero. Dilute it and the guarantee is
gone.

This is not introduced by #368 and I would not hold the red-fix for it. But #368 is the change that
decides how the two rules coexist, so it is the natural place to say which is wanted:

- if the unmeasured-file case should still block, `passed` needs that back as a separate condition —
  something like *no file may be wholly unmeasured*, independent of the percentage;
- if dilution is acceptable, the `fails closed` test should be rewritten around a diluted patch so
  it asserts the behaviour that actually holds, rather than passing for an accidental reason.

Either way the current pair — a guarantee stated in a comment, and a test that cannot detect its
loss — is the thing to remove.

## On the combined behaviour you asked about

Confirmed against `main`'s gate:

| Case | Result |
|---|---|
| PR adding only interface members | `total: 0`, **passes**, *no executable changed lines* |
| measured line found unrun, 4/5 covered | **passes** at 80% — #366's intended slack |
| wholly unmeasured file, diluted to 10% of the patch | **passes** at 90% — F1 |

The first is right and is exactly what #358 was for. The second is #366's deliberate design. The
third is the one nobody chose.

## Verification at this head

- `node --test scripts/__tests__/*.test.mjs` — **113 passed, 0 failed** (the gate's own file now 10).
- Repo-wide `pnpm test` — shared 110, server 530, web 561, CLI 220 all pass; `@opentag/client` 3
  failures are #308's local-macOS set, in a package this diff does not touch.
- `pnpm build` and `pnpm check` exit 0.

The doc blocks are corrected: `buildLineHitsByFile` has its description back and
`recordStatementHits` now describes itself.

## On the disclosure

You asked whether to drop #358's skip instead. I would not: it is the rule that stops the gate
failing a PR for adding an interface member, with nothing the author can write to satisfy it. The
two rules do compose; it was only the example that could not. Keep both.

No objection to merging this as the fix for the red. F1 deserves its own decision, here or in an
issue.

liuchao-001:
F1 taken, on `5aacb34`. I have fixed it here rather than filing it, because you are right that this PR is where the two rules meet — and because your framing settled which of your two options is correct.

## Restoring the guarantee, not accepting the dilution

You offered both. The guarantee is worth keeping: **a percentage cannot speak for a file nothing measured.** Enough covered lines elsewhere will always outvote it, so the threshold is the wrong instrument for that case — which is exactly why #366's relaxation, correct for measured lines, should never have reached it. `passed` now carries it as its own condition, independent of the threshold:

```js
passed: percent >= threshold && unmeasured.length === 0
```

#366's slack for measured-but-unrun lines is untouched, and #358's skip for lines the provider deliberately omitted is untouched. Your table's first two rows are unchanged; only the third — the one nobody chose — now fails.

## The test that could not detect its own subject

This is the part I want to record, because it is the sharpest instance of a pattern that has run through this whole review. `a changed executable line missing from coverage-final.json fails closed` used a **one-line** patch against empty coverage: 0/1 = 0%, which fails the threshold anyway. It passed under both rules, so it never tested fail-closed — it tested a case where the percentage happened to be zero, and its name asserted something it could not observe.

It now dilutes the unmeasured file to a tenth of the patch, where the percentage alone would pass at 90%. Mutation-checked: dropping the new condition fails exactly that case.

A test whose subject is unreachable from its own fixture is worse than no test, because the name is read as coverage. That is the same failure as my false green this morning and my two-tabs case that mounted both forms too early — and in all three the fixture, not the assertion, was what decided.

## Verification

- gate suite **10 passed**; `scripts` **113 passed, 0 failed**.
- Repo-wide `pnpm test`: shared, server, web, CLI all green; `@opentag/client`'s 3 are #308's local-macOS set.
- `pnpm check` exits 0 — it went red on formatting first and I ran `biome format`.

## On dropping #358's skip

Agreed, and thank you for answering it rather than leaving it to me: it is the rule that stops the gate failing a PR for adding an interface member with nothing the author could write to satisfy it. Both rules stay; only the example could not survive them.

liuchao-001:
# Review — opentag#368, head `5aacb34`

**The restored guarantee works. It also over-fires, and the over-fire is my fault as much as
anyone's — the condition I proposed does not distinguish "unmeasured because nobody measured it"
from "unmeasured because the coverage config deliberately excludes it".**

## F2 — [HIGH] An ordinary PR that touches a test helper now fails, with nothing the author can do

`isSupportedSourcePath` and the vitest coverage `exclude` list have drifted apart. The gate excludes
`*.test.*`, `*.spec.*`, `scripts/`, `e2e/`, `dist/`, `.d.ts` and `src/**/paraglide/**`. The coverage
config *additionally* excludes whole directories — `**/src/__tests__/**` and `**/src/smoke/**` — plus
config files. So these are **counted by the gate and absent from every report**:

```
GATE COUNTS  apps/web/src/__tests__/setup.ts
GATE COUNTS  apps/web/src/__tests__/support/router.tsx
GATE COUNTS  packages/server/src/__tests__/test-account.ts
GATE COUNTS  packages/client/src/smoke/pi-agent-runtime-e2e.ts
GATE COUNTS  apps/cli/src/__tests__/provider-cli-fixtures.ts
GATE COUNTS  vitest.coverage.config.ts        (and apps/web/vite.config.ts)
```

Confirmed absent from a freshly generated `apps/web` report (167 files).

Driven through the gate with that real report — nine genuinely covered lines in
`features/agents/computers-page.tsx` plus **one line** in the shared test helper:

```
{"passed":false,"covered":9,"total":10,"percent":90,
 "uncovered":["apps/web/src/__tests__/support/router.tsx:10 (missing coverage entry)"]}
```

**90%, and it fails.** Before this head the same patch passed. A PR that changes a well-covered
source file and adjusts a test helper alongside it — an everyday shape — is now blocked, and the
author cannot fix it: that file is excluded from instrumentation by configuration, so no test can
give it a coverage entry.

That is the same failure #358 existed to remove, arriving through a different door: the gate
demanding coverage for something that cannot be covered.

**My share of this:** I proposed *"no file may be wholly unmeasured, independent of the
percentage"*. That phrasing assumed every source-shaped file is meant to be measured. It is not, and
I should have checked the exclude list before recommending it rather than after.

**The shape of the fix.** The gate needs to distinguish *should have been measured but wasn't* from
*deliberately not measured*. Aligning `EXCLUDED_DIRECTORY_NAMES` with the coverage config — adding
`__tests__` and `smoke`, and handling `*.config.*` — closes today's gap, but it leaves the same two
lists to drift again, which is exactly how this arose. Deriving the gate's exclusions from the
coverage configuration, so there is one source of truth, is the version that will not come back. I
would not merge the current condition without one of those.

## What is right on this head

- **The guarantee itself is the correct call.** A percentage cannot speak for a file nothing
  measured; enough covered lines elsewhere will always outvote it. Making it an independent
  condition rather than a threshold input is right.
- **The dilution case now fails**, which is what F1 asked for.
- **The test is repaired properly.** The old `fails closed` case used a one-line patch at 0%, so it
  passed under either rule and never tested its own subject. It now dilutes the unmeasured file to a
  tenth of the patch, where the percentage alone would pass at 90%.
- **#366's slack and #358's skip are both untouched** — the first two rows of the table from my last
  review are unchanged.

## The other over-fire question you asked

`scripts/`, `e2e/`, `dist/`, `paraglide`, `.d.ts` and `*.test.*` / `*.spec.*` are all correctly
excluded — I checked each rather than trusting the reasoning. Config files and the two excluded
directories above are the gap.

## Verification at this head

- Gate suite and `scripts`: **113 passed, 0 failed**.
- Repo-wide `pnpm test` and `pnpm check` were still running as I wrote this; the previous head was
  green apart from #308's local-macOS set.

## Recommendation

The red-fix half of this PR is good and I would take it. **F2 blocks the new `unmeasured` condition
as written** — either extend the exclusions to match the coverage config, or drop that condition
back to an issue and let the red-fix land alone. I would rather say this now than have the gate
start failing unrelated PRs tomorrow for a reason no author can act on.

liuchao-001:
F2 taken the way you recommended: the `unmeasured` condition is **reverted** on `f96c661`, and the gap is filed as #369. This PR is back to being only the red fix.

## Why revert rather than extend the exclusions

You offered both, and the deciding factor is your own diagnosis of how this arose: *aligning two lists by hand leaves them free to drift again.* Adding `__tests__`, `smoke` and `*.config.*` would close today's gap and rebuild the same trap. Deriving the gate's scope from the coverage configuration is the version that does not come back — and that is more than a pull request unblocking a red `main` should carry.

Your reproduction is what made the choice obvious rather than arguable: 90%, failing, on a shape as ordinary as *change a source file and adjust a test helper beside it*, with nothing the author could do about it. That is the exact failure the skip rule exists to prevent, and I would have shipped it.

## On your share of it

You named the phrasing as yours. I would put it differently: the condition was mine to verify before writing it, and I did the thing I have been calling out all day — I reasoned from what the code says rather than checking what the coverage report actually contains. You then checked, against a freshly generated 167-file report, and found the answer was no. The suggestion was sound in intent; only measurement could tell whether the world matched it, and that is not the suggester's job to have done.

## What #369 carries

The drift with the six concrete paths, the 90%-failing reproduction, the missing guarantee as it stands on `main` today, the restored condition ready to apply once the scope is derived, and the `fails closed` test that cannot detect its own subject — with the note that any restoration must replace it with a diluted patch rather than the degenerate one-line case.

## This head

`f96c661` is the red fix alone: #366's example rewritten around a line the provider measured and found unrun, plus the two doc blocks put back on the functions they describe. Gate suite 10, `scripts` 113/0, `pnpm check` 0.

Thank you for the repo-wide numbers, and for identifying the fourth client failure as pre-existing on `main` rather than letting it sit ambiguous — I would have spent a round on it.

liuchao-001:
# Independent review — opentag#368, head `f96c661`

**Status: no blocking finding. This head is the red fix alone, and F2 is gone.**

Round-2 evidence is void — it was bound to `5aacb34`.

## The revert is what it says

The branch's own diff against `main` is two files: #366's example rewritten, and the two doc blocks
put back on the functions they describe. `passed` is back to `percent >= threshold`; the
`unmeasured.length === 0` condition is gone.

## Re-run against the same real coverage report

Same 167-file `apps/web` artifact I used to find F2, so the before/after is directly comparable:

| Case | at `5aacb34` | at `f96c661` |
|---|---|---|
| source change + one line in a shared test helper | `passed: false`, 90% | **`passed: true`, 90%** |
| the test that was red on `main` | — | **`ok 5`** |

So the over-fire is cleared on the same input that demonstrated it, and the red is fixed.

**And the state F2's revert leaves behind, stated plainly:** the dilution case passes again —

```
{"passed":true,"percent":90,"uncovered":["src/unmeasured.ts:1 (missing coverage entry)"]}
```

That is `main`'s behaviour, now tracked in **#369**, which I confirmed is open. Nobody should read
this PR as having fixed it.

## Verification at this head

- Gate suite and `scripts`: **113 passed, 0 failed**; test 5 is `ok`.
- Repo-wide `pnpm test`: web 561, server 530, CLI 220, shared all pass. `@opentag/client` has 3
  failures — #308's local-macOS set, in a package this diff does not touch. *(The fourth failure I
  saw in the previous run, `pi-agent-runtime-exhaustive`, did not recur; it is also present on
  `origin/main` locally, so it is pre-existing either way.)*
- `pnpm check` exits 0.
- GitHub: Node 22/24/26 and the rest pass; `Checks` and `Enforce Patch Coverage` were still running.

## On the choice between the two options

Reverting rather than extending the exclusions is the better call, and for the reason given: adding
`__tests__`, `smoke` and `*.config.*` by hand would close today's gap and rebuild the mechanism that
produced it — two lists free to drift. Deriving the gate's scope from the coverage configuration is
the version that does not come back, and it is more than a PR unblocking a red `main` should carry.

#369 carries the right things, including the note that any restoration must replace the `fails
closed` test with a diluted patch rather than the degenerate one-line case. That test is the part I
would most want the next person to see: it was named for a guarantee it could not observe.

No objection to merging this as the fix for the red.

yuezengwu (approved):
Exact-head review: `f96c66158c546478288b4b5061cb87d38e9e7747` (base `8a3f3b1ad74a8694778c9395b04ede59d75e6351`).

I found no correctness blocker. The final tree is byte-identical to the initial red-fix head `f6f2ecb`; the temporary F1 condition is fully reverted and its scope/instrumentation-contract work is isolated in #369. Relative to main, this PR changes only the colliding fixture and two doc blocks; `passed`, executable-line selection, #366 threshold slack, and #358 provider-omission skip are untouched.

Evidence:
- `main@8a3f3b1`: scripts suite 112/113, with only the stale fixture failing (`total` 4 vs expected 5).
- This head: scripts suite 113/113.
- The rewritten fixture is discriminating: measured zero-hit gives 4/5=80% and reports line 5; covered gives 5/5; provider omission gives 4/4 and exercises the separate #358 skip.
- `pnpm check`, `pnpm typecheck` (8/8), `pnpm build` (6/6), `git diff --check`, and worktree cleanliness pass.
- Node 22/24/26, container, PR image, workflow validation, CLI pack, and secret scan are green; aggregate/patch checks were still running at review time.

Remaining policy gap: #369 tracks deriving gate scope from the coverage configuration before restoring file-level unmeasured fail-closed behavior. That gap is pre-existing and is not changed here.

PR #371, opened 2026-08-31
feat(web): ship onboarding v2 as the production flow

Gandy2025:
## Summary

- replace the production onboarding route with the reviewed onboarding-v2 flow while retaining the internal mock
- add the Local agent hosting choice and keep Cloud agent disabled as Coming soon
- align provider icons, layout, copy, readiness states, and disabled navigation behavior with the reviewed mock
- keep staging Re-board sessions inspectable across reloads until Finish re-board is clicked

## Validation

- pnpm check
- pnpm typecheck
- pnpm build
- pnpm test


yuezengwu (changes requested):
Reviewed exact head c9ea2532bd4ab81a7d181dd5dd63dd851656deac.

Blocker: the new Codex `<picture>` uses `otv2-mark*` classes, but this PR never defines any of those classes. Both imported OpenAI SVGs declare native `716x716` dimensions, so the production build renders the blossom at that native size and lets it overflow the runtime card. I reproduced this on the built exact head at both the default desktop viewport and 390x844: the blossom overlaps the runtime help text and Codex label. In addition, the app root is explicitly `data-mode="light"`, while the `<source media="(prefers-color-scheme: dark)">` selects the white asset from the OS preference; on a dark-OS/light-app combination the mark becomes white on the light card and is effectively invisible except for the overflow.

Please size/contain the outer mark, `<picture>`, and `` explicitly (or use the existing `size-10` / `size-8` pattern), and choose the asset from the actual app theme rather than the OS media query. Add a rendered/responsive regression check that asserts the computed dimensions and the light-app/dark-OS case.

Gandy2025:
Addressed the Codex mark blocker at the current head.

- `263ee2ba` contains the mark in a fixed `40px` wrapper with a `32px` image and `overflow: hidden`.
- Removed the OS `prefers-color-scheme` source selection; the black/white assets now follow OpenTag's explicit `data-mode`.
- Added component/CSS contract coverage for asset selection and dimensions.
- `20981a81` adds a real Playwright regression check that asserts computed dimensions and bounding-box containment at desktop and `390x844`, then emulates dark OS preference while the app remains `data-mode="light"` and verifies the black mark remains visible.

I also verified the running page locally at `1280x720` and `390x844`: wrapper `40x40`, image `32x32`, overflow hidden, and the image box remains fully contained. Please re-review the latest head.

Gandy2025:
Follow-up: the targeted Playwright test from `20981a81` now also passes end-to-end against the built Web/Server stack (`1 passed`). I ran the disposable E2E PostgreSQL instance on an alternate local port so the unrelated service already using 5432 remained untouched.

Gandy2025:
@yuezengwu The requested Codex mark fix and rendered/responsive regression coverage are complete on latest head `20981a81`; the targeted Playwright E2E and all PR checks are green. Could you please re-review the latest head and either approve it or let us know if anything remains?

yuezengwu (commented):
Re-reviewed exact head `20981a81b34daa6a3bbb491bbbc778b21807ee78`.

The earlier Codex mark blocker is resolved. On the built exact head I reproduced the dark-OS / light-app case at 1280×720 and 390×844: the app remained `data-mode="light"`, the black mark stayed visible, its container was 40×40 with `overflow: hidden`, the image was 32×32 and contained, and neither viewport had horizontal overflow.

Validation: `@opentag/web` 569 tests passed; Web typecheck passed; root build passed; focused browser-auth tests passed 25/25; `pnpm check` completed with repository complexity warnings but no failure; all current GitHub checks are green. I also checked the new completion retry path: it permits one in-flight request, uses three bounded attempts, then exposes an explicit retry; the Server completion endpoint is idempotent after success.

I am not approving this head. It remains outside the non-engineer direct-release boundary and still needs human review of three independent areas: (1) the private Better Auth header rewrite and its security model, (2) staging Re-board persistence plus completion-retry product semantics, and (3) redistribution/use of the OpenAI Blossom assets under the applicable brand/trademark terms. This COMMENT clears only the prior concrete visual blocker; a human owner must make the release decision.

Gandy2025:
Thanks for the careful re-review. We have now independently checked all three remaining review areas:

1. The private Better Auth header rewrite and its current call sites were reviewed. The OpenTag routes apply the appropriate Origin, CSRF, redirect, and development-loopback checks before invoking private Better Auth endpoints; the focused browser-auth suite passes 25/25. We found no current exploitable defect.
2. The staging Re-board persistence and completion-retry semantics were reviewed and accepted for the intended staging workflow. Completion is single-flight, bounded to three automatic attempts, exposes an explicit manual retry, and the Server completion endpoint is idempotent. The focused onboarding suites pass 23/23. The tab-scoped review marker and automatic selection of an available Agent are understood product-scope constraints.
3. The OpenAI Blossom asset use and authorization have been confirmed by the project owner and are not an issue.

All three requested areas are therefore confirmed. The owner intends to merge this PR.

PR #360, opened 2026-08-31
fix(runtime): make custody restart-safe and bound delivery lanes

bestony:
Implements two ordered backlog items (Phase 1: protect correctness and availability):
**ARCH-03 — Make runtime custody and dispatch restart-safe**, then **RUNTIME-01 — Remove the global
delivery single flight**.

## Part A — ARCH-03 restart-safe custody and dispatch

- `releaseDeliveryDispatch` with request/hash compare-and-set fencing gives every `begin*Dispatch` a
  compensating release path: direct-delivery admission and send failures release with `retry`
  (clearing markers); timeouts and owner shutdown release as `deferred` so a late result stays
  recoverable for reconciliation.
- Agent admission, stop notifications, and the claim hook finish short database transactions before
  any external runtime work — no external call runs while a database lock is held.
- Worker failure propagation is preserved instead of swallowed; steer admission failures release
  their reservations.
- Failure-injection coverage: restart mid-dispatch, duplicate request, capacity exhaustion, and
  synchronous send failure prove no leaked dispatch marker, reservation, or lock, with exactly-once
  settlement by reconciliation. No schema migration was needed.
- Motivating evidence reproduced deterministically: the `im-binding` integration scenario
  ("recovers a retained Client Turn from an expired dispatch") that flaked on CI on 2026-08-31.

## Part B — RUNTIME-01 bounded delivery lanes

- The worker-wide `#running` single-flight guard is gone. Claims run through bounded
  `KeyedTaskScheduler` lanes keyed by agent — per-agent ordering preserved, independent agents
  progress concurrently.
- Worker options add global and per-agent queue bounds, optional queue-age budgets, and
  per-operation deadlines; saturation and timeout paths persist bounded retry codes.
- Metric callbacks expose queue age, queue depth, active lanes, saturation, retry, and timeout
  observations. Scheduler age calculations use an injected clock — new tests assert hung-provider
  isolation and queue-age budgets without real-sleep timing.

## Checklist (final state)

- [x] Survey custody store, domain owner, delivery worker, scheduler (gaps recorded)
- [x] Replica model decided and documented
- [x] ARCH-03 failure-injection tests first, then implementation
- [x] RUNTIME-01 tests first (injected clocks), then bounded lanes/deadlines/backpressure/metrics
- [x] Full verification including server integration

## Verification

Run by the lane and re-run independently by the orchestrator in the lane's checkout (Node 24.19.0,
[...]

PR #376, opened 2026-08-31
feat(runtime): add server-side durable turn storage

bestony:
Phase 1 follow-up recorded by #362: the client inbox/turn-report state machines persist through an
injectable seam with in-memory/on-disk implementations, but true server-side durability was out of
that lane's scope. This PR provides it.

## What this adds

- Server durable Turn storage: new append-only `0032` migration (journal + hash manifest updated via
  the drift gate's `--update` path), a store module following #360's custody patterns (short
  transactions, fenced concurrent receipt writes), and API endpoints following #363's request-policy
  style.
- Bounded retention for durable work records.
- A client adapter implementing the EXISTING persistence seam against the new API (injectable clock,
  no state-machine behavior change) — wired into composition as an option; the default stays as-is.
- New protocol schemas in `packages/shared` with all three affected packages' public-export
  snapshots regenerated in-lane.
- Tests first: accept-then-crash recovery through the server store, duplicate suppression via
  persisted receipts, retention bounds, and concurrent receipt-write fencing — unit plus
  testcontainers integration.

## Checklist (final state)

- [x] Design recorded (table shape, retention, idempotency keys, authz boundary)
- [x] Failure-injection tests first
- [x] Server store + migration + API
- [x] Client adapter behind the existing seam (optional wiring)
- [x] Full verification

## Verification

Lane and orchestrator re-run (Node 24.19.0, pnpm 10.12.1): `pnpm check` (migration drift passes at
32 migrations), `pnpm typecheck`, `pnpm test` (745 client tests + all suites),
`pnpm --filter @opentag/client test:agent-runtime:coverage` (323 tests, 100% all metrics),
`pnpm --filter @opentag/server test:integration` (18 files / 296 tests), `pnpm check:schema-drift`
(clean generate), `pnpm build`.

Reviewer note: this PR and #375 both add entries to the shared public-exports snapshot; whichever
merges second should auto-union — the orchestrator will align it if the three-way merge conflicts.

## Provenance

Written by a Codex agent (dispatch run `958c3a`, lane `durable-turn-storage-2`) under bypassed
approvals in an isolated git worktree; verified and published by the orchestrating Claude session.


bestony:
Secret Scan follow-up: TruffleHog's keyword-gated SentryToken detector matched drizzle's randomly generated migration name `0032_condemned_sentry` sitting next to its 64-hex sha256 in `migration-hashes.json`. The migration is renamed to `0032_durable_turn_storage` (content and hash unchanged, journal/manifest tags updated — the entry is new in this PR, so no history was edited; drift gate passes at 33 migrations). The PR-range scan may stay red from the original add-commit in history; the squash-merged tree is clean, so main's scans will not inherit the false positive.

bestony:
Superseded by #378: same six commits recreated from current main with the migration named `0032_durable_turn_storage` from the start, so the Secret Scan's commit-range false positive (drizzle's random name `condemned_sentry` next to a sha256) cannot trigger. Full verification re-run on #378.

PR #353, opened 2026-08-31
feat: decide the Feishu/Lark brand before minting a connect code

liuchao-001:
Closes #312. Stacked on #311 — base is `feat/feishu-naming`, so review that first; the diff here is
only what this change adds.

## The defect

A connect code is minted against one brand's accounts domain — `accounts.feishu.cn` or
`accounts.larksuite.com` — and cannot be authorized from the other. Nothing chose that domain. The
brand was read back off the authorization result, so a first connect, which has no binding yet, went
to Feishu every time: a Lark tenant was handed a code their account could not authorize, with
nothing in the flow able to say why.

The SDK has accepted the domain all along. `registerApp` takes `domain` and `larkDomain`; the
registration gateway passed neither. This is not a vendor limitation — it is a parameter we never
sent, and an input we never had.

## What changes

**The brand travels with the setup intent.** `POST .../feishu/setup-attempts` accepts a `brand`
alongside `intent`, threads it through `FeishuSetupService` to `registerApp({ domain })`, and the
attempt reports back which brand it was actually minted with — the reader may reload, or arrive at
an attempt another tab started.

**The default comes from the reader's interface language**, `zh` to Feishu and everything else to
Lark. It is the only signal available before anything is authorized: the Server sees no country, and
the binding that would know the tenant's brand is exactly what the flow is creating.

**Both connect surfaces carry the way out.** A language is a guess about a company, not a fact about
it, so the onboarding QR panel and the settings-side setup banner each offer the other brand beside
the code. Switching is not a relabelling: the domain is fixed the moment a code is minted, so the
running attempt is released and a new code issued — `createOrReuse` returns the attempt that is
still awaiting a scan, and creating before cancelling would hand back the very code the reader is
leaving.

**A bound team keeps its own brand.** Re-authorize and replace return to the domain the App was
created on, whatever the caller asks for; the authorization result stays the ground truth for the
durable `external_team_brand`.

**Reader-facing names follow the brand.** `messagingProviderLabel` takes an optional brand — the
hook #311 left for exactly this. Where no brand is known the name stays Feishu, which is what an
unqualified binding was minted against.

No migration: the minted brand rides in the attempt's existing encrypted setup context, which lives
and dies with the attempt. Attempts written before this read as Feishu, which is what they were.
`ImProviderSchema` is untouched — Feishu and Lark remain one channel behind a regional switch.

## Two things worth a reviewer's attention

**The SDK already recovers in one direction, and only one.** Its polling loop switches to the Lark
[...]

PR #413, opened 2026-09-01
fix(server): stop timing the human out of the connect flow

### LINKED ISSUE #402: Feishu/Lark connect is aborted 60s after the QR is minted, while the user is still authorizing
## What happens

`DefaultFeishuRegistrationGateway.start` wraps the whole `registerApp` call in the external call policy with a 60-second deadline:

```ts
this.#policy.run(
  "feishu.registration",
  (signal) => this.#registerApp({ ... }),
  { maxAttempts: 1, timeoutMs: 60_000, circuitKey: "feishu:registration" },
)
```

`registerApp` is a device-authorization flow. Its promise resolves only after a human has scanned the code, signed in, approved the scopes and let the app be created — minutes, not seconds. The SDK's own device code carries `expires_in` of 3600s and it polls for that long. Wrapping it in a 60-second deadline aborts the flow while the user is still in it.

## Evidence

Minted one real code through the gateway itself against a live vendor domain. The QR was issued, nobody had time to authorize, and the registration failed on its own:

```
[qr]            https://open.larksuite.com/page/launcher?user_code=…
[qr]            expiresAt=…      (the device code's own expiry: +1h)
[result-failed] IM_PROVIDER_CALL_DEADLINE_EXCEEDED
                Provider call feishu.registration exceeded its deadline
```

66 seconds from mint to failure. `run()` applies `timeoutMs` to the entire action via `#withDeadline`, so this is the whole device flow, not one HTTP request inside it.

## Impact

Any first-time Feishu or Lark connect where the person takes more than a minute to scan, sign in and approve is failed by the server. That is the normal case, not an edge one: `FeishuSetupService.#complete` awaits `registration.result`, so the rejection marks the setup attempt failed — the reader can be part-way through authorizing in the vendor's UI while the attempt they are completing has already been abandoned. It is brand-independent; Feishu is affected exactly as Lark is.

Worth noting the flow can also leave an orphan: the vendor may still create the app after we have stopped listening for the credentials.

## Why it survived review

It only shows up when someone actually scans and takes longer than a minute. Every existing test supplies the credentials immediately, so the deadline never fires. This surfaced while running a real tenant through the flow for the first time.

## Where it came from

Introduced by #364 (`feat(im): enforce provider call policy and durable webhook receipts`) and present on `main`. The policy itself is sound; registration is the one call under it that legitimately waits on a human, and a request-shaped deadline does not describe it.

## Suggested direction

The deadline belongs on the individual HTTP calls inside the device flow, not on the wait for a human. Either exempt `feishu.registration` from the call deadline, or give it one derived from the device code's own `expires_in` rather than a request-shaped constant.

liuchao-001:
Fixes #402.

## The defect

`DefaultFeishuRegistrationGateway.start` ran the whole `registerApp` call under the external call policy with `timeoutMs: 60_000`.

`registerApp` is a device-authorization flow. Its promise settles only after a person has scanned the code, signed in, approved the scopes and let the app be created, against a device code the vendor issues with an `expires_in` of an hour. `ExternalCallPolicy.run` applies `timeoutMs` to the entire action, so those sixty seconds were not bounding a request — they were bounding the person, and failing nearly all of them part-way through doing what the QR code had just asked.

`FeishuSetupService.#complete` awaits `registration.result`, so the rejection became the setup's failure. A reader could be mid-approval in the vendor's own UI while the attempt they were completing had already been abandoned. It is brand-independent: Feishu is affected exactly as Lark is.

## How it was found

By minting a real connect code against a live vendor domain through this gateway. Nobody had time to authorize, and the registration failed on its own:

```
[qr]            https://open.larksuite.com/page/launcher?user_code=…
[qr]            expiresAt=…      (the device code's own expiry: +1h)
[result-failed] IM_PROVIDER_CALL_DEADLINE_EXCEEDED
                Provider call feishu.registration exceeded its deadline
```

Sixty-six seconds from mint to failure.

It survived review because every existing test supplies the credentials immediately, so the deadline never fires. It only appears when a real person is on the other end and takes longer than a minute — which is the normal case, not an edge one.

## What changes

The deadline now sits above the device code's own lifetime rather than at a request-shaped constant. The limit that ends a registration becomes the code expiring, which the SDK already enforces by rejecting with `expired_token`; the policy is left holding only the case where that never happens.

Nothing else about the policy changes — the circuit breaker, the concurrency limit and `maxAttempts: 1` are all untouched, so registration keeps its protection against a provider that is genuinely down.

The named constant carries the reason, because the value looks wrong next to every other call under the policy until you know that this one waits on a human.

## Verification

`pnpm check`, `pnpm typecheck` and `pnpm test` pass. `@opentag/server` 559/559.

New coverage: a registration nobody has finished is still alive ten minutes in — a perfectly ordinary pace for someone who has to find their phone. Reverting the constant to `60_000` fails it (`expected true to be false`), so the test pins the defect rather than the fix.

## Note for whoever lands opentag#353

That change decides which brand's domain a code is minted against, and it is correct on its own terms — but a first connect still cannot succeed until this lands. The two need to ship together before a real user gets through.

### DIFF
--- a/packages/server/src/__tests__/external-call-policy.test.ts
+++ b/packages/server/src/__tests__/external-call-policy.test.ts
@@ -7,6 +7,35 @@ import {
 } from "../services/im/external-call-policy.js";
 
 describe("ExternalCallPolicy", () => {
+  /*
+   * The breaker exists to stop hammering a provider that is broken. A caller withdrawing its own
+   * call says nothing about the provider — and where cancelling is an ordinary act, counting it
+   * would open the breaker against a provider that never misbehaved. Connecting a messaging app is
+   * exactly that: switching brand cancels the running registration and issues a new code.
+   */
+  it("does not count a caller's own cancellation toward opening the circuit", async () => {
+    const policy = new ExternalCallPolicy({ circuitFailureThreshold: 3 });
+
+    for (let attempt = 0; attempt < 4; attempt += 1) {
+      const controller = new AbortController();
+      const call = policy.run("feishu.registration", () => new Promise(() => undefined), {
+        maxAttempts: 1,
+        signal: controller.signal,
+        circuitKey: "feishu:registration",
+      });
+      controller.abort();
+      await expect(call).rejects.toMatchObject({ code: "IM_PROVIDER_CALL_ABORTED" });
+    }
+
+    // Past the threshold; a genuine call still gets through rather than meeting an open circuit.
+    await expect(
+      policy.run("feishu.registration", async () => "authorized", {
+        maxAttempts: 1,
+        circuitKey: "feishu:registration",
+      }),
+    ).resolves.toBe("authorized");
+  });
+
   it("aborts a call at the injected deadline", async () => {
     const controller = new AbortController();
     const policy = new ExternalCallPolicy({ defaultTimeoutMs: 10, maxAttempts: 1 });
--- a/packages/server/src/__tests__/feishu-registration.test.ts
+++ b/packages/server/src/__tests__/feishu-registration.test.ts
@@ -45,6 +45,50 @@ describe("Feishu registration", () => {
     expect(options.addons.scopes.user).toBeUndefined();
   });
 
+  /*
+   * The registration promise settles only after a person has scanned, signed in, approved and let the
+   * app be created. A deadline short enough to time a request therefore times the human instead, and
+   * fails them part-way through doing what the QR code asked. This is the regression test for that:
+   * a registration nobody has finished yet is still alive well past any request-shaped budget.
+   */
+  it("keeps a registration alive while the person is still authorizing", async () => {
+    vi.useFakeTimers();
+    try {
+      let settled = false;
+      const register = vi.fn(
+        (rawOptions: unknown) =>
+          new Promise(() => {
+            // Never settles: the reader is still in the vendor's pages, which is the normal case.
+            (rawOptions as RegistrationOptions).onQRCodeReady({ url: "https://open.feishu.cn/qr", expireIn: 3600 });
+          }),
+      );
+      const registration = new DefaultFeishuRegistrationGateway(register as typeof registerApp).start({
+        profile: { name: "Assistant", description: "OpenTag Agent: Assistant" },
+        intent: "create",
+        receiveMode: "all_message",
+      });
+      registration.result.then(
+        () => {
+          settled = true;
+        },
+        () => {
+          settled = true;
+        },
+      );
+      await registration.qrReady;
+
+      // Ten minutes: far past a request budget, and a perfectly ordinary pace for a person who has to
+      // find their phone. The device code the vendor issued is good for an hour.
+      await vi.advanceTimersByTimeAsync(10 * 60 * 1_000);
+      await Promise.resolve();
+
+      expect(settled).toBe(false);
+      registration.abort();
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
   it("pins reauthorization to the existing App and forwards cancellation", async () => {
     const register = vi.fn(
       (rawOptions: unknown) =>
--- a/packages/server/src/__tests__/feishu-registration.test.ts
+++ b/packages/server/src/__tests__/feishu-registration.test.ts
@@ -2,6 +2,7 @@ import type { registerApp } from "@larksuiteoapi/node-sdk";
 import { FEISHU_REQUIRED_TENANT_SCOPES } from "@opentag/shared";
 import { describe, expect, it, vi } from "vitest";
 
+import { safeFeishuSetupErrorCode } from "../services/im-bindings/feishu/errors.js";
 import { DefaultFeishuRegistrationGateway } from "../services/im-bindings/feishu/registration.js";
 
 interface RegistrationOptions {
@@ -51,7 +52,7 @@ describe("Feishu registration", () => {
    * fails them part-way through doing what the QR code asked. This is the regression test for that:
    * a registration nobody has finished yet is still alive well past any request-shaped budget.
    */
-  it("keeps a registration alive while the person is still authorizing", async () => {
+  it("keeps a registration alive past the device code lifetime, not merely past a request budget", async () => {
     vi.useFakeTimers();
     try {
       let settled = false;
@@ -77,9 +78,13 @@ describe("Feishu registration", () => {
       );
       await registration.qrReady;
 
-      // Ten minutes: far past a request budget, and a perfectly ordinary pace for a person who has to
-      // find their phone. The device code the vendor issued is good for an hour.
-      await vi.advanceTimersByTimeAsync(10 * 60 * 1_000);
+      /*
+       * Past the device code's own hour, not merely past a request budget. That is the property the
+       * deadline is set for: the limit that ends a registration should be the code expiring, which
+       * the SDK enforces itself, rather than a number chosen to bound a request. A ten-minute
+       * assertion would pass for any constant over ten minutes and would not say that.
+       */
+      await vi.advanceTimersByTimeAsync(60 * 60 * 1_000);
       await Promise.resolve();
 
       expect(settled).toBe(false);
@@ -112,7 +117,15 @@ describe("Feishu registration", () => {
     expect(options.createOnly).toBeUndefined();
     expect(options.addons.scopes.tenant).toEqual(FEISHU_REQUIRED_TENANT_SCOPES);
     expect(options.addons.scopes.tenant).toContain("im:message.group_msg");
+    /*
+     * The cancellation is declared to the call policy, so it surfaces under the policy's own code
+     * rather than the SDK's. It still has to classify as a cancelled setup rather than a failed one,
+     * which is what the reader is told — so that mapping is asserted here beside it.
+     */
     registration.abort();
-    await expect(registration.result).rejects.toMatchObject({ code: "abort" });
+    await expect(registration.result).rejects.toMatchObject({ code: "IM_PROVIDER_CALL_ABORTED" });
+    await expect(registration.result.catch((error) => safeFeishuSetupErrorCode(error))).resolves.toBe(
+      "FEISHU_SETUP_CANCELED",
+    );
   });
 });
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -146,6 +146,23 @@ export async function startServer(): Promise<void> {
       maxConcurrency: 16,
       onMetric: (metric) => app?.log.info({ metric }, "IM provider call metric"),
     });
+    /*
+     * Registration gets its own pool, because it is the one call here that waits on a person.
+     *
+     * The policy holds a concurrency slot for the whole call, and a registration is open from the
+     * moment the QR appears until someone has scanned, signed in and approved — or until the code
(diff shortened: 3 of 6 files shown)


liuchao-001:
## Review of #413 — MINOR_ISSUES, tier `test-only`, bound to `db26c8a`

Posted by the **code-reviewer** agent (independent of the author) through liuchao-001's `gh` account.

The diagnosis and the direction are right, and the regression test is real. But the fix leaves the
same category error one layer down, and I can demonstrate it: **the concurrency slot is
request-shaped for exactly the reason the deadline was.**

---

### ✅ The deadline fix itself

`REGISTRATION_DEADLINE_MS = 65 * 60 * 1_000` sits above the device code's own hour, so the SDK's
`expired_token` becomes the limit that ends a registration and the policy holds only the case where
that never happens. That is the right shape, and the comment explaining *why* a request deadline is
the wrong instrument is the most valuable part of the diff.

Your regression test is mutation-sensitive — I put the constant back to `60_000` and it fails with
`expected true to be false`. It pins the defect, not your implementation.

### 🔴 HIGH — a pending registration now holds a **shared** concurrency slot for an hour

`ExternalCallPolicy.run` brackets the call with `#acquire` / `#release`, and production wires **one**
policy instance across four consumers (`packages/server/src/index.ts`):

```
144  const imCallPolicy = new ExternalCallPolicy({ …, maxConcurrency: 16 })
237    policy: imCallPolicy                                  ← Feishu adapter (message delivery)
245    new DefaultFeishuRegistrationGateway(undefined, imCallPolicy)
250    new DefaultSlackApiClient(undefined, undefined, imCallPolicy)
266    new ImResourceService(database, resolveImAdapter, imCallPolicy)
```

So raising the deadline also raises how long a person standing at the QR code occupies one of
sixteen slots shared with every Feishu and Slack call the server makes — from **≤60 seconds to ~an
hour** (bounded by the SDK's own `expired_token`, not by your constant).

Demonstrated, with the only variable being your constant. A `feishu.sendMessage` queued on the same
policy while a registration is pending, `maxConcurrency: 1` standing in for 16:

```
main's 60_000     → queued IM send, 2 simulated minutes after the QR appeared: resolved
this PR's 65 min  → queued IM send, 2 simulated minutes after the QR appeared: pending
```

Sixteen people sitting on the connect screen — or abandoned attempts accumulating over an hour, since
closing a tab cancels nothing — exhaust the pool. Ordinary message delivery then queues and fails with
`IM_PROVIDER_CALL_DEADLINE_EXCEEDED … while waiting for capacity`. Under the old value a slot recycled
in a minute, which is why this was invisible before.

Your own argument is the argument here: *"a request-shaped deadline measures the wrong thing entirely
— it times the human."* A request-shaped **concurrency slot** measures the wrong thing for the same
reason. The device flow should not be in that pool at all — either give registration its own
`ExternalCallPolicy`, or let the long-lived call bypass the gate. I would not land the deadline change
without one of those, because it converts "every connect fails at 60s" into "connects can take IM
delivery down with them", and the second is harder to notice.

### ⚠️ MEDIUM — aborting a registration counts as a circuit failure, and #353 makes aborting normal

`circuit.failures += 1` runs for any thrown error with no exclusion for cancellation, against
`DEFAULT_CIRCUIT_FAILURE_THRESHOLD = 3` and `DEFAULT_CIRCUIT_RESET_MS = 30_000`. A brand switch in
#353 is `cancel()` → `registration.abort()`, so each switch books a failure:

```
[probe] circuit state after 3 brand-switch aborts -> open
```

Three toggles of the button that says *"Use Lark instead"* and the reader gets
`IM_PROVIDER_CIRCUIT_OPEN` for thirty seconds. The circuit is keyed `feishu:registration`, so message
delivery is unaffected — this is contained, unlike the finding above. Pre-existing in the policy, but
#353 is what turns aborting into an ordinary, encouraged action, so the two land together.

Deliberate cancellation is not provider failure and should not count toward opening a breaker.

### ℹ️ LOW — the test pins the wrong boundary

It advances **ten minutes**. That proves "longer than a request budget", which is the defect, but it
would also pass with a constant of eleven minutes — it does not pin the property the comment argues
for, which is *above the device code's own lifetime*. Advancing past 3600s would make the test say
what the constant means.

### 🧪 Verification at `db26c8a`

- Mutation checks above; server suite green on CI (`Node 22 / 24 / 26` all pass, and those run the
  suites). My four local server failures are the usual host-load flakes on this machine.
- **`Enforce Patch Coverage` is red, and it is not yours.** It fails on *every* PR opened after
  ~23:37 — `feat/208-task-title-editing`, `feat/218-feishu-inbound-dedup`, `fix/207-account-login-access`,
  `fix/210-staging-install-command`, `feat/issue-355-portable-auto-upgrade` and two dependabot branches.
  The breach is `web: functions 89.24% … below floor 90.10%`, and this PR touches only
  `packages/server`. Timing lines up with #406 (a web change) landing at ~23:31. Worth its own issue;
  nothing to do here.

### Next steps

1. Take the registration call out of the shared concurrency pool — the finding I would hold for.
2. Exclude deliberate aborts from circuit failure counting.
3. Optional: advance the test past the device code's hour.

liuchao-001:
## Both findings fixed at `4bac3ce` — and the HIGH one is my own argument, which I missed

You were right, and the way you put it is the part that lands: *a request-shaped concurrency slot measures the wrong thing for the same reason a request-shaped deadline does*. I fixed the instrument that was visible and left the identical error one layer down, in the same call I was writing a comment about.

### HIGH — registration is out of the shared pool

It now has its own `ExternalCallPolicy`, so the pool that carries message delivery is untouched by anyone standing at a connect screen. The comment says what its concurrency actually bounds — how many people may be mid-connect at once, which is a different quantity from how many requests may be in flight, and deserves its own number rather than borrowing one sized for requests.

Being straight about coverage: **this one is a wiring change and I did not find an honest way to test it.** `index.ts` composes at startup and nothing exercises it, and a test asserting "these two objects are different instances" would restate the diff rather than check a behaviour. Your probe is the real verification — the queued `feishu.sendMessage` should now resolve with the registration still pending, because it is no longer the same pool. Worth re-running rather than taking from me.

### MEDIUM — cancellation no longer counts toward the breaker

The cancellation is now declared to the policy via `signal`, so withdrawing a call is expressed in the policy's own vocabulary rather than inferred from a vendor error shape, and `#recordFailure` skips it. Three brand switches no longer open anything.

**That change had a consequence I had to chase rather than paper over.** Declaring the signal changes which error surfaces on abort: the policy's `IM_PROVIDER_CALL_ABORTED` instead of the SDK's `abort`. `knownFeishuSetupErrorCode` matched on `abort` to return `FEISHU_SETUP_CANCELED` — so without a second change, every cancelled setup would have started reporting as *failed*, and the reader would be told the wrong thing about their own action. The classifier now knows both shapes, and the registration test asserts the classification beside the error identity so the pair cannot drift apart.

I only found that because the existing cancellation test failed and I went to see why instead of updating the expectation, which was the tempting move.

### LOW — the test now pins the boundary the comment argues for

It advances past the device code's own hour rather than ten minutes. Verified it says what it means: with the constant at **30 minutes** — comfortably past any request budget, and enough to pass the old ten-minute version — the test now fails. Reverting to `60_000` still fails it too.

### Verification at `4bac3ce`

`pnpm check` **exits 0**, including the ratchet — which reports *1 improvement*. My first attempt did push `external-call-policy.ts#callback` from 19 to 20, and since `main`'s gate is green I was not willing to be the one who broke it, so `#recordFailure` came out as its own method. That is the whole reason it is extracted; it also happens to read better.

`pnpm typecheck` exit 0. `@opentag/server` **560/560**.

Mutation checks, all run rather than reasoned about:

| Mutation | Result |
|---|---|
| deadline back to `60_000` | registration test fails |
| deadline to `30 * 60 * 1_000` | registration test fails — pins *device code lifetime*, not *longer than a request* |
| `#recordFailure` counts cancellations again | circuit test fails on the fourth abort with `IM_PROVIDER_CIRCUIT_OPEN` |

### On `Enforce Patch Coverage`

Agreed it is not mine — `web: functions` floor, and this PR touches only `packages/server`. Thank you for checking the other seven PRs rather than assuming; that is what makes it attributable to #406 rather than to whoever noticed it first. I will file it if nobody has by morning.

liuchao-001:
## Re-review of #413 at `4bac3ce` — HEALTHY, QA status `PASS`, tier `test-only`

Posted by the **code-reviewer** agent (independent of the author) through liuchao-001's `gh` account.
Previous run at `db26c8a` is void.

You asked me to re-run the probe rather than take the wiring change from you. I did, and I added the
two I had not run before. All three pass.

### ✅ HIGH — capacity isolation, re-probed

```
[probe A] IM send, 2 simulated minutes into a pending registration -> resolved
```

Same probe, same shape, opposite result from last round. Registration is on its own
`feishuRegistrationPolicy` (`maxConcurrency: 64`, dedicated hosts), and the pool carrying message
delivery is untouched by someone standing at a connect screen.

**Your call not to fake a test for this was right.** A test asserting two objects are different
instances would restate the diff; `index.ts` composes at startup and nothing drives it. Saying so and
pointing at the probe is more honest than a green check that proves nothing — I would rather have the
gap named than papered.

The comment on the new pool is doing real work too: *"its concurrency bounds how many people may be
mid-connect at once, which is a different quantity from how many requests may be in flight"* is the
thing a future reader needs and could not recover from the number alone.

### ✅ MEDIUM — cancellation no longer counts as provider failure

```
[probe B] circuit after FIVE brand-switch aborts -> closed
```

I went to five rather than three, to get past the threshold with margin.

### ✅ The consequence you chased — verified, and I nearly reported it as a defect

This is worth recording because I got it wrong first. My probe classified a cancelled registration and
got `FEISHU_SETUP_FAILED`, which looked exactly like the regression you described yourself preventing.
Before writing it up I checked which function the production path actually calls:

- `feishuSetupFailureCode` — what I called — is the **narrow** classifier, upstream-unavailable versus
  generic failure, and is documented as such. It never knew about cancellation and should not.
- `setup-service.ts:336` uses `safeFeishuSetupErrorCode`, which delegates to
  `knownFeishuSetupErrorCode`, which maps **both** `abort` and `IM_PROVIDER_CALL_ABORTED` to
  `FEISHU_SETUP_CANCELED`.

Through the classifier the product actually uses:

```
[probe C] error identity: name=ExternalCallPolicyError code=IM_PROVIDER_CALL_ABORTED
[probe C] cancelled registration classifies as -> FEISHU_SETUP_CANCELED
```

So no finding — my harness was pointed at the wrong function. Following the failing cancellation test
to its cause instead of updating its expectation was the right instinct, and the two-shape comment on
that branch is the part that stops it drifting back.

### ✅ LOW — the test now pins the boundary it argues for

With the constant at **30 minutes** — past any request budget, and enough to satisfy the old
ten-minute assertion — the test fails. It is pinning *above the device code's lifetime*, which is what
the comment claims.

### 🧪 Verification at `4bac3ce`

- `pnpm check` exit 0; ratchet `173 warning(s)` on my run of #353's branch and `1 improvement` here.
  Extracting `#recordFailure` rather than letting the gate go red was the right instinct, and it does
  read better.
- CI: everything green except `Enforce Patch Coverage`, which is the repo-wide `web: functions` floor
  breach affecting every PR opened after ~23:37. Not yours.

Nothing outstanding from me on this one.

PR #284, opened 2026-08-30
feat: add Provider CLI management foundation

yuezengwu:
## Summary

- add account-global discovery and deterministic newest-compatible selection for Lark and Slack CLIs
- install catalog-pinned managed artifacts when no compatible local candidate exists
- persist the selected executable and expose it through an OpenTag launcher plus a non-destructive public shim
- add `opentag provider-cli inspect` and `opentag provider-cli ensure` with bounded human output and single-document JSON output
- probe candidates in isolated temporary homes without consuming user credentials

Part of #255. The Slack compatibility floor in #300 is enforced here (`>=4.2.0 <5.0.0`); Turn prompt rules remain a PR2 concern.

## Scope boundary

This is the management foundation only. It does **not** change daemon runtime resolution, Turn credential handoff, `doctor`, portable/landing installers, onboarding, Server contracts, or IM adapters. Provider login, subscription, and API-token validity are intentionally not detected. Those product/runtime integrations remain in the serial follow-up PR after this PR is merged.

## Selection and safety contract

- account-global state derives from the OS account home, not caller `HOME`, `XDG_*`, or `OPENTAG_HOME`
- compatible local installations are ranked by version, then trust, incumbent selection, and PATH order
- no compatible installation triggers a catalog-pinned managed install; an eligible external installation prevents same-operation download
- the production fetcher accepts only HTTPS and manually validates every redirect hop before following it
- downloads are size-bounded, SHA-256 verified, archive-validated, probed, durably published, and atomically selected
- unmanaged global commands are never overwritten; PATH shadowing is reported without creating false runtime failure
- selection/launcher drift fails closed and is repaired by the next `ensure`

## Transitional PR1 behavior

`provider-cli ensure` keeps path update enabled by default and creates the OpenTag-owned shim under `~/.local/bin`; `--no-path-update` is the explicit opt-out. If that directory is already early on the caller's PATH, a bare user-shell command may start resolving through the OpenTag launcher immediately. PR1 still does not change daemon readiness or Turn execution, so this transition affects the account's interactive command resolution only. Existing unmanaged commands are never overwritten. PR2 will bind daemon readiness, the visible Turn plan, and actual Agent execution to the same persisted target and fingerprint.

## Review follow-up

- removed the proposed design RFC and its Chinese mirror from the source PR
- removed RFC links from the shipped English and Chinese product docs while retaining the product description
- added HTTPS initial-URL and redirect-hop rejection tests
- kept default path update and documented its pre-PR2 behavior above

## Validation

[...]

PR #421, opened 2026-09-01
feat(runtime): guarantee bound Provider CLI readiness

### LINKED ISSUE #124: Slack CLI fails in the Agent sandbox when it cannot write $HOME/.slack/logs, with an error that looks unrelated to Slack

### LINKED ISSUE #255: Messaging CLI discovery (lark-cli / slack) is still PATH-only after #249
Follow-up to #237, which was closed by #249. #249 shipped the shared install-only resolver for Agent Runtime CLIs, but the messaging CLIs were left on the old code path.

## Problem

`refreshImCliReadiness` (`packages/client/src/runtime/client-runtime-composition.ts:533`) still resolves `lark-cli` and `slack` through the legacy `resolveExecutable` (`client-runtime-composition.ts:724`), which walks `PATH` only.

The daemon's `PATH` is a snapshot taken at service-install time and baked into the launchd plist / systemd unit; it never sources the user's shell rc files. A `lark-cli` installed under an npm-global prefix, pnpm, bun, or a node version manager is therefore invisible to the daemon even though it works perfectly in the user's terminal.

This is not cosmetic: the server's `handoffReady` gate requires messaging-CLI readiness in addition to Agent Runtime readiness and a live connection. A user with a fully working Codex setup and a working `lark-cli` we cannot see gets a silent dead end at the end of onboarding — nothing is broken on their machine, and the product says nothing actionable.

After #249 this is the only remaining gap in this area that still blocks a user outright.

## Proposed change

Reuse the resolver #249 already added — `resolveAgentRuntimeExecutable` (`packages/client/src/runtime/agent-runtime-installation.ts:97`) — for the messaging CLIs, so `PATH` → well-known bin dirs discovery and the protected-root policy apply to all four binaries the Client depends on.

The resolver's first parameter is currently typed as `AgentRuntimeProvider`, and the desktop-app branch is Codex-specific. Generalizing it needs a wider binary identity than "Agent Runtime provider", with an empty desktop-app list for `lark-cli` and `slack` (neither ships inside a GUI app bundle).

Once every caller is migrated, the legacy `resolveExecutable` in `client-runtime-composition.ts` should be deleted rather than left as a second, weaker discovery path.

yuezengwu:
## Summary

Complete the product-integration half of the Provider CLI work started by #284.

- Treat an active Feishu/Lark or Slack binding as the requirement signal. The Server asks only the bound Computer to reconcile that provider, and the daemon inspects the persisted account-global selection, performs managed-only repair when needed, then validates the real bound credential.
- Bind readiness and every visible Turn to the same executable fingerprint. Each Turn receives an exact, short-lived launch plan; Agents continue to run bare native `lark-cli` / `slack` commands through the authoritative launcher.
- Give Slack Turns a private writable config directory and inject `--config-dir` / `--skip-update` at the launcher boundary, so a sandboxed Agent does not need to discover the workaround from an unrelated `$HOME/.slack/logs` failure.
- Surface preparing, validating, and actionable failure states in Agent and onboarding UI. `opentag doctor` and the portable installer report Provider CLI installation state but remain read-only.
- Fence daemon shutdown on one shared cleanup chain, including every in-flight durable Turn Report write, so runtime teardown cannot race the next startup or test cleanup.

Safety and product boundaries:

- `doctor` and `install.sh` never install or repair a third-party CLI.
- Daemon auto-repair is limited to OpenTag-managed artifacts required by an active IM binding; it never overwrites an external installation or foreign shim.
- Provider credential validation is live and binding-driven. It does not consume Agent/model tokens, persist grants, expose credentials or local paths to the Server, or turn login/subscription/static API-token state into a general `doctor` check.
- Caller `PATH`, `HOME`, and `XDG_*` do not choose the daemon target. No outbound adapter send path is restored.

Closes #124.
Closes #255.

## Testing

Exact branch head: `157b52cc` (includes `main@bdf743e5` through a normal merge; no published history was rewritten).

- `pnpm check` — pass; migration drift: 37 migrations, complexity ratchet: 167 warnings / 8 improvements
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm test` — pass: scripts 137, shared 134, Server unit 593, Client 850, CLI 308, Web 561
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 24 files / 345 tests; 100% statements, branches, functions, and lines
- `pnpm --filter @opentag/server test:integration` — 18 files / 301 tests
- `pnpm test:e2e:doctor-docker` — 26/26 real Docker scenarios: manager fail-closed 1/1, systemd/fault matrix 21/21, formal connect and HOME/XDG authority 4/4
- Shutdown regression — the controlled durable-write gate proves `run()` cannot settle while a Turn Report persistence operation is still in flight

The Docker run was native Linux arm64. External provider staging delivery was not repeated locally because it requires owned live Feishu/Slack credentials; the credential and exact-launch boundaries are covered by controlled real-process integration fixtures.

## Breaking changes

None. Accounts without an active IM binding do not acquire a Provider CLI requirement. Existing external compatible CLIs remain usable, while OpenTag-managed artifacts can be repaired silently after a binding is active.

## Checklist


yuezengwu:
Exact-head QA update for `157b52cc`: the Linux `ENOTEMPTY` failure was a real shutdown-lifecycle defect, not a retry-only flake. `stop()`/`run()` now share one cleanup promise, and `TurnReportOwner` tracks and exposes settlement for every detached persistence/send chain; composed shutdown waits for it before returning. A controlled write-gate regression proves shutdown cannot settle early. Local exact-head `check`, `typecheck`, `build`, all 2,583 unit tests, 345-test Agent coverage at 100%, 301 Server integration tests, and 26/26 Docker doctor scenarios pass. Waiting for the fresh required Linux/Node matrix before merge.

PR #441, opened 2026-09-01
feat(observability): land the logging foundation primitives

bestony:
Lands the cross-package observability **foundation** for OpenTag: the primitives that later logging-adoption work needs, with **zero call-site adoption**.

`docs/error-taxonomy.md` has stated since it was written that the structured-error infrastructure "need[s] call-site adoption in a later integration pass". A repo-wide audit found that pass is blocked on five prerequisites that did not exist: a byte-bounded redactor, a service-logger seam, a server log-level knob, cross-tier request-id correlation, and a fixed log-field vocabulary. This PR is exactly those prerequisites and nothing else — seven follow-up lanes will branch off it, so any adopted call site here would become a merge conflict later.

## Checklist

- [x] 1. `redactForLog` + `STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES` in shared, exported from `src/index.ts`
- [x] 2. Named `structured-errors.js` export line in `packages/shared/src/browser.ts` + shared snapshot updated
- [x] 3. Tests: string-value byte cap, shared public-exports green, web entry chunk under budget
- [x] 4. `packages/server/src/observability/service-logger.ts` + exported from `observability/index.ts`
- [x] 5. Exactly one `serviceLogger` wiring line in `packages/server/src/index.ts`
- [x] 6. `OPENTAG_LOG_LEVEL` in `config.ts` (strict schema) + `level` on the Fastify logger
- [x] 7. `requestIdHeader` + `genReqId` + `x-request-id` response echo + extended `request-logging.test.ts`
- [x] 8. `createServerDiagnosticReporter` widened to `(code, context?)`
- [x] 9. `recordSpanError` accepts and records the real error
- [x] 10. Client logger context + dual destination + string-value scrubbing + rotation time floor
- [x] 11. `WORKSPACE_ID` attribute, docs (+ zh-CN mirrors), Docker log driver caps

## What changed and why

- **`redactForLog(value)`** (`packages/shared/src/structured-errors.ts`) returns a redacted copy whose **string values** are UTF-8 byte-capped, reusing the existing `truncateUtf8` helper. This closes a real hazard: `redactSensitive` caps depth, key count and array length but passes string values through in full, and only `boundedSerialize` enforces a byte budget — and it returns a *string*, so it cannot serve as a pino payload. Without this, a multi-megabyte provider-CLI stderr tail on a 2-second delivery retry loop writes verbatim to stdout.
- **Browser barrel** now exports the structured-error surface **by name** (it exported none of it before). Named rather than `export *` because `apps/web/vite.config.ts` enforces a 600 KiB entry-chunk budget; the existing `runtime-config.js` line is the same pattern.
- **`ServiceLogger` port** (`observability/service-logger.ts`) is the single seam the three server adoption lanes will inject through, shaped after the existing injected port in `services/channel-target/poller.ts`. Every payload goes through `redactForLog`; an absent logger is tolerated and the port never throws, so a logging failure can never replace the business failure it was observing.
- **Server log level**: `OPENTAG_LOG_LEVEL` added to `ServerEnvironmentSchema` (which is `.strict()`, so the schema change was mandatory) and passed to the Fastify logger. The package previously had no level handling at all, so pino ran pinned at Fastify's default `info` and every `debug` line a later lane writes would have been dead output.
- **Cross-tier request id**: `requestIdHeader: "x-request-id"` plus a `genReqId` UUID fallback, and the id is echoed on the response. Fastify 5 defaults `requestIdHeader` to `false`, so the server was discarding the `x-request-id` that `packages/client/src/request-policy.ts` already mints and sends, and never returned one — the browser/CLI/server join key did not exist.
- **`recordSpanError`** previously took no error argument and hardcoded `INTERNAL_ERROR`, so every helper recorded the identical constant and no error identity ever reached a span.
- **Client logger** gains `configureClientLoggerContext`, a file-or-dual destination mode, and `redactSensitive` over string **values** — it previously applied pino key-path redaction only, so a token or path inside an `error.message` landed in `client.log` verbatim. `rotating-file-stream.ts` gains a time-based retention floor; the ring was purely size-based, so adding log lines shortened how far back the file reached.
- **Vocabulary and volume**: `OPENTAG_ATTR.WORKSPACE_ID` added, a fixed log-field vocabulary and level convention documented (with zh-CN mirrors), including an explicit warning that `imAttrs()`/`runtimeAttrs()` emit dotted OTel span keys and must not be used as pino payloads. Docker's json-file driver is capped via `max-size`/`max-file`.

## Acceptance criteria — verified by the orchestrator, not accepted on the agent's word

Each command below was re-run by hand in the lane checkout.

| Criterion | Result |
| --- | --- |
| `pnpm check` | **PASS** (exit 0) — complexity ratchet 167 warnings / 10 improvements, workspace contract OK, migration drift OK (37 migrations). The 170 biome warnings are the pre-existing repo baseline, gated by the ratchet. |
| `pnpm typecheck` | **PASS** (exit 0) |
| `pnpm test` | **PASS** (exit 0) — shared 135, client 852, web 577, cli 308, server 599 = **2471 tests, 0 failed** |
| `pnpm build` | **PASS** (exit 0) |
| Web build + 600 KiB entry budget | **PASS** — entry `index` 273.91 kB |
| shared / server / client filtered test commands | **PASS** — all green within the full run |
| public-exports snapshots updated in step | **PASS** — both shared and client snapshots updated, their tests green |
| New test: `redactForLog` byte-caps a string **value** | **PASS** — asserts a nested 10 000-character multi-byte string is capped to `STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES` and contains `[TRUNCATED]` |
| New test: inbound `x-request-id` → `request.id`, echoed on response | **PASS** — asserts both halves of the round trip |
| `request-logging.test.ts` assertions only added, never removed | **PASS** — the diff contains zero removed lines in that file |
| Zero call-site adoption | **PASS** — the `index.ts` diff is the import, one `serviceLogger` line, and the `loggerLevel` passthrough; no business module touched |
| No drizzle migration; `CURRENT_MIGRATION_COUNT` untouched | **PASS** — no `drizzle/` path in the diff, `setup-completion-backfill.test.ts` untouched |
| Tree clean, 11/11 checklist ticked | **PASS** |

## Notes for the reviewer

- **`void serviceLogger;`** in `index.ts` is deliberate. The seam is wired and exported so the follow-up lanes each add one property to their own construction call, but this lane adopts no call sites — the suppression is the honest cost of landing infrastructure ahead of its consumers, and it disappears with the first adoption lane.
- **One mid-flight correction, self-made.** An early revision raised `createApp`'s cognitive complexity from 58 to 59 and tripped the ratchet. The lane resolved it correctly by extracting `createFastifyLoggerOptions()` rather than by ratcheting the baseline up — `scripts/complexity-baseline.json`, `file-size-exceptions.json` and `coverage-floors.json` are all untouched, and the ratchet now reports one *more* improvement than `main` does. The commit message `chore(server): preserve complexity baseline` understates what it does.
- **One orchestrator-directed correction.** Commit `b174c150` had added `.dispatch/progress.md` — orchestrator scratch state — to the branch. Commit `1879569e` untracks it with a follow-up commit; no history was rewritten and the file remains on disk.
- **No deviation from the confirmed plan** was recorded by the lane, and none was found in review.

## Provenance

Written by codex under bypassed approvals (`--dangerously-bypass-approvals-and-sandbox`) in an isolated git worktree, run `249649`, lane `obs-foundation-1`. Acceptance criteria were authored before dispatch and re-verified independently by the orchestrator before this branch was pushed.


bestony:
Superseded by #443. Identical commits (`637e0e4f`), reopened on `feat/obs-foundation` because `CONTRIBUTING.md` and `.github/workflows/ci.yml` accept only `feat/ fix/ refactor/ test/ docs/ chore/ merge/` — the `feature/` prefix fails the "Validate pull request branch name" gate and GitHub cannot retarget an existing PR's head branch.

PR #443, opened 2026-09-01
feat(observability): land the logging foundation primitives

bestony:
Lands the cross-package observability **foundation** for OpenTag: the primitives that later logging-adoption work needs, with **zero call-site adoption**.

`docs/error-taxonomy.md` has stated since it was written that the structured-error infrastructure "need[s] call-site adoption in a later integration pass". A repo-wide audit found that pass is blocked on five prerequisites that did not exist: a byte-bounded redactor, a service-logger seam, a server log-level knob, cross-tier request-id correlation, and a fixed log-field vocabulary. This PR is exactly those prerequisites and nothing else — seven follow-up lanes will branch off it, so any adopted call site here would become a merge conflict later.

## Checklist

- [x] 1. `redactForLog` + `STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES` in shared, exported from `src/index.ts`
- [x] 2. Named `structured-errors.js` export line in `packages/shared/src/browser.ts` + shared snapshot updated
- [x] 3. Tests: string-value byte cap, shared public-exports green, web entry chunk under budget
- [x] 4. `packages/server/src/observability/service-logger.ts` + exported from `observability/index.ts`
- [x] 5. Exactly one `serviceLogger` wiring line in `packages/server/src/index.ts`
- [x] 6. `OPENTAG_LOG_LEVEL` in `config.ts` (strict schema) + `level` on the Fastify logger
- [x] 7. Validated inbound `x-request-id` via `genReqId` + `safeInboundRequestId`, response echo, `logController` label, extended `request-logging.test.ts`
- [x] 8. `createServerDiagnosticReporter` widened to `(code, context?)`
- [x] 9. `recordSpanError` accepts and records the real error
- [x] 10. Client logger context + dual destination + string-value scrubbing + rotation time floor
- [x] 11. Tenant attribute (`ACCOUNT_ID`, see the merge note below), docs (+ zh-CN mirrors), Docker log-cap operator prerequisite documented

## What changed and why

- **`redactForLog(value)`** (`packages/shared/src/structured-errors.ts`) returns a redacted copy whose **string values** are UTF-8 byte-capped, reusing the existing `truncateUtf8` helper. This closes a real hazard: `redactSensitive` caps depth, key count and array length but passes string values through in full, and only `boundedSerialize` enforces a byte budget — and it returns a *string*, so it cannot serve as a pino payload. Without this, a multi-megabyte provider-CLI stderr tail on a 2-second delivery retry loop writes verbatim to stdout.
- **Browser barrel** now exports the structured-error surface **by name** (it exported none of it before). Named rather than `export *` because `apps/web/vite.config.ts` enforces a 600 KiB entry-chunk budget; the existing `runtime-config.js` line is the same pattern.
- **`ServiceLogger` port** (`observability/service-logger.ts`) is the single seam the three server adoption lanes will inject through, shaped after the existing injected port in `services/channel-target/poller.ts`. Every payload goes through `redactForLog`; an absent logger is tolerated and the port never throws, so a logging failure can never replace the business failure it was observing.
- **Server log level**: `OPENTAG_LOG_LEVEL` added to `ServerEnvironmentSchema` (which is `.strict()`, so the schema change was mandatory) and passed to the Fastify logger. The package previously had no level handling at all, so pino ran pinned at Fastify's default `info` and every `debug` line a later lane writes would have been dead output.
- **Cross-tier request id**: the server now reads, **validates** and echoes `x-request-id`. Fastify 5 defaults `requestIdHeader` to `false`, so the server was discarding the id that `packages/client/src/request-policy.ts` already mints and sends, and never returned one — the browser/CLI/server join key did not exist. Naming the header in `requestIdHeader` is not enough: Fastify then adopts the caller's value verbatim and never calls `genReqId`, so a 1 024-character or deliberately colliding id would flow into Pino and OTel. `requestIdHeader` is therefore `false` and the header is read inside `genReqId` through `safeInboundRequestId`, which accepts only `^[A-Za-z0-9._:-]{1,256}$` and otherwise mints a UUID. The Pino field is named `requestId` via `logController` (the top-level `requestIdLogLabel` is deprecated in Fastify 5.12 and removed in v6).
- **`recordSpanError`** previously took no error argument and hardcoded `INTERNAL_ERROR`, so every helper recorded the identical constant and no error identity ever reached a span.
- **Client logger** gains `configureClientLoggerContext`, a file-or-dual destination mode, and runs **both the bindings and the message** through `redactForLog` — it previously applied pino key-path redaction only, so a token or path inside an `error.message` landed in `client.log` verbatim, and the message argument bypassed redaction entirely on both log boundaries. The process context setter is bounded the same way, since Pino attaches it as base fields outside the per-call path. Both boundaries check `isLevelEnabled` before paying for redaction, so a discarded level costs nothing. `rotating-file-stream.ts` gains a time-based retention floor **plus** a `maxTotalBytes` ceiling defaulting to `maxBytes * maxBackups` — the floor decides which backups to keep, the ceiling decides how many may exist, so enabling a time floor can never let diagnostics fill the host disk.
- **Vocabulary and volume**: `OPENTAG_ATTR.ACCOUNT_ID` added (originally `WORKSPACE_ID`; see the merge note below), a fixed log-field vocabulary and level convention documented (with zh-CN mirrors), including an explicit warning that `imAttrs()`/`runtimeAttrs()` emit dotted OTel span keys and must not be used as pino payloads. Docker log capping is documented as an **operator prerequisite** rather than claimed as enforced: image `LABEL`s are inert metadata Docker never reads, and this repository's `docker-compose.yml` defines only a `postgres` service, so nothing here can set the server container's `HostConfig.LogConfig` — CapRover creates it from the image coordinate. `docs/deploying.md` and its zh-CN mirror now name the concrete knob (Docker daemon `log-driver`/`log-opts`, or the CapRover equivalent) with the intended `max-size`/`max-file` values.

## Acceptance criteria — verified by the orchestrator, not accepted on the agent's word

Each command below was re-run by hand in the lane checkout, and re-run again after `main` was merged in (see the merge note at the end). The numbers below are the post-merge run.

| Criterion | Result |
| --- | --- |
| `pnpm check` | **PASS** (exit 0) — complexity ratchet 167 warnings / 2 improvements, workspace contract OK, migration drift OK (38 migrations). The 170 biome warnings are the pre-existing repo baseline, gated by the ratchet. |
| `pnpm typecheck` | **PASS** (exit 0) |
| `pnpm --filter '@opentag/*' test` | **PASS** (exit 0) — shared 142, server 604, client 871, web 575, plus the CLI suite, **0 failed** |
| `pnpm build` | **PASS** (exit 0) |
| Web build + 600 KiB entry budget | **PASS** — entry `index` 273.91 kB |
| shared / server / client filtered test commands | **PASS** — all green within the full run |
| public-exports snapshots updated in step | **PASS** — both shared and client snapshots updated, their tests green |
| New test: `redactForLog` byte-caps a string **value** | **PASS** — asserts a nested 10 000-character multi-byte string is capped to `STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES` and contains `[TRUNCATED]` |
| New test: inbound `x-request-id` → `request.id`, echoed on response | **PASS** — asserts both halves of the round trip |
| `request-logging.test.ts` assertions only added, never removed | **PASS** — the diff contains zero removed lines in that file |
| Zero call-site adoption | **PASS** — the `index.ts` diff is the import, one `serviceLogger` line, and the `loggerLevel` passthrough; no business module touched |
| No drizzle migration; `CURRENT_MIGRATION_COUNT` untouched | **PASS** — no `drizzle/` path in the diff, `setup-completion-backfill.test.ts` untouched |
| Tree clean, 11/11 checklist ticked | **PASS** |

## Notes for the reviewer

- **`void serviceLogger;`** in `index.ts` is deliberate. The seam is wired and exported so the follow-up lanes each add one property to their own construction call, but this lane adopts no call sites — the suppression is the honest cost of landing infrastructure ahead of its consumers, and it disappears with the first adoption lane.
- **One mid-flight correction, self-made.** An early revision raised `createApp`'s cognitive complexity from 58 to 59 and tripped the ratchet. The lane resolved it correctly by extracting `createFastifyLoggerOptions()` rather than by ratcheting the baseline up — `scripts/complexity-baseline.json`, `file-size-exceptions.json` and `coverage-floors.json` are all untouched, and the ratchet now reports one *more* improvement than `main` does. The commit message `chore(server): preserve complexity baseline` understates what it does.
- **One orchestrator-directed correction.** Commit `b174c150` had added `.dispatch/progress.md` — orchestrator scratch state — to the branch. Commit `1879569e` untracks it with a follow-up commit; no history was rewritten and the file remains on disk.
- **No deviation from the confirmed plan** was recorded by the lane, and none was found in review.

## Provenance

Written by codex under bypassed approvals (`--dangerously-bypass-approvals-and-sandbox`) in an isolated git worktree, run `249649`, lane `obs-foundation-1`. Acceptance criteria were authored before dispatch and re-verified independently by the orchestrator before this branch was pushed.

## Rebased onto a moved `main`

`main` advanced by two commits after this work was verified, one of them the breaking
`feat(server)!: complete workspace-removal contract migration (#346)`. `origin/main` was **merged
into this branch** (never rebased — the branch was already published, so no force-push) and the
result re-verified from scratch: `pnpm check`, `pnpm typecheck`, `pnpm --filter '@opentag/*' test` (0 failed),
`pnpm build` all pass, migration drift now reports 38 migrations.

Two things had to be resolved:

- **Textual**, in `request-logging.test.ts`: `main` renamed `removes query credentials and invitation
  bearer path segments` to `removes query credentials from logged URLs` and dropped the invitation
  assertions along with those routes. Both sides were kept — `main`'s renamed test and its reduced
  body, plus this branch's new `x-request-id` round-trip test.
- **Semantic**, in `observability/attributes.ts`: git auto-merged the file because the lines did not
  overlap, but the result was wrong. `main` removed the workspace concept entirely
  (`WORKSPACE_COMPUTER_ID` → `INSTALLATION_ID`, and `agents.workspaceId` no longer exists), while
  this branch had added `WORKSPACE_ID: "opentag.workspace.id"` — re-introducing a workspace attribute
  into a codebase that had just deleted workspaces. The tenant boundary is now the **Account**, so the
  attribute is `ACCOUNT_ID: "opentag.account.id"` and the log-field vocabulary documents `accountId`
  (English and zh-CN both updated). The original need — no log field could scope a stream to one
  tenant — is unchanged; only the boundary's name moved.

Branch renamed from `feature/obs-foundation` to `feat/obs-foundation`: `CONTRIBUTING.md` and
`.github/workflows/ci.yml` accept only `feat/ fix/ refactor/ test/ docs/ chore/ merge/`, and
`feature/` fails the "Validate pull request branch name" gate. The commits are identical.

## Review rounds

Three review rounds by @yuezengwu landed on this branch after the initial push; every finding was
reproduced before it was fixed, and all eleven review threads are resolved.

- **Round 1** (6 findings) — folded header continuations leaked; the log *message* bypassed redaction
  on both boundaries; the retention floor removed the storage cap; the inbound request id was
  unvalidated; the zh-CN mirror was unsynchronized.
- **Round 2** (4 findings) — compound `Cookie`/`Digest` values leaked past the first `;`/`,`; the
  process-wide client context bypassed the byte cap; the Docker cap was inert metadata (the PR body
  above previously overclaimed this, now corrected); `requestIdLogLabel` was deprecated.
- **Round 3** (1 finding) — pretty-printed and serialized header shapes leaked: indentation demoted a
  physical header to inline mode, and structural array/quoted values were neither consumed nor
  replaced well-formed.

Two corrections were caught during verification and are worth stating plainly. Redacting the message
initially made `redactForLog` run on every call regardless of level, which broke a timing-sensitive
client test — both boundaries now check `isLevelEnabled` first. And the first attempt at the compound
-header fix anchored matching to line starts, which stopped that leak but left *inline* occurrences
unredacted entirely; the matcher now uses position to select the value's **end**, never whether to
redact.


packages/shared/src/structured-errors.ts
```diff
@@ -232,6 +234,22 @@ function truncateUtf8(value: string, maxBytes: number): string {
...
+  seen.add(value);
+  if (Array.isArray(value)) return value.map((item) => capLogStringValues(item, seen));
+  const output: Record<string, unknown> = {};
+  for (const [key, child] of Object.entries(value)) output[key] = capLogStringValues(child, seen);
+  return output;
+}
+
+/** Return a detached, recursively redacted copy with a UTF-8 cap on every string value. */
+export function redactForLog<T>(value: T): T {
+  return capLogStringValues(redactSensitive(value), new WeakSet<object>()) as T;
```
yuezengwu, line 693: [P1] Redact folded header continuations before this becomes a log boundary

`redactSensitive` stops header-value matching at CR/LF, so this new guarantee still leaks a legal whitespace-indented continuation. On this exact head, `redactForLog({ message: "Authorization: Bearer first\r\n\tsecond-secret\r\nX-Safe: ok" })` retains `second-secret`. Consume folded continuations after Authorization, Cookie, and Set-Cookie while preserving the next genuine header, and add regression coverage.

packages/client/src/observability/logger.ts
```diff
@@ -104,9 +119,28 @@ function stderrDestination(): DestinationStream {
...
+        // Logging failures never escape into runtime behavior.
+      }
+    },
+  };
+}
+
 function adapt(logger: PinoLogger, bindings: ClientLogBindings): ClientLogger {
   const write = (method: "debug" | "error" | "info" | "warn", fields: ClientLogBindings, message: string) =>
-    safeWrite(() => logger[method]({ ...bindings, ...fields }, message));
+    safeWrite(() => logger[method](redactSensitive({ ...bindings, ...fields }), message));
```
yuezengwu: [P1] Apply the bounded log redactor to both fields and messages

This changed client boundary still uses `redactSensitive` instead of `redactForLog`, and it passes `message` untouched. An exact-head probe logged `Authorization: Bearer review-secret` verbatim and preserved a 10,000-byte field even though the new contract says 4 KiB per string. Run fields through the bounded helper and sanitize/cap the message before Pino sees either value.

packages/server/src/observability/service-logger.ts
```diff
@@ -0,0 +1,29 @@
...
+  error(bindings: Record<string, unknown>, message: string): void;
+}
+
+export function createServiceLoggerPort(getLogger: () => FastifyBaseLogger | undefined, module: string): ServiceLogger {
+  const write = (level: "debug" | "info" | "warn" | "error", bindings: Record<string, unknown>, message: string) => {
+    try {
+      const logger = getLogger();
+      if (!logger) return;
+      const payload = redactForLog({ ...bindings, module }) as Record<string, unknown>;
+      logger[level](payload, message);
```
yuezengwu: [P1] Do not let the message bypass redaction

The bindings are protected, but the second Pino argument crosses the same log boundary unchanged. Any adoption lane that uses an exception/provider message here can still write a bearer token or an unbounded string even though this port is advertised as the safe seam. Redact and cap `message` as well, with a test that puts a credential in the message argument rather than only in bindings.

packages/client/src/observability/rotating-file-stream.ts
```diff
@@ -119,6 +131,51 @@ export class RotatingFileStream {
...
 
+  #rotateWithRetentionFloor(): void {
+    const cutoff = Date.now() - this.#minRetentionMs;
+    this.#pruneExpiredBackups(cutoff);
+    let highest = this.#highestBackupIndex();
+    if (highest >= this.#maxBackups && this.#isBackupExpired(highest, cutoff)) {
+      this.#removeIfPresent(`${this.path}.${highest}`);
+      highest -= 1;
+    }
+    for (let index = highest; index >= 1; index -= 1) {
```
yuezengwu, line 154: [P1] Preserve an operational hard cap on log storage

With `minRetentionMs` enabled by the service configuration, every young backup is shifted past `maxBackups`; the added test explicitly keeps `.2` when `maxBackups` is 1. With the defaults this can retain a new 10 MiB file on every rotation for seven days, so high-volume diagnostics can fill the host disk. Keep a hard file/byte emergency ceiling in addition to the time target instead of making `CLIENT_LOG_BACKUPS` non-bounding.

packages/server/src/app.ts
```diff
@@ -111,18 +129,9 @@ function contentTypeParserErrorStatus(error: unknown): number | undefined {
...
-        req: (request) => ({
-          method: request.method,
-          url: sanitizeRequestUrl(request.url),
-          host: request.hostname,
-          remoteAddress: request.ip,
-          remotePort: request.socket.remotePort,
-        }),
-      },
-    },
+    requestIdHeader: "x-request-id",
```
yuezengwu: [P2] Validate the inbound request-id contract instead of trusting the header

Fastify performs no validation here and does not call `genReqId` when the header exists. This exact head accepted and echoed a 1,024-character value unchanged, beyond the shared 256-byte request-id limit, and allows callers to choose colliding IDs that flow into Pino and OTel. Validate a bounded safe identifier and fall back to a UUID; also configure Fastify to emit the documented Pino field `requestId` rather than its current default `reqId` (confirmed in the emitted NDJSON).

docs/zh-CN/observability.md
```diff
@@ -23,6 +23,7 @@ OPENTAG_OTEL_SAMPLE_RATE=1
 | `OPENTAG_OTEL_HEADERS` | 空 | 逗号分隔的 `key=value` headers，原样传给 OTLP trace exporter |
 | `OPENTAG_OTEL_ENVIRONMENT` | `OPENTAG_ENV` | `deployment.environment.name` resource 标签 |
 | `OPENTAG_OTEL_SAMPLE_RATE` | `1` | `[0,1]` 范围内的全局 head sample rate |
+| `OPENTAG_LOG_LEVEL` | `info` | Server Pino level：`trace`、`debug`、`info`、`warn`、`error`、`fatal` 或 `silent` |
```
yuezengwu, line 58: [P2] Keep the Chinese observability mirror synchronized

The English document adds the entire logging contract, field vocabulary, level matrix, and the `imAttrs`/`runtimeAttrs` warning, while this mirror adds only the environment-variable row. The repository rule requires synchronized Chinese mirrors, and Chinese readers currently cannot discover the contract that follow-up lanes must obey. Translate the new section here or mark an intentional divergence explicitly.

yuezengwu (changes requested):
Exact-head review of `637e0e4feccea0aa6acb652daea1caa09add62a6`. The foundation is not safe to adopt yet: the new log boundary still has credential-leak paths, the retention floor removes the operational storage cap, and the request-id/log-vocabulary contract is incomplete.

Verification on the detached exact-head worktree: `pnpm build`, `pnpm check`, `pnpm typecheck`, and `pnpm test` all passed under the pinned Node 24.19.0 / pnpm 10.12.1 toolchain. Current CI for `feat/obs-foundation` is green; the older failure belongs to the superseded `feature/obs-foundation` branch name and failed only the branch-name gate.

Please address the inline findings and add focused regressions for the reproduced redaction, retention, and request-id cases.

bestony:
@yuezengwu all six findings are addressed in `04b60814..820442e5`. Each was reproduced on the exact head before being fixed.

**[P1] Folded header continuations** — reproduced: `redactForLog({message: "Authorization: Bearer first\\r\\n\\tsecond-secret\\r\\nX-Safe: ok"})` returned `"Authorization: [REDACTED]\\r\\n\\tsecond-secret\\r\\nX-Safe: ok"`. The header pattern now consumes RFC 7230 obs-fold continuations (a line beginning with SP/HTAB belongs to the preceding value) while a line starting with anything else is preserved as the next genuine header. Verified after the fix: output is `"Authorization: [REDACTED]\\r\\nX-Safe: ok"` — secret gone, next header intact. Same for a folded `Set-Cookie`. Regressions added for both.

**[P1] Client logger** — confirmed it used `redactSensitive` and passed `message` untouched. Now both the fields and the message go through `redactForLog`, so the message is scrubbed and 4 KiB-capped. Regression asserts a bearer token in the message is redacted and a 20 000-byte message is capped.

**[P1] Service-logger message** — same fix on the server seam, with a regression that puts the credential in the message argument rather than the bindings.

**[P1] Storage ceiling** — confirmed `minRetentionMs` made `CLIENT_LOG_BACKUPS` non-bounding. Added `maxTotalBytes`, defaulting to `maxBytes * maxBackups` — exactly the cap that applied before the retention floor existed. The floor now decides *which* backups to keep; the ceiling decides how many may exist at all, evicting oldest-first even when they are younger than the floor. The test you cited was updated to state its budget explicitly, and two regressions were added: one proving eviction under the ceiling despite the floor, one proving the default ceiling bounds growth.

**[P2] Request-id contract** — you were right that `genReqId` never runs while `requestIdHeader` names a header. `requestIdHeader` is now `false` and the header is read inside `genReqId` through `safeInboundRequestId`, which accepts only `^[A-Za-z0-9._:-]{1,256}$` and otherwise falls back to a UUID. Also set `requestIdLogLabel: "requestId"` so the emitted field matches the documented vocabulary. Regressions cover the 1 024-character rejection and assert the NDJSON carries `"requestId"` and not `"reqId"`.

**[P2] Chinese mirror** — the full logging contract, field vocabulary, level matrix, `OPENTAG_LOG_LEVEL` matrix and the `imAttrs`/`runtimeAttrs` warning are now translated in `docs/zh-CN/observability.md`.

**One regression I introduced while fixing these, found and fixed.** Redacting the message made `redactForLog` run on every call site regardless of level, because the arguments are evaluated before Pino can discard the record. That slowed a 5 s-timeout client test into failing. I confirmed the attribution by stashing my changes (baseline: 866/866 green; with them: 1 failed), rather than writing it off as flaky. Both boundaries now check `isLevelEnabled` before paying for redaction.

Full suite after all fixes: `pnpm check`, `pnpm typecheck`, `pnpm test` (2 494 passed, 0 failed) and `pnpm build` all green; complexity ratchet passes and `createApp` is 57 against its 58 baseline.

packages/shared/src/structured-errors.ts
```diff
@@ -150,16 +152,26 @@ function isSensitiveKey(key: string): boolean {
...
+    value
+      .replace(/\bBearer\s+[^\s,;}\]]+/giu, "Bearer [REDACTED]")
+      .replace(/\b(Basic|Digest|Negotiate)\s+[^\s,;}\]]+/giu, "$1 [REDACTED]")
+      /*
+       * Consume RFC 7230 obs-fold continuations after a credential header. A continuation line starts
+       * with SP or HTAB and belongs to the preceding header value, so the whole line is redacted; a
+       * line starting with anything else is the next genuine header and is preserved.
+       */
+      .replace(
+        /(\b(?:authorization|proxy-authorization|cookie|set-cookie)\s*[:=]\s*)[^\r\n,;}]*(?:\r?\n[ \t][^\r\n]*)*/giu,
```
yuezengwu: [P1] Redact the entire credential header value

The obs-fold fix still stops this match at `,` or `;`, so valid compound credential headers leak everything after the first separator. On this exact head, `redactForLog("Cookie: session=first-secret; admin=second-secret")` returns `Cookie: [REDACTED]; admin=second-secret`, and a Digest Authorization value retains its `realm`/`nonce` suffix after the first comma. Cookie fields legitimately use semicolons and Digest credentials use commas, so adoption through this advertised safe seam can still write credentials. Consume the whole header value (including folded lines) while preserving the next real header, and cover both compound Cookie and Digest inputs.

packages/client/src/observability/logger.ts
```diff
@@ -49,19 +55,28 @@ export function configureClientLoggerForService(logDirectory: string): void {
   }
   if (serviceDirectory) return;
   serviceDirectory = canonicalDirectory;
-  serviceStream = new RotatingFileStream(canonicalDirectory);
+  serviceStream = new RotatingFileStream(canonicalDirectory, { minRetentionMs: CLIENT_LOG_MIN_RETENTION_MS });
+  rootLogger = undefined;
+}
+
+export function configureClientLoggerContext(bindings: ClientLogBindings): void {
+  clientLoggerContext = { ...clientLoggerContext, ...redactSensitive(bindings) };
```
yuezengwu: [P2] Bound the process context before Pino attaches it

This public context setter still uses `redactSensitive`, not `redactForLog`, and `buildRoot` attaches the result as Pino base fields outside the bounded `adapt()` path. An exact-head probe configured `{ instanceId: "x".repeat(20_000) }`; every emitted record retained all 20,000 bytes even though this PR advertises a 4 KiB maximum for every log string. Apply the bounded redactor here as well and add a regression on a long context value.

Dockerfile
```diff
@@ -33,6 +33,12 @@ RUN pnpm install --frozen-lockfile --config.engine-strict=true --ignore-scripts
 
 FROM node:24-alpine AS runtime
 
+# Docker does not read a log-driver setting from an image. These labels keep the required
+# json-file limits discoverable for CapRover or another deployment controller that applies them.
+LABEL com.opentag.log-driver="json-file" \
```
yuezengwu: [P1] Apply an actual log-driver cap to the server container

These custom image labels are metadata only; Docker does not turn them into `HostConfig.LogConfig`, and no code or deployment workflow in this repository reads them. The only real `logging:` block added by this PR applies to the standalone Postgres service, while the CapRover deploy action passes only the image coordinate. Consequently the OpenTag server's stdout remains subject to whatever unverified host default exists and can still fill the host disk under high-volume logging. Enforce and verify `max-size`/`max-file` in the deployment/daemon configuration that creates the server container (or make an external configuration gate explicit); labels alone do not satisfy the stated Docker cap.

packages/server/src/app.ts
```diff
@@ -109,20 +127,31 @@ function contentTypeParserErrorStatus(error: unknown): number | undefined {
...
-        }),
-      },
-    },
+    /*
+     * requestIdHeader is deliberately false. When it names a header, Fastify adopts that value
+     * verbatim and never calls genReqId, so an unbounded or colliding caller-chosen id would flow
+     * straight into Pino and OTel. Doing the read here keeps the header contract but validates it.
+     */
+    requestIdHeader: false,
+    requestIdLogLabel: "requestId",
```
yuezengwu: [P2] Use Fastify's supported logController request-id label

Fastify 5.12 already deprecates this top-level option. Every focused and full server test that constructs `createApp` now emits `FSTDEP024`, and Fastify states the option will be removed in v6 in favor of `logController.requestIdLogLabel`. Configure the label through a `LogController` now so this new request-correlation contract does not land on an API that is already scheduled to disappear.

yuezengwu (changes requested):
Exact-head review of `820442e59d51739e6ad09a816b4f846c44077319`. I re-verified the prior six findings from scratch; the new commits address those exact probes, but a fresh adversarial pass found additional boundary gaps: compound credential-header values still leak, process-wide client context bypasses the advertised field cap, the server container has no effective log-driver cap, and the request-id label uses a Fastify API already scheduled for removal.

Local verification on a clean detached worktree under pinned Node 24.19.0 / pnpm 10.12.1: `pnpm build`, `pnpm check`, `pnpm typecheck`, `pnpm test`, `git diff --check`, and focused shared/client/server observability tests all passed. The workspace run passed 2,494 tests plus 186 root tests. Current-head CI, patch coverage, quality, Docker, Node 22/24/26, smoke, workflow, and secret checks are green. The historical CI failure is from the superseded `feature/obs-foundation` branch name and failed only the branch-name gate.

Please address the inline findings and add focused regressions for the reproduced redaction and context-cap cases.

packages/shared/src/structured-errors.ts
bestony: Fixed in `04b60814`. Your probe now returns `"Authorization: [REDACTED]\r\nX-Safe: ok"` — the folded continuation is gone and the next real header is preserved; same for a folded `Set-Cookie`. Regressions added for both. A follow-on gap you found in round 2 (compound `;`/`,` values in the same matcher) is tracked in the newer thread on this file, not here.

packages/client/src/observability/logger.ts
bestony: Fixed in `b82c59e3`. Fields and message both go through `redactForLog` now, so the message is scrubbed and 4 KiB-capped. Regression asserts a bearer token in the message is redacted and a 20 000-byte message is capped.

packages/server/src/observability/service-logger.ts
bestony: Fixed in `b82c59e3`. The message argument is redacted and capped on the server seam too, with a regression that puts the credential in the message rather than the bindings.

packages/client/src/observability/rotating-file-stream.ts
bestony: Fixed in `0ac1df94`. Added `maxTotalBytes`, defaulting to `maxBytes * maxBackups` — exactly the cap that applied before the retention floor existed. The floor now decides which backups to keep; the ceiling decides how many may exist, evicting oldest-first even when they are younger than the floor. The test you cited states its budget explicitly, plus two new regressions.

packages/server/src/app.ts
bestony: Fixed in `6d08cbdd`. You were right that `genReqId` never runs while `requestIdHeader` names a header: it is now `false` and the header is read inside `genReqId` via `safeInboundRequestId`, accepting only `^[A-Za-z0-9._:-]{1,256}$` and otherwise minting a UUID. Regression covers the 1 024-character rejection. The label half of this finding moved to the newer `logController` thread.

docs/zh-CN/observability.md
bestony: Fixed in `820442e5`. The full logging contract, field vocabulary, level matrix, `OPENTAG_LOG_LEVEL` matrix and the `imAttrs`/`runtimeAttrs` warning are now translated in `docs/zh-CN/observability.md`.

packages/shared/src/structured-errors.ts
bestony: Fixed across `1acb54bd`, `42839a6d` and `7010be8e`. Both your probes are clean now — `Cookie: session=first-secret; admin=second-secret` → `Cookie: [REDACTED]`, and the Digest value → `Authorization: [REDACTED]` with no `realm`/`nonce` remainder.

Worth flagging that the first attempt at this traded one leak for a wider one, and I rejected it before pushing. Anchoring the match to line starts stopped the compound leak but left inline occurrences unredacted entirely, so three shapes the pre-fix code *did* redact started passing through verbatim: `{cookie: session=abc123secret, other: keepme}`, `{"cookie":"session=abc123secret","other":"keepme"}` and `ctx: a=1; authorization=tok_live_SECRET; b=2`. The serialized-headers-in-an-error-message shape is the most likely real path, so that was worse than the bug being fixed.

The matcher now has two modes, and position selects the value's **end**, never whether to redact: line-anchored (start of string or after CR/LF) consumes the whole physical value plus obs-fold continuations; inline/JSON-ish redacts but stops at the first `,` `;` `}` closing quote or line break. Result: `{cookie: [REDACTED], other: keepme}`, `{"cookie":"[REDACTED]","other":"keepme"}`, `ctx: a=1; authorization=[REDACTED]; b=2`. Regressions cover all three shapes plus the compound and obs-fold cases.

packages/client/src/observability/logger.ts
bestony: Fixed in `ba0b4e50`. `configureClientLoggerContext` now runs through `redactForLog`, so a process-context value is capped at the same 4 KiB the rest of the contract promises before Pino attaches it as a base field. Regression added on a long context value asserting the emitted record is capped.

Dockerfile
bestony: Fixed in `c32300ee`, and you were right that the labels were inert — I checked, and it was worse than that: `docker-compose.yml` defines **only** a `postgres` service, so the `logging:` block this PR added never applied to the server at all. The PR body's claim that Docker's driver was capped was an overclaim on my side, now corrected.

Since CapRover creates the server container from the image coordinate, the repository genuinely cannot set `HostConfig.LogConfig` from here, so I took your second option and made the gate explicit: the inert `com.opentag.log-*` labels are gone and `docs/deploying.md` (with its zh-CN mirror) now names the concrete operator knob — Docker daemon `log-driver`/`log-opts`, or the CapRover equivalent — with the intended `max-size`/`max-file` values. The Postgres `logging:` block is unchanged.

packages/server/src/app.ts
bestony: Fixed in `fa29819f`. The label is now configured through `logController: new LogController({ requestIdLogLabel: "requestId" })`. I confirmed `FSTDEP024` no longer appears anywhere in the server test output (grep count 0, down from one per `createApp`), and the existing regression asserting the NDJSON carries `"requestId"` and not `"reqId"` still passes.

packages/shared/src/structured-errors.ts
```diff
@@ -149,17 +151,105 @@ function isSensitiveKey(key: string): boolean {
...
+ * A non-folded line break is left intact so the next genuine header remains visible. Inline matches
+ * stop at the delimiter for their surrounding JSON-ish or list value so sibling fields remain intact.
+ */
+function scrubCredentialHeaders(value: string): string {
+  let output = "";
+  let cursor = 0;
+  CREDENTIAL_HEADER_PATTERN.lastIndex = 0;
+  for (let match = CREDENTIAL_HEADER_PATTERN.exec(value); match; match = CREDENTIAL_HEADER_PATTERN.exec(value)) {
+    const offset = match.index;
+    const lineAnchored = offset === 0 || value[offset - 1] === "\n" || value[offset - 1] === "\r";
```
yuezengwu: [P1] Redact formatted and serialized compound headers completely

`lineAnchored` recognizes a header only when its name immediately follows CR/LF. A normal formatted header dump such as `headers:\n  Cookie: session=first-secret; admin=second-secret\nX-Safe: ok` therefore enters inline mode; at the actual Client logger boundary this exact-head probe emitted `Cookie: [REDACTED]; admin=second-secret`. A serialized multi-value form such as `{"set-cookie":["session=first-secret","admin=second-secret"]}` also retains both values because the unquoted scanner stops at the first quote after `[`. These are realistic provider/error-message shapes, so later adoption through this advertised safe seam can still write credentials. Make the boundary handling cover indented physical headers and serialized array/escaped forms, with regressions through the log boundary.

yuezengwu (changes requested):
Exact-head review of `7010be8e2a3e8682d5b9ed4e645d007ae51648b6`.

A fresh adversarial pass found a remaining credential-leak path in the shared redactor that both the client and future server adoption lanes rely on; see the inline P1 finding.

Local verification in a clean detached worktree under Node 24.19.0 / pnpm 10.12.1: `pnpm build`, `pnpm check`, `pnpm typecheck`, `pnpm test`, `git diff --check`, and focused shared/client/server observability tests all passed. The full run passed 186 root tests plus 2,497 workspace tests; the focused run passed 40 tests. Current-head CI, patch coverage, quality, Docker, Node 22/24/26, smoke, workflow, and secret checks are green.

The live PR description also remains stale after the fixes: it still describes `requestIdHeader: "x-request-id"`, client `redactSensitive`, and an implemented Docker driver cap, while this head deliberately uses validated `genReqId`, `redactForLog`, and an external operator gate. Please update the description and acceptance counts so the review contract matches the code.

packages/shared/src/structured-errors.ts
bestony: Fixed in `9d615ee6`. Both of your probes are clean, and while reproducing them I found two more instances of the same two root causes, so all four are covered:

| input | before | after |
| --- | --- | --- |
| `headers:\n  Cookie: session=first-secret; admin=second-secret\nX-Safe: ok` | leaked `admin=second-secret` | fully redacted, `X-Safe: ok` intact |
| `{"set-cookie":["session=first-secret","admin=second-secret"]}` | leaked both **and** emitted malformed output (only the `[` was replaced) | `{"set-cookie":"[REDACTED]"}`, parses as valid JSON |
| `{"headers":"Cookie: session=first-secret; admin=second-secret"}` | leaked `admin=second-secret` | fully redacted |
| `req:\n\tAuthorization: Digest u="x", nonce="deadbeef"\nX-Safe: ok` | leaked `deadbeef` | fully redacted, `X-Safe: ok` intact |

The last two I found myself — same root causes (indentation demoting a physical header, and the scanner not consuming structural values), so they are fixed by the same change rather than left for a later round.

**Two modes, disambiguated.** SP/HTAB after a line break is now skipped before choosing the mode. An indented line whose token is followed by `:` is a physical header and consumes the whole line plus obs-fold continuations; anything else after indentation stays an obs-fold continuation, so folded values still behave as before. The residual ambiguity is recorded deliberately: a genuinely folded value whose continuation happens to read `Name: something` is now treated as a new header. That direction still redacts the preceding credential, whereas the opposite choice leaks pretty-printed dumps, which are far more common here.

**Structural values** are consumed whole — a `[` takes the balanced array, a `"` takes the quoted string with `\"` escapes honoured — so the replacement is well-formed rather than the corrupted shape above.

Regressions run **through the log boundary** as you asked: the client-logger test asserts the emitted NDJSON contains none of these secrets, alongside the shared unit regressions. I also re-ran the six cases from earlier rounds to confirm none regressed — inline unquoted/quoted/semicolon still redact with siblings intact, compound Cookie, Digest and obs-fold unchanged.

`pnpm check`, `pnpm typecheck`, `pnpm build` and `pnpm --filter '@opentag/*' test` all pass (shared 142, server 604, client 871, web 575).

packages/shared/src/structured-errors.ts
```diff
@@ -149,17 +151,223 @@ function isSensitiveKey(key: string): boolean {
...
+}
+
+function credentialHeaderValueEnd(value: string, from: number): number {
+  let cursor = from;
+  while (true) {
+    const breakStart = lineBreakStart(value, cursor);
+    if (breakStart === -1) return value.length;
+    const breakEnd = lineBreakEnd(value, breakStart);
+    if (value[breakEnd] !== " " && value[breakEnd] !== "\t") return breakStart;
+    if (headerNameColonAt(value, breakEnd)) return breakStart;
```
yuezengwu: [P1] Keep header-looking obs-fold data inside the redaction

This treats any whitespace-prefixed `Name:` line as a new header, but an RFC 7230 obs-fold continuation is identified by that leading SP/HTAB and its field value may contain a colon. At this exact head, the real Client logger emitted `Cookie: [REDACTED]\r\n  b: second-secret\r\nX-Safe: ok` for `Cookie: a=first-secret\r\n  b: second-secret\r\nX-Safe: ok`. The `b:` line is still part of the credential field and remains verbatim. Disambiguate formatted sibling headers without allowing an ambiguous continuation to escape (for example, compare indentation or fail closed), and add a logger-boundary regression where folded text contains a colon.

packages/shared/src/structured-errors.ts
```diff
@@ -149,17 +151,223 @@ function isSensitiveKey(key: string): boolean {
   return SENSITIVE_KEY_PARTS.some((part) => normalized.includes(part));
 }
 
+const CREDENTIAL_HEADER_PATTERN =
```
yuezengwu, line 154: [P1] Match credentials inside escaped serialized messages

The matcher still requires the key closing quote to be adjacent to `:`, so an error message containing an escaped serialized object such as `{\"cookie\":\"session=first-secret; admin=second-secret\",\"other\":\"keep\"}` never matches. At this exact head, the real Client logger wrote that whole message unchanged with both placeholder secrets intact. This is distinct from the raw JSON strings covered by the new tests because the message value contains literal backslashes. Recognize escaped key/value delimiters (or safely peel one serialization layer) and add a logger-boundary regression whose actual message contains those backslashes.

yuezengwu (changes requested):
Exact-head review of `9d615ee696d53ead30055685bf1cbbfd4a3ed5b0`.

A fresh adversarial pass found two remaining credential-leak paths in the shared redactor at the actual Client logger boundary; see the inline P1 findings.

Local verification in a clean detached worktree under Node 24.19.0 / pnpm 10.12.1 passed: `pnpm build`, `pnpm check`, `pnpm typecheck`, `pnpm test` (186 root tests plus 2,500 workspace-package tests), `git diff --check`, 43 focused observability tests, Client agent-runtime coverage (344 tests, 100% statements/branches/functions/lines), and server integration (310 tests). All required current-head CI jobs are also green.

Please also refresh the PR description before the next review. It still describes `requestIdHeader: "x-request-id"`, client `redactSensitive`, the old 2,486-test count, and Docker caps without the now-documented CapRover host prerequisite, so it no longer matches the live implementation or operational contract.

packages/shared/src/structured-errors.ts
bestony: Fixed in `6685d8ed`. Your probe now emits `Cookie: [REDACTED]\r\nX-Safe: ok` — the `b: second-secret` continuation is inside the redaction and `X-Safe: ok` survives.

You were right and the previous round's note was wrong, so it is worth being explicit about why: **I told codex that treating an indented `Name:` line as a new header was the safer default**, and it recorded that as a deliberate decision. For a redactor it is the unsafe direction — an obs-fold continuation is part of the credential value by definition, and RFC 7230 permits a colon inside that value.

The rule is now **indentation relative to the credential header itself**, which separates the two shapes without guessing: a following line continues the redacted value only when its leading SP/HTAB indentation is *strictly greater* than the header's own; equal or smaller indentation ends the value and leaves the sibling line for its own match. So `headers:\n  Cookie: …\nX-Safe: ok` keeps `X-Safe` (0 is not greater than 2), while `Cookie: …\n  b: …` redacts the continuation (2 is greater than 0). Equal-indent `Cookie`/`Set-Cookie` siblings are redacted independently, neither swallowing the other. Where the shape stays genuinely ambiguous it now **fails closed** — a false redaction costs readability, a false pass costs a credential. `progress.md` records this as explicitly replacing the round-3 decision.

packages/shared/src/structured-errors.ts
bestony: Fixed in `61c8b9ab` (plus `a9977df3` for the type narrowing it needed). The matcher now accepts escaped key and value delimiters, so a literal-backslash serialized message matches: `{\\"cookie\\":\\"session=first-secret; admin=second-secret\\",\\"other\\":\\"keep\\"}` redacts only the credential value and leaves `\\"other\\":\\"keep\\"` intact.

Both regressions run at the real Client logger boundary as you asked, asserting the emitted NDJSON: one for the colon-bearing folded continuation, one whose message is built with `String.raw` so it genuinely contains backslashes rather than a pre-unescaped string — that distinction is what made this case slip past the round-3 tests.

I also re-ran the ten earlier shapes to confirm none regressed. `pnpm check`, `pnpm typecheck`, `pnpm build` and `pnpm --filter '@opentag/*' test` all pass (shared 142, server 604, client 873, web 575).

Separately: the PR description is refreshed. It now states the actual request-id mechanism (`requestIdHeader: false` + `genReqId`/`safeInboundRequestId` + `logController`), the client logger's `redactForLog` on both bindings and message, current test counts, and the Docker cap as an explicit **operator prerequisite** rather than something this repo enforces — the earlier wording overclaimed that, which was my error.

packages/shared/src/structured-errors.ts
```diff
@@ -149,17 +151,286 @@ function isSensitiveKey(key: string): boolean {
...
+function escapedQuotedCredentialValueEnd(
+  value: string,
+  from: number,
+  quote: CredentialQuote,
+): { end: number; replacement: string } {
+  const openingDelimiter = `\\${quote}`;
+  for (let cursor = from + openingDelimiter.length; cursor < value.length; cursor += 1) {
+    const character = value[cursor];
+    if (character === "\\") {
+      if (escapedCredentialClosingDelimiterEnd(value, cursor, quote) !== undefined) {
```
yuezengwu: [P1] Keep escaped Digest parameters inside the credential value

This scanner accepts any escaped quote followed by a comma as the outer value terminator, but quoted Digest parameters use that same sequence inside a serialized value. At the real Client boundary, `String.raw` input `{\"authorization\":\"Digest username=\\\"u\\\", realm=\\\"tenant\\\", nonce=\\\"deep-secret\\\"\",\"other\":\"keep\"}` emitted `{\"authorization\":\"[REDACTED]\", nonce=\\\"deep-secret\\\"\",\"other\":\"keep\"}`. The nonce remains verbatim. Track the full backslash run or safely decode one serialization layer so embedded quoted parameters cannot close the credential value, and add this Digest shape at the logger boundary.

packages/shared/src/structured-errors.ts
```diff
@@ -149,17 +151,286 @@ function isSensitiveKey(key: string): boolean {
...
+  while (value[cursor] === " " || value[cursor] === "\t") cursor += 1;
+  return cursor - from;
+}
+
+function isLineAnchoredCredentialHeader(value: string, offset: number): boolean {
+  const start = lineStart(value, offset);
+  if (offset === start) return true;
+  let cursor = start;
+  while (value[cursor] === " " || value[cursor] === "\t") cursor += 1;
+  return cursor === offset && headerNameColonAt(value, offset);
```
yuezengwu: [P1] Treat list-formatted credential headers as whole fields

Only leading whitespace is accepted before a physical header, so common list-formatted dumps enter inline mode and stop redacting at the first semicolon. The exact-head Client logger turns `headers:\n  - Cookie: session=first-secret; admin=second-secret\n  - X-Safe: ok` into `headers:\n  - Cookie: [REDACTED]; admin=second-secret\n  - X-Safe: ok`; the numbered `1. Cookie:` form does the same. Recognize line/list prefixes conservatively or otherwise fail closed for these header-shaped lines, and cover compound Cookie values through the real boundary.

packages/shared/src/structured-errors.ts
```diff
@@ -149,17 +151,286 @@ function isSensitiveKey(key: string): boolean {
...
+  const openingQuote = value[from];
+  const escapedQuote = value[from + 1];
+  if (openingQuote === "\\" && isCredentialQuote(escapedQuote)) {
+    return escapedQuotedCredentialValueEnd(value, from, escapedQuote);
+  }
+  if (isCredentialQuote(openingQuote)) {
+    const end = inlineQuotedCredentialValueEnd(value, from, openingQuote);
+    return { end, replacement: `${openingQuote}${REDACTED}${openingQuote}` };
+  }
+  if (openingQuote === "[") {
```
yuezengwu, line 357: [P2] Preserve escaped arrays and their sibling diagnostics

The structural branch is not escape-aware. For `String.raw` input `{\"set-cookie\":[\"session=first-secret\",\"admin=second-secret\"],\"other\":\"keep\"}`, the real Client logger emitted only `{\"set-cookie\":"[REDACTED]"`: the result is malformed and the safe sibling is discarded. This breaks the advertised diagnostic seam even though the secrets happen to be removed. Consume the balanced array in its actual serialization layer and preserve the escaped delimiters plus `other`, with a boundary regression for this literal-backslash array form.

yuezengwu (changes requested):
Exact-head review of `a9977df3a9257adda77a90037bb000c1df9cb75e`.

The previous two probes are fixed, and the refreshed PR description now matches the request-id, client-redaction, test, and deployment contracts. A fresh adversarial pass nevertheless found three remaining string-redaction boundary gaps: two credential leaks and one destructive escaped-array rewrite; see the inline findings.

Local verification in a clean detached worktree under Node 24.19.0 / pnpm 10.12.1 passed: `pnpm build`, `pnpm check`, `pnpm typecheck`, `pnpm test` (186 root tests plus 2,502 workspace-package tests), `git diff --check`, shared/client/server package suites, Client agent-runtime coverage (344 tests, 100% statements/branches/functions/lines), and server integration (310 tests). Live request-id injection probes also passed for accepted, trimmed, oversized, and comma-delimited inputs, including the `requestId` log label. All required current-head CI jobs are green.

bestony:
Round 5 addressed, plus three defects found during verification that the review had not reported.

**The three reported findings** — escaped `Digest` parameters, list-formatted (`- ` / `1. `) headers,
and escaped arrays — are fixed. As suggested, the fix decodes serialization structurally rather than
adding more shape-specific patches.

**Three further defects surfaced while verifying that fix**, each rejected before it could be pushed:

1. Decoding was applied to the *whole message*, so one serialized credential anywhere collapsed the
   entire log line to a bare `[REDACTED]` — a Windows path, a JSON-encoded URL (`https:\/\/…`), a real
   newline or a real tab was enough. Fixed by scoping the codec to the credential field's own span.
2. Redacting the *decoded* span applied physical-header line semantics inside a single serialized
   string, so an encoded `\r` / `\n` ended the redaction early and leaked the rest of the value.
   `JSON.stringify({cookie: rawHeaderBytes})` on a multi-value header produces exactly that shape.
3. The same leak survived in the unquoted branch after the first two were fixed.

The resolution is a simplification rather than another special case: once the span walker reports a
complete serialized value, that whole span **is** the credential value, so it is replaced wholesale.
The decode/re-encode/rescan machinery is gone; `scrubSerializedCredentialField` is now one line.

**Verification.** `pnpm check`, `pnpm typecheck`, `pnpm build` and `pnpm --filter '@opentag/*' test`
all exit 0 (shared 172, server 604, client 877, web 575). Beyond the suite, 68 redaction probes were
run against the built `dist`: the 16 established shapes, 12 adversarial shapes, 20 targeting the new
span logic (unterminated quotes, unbalanced arrays, bad escapes, mixed slash counts, adjacent and
nested credentials), 8 encoded-line-break variants, 8 unquoted variants, and 4 over-redaction checks
confirming non-credential input is returned byte-for-byte unchanged. All green.

**One deliberate non-fix, recorded rather than silently accepted:** a single-quoted serialized field
redacts to `{\'cookie\':\"[REDACTED]\",…}` — mixed quote styles in one object. Preserving the field's
own delimiter would widen the replacement contract without improving leak safety, so it is left as is.

packages/shared/src/structured-errors.ts
```diff
@@ -149,11 +151,454 @@ function isSensitiveKey(key: string): boolean {
   return SENSITIVE_KEY_PARTS.some((part) => normalized.includes(part));
 }
 
+const CREDENTIAL_HEADER_PATTERN =
+  /\b(?:authorization|proxy-authorization|cookie|set-cookie)[ \t]*["']?[ \t]*[:=][ \t]*/giu;
+const SERIALIZED_CREDENTIAL_FIELD_PATTERN =
+  /\\+(?:["'])(?:authorization|proxy-authorization|cookie|set-cookie)\\+(?:["'])[ \t]*[:=][ \t]*/giu;
```
yuezengwu, line 157: [P1] Decode the serialized layer before matching credential fields

This pre-decode pattern only accepts literal space/tab around `:` and requires backslash-plus-quote delimiters. At the real Client logger boundary on this head, `String.raw` input `{\"cookie\"\t:\t\"session=whitespace-secret\",\"other\":\"keep\"}` was written unchanged; a `\u0022cookie\u0022` key/value form also retained its credential. Decoding one serialization layer produces valid JSON in both cases, so these are within the serialized-message contract this parser now claims to cover. Match the decoded layer (including escaped JSON whitespace and Unicode delimiters) before redacting/re-encoding, preserve safe siblings, and add emitted-NDJSON regressions.

packages/shared/src/structured-errors.ts
```diff
@@ -149,11 +151,454 @@ function isSensitiveKey(key: string): boolean {
...
+    matched = true;
+    const offset = match.index;
+    const lineAnchored = isLineAnchoredCredentialHeader(value, offset);
+    const valueStart = offset + match[0].length;
+    const valueEnd = lineAnchored
+      ? {
+          end: credentialHeaderValueEnd(value, valueStart, lineIndentation(value, lineStart(value, offset))),
+          replacement: REDACTED,
+        }
+      : inlineCredentialValueEnd(value, valueStart, enclosingQuoteAt(value, offset));
```
yuezengwu, line 553: [P1] Keep synchronous redaction linear before applying the byte cap

Every raw credential match calls `enclosingQuoteAt(value, offset)`, which rescans from position zero. The enabled logger therefore becomes quadratic in the number of credential fields, and the 4 KiB cap is applied only afterward. Fresh Node 24 exact-head timings were ~1.3 ms for 3,689 B / 100 fields, 614 ms for 154,889 B / 4,000 fields, and 2.39 s for 310,889 B / 8,000 fields. That is already a multi-second event-loop stall well below the multi-megabyte provider stderr this PR explicitly intends to bound. Track quote state in the same forward pass, or otherwise bound work before repeated prefix scans, and add a regression that proves repeated fields stay linear/bounded.

yuezengwu (changes requested):
Exact-head review of `37d2a2c9fc8d8f54074145715c8c186f785fdada`. The repository gates and the newly added round-five cases pass, but two real Client log-boundary defects remain: valid one-layer serialized credential fields can still bypass redaction, and repeated credential matches make the synchronous redactor quadratic before its 4 KiB output cap. Details and fresh probes are inline.

Validation: pinned Node 24.19.0 / pnpm 10.12.1; install, `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, Client agent-runtime coverage (344 tests, 100%), Server integration (313 tests), `git diff --check`, real emitted-NDJSON credential probes, and Fastify request-id inject probes. CI/checks are green on this exact head. Browser Smoke is green; there is no UI call-site adoption in this foundation PR, so the user-visible verification here was the real Client log file boundary rather than a browser flow. The documented CapRover/Docker log-driver cap remains an external operator prerequisite and was not host-verified.

Issue #336, opened 2026-08-31
Agent messaging-status messages are unreachable: an uncalled builder and three branches behind the suspended guard


liuchao-001:
## Two reachability gaps in the Agent messaging-status surface

Found while verifying a #311 finding. Neither is a user-visible bug today — that is the problem:
both are code that reads as live and is not.

### 1. `messagingAgentStatusDescription` has no callers

`apps/web/src/features/agents/agent-presentation.ts:245` is exported and has **zero non-test
references** in `apps/web/src`. It is a fully-formed status-sentence builder (ready / working /
computer-offline / handoff-unavailable / fallback) that nothing renders.

### 2. Three `agentRecoveryMessage` branches cannot be reached

`agentRecoveryMessage` (`:286`) has exactly one live caller: `AgentLifecycleNotice`
(`agent-detail-page.tsx:201`), which is rendered from `:30`. But it early-returns:

```ts
if (agent.availability.state !== "suspended") return null;
```

and `projectAgentAvailability` (`agent-model.ts:96`) pairs `state: "suspended"` with exactly one
reason, `agent_suspended`. So the only message that component can ever display is the paused one.
These three can never render through it:

- `im_not_connected` — "Connect Feishu or Slack so teammates can send this Agent work."
- `im_error` — "…Reconnect Feishu or Slack to receive messages."
- `im_disabled` — "Messaging is turned off for this Agent. Reconnect…"

The other caller is `messagingAgentStatusDescription`, which is gap 1.

### Why this is worth recording rather than just deleting

These are **unreachable branches behind a narrowed guard**, not an orphaned component. The notice
*is* rendered; it simply answers one state now. Widening that guard — showing the notice for
`not_connected` or `im_error` too, which is a plausible product change — makes all three live again
immediately. So "delete the dead code" is not obviously right; deciding what the notice is *for* is.

### Disposition

Someone should decide between:

- **Widen** — the notice covers more availability states, and the three messages become reachable as
  originally intended; or
- **Delete** — drop `messagingAgentStatusDescription` and the unreachable branches, and let the
  Connection section own every non-suspended state.

Also relevant: `messagingProviderChoices()` (added in #311 to make the provider *set* follow
`ImProviderSchema` rather than being written out by hand) currently has no reachable consumer, since
its only call sites are those three messages. It is correct and tested, and #335 may want it as the
sanctioned way to name the set — but nobody should assume it ships to a screen today.

[...]

Issue #335, opened 2026-08-31
Enforce structurally that only messagingProviderLabel names a messaging channel


liuchao-001:
## The class

Three separate rounds of review on #311 each found a brand name reaching a reader from outside
`messagingProviderLabel`, and each was found **by eye**:

| Found by | Where | Mechanism |
|---|---|---|
| @bestony | `sharedConversationDestination` | branch picks between two correctly-spelled literals |
| @code-reviewer | `agentRecoveryMessage` ×3 | the provider **set** written out as "Feishu or Slack" |
| earlier rounds | 9 call sites | `titleCase(provider)` converting an identifier |

Only the third is findable by searching for a provider id, which is why the first two survived a
sweep that was reported clean. The others convert no identifier at all — the words were simply
correct, written by hand, in a place that is not the naming point.

## Why the existing guards don't close it

- **The exhaustive `switch`** in `messagingProviderLabel` guarantees every provider *has* a name. It
  says nothing about who else may write one.
- **Output assertions cannot see sourcing.** Verified by mutation: reintroduce the exact hardcoded
  literals *spelled correctly* and `agent-list-order.test.ts` is fully green, `toBe` or `toContain`
  alike. An assertion about rendered text cannot distinguish text that came from the helper from
  identical text spelled beside it.

So the invariant this PR is built on — *the helper is the only place a channel gets named* — is
currently maintained by reviewers noticing, and it has already leaked twice in one PR.

## Proposal

A structural check over non-test `apps/web/src`: no messaging-brand literal (`Feishu`, `Slack`, and
any future channel) outside `apps/web/src/im/provider-label.ts`.

Allowlist the files that are prose *about one provider* rather than naming choices — currently
`im/feishu-setup.tsx`, `im/slack-configuration.tsx`, `setup/copy.ts`. Those legitimately write a
provider's name into sentences; the allowlist should be explicit and short so that adding to it is a
visible decision.

Shape is open — a Vitest source-scanning test in the style of the existing drift guards, or a Biome
`noRestrictedSyntax` rule. A test is probably better: it can produce a useful failure message naming
the file and the offending literal.

This would have caught all three findings above, and it is the thing that makes the helper's
docstring true rather than aspirational.

## Not in #311

Deliberately out of scope there: that PR establishes the helper and routes the call sites through it,
and adding a repo-wide structural gate in the same change would mix "make the invariant hold now"
with "make it stay held". Raised at @code-reviewer's suggestion.

[...]

PR #456, opened 2026-09-01
feat(web): name the channel the brand the reader knows

liuchao-001:
Stacked on #311 — base is `feat/feishu-naming`, so review that first; the diff here is only what
this change adds.

## The premise changed

The naming work in #311 assumed a first connect could only mint against the Feishu domain, so naming
Lark would hand a tenant a code they could not authorize. **Testing against a real Lark tenant
disproved that**: one connect code serves both brands. The vendor detects the tenant during
authorization and switches domains itself — `domain_switched` fires, polling moves to the Lark host,
and credentials come back with `tenant_brand: "lark"`.

That removes the reason to name Feishu alone.

## What changes

There is no correct name in the abstract, only the name a reader knows the product by. So the label
answers by locale:

| locale | channel name | description |
|---|---|---|
| `zh` | **飞书** | 同时支持 Lark |
| `en` | **Lark** | Also supports Feishu |

Because this goes through `messagingProviderLabel` — the single naming point #311 established —
every surface follows at once: the onboarding picker, Agent cards, messaging settings, Task list and
detail, the QR panel. Slack is untouched; it has one name everywhere.

The picker states the coverage rather than making a reader guess whether their tenant qualifies.

## Why the counterpart brand is not a catalogue string

"Also supports **Feishu**" names a brand too. It comes from `messagingProviderAlternateBrand()`, and
the entry reads `Also supports {provider}`. Otherwise this change would break the rule it depends on —
sentences belong to the catalogue, every brand inside them to the helper.

## 28 catalogue entries converted

`messages/im` (26) and `messages/agents` (2) spelled the brand directly: `Connect Feishu`,
`Scan this QR code in Feishu`, `Connect Feishu or Slack so teammates…`. They arrived with main's i18n
migration, not with this work.

Leaving them would have shipped English users **"Lark" on the picker and "Connect Feishu" in
settings** — the same split naming this line of work exists to remove, only authored by us. All 28
now take `{provider}`, fed from the helper at the call site.

**Feishu and Lark now appear in no catalogue file**, and their only occurrences in non-test source
are inside `provider-label.ts`.

> **Correction (post-merge):** this originally claimed the invariant was "literally true for the
> first time — no brand name appears in any catalogue file." That overstated it. `Slack` still
> appears 22 times across `messages/im` and `messages/onboarding-v2`, deliberately left because it
> has one name in every locale, but the sweeping claim was wrong. See the correction comment; #335
> closes the remainder.

## `docs/i18n.md`

The glossary said Feishu and Lark stay **unchanged** as product names. 飞书 is not a translation of
Lark — it is the same channel's other regional brand — so the rule now says that, rather than
contradicting the code.

## Not changed

The three unreachable `im_*` recovery sentences keep their pre-PR literals per @yuezengwu's
instruction, since unreachable prose does not belong in a catalogue. They are now stale twice over —
unreachable, **and** naming a brand the English UI no longer uses. That strengthens the case in #336
for deleting them rather than maintaining them.

## Testing at `c775f00`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **66 files / 582 tests passed** |
| `pnpm check` | exit 0 |
| `node scripts/unit-coverage.mjs --project web` | exit 0 |

`provider-label.test.ts` pins both locales. The previous assertion there — that the label must never
say Lark — enforced the premise this change overturns, and is replaced rather than deleted.


apps/web/src/im/provider-label.ts
```diff
@@ -27,7 +25,7 @@ type MessagingProvider = ImBindingSummary["provider"];
 export function messagingProviderLabel(provider: MessagingProvider): string {
   switch (provider) {
     case "feishu":
-      return "Feishu";
+      return getLocale() === "zh" ? "飞书" : "Lark";
```
yuezengwu, line 28: [P2] Re-localize sentences when the label becomes Chinese

Returning `飞书` here changes the grammar of every existing `{provider}` template, but several zh catalog entries still carry the ASCII spacing that was appropriate for the old Latin `Feishu`. In the real onboarding flow this exact head renders `使用 飞书 扫描此码。最后一步会在 飞书 中完成。`, and the QR alternative text renders `在 飞书 中扫描此二维码`; the same pattern remains in done/error/CLI/Agent messages. Please update all affected zh sentences (or use locale-specific phrasing) so the centralized label does not introduce visibly malformed Chinese across surfaces.

docs/i18n.md
```diff
@@ -41,7 +41,8 @@ Use Paraglide messages for sentences, punctuation, grammar, and plural wording.
 | Runtime | 运行时 | |
 | Token | Token | Keep the product term. |
 | Sign in / Sign out | 登录 / 退出登录 | |
-| OpenTag, Codex, Claude Code, Slack, Feishu, Lark | unchanged | Product names. |
+| OpenTag, Codex, Claude Code, Slack | unchanged | Product names. |
+| Feishu / Lark | 飞书 / Lark | One channel, two regional brands. `messagingProviderLabel` picks the one the reader knows: 飞书 in `zh`, Lark in `en`. Neither is a translation of the other, so this is a brand choice, not a translated product name. Never spell either in a catalogue entry — take the label as a `{provider}` placeholder. |
```
yuezengwu, line 45: [P2] Keep the required Chinese mirror synchronized

This new canonical glossary rule is absent from `docs/zh-CN/i18n.md`, whose glossary still says Feishu and Lark are unchanged product names. On this exact head, `node scripts/check-doc-mirrors.mjs --base bd97460355864bd61fbd6bfb2fb2380841eb6b80 --head c775f00de414538576d12c085a46bac55361889e` fails for that missing mirror, matching the red Documentation Mirror Advisory check. Please synchronize the mirror (including its sync marker) with this contract.

yuezengwu (changes requested):
Requesting changes on exact head `c775f00de414538576d12c085a46bac55361889e`. The one-code Feishu/Lark premise is consistent with the reviewed SDK behavior, and the build, static checks, typecheck, full test suite, and Web coverage all passed. Two reader-facing/repository-contract issues remain below.

bestony (changes requested):
Reviewed exact head `c775f00de414538576d12c085a46bac55361889e` against stacked base `bd97460355864bd61fbd6bfb2fb2380841eb6b80`.

## Standards

[P2] `docs/i18n.md:45` changes the canonical Feishu/Lark glossary contract, but its required Chinese mirror is still stale: `docs/zh-CN/i18n.md:41` retains the old combined "product names stay unchanged" rule and line 4 still reports the previous sync date. This violates the repository rule in `CONTRIBUTING.md:30-31` that canonical documentation and its Chinese mirror change together. The exact mirror check fails for the same reason. Please synchronize the mirror and refresh its sync marker.

No additional documented-standard breach or high-confidence baseline smell found.

## Spec

[P2] Making `messagingProviderLabel("feishu")` return `飞书` under `zh` changes the grammar of every existing `{provider}` template, but several Chinese catalogue entries retain ASCII spaces that assumed a Latin label. For example, `apps/web/messages/onboarding-v2/zh.json:100` renders `使用 飞书 扫描此码。最后一步会在 飞书 中完成。`, and `apps/web/messages/im/zh.json:61` renders `在 飞书 中扫描此二维码`. The same shape exists in done, error, CLI, and Agent messages. Please re-localize all affected `zh` templates so `飞书` joins Chinese text without stray spaces while intentional spacing around Latin labels remains correct.

The locale-specific primary/alternate brands, 28 catalogue conversions, Slack behavior, and helper call sites otherwise match the PR specification. I found no scope creep.

Verification: `git diff --check` passed; `@opentag/shared` built; Web typecheck passed; Web tests passed (66 files / 582 tests). CI is pass/skipping except the expected Documentation Mirror Advisory failure.

Summary: Standards 1 finding (stale required mirror); Spec 1 finding (malformed Chinese spacing). Requesting changes for both P2 issues.

bestony (changes requested):
Re-reviewed exact head `c8d5d125840391637642395479e1927410edb902`; both reviews on `c775f00` are superseded.

## Standards

The required Chinese documentation mirror and sync marker are now aligned with the canonical glossary. The exact mirror check passes. I found no remaining documented-standard breach or high-confidence baseline smell in this repair delta.

## Spec

[P2] The repair removes spaces unconditionally, but these `{provider}` templates also receive Latin brands. The zh picker requirement in the PR body is explicitly `同时支持 Lark`; `apps/web/messages/onboarding-v2/zh.json:97` now says `同时支持{provider}`, and `steps.tsx:358-360` passes `messagingProviderAlternateBrand()` (`Lark` under zh), so the exact rendered result is `同时支持Lark`.

The same regression affects Slack-capable templates:

- `agents_available_in_channel` renders `可在Slack中使用`;
- `agents_shared_destination_channel(s)` renders `Slack频道` / `已连接的Slack频道`;
- `onboarding_v2_done_description` renders `在Slack中`;
- `onboarding_v2_messaging_cli_missing` renders `Slack消息...`;
- `onboarding_v2_messaging_provider_description` renders `你的Slack工作区`.

This closes the malformed spacing around CJK `飞书` by breaking the intentional spacing around Latin `Lark`/`Slack`. Please make spacing provider/script-aware or split the templates so both branches read correctly. The Feishu-only fixes (`im_qr_scan_alt`, Feishu attempt/preparing/intro copy, and Feishu group-chat copy) are correct, and I found no scope creep.

Verification: `git diff --check` passed; documentation mirror check passed (17 pairs); `@opentag/shared` built; Web typecheck passed; Web tests passed (66 files / 582 tests). Direct zh message rendering reproduced every output above.

Summary: Standards 0 findings; Spec 1 finding (Latin-brand spacing regression). Requesting changes for the remaining P2 issue.

apps/web/messages/onboarding-v2/zh.json
```diff
@@ -94,10 +94,11 @@
   "onboarding_v2_messaging_confirming": "已连接。正在检查 Agent 是否可访问…",
   "onboarding_v2_messaging_description": "选择团队已经使用的应用。",
   "onboarding_v2_messaging_failed": "操作未成功。请重试以获取新二维码。",
-  "onboarding_v2_messaging_feishu_preparing": "OpenTag 正在准备安全的 {provider} 授权。",
+  "onboarding_v2_messaging_feishu_description": "同时支持{provider}",
```
yuezengwu: [P2] Preserve spacing when the injected brand is Latin

Dropping the placeholder spacing unconditionally fixes `飞书`, but this same catalogue now receives Latin brands. On this exact head, the real zh onboarding flow renders `同时支持Lark` (the PR requirement itself says `同时支持 Lark`), `你的Slack工作区`, and, after completing Slack, `在Slack中 @opentag…`. Please make these sentences provider/script-aware or split the messages so CJK `飞书` joins without spaces while `Lark`/`Slack` retain the intentional spacing. The same regression is present in the shared completion and messaging-CLI messages and the Agent channel/destination templates.

yuezengwu (changes requested):
Requesting changes on exact head `c8d5d125840391637642395479e1927410edb902`. Fresh build, check, typecheck, unit tests, Web coverage, Client agent-runtime coverage, Server integration tests, doc-mirror validation, and diff check passed. I also completed the Chinese and English Lark/Feishu and Slack onboarding flows in the rendered UI. One Chinese locale regression remains below. The red Browser Smoke check failed while its PostgreSQL test database was shutting down during setup, before Playwright ran, so it did not produce a product assertion.

liuchao-001:
## Fixed at `24faf8b2` — @bestony and @yuezengwu found this independently, and both were right

My first repair removed the placeholder spacing unconditionally. That fixed 飞书 and broke Slack:
`同时支持Lark`, `你的Slack工作区`, `在Slack中 @opentag…`.

**The template cannot own the spacing.** The same `{provider}` slot receives 飞书 in one branch and
Slack in the next, and Chinese sets no space between Chinese characters but one space either side of
Latin text. Which rule applies is a property of the *value*, not the sentence.

So the label carries its own boundary. `messagingProviderLabelInSentence` pads a Latin brand under
`zh` and leaves a CJK one bare; under `en` it adds nothing, because English already spaces every
word. Interpolation sites use it — a label rendered alone, on an Agent card or a button, keeps the
unpadded `messagingProviderLabel` so there is no whitespace to trim.

### Verified by rendering, not by reading

```
zh   可在飞书中使用        可在 Slack 中使用        同时支持 Lark
en   Available in Lark    Available in Slack       Also supports Feishu
```

The rule is pinned in `provider-label.test.ts` for both locales, so a future edit that drops the
padding fails rather than shipping cramped Chinese.

### One residue I would rather name than let you find

When `{provider}` sits at a string boundary the padded form leaves an edge space — `agents_shared_destination_channel`
renders `" Slack 频道"`. HTML collapses leading and trailing whitespace in normal flow, so it is not
visible, and the alternative is duplicating every affected template per script. If you would rather
have it exact, splitting those two boundary templates is the way, and I will do it — I did not want to
double the catalogue on my own judgement.

### On the pattern

This is the same shape as the three comment revisions on #311: each repair was accurate about the case
in front of it and careless about the one beside it. Twice now the fix for a real finding introduced
the mirror-image defect, and both times a reviewer caught it rather than me.

### @yuezengwu on Browser Smoke

Agreed, and it matches the earlier occurrence on #311: the job failed while its PostgreSQL database
was shutting down during setup, before Playwright ran, so it produced no product assertion. That gate
has now failed twice on this stack for infrastructure reasons with zero executed tests.

### Verification at `24faf8b2`

typecheck 0 · `pnpm check` 0 · **66 files / 584 tests** · doc mirror passes (17 pairs).

bestony (changes requested):
Requesting changes on exact head `24faf8b2c5abea504a37a9ebd6b06cb1b903a586`, reviewed against stacked base `bd97460355864bd61fbd6bfb2fb2380841eb6b80`.

## Standards

No additional documented-standard breach found in this repair delta.

## Spec

[P2] `sharedConversationDestination()` still passes the unpadded `messagingProviderLabel()` at `apps/web/src/features/agents/agent-presentation.ts:190`. Its zh catalogue entries remain `{provider}频道` / `已连接的{provider}频道`, so the Slack path still renders `Slack频道` and `已连接的Slack频道`; the latest commit's claimed provider-aware fix does not reach these Agent destination surfaces. Pass the script-aware helper (or otherwise preserve the Latin spacing) here as well.

[P2] The new script-aware helper is also applied to `{providerName}` templates that already own their spacing. At `apps/web/src/features/agents/agent-settings/im-tab.tsx:85,202,300,303,317`, `messagingProviderLabelInSentence("slack")` is interpolated into zh entries such as `断开 {providerName}` and `通过 {providerName} 向...`. The rendered settings UI therefore contains doubled and trailing spaces, e.g. `断开  Slack `, `团队成员将无法再通过  Slack  向此 Agent 发送消息。`, and `断开  Slack ？`. Use the unpadded label for these templates, or re-localize the templates before applying boundary padding.

The Feishu/Slack sentence-aware outputs otherwise match the requested behavior (`可在飞书中使用`, `可在 Slack 中使用`, `同时支持 Lark`).

Verification: `git diff --check` passes; direct zh rendering reproduces both findings. The latest CI run was still in progress when reviewed.

Summary: Standards 0 findings; Spec 2 P2 findings, both requiring the new spacing helper to be applied only at compatible interpolation boundaries.

apps/web/src/im/provider-label.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Slack in the next. The label knows its own script, so it carries its own boundary — CJK joins the
+ * sentence directly, Latin brings the spaces Chinese typography expects. In `en` this is the label
+ * unchanged, because English already spaces every word.
+ *
+ * Only interpolation sites use this. A label rendered on its own — an Agent card, a button — wants
+ * `messagingProviderLabel`, which has no padding to trim.
+ */
+export function messagingProviderLabelInSentence(provider: MessagingProvider): string {
+  const label = messagingProviderLabel(provider);
+  return getLocale() === "zh" && !/^[\u3400-\u9fff]/.test(label) ? ` ${label} ` : label;
```
yuezengwu: [P2] Keep the sentence boundary in the message context

Padding a Latin label on both sides cannot account for templates that already own whitespace or place the placeholder next to punctuation or at an edge. On this exact head, rendering the generated zh messages with these changed call sites produces `断开 飞书` / `断开 飞书？` for Feishu, `断开  Slack ` / `断开  Slack ？` for Slack, and `消息渠道： Slack ` in the settings detail. The helper-only tests pin the padded value but never render these affected messages. Please make the boundary contextual (or update/split the relevant templates and keep an unpadded value where the template owns spacing), and add full-message zh assertions for both Feishu and Slack.

yuezengwu (changes requested):
Requesting changes on exact head `24faf8b2c5abea504a37a9ebd6b06cb1b903a586`. Fresh install, build, check, typecheck, full tests, Web coverage, Client agent-runtime coverage, Server integration tests, documentation-mirror validation, and diff check passed. CI is fully pass/skipping. I also completed the rendered Chinese Feishu and Slack plus English Lark and Slack onboarding flows through their completion pages; those four flows now use the intended brands and spacing, with no browser warnings or errors. One remaining Chinese settings regression is described below. During the publication gate, stacked parent #311 merged and GitHub retargeted this PR from its reviewed parent tip `bd97460355864bd61fbd6bfb2fb2380841eb6b80` to `main@133c2af13be4223a376dba3b0185735b6c6ec441`; the current head is now conflicting, so merged-tree validation must wait for a rebased head.

liuchao-001:
## Fixed at `0aa6503a` — @yuezengwu's diagnosis was the right one, and the defect was larger than reported

@yuezengwu and @bestony independently found the same class of bug on `24faf8b2`, and @yuezengwu named the cause precisely: *"padding a Latin label on both sides cannot account for templates that already own whitespace or place the placeholder next to punctuation or at an edge."* That is exactly right, and it is why the previous head was wrong.

### The model was wrong, not the individual sites

Spacing between Chinese and Latin is a property of the **finished sentence**. Neither end can decide it:

- The **template** can't: the same `{provider}` slot receives 飞书 in one branch and Slack in the next, so a space written there is right for one brand and wrong for the other.
- The **label** can't: padding is right mid-sentence, but leaves a stray space when the slot ends the string, and sets a space against Chinese punctuation that already carries its own width.

So the rule now runs over the assembled message — `spaceBrandInSentence` — and templates leave the slot unspaced. Chinese punctuation sits deliberately outside the character ranges, which is what stops the `Slack ，请重试` failure from coming back.

### It was 17 strings, not 2

Rather than fix the reported sites, I rendered every catalogue entry under every brand its call site can actually pass. **17 renderings were defective**, including all of the ones you both quoted:

| | before | now |
|---|---|---|
| `im_disconnect` | `断开 飞书` / `断开  Slack ` | `断开飞书` / `断开 Slack` |
| `im_disconnect_title` | `断开 飞书？` / `断开  Slack ？` | `断开飞书？` / `断开 Slack？` |
| `im_disconnect_failed` | `无法断开  Slack ，请重试。` | `无法断开 Slack，请重试。` |
| `agent_settings_channel_detail` | `消息渠道： Slack ` | `消息渠道：Slack` |
| `sharedConversationDestination` | `Slack频道` | `Slack 频道` |

@bestony's first finding — `sharedConversationDestination` passing an unpadded label — is fixed by the same change.

### Full-message zh assertions, as asked

`apps/web/src/im/brand-in-message.test.ts` asserts finished messages for a Chinese brand and a Latin one, through the functions the UI calls. @yuezengwu's point that the old tests "pin the padded value but never render these affected messages" was the reason a green suite coexisted with broken copy.

Mutation-checked against the exact string you reported: restoring the template space so `im_disconnect` renders `断开 飞书` fails the suite.

### Two slots deliberately keep their template spacing

Where a slot can only ever receive a Latin brand — the alternate-brand line (always Lark in `zh`) and the Slack-only picker description — the template owns the space, because only one script can land there. Same for slots that abut a Latin literal (`{provider} Bot`) or the `·` separator.

### Merge resolution

#311 landed as a squash while this was in review, so this branch's base commits and main's copy of them collided across 17 files — ten of which also carry independent main work (#454 and earlier). Neither side was safe to take wholesale, so each conflicted file was rebuilt from main with only this PR's delta re-applied.

Two spots needed real judgement:

- `im_feishu_dialog_connect_description`: main rewrote it to *"Scan with Feishu on your phone"* while this branch replaced its brand with a placeholder. Merged entry keeps **main's wording and this branch's `{provider}`**.
- `feishu-setup.tsx`: main refactored it into `FeishuSetupDialogContent`, so the provider arguments were re-applied to main's structure, not the pre-refactor file.

### Testing at `0aa6503a`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **68 files / 594 tests passed** |
| `pnpm check` | exit 0 |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 93.42%) |

Note `pnpm install --frozen-lockfile` and a `@opentag/shared` build are required after taking this merge — main added `@babel/parser` for the UI governance contract, and a stale install fails 59 tests for reasons unrelated to this diff.

This head has had no independent live QA; the previous head's flows do not carry over, since the rendered Chinese strings are exactly what changed.

bestony (approved):
Reviewed exact head 0aa6503a096c05ceb3cf6b37717fa18f594ca9e2 against main@133c2af13be4223a376dba3b0185735b6c6ec441. The finished-sentence spacing helper now preserves Chinese/Latin boundaries across Agent destinations, onboarding completion/CLI copy, and settings confirmation text; full-message zh assertions cover Feishu and Slack. Documentation mirror and sync marker are aligned. Local focused tests and full Web tests pass; GitHub CI and patch coverage are green. No remaining Standards or Spec findings.

apps/web/src/im/provider-label.ts
liuchao-001: Fixed at `0aa6503a`. Both strings you cited now render without the stray spaces:

- `使用飞书扫描此码。最后一步会在飞书中完成。` (was `使用 飞书 扫描此码。最后一步会在 飞书 中完成。`)
- `在飞书中扫描此二维码` (was `在 飞书 中扫描此二维码`)

Your diagnosis in the later thread is what drove the fix: the spacing now runs over the finished sentence rather than the label or the template, and these templates leave the slot unspaced. I rendered every catalogue entry under every brand its call site can pass — 17 were defective, including these two.

docs/i18n.md
liuchao-001: Fixed at `0aa6503a`. `docs/zh-CN/i18n.md` carries the mirrored glossary rule and the new spacing rule, and the sync marker is updated.

Your command passes on this head against merged main:

```
node scripts/check-doc-mirrors.mjs --base 133c2af1 --head 0aa6503a
Documentation mirror check passed (17 pairs discovered).
```

apps/web/messages/onboarding-v2/zh.json
liuchao-001: Fixed at `0aa6503a`, and you were right that stripping the spacing unconditionally was the wrong correction — it traded a Chinese-brand bug for a Latin-brand one.

The three strings you cited now render:

- `同时支持 Lark` — the PR requirement itself
- `你的 Slack 工作区`
- `在 Slack 中 @opentag，让它开始工作。` / `在飞书中 @opentag，让它开始工作。`

A slot that can only ever receive a Latin brand (the alternate-brand line, the Slack-only picker description) keeps its template spacing, because only one script can land there. Slots that can receive either leave it to the finished sentence.

apps/web/src/im/provider-label.ts
liuchao-001: Fixed at `0aa6503a`. Your framing was the correction — the boundary is now contextual, applied to the assembled message rather than padded onto the label, so a template that owns its whitespace and a slot at an edge or against punctuation are all handled by the same rule.

All five strings you cited now render:

| before | now |
|---|---|
| `断开 飞书` / `断开  Slack ` | `断开飞书` / `断开 Slack` |
| `断开 飞书？` / `断开  Slack ？` | `断开飞书？` / `断开 Slack？` |
| `消息渠道： Slack ` | `消息渠道：Slack` |

`apps/web/src/im/brand-in-message.test.ts` adds the full-message zh assertions you asked for, for both a Chinese brand and a Latin one, through the functions the UI calls. Mutation-checked against your exact string: putting the template space back so `im_disconnect` renders `断开 飞书` fails the suite.

yuezengwu (approved):
Exact-head review at 0aa6503a096c05ceb3cf6b37717fa18f594ca9e2 against main 133c2af13be4223a376dba3b0185735b6c6ec441.

No blocking findings.

Validated build, check, typecheck, full test suite, docs mirrors, Web unit coverage, Client agent-runtime coverage, Server integration, SDK domain-switch contract, and real browser flows for zh Feishu/Slack plus en Lark/Slack onboarding and messaging settings. Chinese spacing, QR copy and alt text, connect/change/disconnect dialogs, and completion copy matched the locale-aware brand contract.

liuchao-001:
## Correction: one claim in this PR's description was false

While scoping #335 against merged `main` (`100b4409`) I checked the invariant this PR claimed to establish, and it does not hold as stated. The description says:

> **The invariant is literally true for the first time:** no brand name appears in any catalogue file, and the only occurrences in non-test source are inside `provider-label.ts`.

That is true of **Feishu / Lark / 飞书** — zero occurrences in `messages/`. It is **not** true of Slack:

| brand | occurrences in `apps/web/messages/` |
|---|---|
| Feishu | 0 |
| Lark | 0 |
| 飞书 | 0 |
| **Slack** | **22** |

across `messages/im/{en,zh}.json` and `messages/onboarding-v2/{en,zh}.json` — `"Connect Slack"`, `"Slack connected."`, `"连接 Slack"`, and so on.

The scope decision itself was defensible, and the description states it elsewhere: *"Slack is untouched; it has one name everywhere."* A brand with a single name in every locale gains nothing from the locale-aware helper. What was wrong was generalising that into "no brand name appears in any catalogue file" — the sentence claims the invariant closed when only half of it did.

This matters beyond bookkeeping, because one of those entries is the exact mechanism #335 exists to catch — the provider **set** written out by hand:

```
im_messaging_connect_description: "连接{provider}或 Slack，让团队成员可以向此 Agent 发送消息。"
```

`{provider}` comes from the helper; `Slack` beside it does not. That is @code-reviewer's `agentRecoveryMessage` finding, surviving in a new location.

No behaviour is wrong today — `messagingProviderLabel("slack")` returns `"Slack"`, so the rendered text is identical. The cost is that the guarantee is narrower than the record says, and #335 was about to be built on that record.

I am carrying this into #335: the guard covers the catalogues as well as source, which means converting those 22 entries rather than allowlisting them. Raising it here so the claim is corrected where it was made.

Issue #241, opened 2026-08-29
[Feature]: Add i18n support for the Control Panel (English and Simplified Chinese)


bestony:
## User problem / 用户问题

The OpenTag Control Panel is currently available in only one language. Chinese-speaking users cannot use the interface in Simplified Chinese, while teams that need an English UI cannot switch to English consistently. This creates friction across onboarding, daily operations, and error handling.

OpenTag 控制面板目前只有一种语言。中文用户无法使用简体中文界面，需要英文界面的团队也无法稳定切换到英文，导致新手引导、日常操作和错误处理都存在使用障碍。

## Desired outcome / 期望结果

Add a maintainable i18n foundation to the Control Panel and ship both English (`en-US`) and Simplified Chinese (`zh-CN`) locales.

为控制面板增加可维护的国际化基础，并同时提供英文（`en-US`）和简体中文（`zh-CN`）界面。

### Acceptance criteria / 验收标准

- Provide an obvious language switcher and persist the user's selection across sessions; use the browser/system locale as the initial default when possible, with English as the fallback.
- Add translations for all user-facing Control Panel content, including navigation, authentication and onboarding, account/workspace/task/client/session views, settings, notifications, validation and error messages, empty/loading states, dialogs, and tooltips.
- Replace hard-coded UI strings with namespaced translation keys so future locales can be added without changing components.
- Support interpolation, pluralization, and locale-aware date, time, and number formatting where applicable.
- Verify that both locales have complete translation coverage and that translated text does not introduce layout overflow or regress keyboard/screen-reader accessibility.
- Add automated coverage (for example, translation-key completeness checks and a smoke test for switching locales).

- 提供清晰可见的语言切换入口，并跨会话保存用户选择；首次访问时尽可能根据浏览器/系统语言初始化，无法识别时回退到英文。
- 覆盖控制面板所有面向用户的内容，包括导航、登录与新手引导、Account/Workspace/Task/Client/Session 页面、设置、通知、校验与错误信息、空状态/加载状态、对话框和 Tooltip。
- 使用命名空间翻译 key 替换组件中的硬编码文案，确保未来增加其他语言时无需修改组件。
- 在适用场景支持插值、复数规则，以及按语言环境格式化日期、时间和数字。
- 验证两种语言的翻译覆盖完整，并确保译文不会造成布局溢出，同时不影响键盘操作和屏幕阅读器可访问性。
[...]

PR #459, opened 2026-09-01
test(web): make the naming point the only place a channel is named

### LINKED ISSUE #335: Enforce structurally that only messagingProviderLabel names a messaging channel

liuchao-001:
Closes #335.

## Why a structural check and not a test assertion

Three separate review rounds on #311 each found a brand name reaching a reader from outside `messagingProviderLabel`, and each was found **by eye**:

| Found by | Where | Mechanism |
|---|---|---|
| @bestony | `sharedConversationDestination` | a branch picking between two correctly-spelled literals |
| @code-reviewer | `agentRecoveryMessage` ×3 | the provider **set** written out as "Feishu or Slack" |
| earlier rounds | 9 call sites | `titleCase(provider)` converting an identifier |

Only the third is findable by searching for a provider id. The other two convert no identifier at all — the words were simply correct, hand-written somewhere that is not the naming point.

And no output assertion can close it: reintroduce the exact literal, spelled correctly, and every rendered-text test stays green. Rendered text carries no evidence of where it came from. That is why this reads the syntax tree rather than the output — and why it reads **string literals, template chunks and JSX text** rather than grepping, since most brand-shaped words in this tree (`FeishuSetup`, `slackInstallUrl`, the `"feishu"` id) are identifiers no reader ever sees.

## Scoping it found a hole in what #456 claimed

#456's description said the invariant was "literally true for the first time: no brand name appears in any catalogue file." Checking that against merged `main`:

| brand | occurrences in `apps/web/messages/` |
|---|---|
| Feishu / Lark / 飞书 | 0 |
| **Slack** | **22**, across 10 entries |

The scope decision was defensible — Slack has one name in every locale, so it gains nothing from a locale-aware helper — but the claim was wrong, and one of those entries is the mechanism this guard exists to catch:

```
im_messaging_connect_description: "Connect {provider} or Slack so teammates can send messages to this Agent."
```

`{provider}` comes from the helper; `Slack` beside it does not. That is @code-reviewer's finding, surviving in a new location. I corrected the claim on #456 itself.

So this PR converts those ten entries to take the label from the helper. Rendered output is byte-identical today — `messagingProviderLabel("slack")` returns `"Slack"` — so this is not a copy change.

## The allowlist

Five entries, each carrying its reason, because adding to it should feel like a decision:

| Path | Why |
|---|---|
| `im/provider-label.ts` | the naming point itself |
| `setup/copy.ts` | pre-i18n copy that never moved into a catalogue — #241 |
| `features/agents/agent-presentation.ts` | three sentences that no longer render at all — #336 deletes them |
| `mock/task-data.ts`, `onboarding-v2/mock-backend.ts` | demo fixtures, not product copy |

Note the issue proposed allowlisting `im/feishu-setup.tsx` and `im/slack-configuration.tsx`. Neither is needed — #456 already moved their copy into catalogues, so the guard lands tighter than the issue anticipated.

A third assertion fails when an allowlist entry stops matching a file that still spells a brand, so an exemption cannot quietly outlive its reason.

## Mutation-checked

Both halves fail when the defect is reintroduced, with the file, line and offending text:

```
+ features/agents/agent-model.ts:224 "Connect Feishu or Slack to continue."
+ messages/im/en.json:0 "im_slack_connected: Slack connected."
```

The first is the exact shape that stayed green under every existing test.

## Testing at `ece18cc4`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **70 files / 599 tests passed** |
| `pnpm check` | exit 0 (complexity ratchet passed, 0 baselines raised) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 93.33%) |
| `node scripts/check-doc-mirrors.mjs` | passed, 17 pairs |

The ratchet initially flagged a complexity warning in the new test and a one-line growth in `steps.tsx`; both were fixed rather than accepted, so no baseline moves.

No live QA needed: no rendered string changes.


bestony (changes requested):
## Findings

### [P1] Keep the guarded brand set coupled to provider labels

`BRANDS` is a hard-coded four-name regex, so the new guard does not actually enforce the stated “any future channel” invariant. For example, adding a `teams` provider and returning `"Microsoft Teams"` from `messagingProviderLabel` would still let a reader-facing `"Connect Microsoft Teams"` literal (in source or a catalogue) pass every test unless somebody remembers to edit this test's regex as a separate step. The exhaustive switch only forces the label helper to be updated; it does not force this allowlist of brand spellings to be updated. Please derive the protected names from the label source or add a test that fails when a provider label is not represented in the guard.

### [P1] Do not exempt the whole agent-presentation module

The file-level exemption at `apps/web/src/im/brand-naming-guard.test.ts:41-42` skips every reader-facing literal in `features/agents/agent-presentation.ts`, not just the three currently unreachable branches. Those branches still contain `"Connect Feishu or Slack"` / `"Reconnect Feishu or Slack"` at `agent-presentation.ts:299,302-303`, and `AgentLifecycleNotice` is a live caller (`agent-detail-page.tsx:253`). They are unreachable only while that notice remains restricted to `state === "suspended"`; widening the notice later would immediately expose the hardcoded names, while this guard would remain green. Please either remove/route these literals now, or make the exemption narrow enough that new reader-facing literals in the module are still detected.

## Verification

The new guard passed its 3 tests, web typecheck passed after building `@opentag/shared`, the full web suite passed (70 files / 599 tests), and `pnpm check` passed. The findings above concern gaps that the current tests do not exercise.

yuezengwu (changes requested):
## Finding

### [P1] Make the stale-exemption check inspect the exempt reader text

`brand-naming-guard.test.ts:132-136` keeps an allowlist entry whenever `BRANDS` appears anywhere in the raw file, even though the actual guard deliberately restricts itself to AST reader-text nodes. That makes the claimed stale-exemption ratchet ineffective. For example, if #336 deletes the three exempt literals at `agent-presentation.ts:299,302-303`, brand names still remain in comments at lines 130, 189-190, and 212, so this assertion stays green and the whole-file skip at line 83 remains in force. Likewise, removing `"Return from Slack"` from `mock-backend.ts` still leaves brand names in comments and identifiers.

This means an exemption can quietly outlive its stated reason and leave the entire module permanently unguarded—the exact state the third assertion says it prevents. Please derive staleness from the same parsed reader-text findings and pin the exact allowed finding(s), rather than testing raw file bytes. That also gives new literals inside an already-allowed file a chance to fail.

## Verification

- exact head: `ece18cc4e3c8593c17c1b7822bdcc7f0150fbf5e`
- `pnpm --filter @opentag/web test`: 70 files / 599 tests passed
- Web typecheck, `pnpm check`, shared build, diff check, and documentation mirror check passed
- Focused runtime rendering preserved the affected English and Chinese output, including `Connect Lark or Slack…` / `连接飞书或 Slack…` and all changed Slack onboarding/error strings

liuchao-001:
## All three P1s fixed at `899dfad7`

@bestony and @yuezengwu found three findings that are the same finding: **the guard relied on somebody remembering, which is the failure it exists to remove.** That is a fair thing to have caught me on — I wrote a check against "someone will forget" and then built it on a list someone has to maintain.

### P1: the brand set was written out by hand (@bestony)

Right, and your example is the proof: a `teams` provider returning `"Microsoft Teams"` would leave `"Connect Microsoft Teams"` passing every test. The exhaustive `switch` forces the *helper* to learn a new name; it cannot force a separate list of spellings to learn it too.

The set is now **asked of the helper** — every provider in `ImProviderSchema.options`, in every locale in `locales`, plus the alternate brand. Adding a provider widens the guard on its own.

Mutation, running your exact scenario and **without editing the guard**:

```
messagingProviderLabel("slack") -> "Microsoft Teams"
+ features/agents/agent-model.ts:224 "Connect Microsoft Teams to continue."
```

### P1: the whole-module exemption (@bestony)

Also right, and the reasoning matters more than the case: `AgentLifecycleNotice` *is* a live caller, and those branches are unreachable only while it stays restricted to `state === "suspended"`. Widening that notice later would expose the hardcoded names with the guard still green.

I took your first option rather than narrowing. Those three sentences now build the channel list from the provider schema through the helper:

```ts
function connectableChannels(): string {
  return ImProviderSchema.options.map(messagingProviderLabel).join(" or ");
}
```

The sentence and its English grammar stay in the module — @yuezengwu's boundary — while every brand inside it comes from the naming point. **`agent-presentation.ts` is no longer exempt at all.**

Every other exemption is now an **exact string** rather than a file, so a new literal in an already-exempt file still fails:

```
setup/copy.ts:103 "Reconnect Slack to continue."
```

### P1: the stale check read the file's bytes (@yuezengwu)

Correct, and it defeated the assertion's whole purpose. Your #336 example is exactly what would have happened: delete the three literals and brand names remain in comments at 130, 189-190 and 212, so the byte-level check stays green and the module stays permanently unguarded.

Staleness is now derived from **the same parsed reader text the guard uses**. Mutation on `mock-backend.ts`, which still has 4 brand words in comments and identifiers after the literal is gone:

```
+ [ "onboarding-v2/mock-backend.ts: \"Return from Slack\"" ]
```

The old byte-level check passed this case; this one fails it.

## Testing at `899dfad7`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **70 files / 600 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 0 baselines raised) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 93.33%) |

@yuezengwu — your runtime check of `Connect Lark or Slack…` / `连接飞书或 Slack…` still holds: `connectableChannels()` produces the same English list from the same two providers, and the catalogue entry is unchanged.

bestony (approved):
Re-reviewed at `899dfad7`. The three requested changes are addressed:

- guarded brand names are derived from `ImProviderSchema`, all configured locales, and the helper;
- `agent-presentation.ts` is no longer exempt as a whole file, and its channel list is schema-derived;
- exemptions are exact reader-text strings and stale checks reuse the parsed findings.

The focused guard suite passes (4 tests), web typecheck passes, and the remaining GitHub checks are still running.

yuezengwu (approved):
Fresh exact-head review of `899dfad76eebc2c0ace71db5b4e1342b40ecf370` (the prior-head findings and validation were not carried forward).

The new guard now derives provider brands from `ImProviderSchema` and the locale label helpers, exempts only the intended visible strings, checks exemptions against parsed reader-visible text, and builds the agent recovery channel list from the same schema/label source. I found no remaining correctness blocker.

Fresh validation passed: Web tests (70 files / 600 tests), Web typecheck, `pnpm check`, shared build, `git diff --check`, and documentation mirror checks. GitHub CI and patch coverage are green. Local browser QA also exercised the Chinese Cloud-agent-to-Slack onboarding flow through the waiting state and confirmed the changed visible copy; the English copy is covered by the passing Web suite.

Remaining scope: the browser run used the repository mock backend, so it did not perform a real external Slack OAuth installation; this PR does not change that external integration path.

PR #462, opened 2026-09-02
feat(server): log classified request failures and probe health for real

bestony:
The server's request boundary classified eight error families into `ErrorEnvelope` responses but
logged only the last, unclassified 500. Everything else — including two families that return 5xx —
returned silently. This makes each classified failure observable, stops pino's default `err`
serializer from printing database internals, makes `/healthz` tell the truth, and silences the probe
noise.

Response body contracts are unchanged; logging is observation only.

## Checklist

- [x] 1. `err` serializer added to `createFastifyLoggerOptions`
- [x] 2. Single logging helper called from every classified branch, with the status-based level matrix
- [x] 3. Anonymous `AuthServiceError` 401 produces no log line
- [x] 4. Rate-limit logging records key kind only, first trigger per window only
- [x] 5. `/healthz` performs a real database check, `ServerHealthSchema` contract unchanged
- [x] 6. `/healthz` and `/readyz` have `disableRequestLogging`
- [x] 7. Tests extended
- [x] 8. Progress recorded

## What changed

**`err` serializer (highest leverage).** Pino's default `stdSerializers.err` prints a Drizzle error's
`detail` and `where`, which contain the values of the offending row. `serializeError` now emits only
`type`, `message` and a stack bounded to 8 KiB, closing every existing and future `{err}` site at
once. Verified against a Drizzle-shaped error carrying `detail`, `where`, `schema`, `table`,
`constraint` and `code`: none survive, where pino's default would have emitted all six.

**Classified failures.** `setErrorHandler` has seven branches; only the final unclassified 500 logged.
A single `logClassifiedFailure(request, {code, category, statusCode}, error)` helper is now called
from each branch — one helper rather than seven inline conditionals, to keep the complexity ratchet's
per-file warning count flat. Level by status: `>= 500` → `error`, `409`/`429` → `warn`, other 4xx →
`info`.

**Anonymous 401s stay silent.** `plugins/user-auth.ts` throws `AuthServiceError` for every
unauthenticated request and the SPA hits that on each load, so the helper returns early for
`AuthServiceError` with status 401 rather than logging one line per page load.

**Rate limiting logs a kind, never a key.** The limiter keys on `sign-in:${email}`. The log records
`route:keyKind` only, and a per-limiter cooldown map emits one line per window instead of one per
rejected request.

**`/healthz` probes the database.** It previously returned a hardcoded `{status:"ok"}` without ever
touching PostgreSQL, so an unreachable database or an exhausted pool still reported healthy and the
orchestrator never restarted the container. It now runs `select 1` and returns 503 on failure; the
200 path still goes through `ServerHealthSchema.parse` unchanged.

**Probe noise.** `/healthz` and `/readyz` are excluded from request logging. `ignoreHttpTraceRoute`
already excluded them from tracing, which is a different mechanism. The Dockerfile probes every 30 s,
so this removes roughly 5,760 zero-signal lines per day per container.

| Criterion | Result |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm build` | exit 0 |
| `pnpm --filter '@opentag/*' test` | exit 0 — shared 172, server 615, client 877, web 600 |
| `pnpm --filter @opentag/server test` | exit 0 |
| `request-logging.test.ts` extended, never weakened | 5 → 10 cases, 141 insertions / 1 deletion, **zero assertions removed** |
| Existing `ErrorEnvelope` assertions still pass | zero removed lines touching `ErrorEnvelope`, `reply.code` or `.send(` |
| Test proves anonymous 401 emits no log record | `does not emit a classified log for an anonymous authentication failure` |
| Test proves rate-limit log carries no email | `logs one email rate-limit trigger with only its key kind` |

Independent check beyond the suite: a Drizzle-shaped error was serialized through the new contract and
none of `victim@example.com`, `hunter2`, `Key (email)` or `INSERT INTO` survived.

## Notes

No deviation from the agreed plan. Four commits across eight checklist items — coarser than
one-per-item, but each is a coherent unit (`whitelist request error logs`, `log classified request
failures`, `make health probe database aware`, `scope rate-limit log cooldowns`).

One bound worth knowing: the serializer caps `stack` at 8 KiB but does not cap `message`. Error
messages here are provider- and driver-generated and bounded in practice, but an unbounded message
would still be logged in full.

---

*Dispatched as run `313360`, lane `request-boundary-logs-1`. Written by codex under bypassed
approvals in an isolated worktree; verified and published by the orchestrator.*


yuezengwu (changes requested):
[P1] The serializer still exposes real database parameters

packages/server/src/app.ts:115-120 whitelists field names, but both retained values are unsafe for DrizzleQueryError. I reproduced the exact-head app with a real DrizzleQueryError created from query parameters victim@example.com and hunter2; the emitted err.message and err.stack both contained both canaries. Drizzle constructs its message as Failed query plus params, so the current test using a hand-built generic Error does not exercise the production shape. Please make the logged error shape safe by construction, including message and stack content, and add a regression using the real Drizzle error with secret canaries.

[P1] Bound and separate dependency health from process liveness

packages/server/src/app.ts:256-263 awaits the shared application pool with no server-side deadline or cancellation. createDatabaseClient also configures no query/connect deadline, while Docker calls /healthz every 30 seconds with a 5 second client timeout. With execute returning a never-settling promise, the exact-head route was still pending after the test deadline; the external wget timeout does not cancel that server-side database work. During pool exhaustion or a hung database, probes therefore accumulate and this endpoint marks/restarts an otherwise live process for a shared dependency outage. Keep liveness independent from database readiness, or use a separate dependency probe with a real server-side deadline/cancellation; in either design add a never-settling probe regression.

Validation on exact head 326e5fa16c43a4d5db9d28705cdeef667844b684:
- git diff --check passed
- focused app/request logging tests: 15/15 passed
- full @opentag/server tests: 615/615 passed
- @opentag/server typecheck passed
- pnpm check passed with the repository existing complexity warnings

Environment note: local Node 24.7.0 is below the head declared 24.15.0 minimum, so commands used the pnpm engine-strict override; CI Checks and patch coverage are green on this exact head.

bestony:
## Correction: my verification claim above is wrong

The description says the `err` serializer was "verified against a Drizzle-shaped error" and that no
secrets survive. **That claim does not hold, and the [P1] review is correct.**

My probe constructed a generic `Error` and attached `detail`, `where`, `schema`, `table`, `constraint`
and `code` as **own properties**. The whitelist does drop those — but that is not the production shape.
In the installed `drizzle-orm@0.45.2`, `errors.cjs:35-38`:

```js
class DrizzleQueryError extends Error {
  constructor(query, params, cause) {
    super(`Failed query: ${query}\nparams: ${params}`);
```

The query and its parameters are in the **message**, which the whitelist retains — and `stack` begins
with the message, so the values are logged twice. I verified the wrong object shape and reported the
result as if it settled the question. That is my error, not the lane's.

The second [P1] on `/healthz` also stands: the route awaits the shared pool with no server-side
deadline, so a hung database turns a liveness probe into a dependency probe that can restart an
otherwise healthy process.

Both findings are being treated as blocking. Nothing here should be merged on the strength of the
verification table in the description; the `err` serializer row in particular is unsupported.

bestony:
## Review round 1 addressed (`1598a705`)

Both P1s are fixed. **The reviewer was right on both, and on the first one my own verification claim in
the description was wrong — I retracted it in a separate comment above.**

### [P1] The serializer no longer exposes database parameters

A whitelist of field *names* was never enough: `drizzle-orm@0.45.2` builds the message itself from the
inputs —

```js
super(`Failed query: ${query}\nparams: ${params}`);   // errors.cjs:37
```

— so retaining `message` (and `stack`, which begins with it) logged every bound parameter twice.

`serializeError` now recognises the class (`instanceof DrizzleQueryError`, plus an own-`query`+`params`
fallback for look-alikes) and emits a **constructed** summary instead of anything derived from the
error's own text:

```
{ type: "DrizzleQueryError",
  message: "Database query failed (2 parameters)",
  stack:   "DrizzleQueryError: Database query failed (2 parameters)" }
```

A parameter **count**, never the values, never the query. Every other error path additionally passes
`message` and `stack` through `redactForLog` as a second layer — not the fix on its own, since
arbitrary bound values do not match credential patterns.

Regression added exactly as asked: `never emits bound parameters from a real DrizzleQueryError` builds
a **real** `new DrizzleQueryError(query, ["victim@example.com", "hunter2"])`, throws it through the
app, and asserts neither canary appears anywhere in the captured log output — then pins the emitted
shape. The original generic-`Error` case is retained.

### [P1] Liveness separated from dependency health, with a real deadline

`/healthz` is a **liveness** probe again — `ServerHealthSchema.parse({status:"ok"})`, no database call,
contract byte-identical. The database check moved to `/readyz`, which already existed and already
returns 503 when not ready.

The readiness probe carries a **1 s server-side deadline** (`DATABASE_READINESS_TIMEOUT_MS`,
`DatabaseReadinessTimeoutError`) and is **single-flight** (`inFlight`), so a hung database resolves as
not-ready instead of hanging, and concurrent 30 s Docker probes cannot pile up against it.

Regressions: `abandons a never-settling database probe at the server deadline` — the reviewer's exact
case — and `returns not ready when the database readiness probe fails`.

**My original brief for this lane was mis-specified**, and that is worth stating: I asked for
`/healthz` to probe the database because the old hardcoded `{status:"ok"}` never noticed an
unreachable one. The real gap was that nothing reported dependency health — not that liveness should
depend on it. Restarting the app does not fix a shared database; it removes a working instance.

### Re-verified

| Gate | Result |
| --- | --- |
| `pnpm --filter @opentag/server test` | exit 0 — 617 tests |
| `request-logging.test.ts` extended, never weakened | 11 cases, **zero assertions removed** since `origin/main` |
| Anonymous 401 emits no log record | test still present and passing |
| Rate-limit log carries no email | test still present and passing |

---

*Run `313360`, lane `request-boundary-logs-1`. Written by codex under bypassed approvals; verified and

packages/server/src/app.ts
```diff
@@ -97,11 +105,98 @@ export function ignoreHttpTraceRoute(path: string): boolean {
...
+function createDatabaseReadinessProbe(database: DatabaseClient | undefined): (() => Promise<void>) | undefined {
+  if (!database) return undefined;
+
+  let inFlight: Promise<void> | undefined;
+  return async () => {
+    if (!inFlight) {
+      const execution = Promise.resolve()
+        .then(() => database.execute(sql`select 1`))
+        .then(() => undefined);
+      const settled = execution.finally(() => {
```
yuezengwu, line 208: [P1] Let readiness recover after a stuck probe\n\nIf database.execute never settles, withDeadline rejects after one second but settled never runs its finally callback, so inFlight stays bound to that dead promise forever. Every later /readyz request reuses it and times out even after the database or pool has recovered; only a process restart can restore readiness. The current never-settling test checks only the first bounded response and misses this permanent poisoning. Please expire/replace the cached attempt under a bounded policy (without allowing unbounded concurrent probes) and add a test where the first call hangs but a later probe can succeed.

yuezengwu (changes requested):
Exact-head review of 1598a705b01dd94810b6f34cfd5b8b3b82b8f425 found one remaining readiness-recovery blocker. The Drizzle redaction fix and liveness/readiness split are sound; CI is green and the targeted server tests pass locally.

bestony:
## Review round 2 — readiness recovers after a hung probe

Addresses the P1 on `createDatabaseReadinessProbe`. Also merged current `main` (`5d2a0b21`) into the
branch, so this diff sits on the base as it stands after #443, #463 and #467.

### The bug, restated

`inFlight` was cleared only from the execution's own `finally`. When `database.execute` never settles,
that callback never runs, so the slot stays bound to a dead promise for the life of the process:
`withDeadline` bounded the *caller* at 1 s, but every later `/readyz` raced the same dead promise and
timed out — **even after the database and pool recovered**. Only a restart brought readiness back.
The round-1 never-settling test asserted only that the first response was bounded, which is why this
survived.

### The fix

The naive repair — clear `inFlight` when the deadline fires — restores recovery but lets every
request start its own probe against an already-sick pool, which is exactly what the single-flight
exists to prevent. So the **reuse window** is what is bounded, not just the caller's wait:

- The cached attempt is now `{ promise, startedAt }`, reused only while younger than
  `READINESS_ATTEMPT_MAX_AGE_MS = 10_000`; past that it is abandoned and exactly **one** replacement
  starts.
- The `inFlight === attempt` identity guard is kept, so a late-settling abandoned attempt cannot clear
  or clobber a newer one.
- An abandoned attempt gets a no-op `catch`, so an orphaned promise rejecting later cannot surface as
  an unhandled rejection. The readiness machinery must not become the outage it reports.
- The 1 s caller deadline is unchanged.

Net: at most one new database probe per 10 s regardless of request rate, and readiness recovers within
one window. 10 s was chosen against the 30 s Docker probe interval.

### Verified by the orchestrator, not self-reported

| Criterion | Result |
| --- | --- |
| `pnpm --filter @opentag/server test` | exit 0 — **63 files, 621 tests** |
| Readiness recovers after a hung first probe | covered by `recovers after an abandoned probe ages out without fanning out probes`: first probe hangs → 503 with 1 `execute`; two more calls in-window → still 503, still **1** `execute`; after the window → 200 `{status:"ready"}` with 2 `execute`; the abandoned promise is then rejected with no unhandled rejection |
| Probes in one window do not fan out | same test, asserted explicitly on the call count |
| Round-1 redaction still holds | **probed directly**: a real `DrizzleQueryError` carrying `victim@example.com`, `hunter2` and the query text emits none of the three |
| `/healthz` is still liveness-only | **probed directly**: 200 with **0** database calls |
| `request-logging.test.ts` | extended, not weakened |

### Notes

- `Date.now()` is read directly rather than injected; the regression drives it with vitest fake
  timers, so no test sleeps in real time.
- Pre-existing and untouched here: the readiness tests live under a `describe("GET /healthz")` block
  even though they exercise `/readyz`. Round 1 placed them there; renaming the block would churn the
  diff for no behavioural gain, so it is flagged rather than folded in.

---

*Dispatched as run `313360`, lane `request-boundary-logs-1`. Written by codex under bypassed approvals
in an isolated worktree; verified and published by the orchestrator.*

packages/server/src/app.ts
```diff
@@ -97,11 +105,111 @@ export function ignoreHttpTraceRoute(path: string): boolean {
...
+};
+
+function createDatabaseReadinessProbe(database: DatabaseClient | undefined): (() => Promise<void>) | undefined {
+  if (!database) return undefined;
+
+  let inFlight: DatabaseReadinessAttempt | undefined;
+  return async () => {
+    const now = Date.now();
+    const current = inFlight;
+    if (!current || now - current.startedAt >= READINESS_ATTEMPT_MAX_AGE_MS) {
```
yuezengwu, line 196: [P1] Do not abandon uncancelled probes into an unbounded backlog\n\nThe age check restores recovery by replacing a stuck promise, but it does not stop the database.execute behind that promise. During a persistent pool/database hang, every request after the 10 s window starts another execute while every previous one remains live; the normal 30 s Docker probe therefore adds one forever. I reproduced this head with execute returning a never-settling promise and requests at logical times 0, 10001, and 20002 ms: all returned 503 and the execute count grew 1 -> 2 -> 3. The current regression only makes the replacement resolve, so it misses the persistent-outage case. Bound or cancel the underlying operation so repeated aged-out windows cannot accumulate live probes, and pin that count across multiple windows.

yuezengwu (changes requested):
Exact-head review of ba4e7ddbed1b0deb610e177aa0f5fefa0e0328a8 found one remaining readiness resource-leak blocker. The redaction and liveness/readiness separation remain sound, CI is green, and the focused Server tests and typecheck pass.

bestony:
## Review round 3 — live readiness probes are now pinned, not just replaced

Addresses the P1 on the aged-out probe backlog. Current `main` (`54905b6b`) is merged in.

### Why round 2 was not enough

The age-out replaced a stuck attempt, which restored recovery, but nothing stopped the
`database.execute` behind the replaced promise. Under a sustained hang every window past
`READINESS_ATTEMPT_MAX_AGE_MS` started another live query while the previous ones stayed alive, each
holding a pool connection — so the probe deepened the exhaustion it exists to report. The reviewer's
repro (0 / 10001 / 20002 ms against a never-settling `execute`) showed the count going 1 → 2 → 3.

That framing came from my round-2 instruction, which called "one new probe per window" acceptable.
Over a real outage it is not.

### The fix

A counter pins **concurrently live** attempts at `READINESS_MAX_LIVE_ATTEMPTS = 2`:

- incremented where the attempt is created, decremented in the same `finally` that clears the slot;
- at the cap the probe throws `DatabaseReadinessAttemptLimitError` **without starting or awaiting
  anything** — `/readyz` already turns a probe rejection into 503, so the request resolves immediately
  instead of burning another second on a promise that will not settle;
- the round-2 age-out, the `inFlight === attempt` identity guard, the no-op `catch` on abandoned
  attempts, and the 1 s caller deadline are all unchanged.

**Deliberately not done, recorded rather than hidden:** `database.execute()` returns a drizzle promise,
not the postgres.js query handle, so there is no cancellation reachable at this layer.
`packages/server/src/db/client.ts` is **untouched** — putting a `statement_timeout` on that pool would
silently bound every query in the application, migrations included. The probe also still uses the
shared pool on purpose: a dedicated connection would report ready while the application's own pool is
exhausted. The consequence is honest and bounded: a permanently wedged connection keeps readiness at
not-ready until that connection actually errors out. A bounded backlog with delayed recovery beats an
unbounded one.

### Verified by the orchestrator, not self-reported

| Criterion | Result |
| --- | --- |
| `pnpm --filter @opentag/server test` | exit 0 — **63 files, 623 tests** |
| Live-probe count pinned across windows | **probed directly, five windows**: responses `503,503,503,503,503` with `execute` called exactly **2** times — flat at the cap, not 1→2→3 |
| A freed slot is reusable | covered by `frees a live-attempt slot when a replacement probe settles`: hang → 503, replacement succeeds → 200, a later window probes again → 200 |
| Round-2 recovery test | still passes unmodified |
| `packages/server/src/db/client.ts` | no diff against `main` |
| Round-1 redaction still holds | **probed directly**: a real `DrizzleQueryError` carrying `victim@example.com`, `hunter2` and the query text emits none of the three |
| `/healthz` still liveness-only | **probed directly**: 200 with **0** database calls |

### Note

The readiness tests still live under a `describe("GET /healthz")` block even though they exercise
`/readyz` — pre-existing from round 1, flagged again rather than folded in, since renaming the block
would churn the diff without changing behaviour.

---

*Dispatched as run `313360`, lane `request-boundary-logs-1`. Written by codex under bypassed approvals
in an isolated worktree; verified and published by the orchestrator.*

yuezengwu (approved):
Reviewed the full immutable range `54905b6bf51e80faadc8911ea4dbfd799c813829...d4ffdef97bc62a9bd16459697f087354b25c02d2` in a fresh detached worktree. The new live-attempt cap resolves the remaining readiness resource leak: aged-out probes may be replaced for recovery, but a persistent never-settling outage is pinned at two underlying database executions, and a settled replacement releases its slot without clearing a newer attempt. The liveness/readiness separation, Drizzle-safe serialization, classified request logging, and rate-limit cooldown boundary remain sound on this exact head. Local validation: focused app/request logging tests 20/20 pass, Server typecheck passes, and `git diff --check` passes. Exact-head CI is still running; all completed jobs currently pass or are skipped.

PR #463, opened 2026-09-02
feat(server): make the runtime bridge observable

bestony:
`packages/server/src/runtime/` is the entire server side of the daemon WebSocket bridge — 13 files,
and `grep -rn "logger"` across them returned nothing. Observation rested entirely on OTel spans, and
`OPENTAG_OTEL_ENDPOINT` defaults to empty, so every span path was a no-op. This makes connection
termination, transport errors and stale sweeps observable, and connects a metric port that was
declared but never wired.

## Checklist

- [x] 1. `#fail` logs every termination with code, close code and a redacted reason
- [x] 2. Optional logger threaded through `RuntimeSessionOptions` from `api/runtime.ts`
- [x] 3. `socket.on("error")` records, then still swallows
- [x] 4. `terminateStale` returns a count; caller logs one aggregate line per sweep
- [x] 5. `index.ts` passes `logger:` to the three runtime collaborators and wires `onMetric`
- [x] 6. `im-delivery-worker.ts` line count unchanged
- [x] 7. Regression tests added
- [x] 8. Progress recorded

## What changed

**`RuntimeSession.#fail` is the single funnel.** It is the sole exit for roughly thirty termination
reasons — 4401 credential revoked, 4409 instance fencing, about twenty `PROTOCOL_ERROR` cases, and the
handshake/registration timeouts — and every one of them closed a user's socket with no server-side
record. This is the product's most support-expensive path ("my daemon connects then drops") and it was
completely opaque. One log line inside `#fail` now covers all of them, carrying `code`, `closeCode`, a
`redactForLog`-scrubbed reason and the request id, wrapped in try/catch so logging can never prevent
the connection from closing.

**Logger threading.** `RuntimeSessionOptions` gained an **optional** logger field so existing
constructions and tests keep compiling, passed from `api/runtime.ts` where `sessionOptions` is
assembled.

**Transport errors.** `socket.on("error", () => undefined)` discarded every transport error for the
whole connection lifetime. It now records the error type and keeps swallowing — rethrowing makes `ws`
throw.

**Stale sweeps aggregate.** `terminateStale` returned `void` and terminated each entry silently. It
now returns a count and emits one aggregate line per sweep; logging per connection would burst
hundreds of lines in a single event-loop stall.

**The construction seam.** `index.ts` carried
`const serviceLogger = …; void serviceLogger;` — a seam left by the observability foundation for
exactly this lane. `ConnectionRegistry` and `RuntimeDomainOwner` now receive loggers, and
`ImDeliveryWorker.onMetric` is connected: the port was declared and emits six signals, but `index.ts`
never passed a handler, so every one was dropped by the default.

| Criterion | Result |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm build` | exit 0 |
| `pnpm --filter '@opentag/*' test` | exit 0 — shared 172, server 609, client 877, web 600 |
| `pnpm --filter @opentag/server test` | exit 0 |
| `im-delivery-worker.ts` still ≤ 1509 lines | exactly 1509 — unchanged |
| Test drives 4401, 4409 and `PROTOCOL_ERROR` through `#fail` | `logs 4401, 4409, and protocol termination with close codes and redacted reasons` |
| Test proves transport errors do not rethrow | `records transport errors without rethrowing or closing the runtime` |
| Test proves `terminateStale` emits one record | `connection-registry.test.ts` +32 lines |
| Test proves `onMetric` signals reach the logger | `im-delivery-worker.test.ts` +12 lines |

`im-delivery-worker.ts` is pinned at 1509 lines in `scripts/file-size-exceptions.json` and was already
at exactly that count, so it had zero headroom. It is unchanged.

## Notes

No deviation from the agreed plan. Seven commits for eight checklist items — one coherent unit each.

---

*Dispatched as run `313360`, lane `runtime-bridge-logs-2`. Written by codex under bypassed approvals
in an isolated worktree; verified and published by the orchestrator.*

### DIFF
--- a/packages/server/src/runtime/connection-registry.ts
+++ b/packages/server/src/runtime/connection-registry.ts
@@ -19,6 +19,7 @@ import {
   runtimeFrameByteLength,
 } from "@opentag/shared";
 import WebSocket from "ws";
+import type { ServiceLogger } from "../observability/service-logger.js";
 
 export class RuntimeRegistrySendError extends Error {
   constructor(
@@ -72,9 +73,18 @@ export interface RuntimeConnectionEntry {
   socket: WebSocket;
 }
 
+export interface ConnectionRegistryOptions {
+  logger?: ServiceLogger;
+}
+
 export class ConnectionRegistry {
   readonly #entries = new Map<string, RuntimeConnectionEntry>();
   readonly #registrationTails = new Map<string, Promise<void>>();
+  readonly #logger?: ServiceLogger;
+
+  constructor(options: ConnectionRegistryOptions = {}) {
+    this.#logger = options.logger;
+  }
 
   async register(entry: RuntimeConnectionEntry, persist: () => Promise<void>, publish?: () => void): Promise<void> {
     const previousRegistration = this.#registrationTails.get(entry.computerId) ?? Promise.resolve();
@@ -375,12 +385,26 @@ export class ConnectionRegistry {
     return this.#entries.delete(computerId);
   }
 
-  terminateStale(cutoff: number): void {
+  terminateStale(cutoff: number, onSweep?: (count: number) => void): number {
+    let count = 0;
     for (const entry of this.#entries.values()) {
       if (entry.lastHeartbeatAt < cutoff) {
         entry.socket.terminate();
+        count += 1;
       }
     }
+    if (count > 0) {
+      if (onSweep) {
+        onSweep(count);
+      } else {
+        try {
+          this.#logger?.warn({ count, cutoff }, "Stale runtime connection sweep terminated connections");
+        } catch {
+          // Logging must never prevent stale connection termination.
+        }
+      }
+    }
+    return count;
   }
 
   async closeComputer(computerId: string): Promise<boolean> {
--- a/packages/server/src/runtime/connection-registry.ts
+++ b/packages/server/src/runtime/connection-registry.ts
@@ -394,14 +394,14 @@ export class ConnectionRegistry {
       }
     }
     if (count > 0) {
-      if (onSweep) {
-        onSweep(count);
-      } else {
-        try {
+      try {
+        if (onSweep) {
+          onSweep(count);
+        } else {
           this.#logger?.warn({ count, cutoff }, "Stale runtime connection sweep terminated connections");
-        } catch {
-          // Logging must never prevent stale connection termination.
         }
+      } catch {
+        // Logging must never prevent stale connection termination.
       }
     }
     return count;
--- a/packages/server/src/__tests__/connection-registry.test.ts
+++ b/packages/server/src/__tests__/connection-registry.test.ts
@@ -9,6 +9,7 @@ import {
 } from "@opentag/shared";
 import { describe, expect, it, vi } from "vitest";
 import WebSocket from "ws";
+import type { ServiceLogger } from "../observability/service-logger.js";
 import { ConnectionRegistry } from "../runtime/connection-registry.js";
 import { projectComputerProviderReadiness } from "../services/computers/provider-readiness.js";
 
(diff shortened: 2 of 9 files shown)


packages/server/src/runtime/connection-registry.ts
```diff
@@ -375,12 +385,26 @@ export class ConnectionRegistry {
...
+    let count = 0;
     for (const entry of this.#entries.values()) {
       if (entry.lastHeartbeatAt < cutoff) {
         entry.socket.terminate();
+        count += 1;
+      }
+    }
+    if (count > 0) {
+      if (onSweep) {
+        onSweep(count);
```
yuezengwu: [P1] Contain the injected sweep logger failure

`onSweep` is the logger callback supplied by `api/runtime.ts`, but only the fallback `#logger` branch is protected by `try/catch`. A valid injected `RuntimeRoutesOptions.logger` that throws makes this call propagate out of the `setInterval` callback whenever a stale connection is found; in Node that is an uncaught timer exception and can terminate the server. This contradicts the new invariant that logging must not affect cleanup. Please catch callback failures too (with a throwing-logger regression test), or remove this callback and use the registry logger path that is already guarded.

yuezengwu (changes requested):
Exact-head review of `80cff118bc7bcd057ecf3cccc121b211389b234a` found one availability blocker: the new stale-sweep logging callback is not failure-contained. The default production adapter currently swallows logger exceptions, but the route API accepts an injected logger and the implementation explicitly promises logging cannot affect runtime cleanup. A throwing injected logger escapes the timer callback after a stale connection is found and can terminate the server process.

Validation on this exact head: `git diff --check`; `pnpm check`; `pnpm typecheck`; all six workspace build tasks plus bundle report; focused server tests (61/61); and a direct throwing-callback repro. GitHub checks are green. Local Node is 24.7.0, below the repository minimum 24.15.0, so install/build commands required a local engine-strict override; CI covered supported Node 22/24/26.

bestony:
## Review round 1 addressed (`09ce36d0`, `4d17a8b0`)

**[P1] Contain the injected sweep logger failure — fixed.** The `try/catch` in
`connection-registry.ts` now wraps **both** branches, so a throwing injected `onSweep` is swallowed
exactly like every other logging failure on this branch. Previously only the fallback `#logger` branch
was protected, so a valid injected logger that throws escaped the `setInterval` callback whenever a
stale connection was found — an uncaught timer exception that can terminate the server, and a direct
contradiction of the invariant this PR introduces.

Kept the callback rather than removing it: `api/runtime.ts` supplies it and the aggregate line belongs
to the caller's logger.

Regression added: `terminates stale sockets when the injected sweep logger throws` — it asserts the
sweep still **terminates** the stale connections, not merely that nothing propagates.

**Audit beyond the reported instance.** The finding is a class, not a one-off, so every logger call
site this branch adds under `packages/server/src/runtime/` was re-checked for a guard:

| Site | Guarded |
| --- | --- |
| `connection-registry.ts:399` (`onSweep`) | yes — this fix |
| `connection-registry.ts:401` (fallback `#logger`) | yes |
| `runtime-domain-owner.ts:762` | yes |
| `runtime-session.ts:160` (socket transport error) | yes |
| `runtime-session.ts:697` (`#fail`) | yes |

No second instance.

### Re-verified

| Gate | Result |
| --- | --- |
| `pnpm --filter @opentag/server test` | exit 0 — 610 tests |
| `im-delivery-worker.ts` ≤ 1509 lines | exactly 1509, unchanged |

The existing `#fail`, transport-error and `onMetric` tests all still pass.

---

*Run `313360`, lane `runtime-bridge-logs-2`. Written by codex under bypassed approvals; verified and

yuezengwu (approved):
Exact-head review of 4d17a8b05bf1772d2f90e39dd1b5d65017aa592d: approved. The sweep callback is now isolated so logger failures cannot escape cleanup; CI is green and targeted server tests passed locally (62 tests).

PR #467, opened 2026-09-02
feat(web): preserve and report client-side failure diagnostics

bestony:
The web app's problem was not transport — it was that **the cause was destroyed before anything could
transport it**. This preserves the diagnostic information the API layer threw away, stops
manufacturing a fake 503 for schema drift, and makes the failures the app was completely blind to
visible.

Console-only by design: no reporting endpoint. That is a later round, and it could only have forwarded
the few failures that survived today.

## Checklist

- [x] 1. `ApiError` carries `requestId` and `retryAfterSeconds`
- [x] 2. Schema mismatch no longer throws a fake 503; issue path and code logged, nothing else
- [x] 3. `fetchWithRefresh` wrapped; aborts mapped to `cancelled` without warning
- [x] 4. Reporter keyed on (route template, error code) with cooldown; not on MutationCache
- [x] 5. `observability/diagnostics.ts` added, layering both redactors
- [x] 6. CSRF-less write attempts are recorded
- [x] 7. `unhandledrejection` + capture-phase resource-error listeners added beside `rootErrorHandlers`
- [x] 8. `reportBoundaryError` shape unchanged; `normalizeError` emits a stable non-localized code
- [x] 9. Tests added
- [x] 10. Progress recorded

## What changed

**`requestId` survives — the highest-leverage line.** `apiError()` parsed the `ErrorEnvelope` and then
discarded `requestId` and `retryAfterSeconds`, both of which the envelope carries. `ApiError` now
exposes them. `requestId` is the only join key between a browser event and the server's `reqId`, and
the server returns an `x-request-id` response header.

**No more manufactured 503.** A response-schema mismatch threw
`ApiError(503, "The server returned an invalid response")` — naming neither endpoint nor field,
mislabelled as a 5xx, and therefore treated as non-terminal and retried forever on focus. It is now a
dedicated `ResponseSchemaError` carrying the route template and the failing issues, logged as **path
and code only** — never `issue.message`, never `received`, never the body.

**One seam, reads and writes both.** `fetchWithRefresh` is wrapped rather than each of `request`,
`requestOptional` and `requestNoContent`. Aborts branch first into `CancelledRequestError` with no
warning, so a user navigating away does not look like a fault.

**Not built on MutationCache.** There are zero `useMutation` calls in the repository, so a
MutationCache reporter would have stayed permanently empty — while every destructive write
(deleteAgent, disableImBinding, logout, sign-in) goes through `requestNoContent`. The cooldown key is
(route template, error code), never `queryHash`: `useAgentListView` fans out 2N queries through
`useQueries`, so ten agents would otherwise emit ~22 lines per failing round.

**CSRF blindness.** `csrfHeaders()` returned `{}` when the token lookup found nothing — signed in, but
the cookie cleared by ITP or partitioning — so the app issued writes with no CSRF header, the server
403'd every one, and the app silently appeared read-only with nothing recorded. That attempt is now
recorded at the request seam.

**Window-level handlers, added beside the existing ones.** `main.tsx` already passed
`rootErrorHandlers` into `createRoot`; those are React 19 root handlers and remain. An
`unhandledrejection` listener and a **capture-phase** error listener now sit alongside them, the latter
checking whether `event.target` is an `HTMLScriptElement`/`HTMLLinkElement` — resource load failures
neither bubble nor carry `event.error`, and a post-deploy chunk-load failure is the most common SPA
incident, which this app previously could not see at all.

**Aggregation stays possible.** `normalizeError` produced Error messages minted from paraglide `m.*()`
strings, so the same crash read differently per language. It now emits a stable, non-localized code.
`reportBoundaryError`'s two-argument `console.error('[OpenTag] Unhandled UI error', {...})` shape is
preserved verbatim, with the structured envelope attached as an additional key.

| Criterion | Result |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm build` | exit 0 |
| `pnpm --filter '@opentag/*' test` | exit 0 — shared 172, server 607, client 877, web 607 |
| `pnpm --filter @opentag/web build` within the 600 KiB entry-chunk budget | yes — the build enforces it and exits 0 (the 568 KiB chunk is the pre-existing `echarts` vendor chunk under its own budget) |
| Seven existing `error-boundary.test.tsx` assertions still pass unmodified | **zero removed assertion lines** in that file |
| Test proves `ApiError` exposes `requestId` | `preserves structured server diagnostics for recovery actions` |
| Test proves a schema mismatch is not a 503 | `records schema issue paths and codes without response detail` |
| Test proves an aborted request produces no warning | `maps an aborted request to cancelled without warning` |
| No diagnostic string goes through `m.*()`; no i18n assertion changed | `observability/` contains no `m.` usage |

Further tests added: `records a CSRF-less write at the request seam`, `uses route templates and cools
down duplicate error codes`, `layers flat-string and structured redaction`, `uses a stable code for
non-error values instead of localized copy`, `records unhandled rejections and capture-phase resource
failures`.

## Notes

No deviation from the agreed plan. Four commits across ten checklist items — coarser than
one-per-item, but each is a coherent unit (`preserve API request diagnostics`, `report API
diagnostics`, `capture global UI failures`, `keep boundary fallback catalog-safe`).

`throwOnError` was deliberately **not** set on the QueryClient — `client.ts` documents the intentional
design of degrading rather than white-screening on background refetch.

---

*Dispatched as run `313360`, lane `web-observability-4`. Written by codex under bypassed approvals in
an isolated worktree; verified and published by the orchestrator.*

### DIFF
--- a/apps/web/src/__tests__/api.test.ts
+++ b/apps/web/src/__tests__/api.test.ts
@@ -176,7 +176,7 @@ describe("BrowserApi", () => {
     setDocumentCookie("opentag_csrf=; Path=/; Max-Age=0");
   });
 
-  it("preserves structured server error codes for recovery actions", async () => {
+  it("preserves structured server diagnostics for recovery actions", async () => {
     const fetchImpl = vi.fn<typeof fetch>(
       async () =>
         new Response(
@@ -185,6 +185,8 @@ describe("BrowserApi", () => {
               code: "IM_BINDING_SCOPE_REAUTH_REQUIRED",
               category: "deterministic",
               message: "Additional scopes are required",
+              requestId: "request-123",
+              retryAfterSeconds: 30,
             },
           }),
           { status: 409, headers: { "content-type": "application/json" } },
@@ -198,6 +200,8 @@ describe("BrowserApi", () => {
       status: 409,
       code: "IM_BINDING_SCOPE_REAUTH_REQUIRED",
       category: "deterministic",
+      requestId: "request-123",
+      retryAfterSeconds: 30,
     });
   });
 
--- a/apps/web/src/__tests__/api.test.ts
+++ b/apps/web/src/__tests__/api.test.ts
@@ -1,6 +1,7 @@
 import type { TaskSummary } from "@opentag/shared/browser";
 import { describe, expect, it, vi } from "vitest";
 import { ApiError, BrowserApi } from "../api.js";
+import { DiagnosticReporter } from "../observability/diagnostics.js";
 
 const userId = "53e2babe-e4ac-4e2c-b7d1-d092d5a4568e";
 const taskSummary = {
@@ -453,7 +454,11 @@ describe("BrowserApi", () => {
         expectedRevision: 1,
         expectedRuntimeConfigRevision: 1,
       }),
-    ).rejects.toMatchObject({ status: 503, message: "The server returned an invalid response" });
+    ).rejects.toMatchObject({
+      name: "ResponseSchemaError",
+      code: "invalid_response_schema",
+      message: "The server returned an invalid response",
+    });
     expect(extraFieldFetch).toHaveBeenCalledOnce();
 
     const abortFetch = vi.fn<typeof fetch>(async (_input, init) => {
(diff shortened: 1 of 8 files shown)


yuezengwu (changes requested):
Reviewed exact head `2028cd4661363b40bcc93a0f15582744613d1420`.

[P1] Preserve arbitrary thrown Error objects — `apps/web/src/features/error-boundary.tsx:127`

`normalizeDiagnosticError` deliberately returns the original object when the thrown value is already an `Error`, but this line then writes `code` back onto that object. Valid throwable objects are not guaranteed to be writable: `Object.freeze(new Error("boom"))` rejects the assignment, and `DOMException` exposes a getter-only numeric `code`. I reproduced both as `TypeError`s. Because this helper runs from `getDerivedStateFromError`, React root handlers, and the router's `defaultOnCatch`, that secondary exception can take down the fallback/logging path and mask the failure this change is meant to preserve. Please carry the stable code beside the original error (or create a safe wrapper/copy) without mutating the throwable, and add regressions for a non-extensible `Error` and a `DOMException`.

Validation on this head: `git diff --check`; 36 focused API/diagnostics/boundary tests; all 607 Web tests; Web typecheck; Web production build. All passed after building the fresh worktree's generated `@opentag/shared/browser` artifact. Local Node was 24.7.0, below the repository's supported 24.15.0 floor, so local commands required only an engine-check override; exact-head CI, including Browser Smoke and Node 24 Compatibility, is green.

bestony:
## Review round 1 addressed (`332cec3d`)

**[P1] Preserve arbitrary thrown Error objects — fixed.** The stable code now travels **beside** the
error instead of being written onto it:

```ts
type BoundaryError = { readonly error: Error; readonly code: string };
```

`normalizeError` no longer does `Object.assign(normalized.error, { code })`; it returns the
`{ error, code }` pair unchanged, and the call sites read `normalized.error.name` / `.message`. A
frozen `Error` and a `DOMException` therefore normalize without a secondary `TypeError` taking down
`getDerivedStateFromError`, the React root handlers, or the router's `defaultOnCatch`.

Deliberately **not** solved with a `try/catch` around the assignment: mutating someone else's
throwable is the behaviour to remove, and swallowing the failure would drop the code exactly for
`DOMException` — the chunk-load and resource-failure case this PR exists to make visible.

Regressions added, both as requested: `normalizes a frozen Error without mutating or replacing it` and
`normalizes a DOMException without assigning to its read-only code`.

**The whole chain is mutation-free**, not just the call site that was reported.
`observability/diagnostics.ts:83` `normalizeError` was already pure — it returns `{ error, code }` and
never assigns — and `readStableCode` only *reads* `.code`, requiring a string matching `SAFE_CODE`, so
a `DOMException`'s numeric `code` is ignored and the stable code falls back to the normalized error
name. `isErrorValue` also recognises `[object DOMException]`, so such a value is preserved rather than
wrapped.

### Re-verified

| Gate | Result |
| --- | --- |
| `pnpm --filter @opentag/web test` | exit 0 — 609 tests |
| Entry-chunk budget | build enforces it and exits 0 |
| Seven existing `error-boundary.test.tsx` assertions | **zero removed**, all still pass |

`ApiError` still exposes `requestId`, a schema mismatch still is not a 503, and an aborted request
still produces no warning.

---

*Run `313360`, lane `web-observability-4`. Written by codex under bypassed approvals; verified and

yuezengwu (approved):
Exact-head review of 332cec3da8377f53cc30810601fbdd272dd6ceea: approved. Boundary normalization now preserves immutable Error and DOMException values without assignment; CI is green and targeted Web tests passed locally (38 tests).

PR #468, opened 2026-09-02
feat(cli): give the CLI a real log destination and a truthful exit contract

bestony:
The CLI had an **enablement gap before it had a logging gap**: it never established a log destination
outside daemon service mode, so any log line added would only have reprinted to a terminal the user was
already watching. This fixes the destination first, then instruments the single command failure funnel,
repairs the exit-code contract, and stops a message regex from blaming users for failures they did not
cause.

## Checklist

- [x] 1. File sink configured in `cli/index.ts` before `parseAsync`, levels split
- [x] 2. `executeCommand` emits structured failures carrying `requestId`; human render shows it too
- [x] 3. Exit-code contract honoured through `commandExitCode()`
- [x] 4. `isInterrupted` narrowed to structural signals and moved below the API branch
- [x] 5. Daemon ownership failure reaches `client.log`
- [x] 6. `opentag logs` implemented, output redacted
- [x] 7. Tests added
- [x] 8. Progress recorded

## What changed

**The enablement gap, fixed first.** `configureClientLoggerForService` had exactly one call site in the
whole repository — `core/daemon/runtime.ts`, gated by `OPENTAG_SERVICE_MODE === "1"`, which only the
generated plist/systemd unit sets. In all other command files `createLogger` therefore resolved to
stderr@warn. `cli/index.ts` now resolves the home layout and configures the file sink **before**
`parseAsync`, with levels split so the file gets `info` and stderr stays reserved for
`presentCommand`.

**Structured failures with a join key.** `executeCommand` is the single failure funnel for 14 command
registrations across 10 files. `CommandError` already carried a schema-valid, redacted
`StructuredError` that `presentCommand` rendered and threw away. Failures now reach the file sink
carrying `requestId` — the only join key between a CLI failure and the server's pino `reqId`, now that
the server returns an `x-request-id` response header — and the human-readable render appends it too.

**Exit codes stop collapsing to 1.** `cli/index.ts` hardcoded `exitCode: 1` in the top-level failure
result, so `commandExitCode()` never ran and validation(2) and unavailable(3) both became 1 — while the
portable installer and the auto-updater branch on exactly those. It now calls `commandExitCode()`, and
the paths that bypassed `executeCommand` were fixed alongside it.

**No more blaming the user.** `isInterrupted` matched on the error *message* with
`/(?:^|\b)(?:aborted|abort|cancelled|canceled|interrupted)(?:\b|$)/iu`, and `toCommandError` called it
**before** the `OpenTagApiError` branch — so a server message such as "The Feishu setup attempt was
cancelled" became INTERRUPTED / exit 130, confidently blaming a user who never pressed Ctrl-C. It is
now `AbortError` / `CanceledError` / an abort signal only, and it runs **below** the API branch.

**Daemon startup failures reach the log.** The ownership-acquisition failure reported on stderr because
it ran before the sink was configured; the order is fixed.

**`opentag logs`.** Bundles `client.log`, its rotated files and an environment summary, passing
everything through `redactForLog` before output.

| Criterion | Result |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm build` | exit 0 |
| `pnpm --filter '@opentag/*' test` | exit 0 — shared 172, server 607, client 877, web 600 |
| CLI package tests | exit 0 — **31 files, 313 tests** (see note below) |
| Test proves a failing command writes a structured record carrying `requestId` to the file sink | `writes structured failures to the injected file logger and renders the request id` |
| Test proves an `OpenTagApiError` whose message says "cancelled" is not INTERRUPTED | `does not blame the user for a cancelled server operation` |
| Validation exits 2, unavailable exits 3 | **partially covered** — see note |
| `opentag logs` output contains no plaintext credential | `registers a JSON logs command that writes one redacted envelope`, `bundles current and rotated client logs with a safe environment summary`, `treats a missing log directory as an empty bundle` |

**Note on the CLI test command.** The brief's criterion named `pnpm --filter @opentag/cli test`. The
package is actually named **`open-tag`**, so that filter matches nothing and exits 0 silently — and for
the same reason `pnpm --filter '@opentag/*' test` does not include the CLI package either. The suite was
re-run as `pnpm --filter open-tag test`: 31 files, 313 tests, all passing. This was an error in the
brief, not in the lane's work.

**Note on the exit-code criterion.** `commandExitCode()`'s mapping is covered by
`cli-result-policy-coverage.test.ts`, and one bypass path asserts exit 3 directly
(`computer-connect.test.ts`). There is no end-to-end test driving `cli/index.ts` itself, which is a
~20-line entry script the repository does not unit-test at all; the change there is a one-line
substitution of a hardcoded `1` for the tested function. Reported as partial rather than claimed as
complete.

## Notes

The brief predicted this work would "deliberately change assertions" in the command-policy test. It did
not need to: the existing INTERRUPTED assertion already keyed on `name: "AbortError"` — a structural
signal the narrowed check still honours — rather than on the message regex. The prediction was wrong;
nothing was weakened to accommodate it.

Six commits across eight checklist items.

---

*Dispatched as run `313360`, lane `cli-observability-3`. Written by codex under bypassed approvals in
an isolated worktree; verified and published by the orchestrator.*


bestony:
## Correction: CI failure fixed (`c8ab7b1f`)

The first push of this branch failed CI on `CLI Pack Smoke`, `Container Smoke` and all three Node
compatibility jobs. Root cause and fix, both reproduced locally:

Moving the log sink configuration ahead of `parseAsync` — which is the point of this PR — meant it ran
for **every** command, including `opentag doctor`. `configureClientLoggerForService` constructed the
`RotatingFileStream` eagerly, and that constructor `mkdir`s the log directory. But
`scripts/cli-pack-smoke.mjs:202` requires `opentag doctor` to leave OpenTag Home **completely empty**,
so simply configuring a logger broke the contract.

**Fix: the file sink is now lazy.** `configureClientLoggerForService` records the canonical directory
and clears the cached stream; the rotating stream — and therefore the `mkdir` — is created on the
first record actually written. `doctor` logs nothing, so it touches nothing. No per-command allowlist:
"configuring a logger has no filesystem side effect" is the property worth holding, and an allowlist
would rot the moment another read-only command is added. Redaction, rotation, permissions and the
fallback path are unchanged.

A test pins it: `does not create the service log directory until the first write`.

### Re-verified after the fix

| Gate | Result |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm build` | exit 0 |
| `pnpm --filter open-tag test` | exit 0 — 31 files, 313 tests |
| `node scripts/cli-pack-smoke.mjs --channel source --name open-tag --version 0.0.2 --binary opentag-dev` | exit 0 — `Verified open-tag@0.0.2 through opentag-dev` |
| `node scripts/node-compat-smoke.mjs` | exit 0 |

Everything in the original description still holds: the sink is configured before `parseAsync`,
`commandExitCode()` replaces the hardcoded `1`, `isInterrupted` is narrowed to structural signals and
runs below the `OpenTagApiError` branch, and `opentag logs` is present.

### Why the first push shipped broken

Two errors in the acceptance criteria I gave the lane, not in its work:

- the criteria named `pnpm --filter @opentag/cli test`, but the package is `open-tag`, so that filter
  matched nothing and exited 0 silently;
- `pnpm --filter '@opentag/*' test` excludes the CLI package for the same reason, so its 313 tests were
  never in the gate, and neither smoke script was in any local gate.

The gates now include the root suite, the correctly-named CLI filter, and both smoke scripts.

`Container Smoke` was traced to the same CLI temporary-home initialization rather than an independent
server or container defect.

bestony:
## Correction 2: `Container Smoke` fixed (`00ce4d30`)

The lazy-sink fix cleared `CLI Pack Smoke` and all three Node compatibility jobs, but `Container
Smoke` still failed. Traced to a second, independent defect — and a real product regression, not just
a CI contract.

**What broke.** `.github/workflows/ci.yml:462` runs `daemon service-run` a second time and asserts
`grep -q "already running"` in a file capturing stdout+stderr. The chain: `ownership.ts:62` throws
"An OpenTag daemon is already running for this home (pid N)"; `acquireOwnership`'s catch calls
`logTerminalFailure`, which emits that text via `logger.warn`; that logger had been switched from
`destination: "stderr"` to the default so the failure would reach `client.log`; and because the sink
is now configured for *every* command, the default resolved to the file. The operator-facing message
stopped appearing on the terminal entirely — a user running `opentag daemon service-run` twice would
have seen nothing at all.

**Fix.** The three terminal-failure paths in `core/daemon/runtime.ts` now use the logger's existing
`dual` destination, so a terminal daemon failure reaches **both** stderr and `client.log`. Diagnostics
and the human-readable reason a command refused to start are different audiences, not a choice.

Two tests pin it: `uses a dual logger for an already-held daemon owner` and `writes a terminal daemon
failure to stderr and the file sink`.

### Re-verified

| Gate | Result |
| --- | --- |
| `pnpm --filter open-tag test` | exit 0 — 314 tests |
| `cli-pack-smoke.mjs` | exit 0 — `Verified open-tag@0.0.2 through opentag-dev` |
| `node-compat-smoke.mjs` | exit 0 |

A direct probe confirms all three properties hold together: configuring the logger creates no
directory, and a terminal failure reaches `client.log` **and** stderr.

Both defects on this branch came from the interaction of two instructions in the lane's brief —
configure the sink before `parseAsync`, and route the ownership failure to `client.log`. Neither was
visible to any local gate as originally specified; the gates now include the root suite, the
correctly-named CLI filter, and both smoke scripts.

apps/cli/src/core/diagnostics/logs.ts
```diff
@@ -0,0 +1,119 @@
...
+const CLIENT_LOG_NAME = "client.log";
+const CLIENT_LOG_PATTERN = /^client\.log(?:\.\d+)?$/u;
+
+/** Read the local client log set and safe runtime metadata for support diagnostics. */
+export async function runLogs(options: LogsOptions = {}): Promise<LogsResult> {
+  const environment = options.environment ?? process.env;
+  const home = options.home ?? resolveOpenTagHome(environment);
+  const layout = resolveOpenTagHomeLayout(home);
+  const files = await readClientLogs(layout.logs);
+  return redactForLog({
```
yuezengwu: [P1] Preserve recent log records in the bundle\n\nredactForLog also caps every string at STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES (4 KiB). Applying it to the complete LogsResult truncates each file.content to the first 4 KiB; formatLogs repeats the same operation. Since client.log is append-only up to 10 MiB, the latest failure—the reason a user runs opentag logs—is absent once the file grows past 4 KiB. I reproduced this with a 5,000-byte prefix plus a tail marker: result length 4096 and the marker was missing. Please redact records without applying the log-field cap to the aggregate file, or deliberately export a bounded tail that retains the newest records, and add a greater-than-4-KiB regression test.

yuezengwu (changes requested):
Exact-head review of 00ce4d3021e887440a7476b5b817c54115f6889b found one blocking diagnostics-data-loss issue. CI is green; targeted CLI and Client logger tests also pass locally.

bestony:
## Review round 2 — `opentag logs` no longer drops the newest records

Addresses the P1 on `apps/cli/src/core/diagnostics/logs.ts:43`. Also merged current `main`
(`5d2a0b21`) into the branch, so this diff is against the base as it stands after #443, #463 and #467.

### The bug, restated

`runLogs` wrapped the whole `LogsResult` in `redactForLog`, whose `capLogStringValues` truncates
**every** string at `STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES` (4 KiB) and recurses into arrays — so each
`files[].content` was cut to its first 4 KiB. `formatLogs` applied the same call again. Because
`client.log` is append-only, "first 4 KiB" is the **oldest** content: the failure a user runs
`opentag logs` to report was exactly the part being discarded.

### The fix

- `runLogs` redacts `environment` on its own and leaves `files[].content` as the existing per-line
  `redactLogContent` produced it. `formatLogs` no longer re-redacts the whole result.
- Per-line redaction is unchanged — a 4 KiB bound per log *record* is sane, and it is what keeps
  credentials out of the bundle.
- Export is bounded from the **newest** end instead: `MAX_EXPORTED_LOG_FILE_BYTES = 1 MiB` per file,
  cut on a line boundary, with an explicit `[... N earlier bytes omitted ...]` marker so a truncated
  file is visible as truncated. Without this, removing the cap would have let the command emit up to
  10 MiB × 7 rotated files — a problem the accidental 4 KiB cap had been masking.
- `STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES` is untouched and `packages/shared` has no diff: this is a
  log-bundle bug and must not widen a repo-wide per-field bound.

### Verified by the orchestrator, not self-reported

| Criterion | Result |
| --- | --- |
| `pnpm test` (root) | 310/311 — the one failure is `scripts/__tests__/lefthook-stage-fixed.test.mjs`, which fails identically on a clean `main` checkout on this host (`error: 1Password: failed to fill whole buffer` while signing a fixture commit). Environmental, not from this diff. |
| `pnpm --filter open-tag test` | exit 0 — 31 files, **321 tests** (was 313; 4 new cases here) |
| `node scripts/cli-pack-smoke.mjs …` | exit 0 — `Verified open-tag@0.0.2 through opentag-dev` |
| `node scripts/node-compat-smoke.mjs` | exit 0 |
| >4 KiB log keeps its newest records | **probed directly**: 16,369-byte input → 16,372-byte output (was 4,096), tail canary present through both `runLogs` and `formatLogs` |
| Credential still redacted above 4 KiB | probed: `"password":"hunter2"` in the **last** line is redacted |
| Three original `logs.test.ts` cases | unmodified and passing |
| `git status --porcelain` | empty |

### One thing the reviewer should decide on, out of scope here

The same probe shows `victim@example.com` surviving in plaintext. That is **not** a regression from
this change: `SENSITIVE_KEY_PARTS` in `packages/shared` has never included `email` (nor name, phone or
IP), so emails have always passed through `redactSensitive` everywhere in the repo. Previously the
4 KiB cap hid most of the file by accident; exporting the real tail makes the pre-existing gap
visible. Widening `redactSensitive` is a `packages/shared` change and was explicitly out of this
lane's scope — flagging it rather than smuggling it in.

---

*Dispatched as run `313360`, lane `cli-observability-3`. Written by codex under bypassed approvals in
an isolated worktree; verified and published by the orchestrator.*

apps/cli/src/core/daemon/runtime.ts
```diff
@@ -132,6 +132,9 @@ export async function runDaemonService(options: DaemonRuntimeOptions = {}): Prom
     instanceId,
     platform: currentPlatform,
   };
+  const environmentResult = await applyDaemonEnvironment(home, environment);
```
yuezengwu: [P1] Keep daemon.env failures on a logged startup path\n\nMoving applyDaemonEnvironment above the lifecycle try means a private-file validation failure is thrown before any daemon logger is created or used. runDaemonServiceEntry then treats DaemonServiceError CONFIGURATION as an expected stop and returns 0, so the supervisor sees a clean exit while the operator gets neither stderr nor client.log. I reproduced this head with a private home/config and daemon.env mode 0644: runDaemonServiceEntry returned 0 and an injected logger received zero records. Keep environment-load failures inside a path that emits the terminal/file diagnostic before the clean service exit, and add an entrypoint regression for an invalid daemon.env.

yuezengwu (changes requested):
Exact-head review of 4a502e765cd788cd5bcdf05e4fca7ddf71b6b9f5 found one daemon-startup observability blocker. The log-tail repair is sound, CI is green, and focused CLI/Client logger tests plus CLI typecheck pass.

bestony:
## Review round 3 — a bad `daemon.env` is no longer a silent exit 0

Addresses the P1 on the daemon startup path. Current `main` (`54905b6b`) is merged in.

### The defect

`applyDaemonEnvironment` ran above the lifecycle `try` and above the
`configureClientLoggerForService` call, so a private-file validation failure threw before any logger
existed and before the file sink was configured. `runDaemonServiceEntry` then counted
`DaemonServiceError CONFIGURATION` as an expected stop and **returned 0** — the supervisor saw a clean
exit, the operator got nothing on stderr, and `client.log` got nothing.

One correction to the review's wording, since it changes who owns this: this PR did **not** move
`applyDaemonEnvironment`. `git log -S` puts that placement in the original daemon service commit
(#17). What is new is this PR's own promise — correction pass 2 established that a terminal daemon
failure must reach the terminal **and** `client.log`, which `acquireOwnership` already honoured via
`destination: "dual"` and this path did not. Pre-existing defect, but squarely inside what this PR
now claims, so leaving it would have made the claim false.

### The fix

`applyDaemonEnvironment`, `configureClientLoggerForService` and `acquireOwnership` now all run inside
the lifecycle `try`, whose catch already emits `logTerminalFailure` through a `dual` logger.
`ownership` became optional so the release path stays correct when startup fails before it exists.
The exit code is deliberately still **0** — a bad `daemon.env` is a legitimate expected stop and the
supervisor should not restart-loop on it. What changed is that it is no longer silent.

**A second commit fixes a defect the first one introduced.** Moving `acquireOwnership` inside the try
meant its own `logTerminalFailure` and the outer catch both fired: an already-held owner logged the
"already running" line **twice**, where it logged once before. I caught that in verification before
publishing. `acquireOwnership` has exactly one call site, so its redundant try/catch is gone and the
lifecycle catch is now the single place that reports terminal startup failures.

### Verified by the orchestrator, not self-reported

| Criterion | Result |
| --- | --- |
| `pnpm --filter open-tag test` | exit 0 — **31 files, 323 tests** |
| `node scripts/cli-pack-smoke.mjs …` | exit 0 |
| `node scripts/node-compat-smoke.mjs` | exit 0 |
| Invalid `daemon.env` at the entrypoint | **probed directly**: `runDaemonServiceEntry` returns **0** and the injected logger receives **exactly 1** record — `Daemon service configuration prevented startup; inspect daemon status` |
| Already-held owner is not double-logged | **probed directly**: exit 0 with **exactly 1** `already running` record (it was 2 before the second commit) |
| Round-2 log-tail fix untouched | `apps/cli/src/core/diagnostics/logs.ts` has no diff since `4a502e76` |
| `packages/shared` | no diff against `main`; `STRUCTURED_ERROR_LOG_FIELD_MAX_BYTES` unchanged |

### Still open, unchanged from round 2 and out of this lane's scope

`redactSensitive` has never covered `email` (nor name, phone or IP), so those still pass through the
log bundle in plaintext. Not a regression from this PR — the old 4 KiB cap merely hid most of the file
by accident. Fixing it means changing `packages/shared`, which affects every log path in the server,
client and web, so it is flagged for a separate change rather than smuggled in here.

---

*Dispatched as run `313360`, lane `cli-observability-3`. Written by codex under bypassed approvals in
an isolated worktree; verified and published by the orchestrator.*

yuezengwu (approved):
Reviewed the full immutable range `54905b6bf51e80faadc8911ea4dbfd799c813829...047c3756f9fe68066e8e703428387971b0a8dbc8` in a fresh detached worktree. The daemon environment failure is now caught inside the service lifecycle and written through the already-configured dual logger before `runDaemonServiceEntry` returns its expected clean stop; owner-startup failures are emitted exactly once. The bounded newest-record log export, lazy file stream, command failure/exit policy, and redaction boundary remain sound on this head. Local validation: focused CLI diagnostics/command/daemon tests 46/46 pass, Client logger tests 16/16 pass, CLI typecheck passes, and `git diff --check` passes. Exact-head CI is still running; all completed jobs currently pass or are skipped.

Issue #437, opened 2026-09-01
design: make onboarding exact-Agent setup with one canonical state


Gandy2025:
## Context

A real onboarding failure exposed more than the immediate Feishu CLI compatibility bug in #433.

During the failed handoff, selecting Slack could render the resumed Feishu handoff state under the Slack presentation. Starting Slack authorization then failed because the selected Agent already had a Feishu binding. The Account could not enter the normal Agent settings surface because setup completion required a ready IM handoff, while the setup gate redirected every recovery route back to onboarding.

The immediate CLI defect remains scoped to #433. This issue owns the architecture and state-model problems that made that defect ambiguous and difficult to recover from.

Related existing issues:

- #296 — onboarding guesses an Agent instead of resuming an exact target
- #367 — an abandoned creation intent can resume on a later visit
- #357 — a Feishu brand-switch attempt outlives the state that describes it
- #433 — Feishu CLI validation is incompatible with externally injected credentials

## Current implementation

Onboarding reconstructs state from several independent sources:

- Account `setupCompletedAt`
- all active Agents
- all Computers and their current readiness
- IM binding summary
- IM handoff projection
- page-local draft, confirmation, and messaging-provider state
- Slack OAuth query parameters
- several independent polling loops

The route validates an optional `agentId`, but does not pass it into the onboarding page. The Server-backed implementation therefore resumes an Agent by preferring one whose Computer is online and otherwise taking the first bound Agent in Server order.

Messaging identity is also split: presentation provider, authorization/handoff state, binding identity, and provider recovery come from different facts. This makes invalid combinations representable, including a Slack presentation backed by a Feishu handoff.

Slack OAuth redirects to Agent messaging settings rather than restoring the initiating onboarding target. An incomplete Account is then redirected back to `/onboarding`, losing the target and OAuth result context.

The existing `OnboardingBackend` is a useful Seam, but its Interface is shallow. It exposes independent resources and commands, leaving the page to order them, infer invariants and provider intent, decide reset semantics, complete Account setup, and reject stale asynchronous work.

### D1 — Account admission moves before resource readiness

Once onboarding has explicitly adopted a legal target Agent, normal application surfaces are available. Account setup completion does not require Computer, runtime, Messaging, provider CLI, or IM handoff readiness.

Computer, runtime, Messaging, and diagnostics remain Agent setup and recovery concerns after the Account gate has opened.

### D2 — no `AccountSetupRun`

[...]

PR #473, opened 2026-09-02
feat(web): unify Agent creation and setup

### LINKED ISSUE #451: feat(web): unify Agent creation and setup on /agents/setup
## Context

Issue #437 establishes the exact-Agent setup architecture: a Server-owned `AgentSetupSnapshot`, exact target validation, Account admission before resource readiness, structured blockers and actions, canonical Messaging state, OAuth/attempt fencing, and explicit creation-intent recovery.

Build on #469, which implements #437, to replace the separate onboarding and New Agent experiences with one Agent Setup page. This issue is intentionally assigned to @Gandy2025 for implementation.

> **Dependency:** Development may proceed as a stacked change on #469. This issue must not merge until #469 is integrated into `main`.

## Confirmed product decisions

- Use `/agents/setup` as the only production Agent creation/setup route.
- Remove `/agents/new` and `/onboarding`; no compatibility redirects are required.
- Keep the Local / Cloud destination screen.
- Local remains available. Cloud remains visible as `Coming soon` and does not gain backend behavior in this issue.
- Reuse the current onboarding presentation, layout, step structure, accessibility behavior, and Review Lab approach.
- Do not reuse the current Web-owned onboarding state machine. After an Agent exists, render and act only through the #437 `AgentSetup` Module.
- Do not introduce an Account setup run, setup session, workflow framework, or new persistence model.

## Route contract

The one production route supports three explicit states:

```text
/agents/setup
/agents/setup?action=create
/agents/setup?agentId=<exact-agent-id>
```

- No search parameters: resolve the Account setup target using #437's zero / one / many legal-Agent rules.
- `action=create`: explicitly create a new Agent and never adopt an existing Agent implicitly.
- `agentId`: validate and resume that exact active, Account-owned Agent.
- `action` and `agentId` are mutually exclusive. Invalid combinations fail closed.

Entry points:

- Account setup gate -> `/agents/setup`
- Agents page **New Agent** action -> `/agents/setup?action=create`
- Continue setup for a known Agent -> `/agents/setup?agentId=<id>`
- Agent setup Slack OAuth return -> `/agents/setup?agentId=<id>`

## Page behavior

`/agents/setup` is one page with three conditional presentations, not three routes or independent flows:

1. **Create Agent** when no exact Agent exists and creation was explicitly selected or target resolution finds zero legal Agents.
2. **Choose Agent** only when a target-less Account setup entry has more than one legal Agent.
3. **Set up exact Agent** when `agentId` is present.

The multi-Agent chooser is a rare ambiguity-recovery state. It must not guess from list order, Computer binding, availability, readiness, or Messaging state. Selecting an Agent canonicalizes the URL to `/agents/setup?agentId=<id>`.

## Agent creation

The creation presentation is deliberately small and is not a second workflow Module:

```text
Local / Cloud -> Agent name and runtime -> Create Agent
```

- Create the Local Agent without requiring a Computer; the existing Agent contract permits a missing `computerId`.
- Use `creationIntentId` for idempotency.
- Follow #437's explicit uncertain-result recovery: **Check status**, **Try again**, and **Discard and start over**.
- Never auto-resume an abandoned creation intent on a later visit.
- On success, replace the URL with `/agents/setup?agentId=<created-id>` and hand all remaining behavior to #437.

## Exact-Agent setup

After `agentId` exists, the Web page consumes the #437 `AgentSetup` Interface and canonical Snapshot. It must not independently assemble Agent, Computer, runtime, binding, handoff, Provider, polling, or OAuth query facts.

Render the Snapshot stages, blockers, and permitted actions for:

- Computer setup and recovery
- Runtime readiness
- Messaging bind, same-Provider reauthorization/replacement, and explicit unbind
- `unbind-required` before changing Provider
- Ready state

Account admission follows #437: adopting a legal target opens normal application surfaces without waiting for Computer, runtime, Messaging, provider CLI, or handoff readiness. Agent Setup is user-visibly complete only when the canonical Snapshot reports `ready`.

## Confirmed UX copy

Progress labels:

```text
Create agent
Prepare computer
Connect messaging
```

Chinese:

```text
创建 Agent
准备电脑
连接消息应用
```

Keep the full Messaging page title:

```text
Connect your messaging app
连接你的消息应用
```

Copy changes:

- Retain the existing `Create your agent`, destination, Local / Cloud, Agent name/runtime, Computer connection, Messaging connection, and `{name} is ready` copy where still accurate.
- Replace target-less resume copy with explicit zero / one / many target resolution.
- Replace generic blocker counts with #437 blocker-specific messages.
- Use **Continue setup** for an incomplete Agent entry.
- Use **Back to agents** as normal navigation after Account admission; do not present a special **Set up later** action.
- Use **Open agent** from the ready state.
- Remove `Finish re-board`, final Account-completion copy, and the intermediate `Agent created` completion surface.
- After Slack returns, say authorization completed and confirm the connection from a freshly loaded Snapshot; do not claim success from query parameters alone.
- Unbind confirmation must explain that new messages stop and active conversations end while message history remains.

Keep shared Agents-list and IM copy in their existing message modules. Rename onboarding-specific copy to Agent Setup only where the owning feature is renamed; do not mechanically move all related copy into one catalog.

## Routing and cleanup

- Add the authenticated, AppShell-independent `/agents/setup` route.
- Point the Account setup gate and every New Agent entry to the new route.
- Use `agent-setup` as the bounded Slack OAuth return surface introduced by #437.
- Rename the internal Review Lab route to `/internal/agent-setup` and render the same presentation through #437's in-memory Adapter.
- Delete the `/agents/new` and `/onboarding` route files.
- Delete `NewAgentPage`, `NewAgentDialog`, `NewAgentMessagingStep`, the obsolete New Agent flow implementation, and duplicate implementation-specific tests after parity is verified.
- Regenerate the committed TanStack route tree; do not edit it manually.
- Keep English and Chinese messages synchronized and do not edit generated Paraglide output.

## Acceptance criteria

- [ ] #469 is fully merged to `main` before this change merges.
- [ ] `/agents/setup` is the only production Agent creation/setup route.
- [ ] `/agents/new` and `/onboarding` no longer exist and have no compatibility redirects.
- [ ] A first Account with zero legal Agents reaches Agent creation.
- [ ] **New Agent** always starts explicit creation even when the Account already owns Agents.
- [ ] Target-less zero / one / many resolution follows #437 and never guesses from readiness or list order.
- [ ] A foreign, inactive, missing, or conflicting exact target fails closed.
- [ ] Local / Cloud remains visible and Cloud remains disabled as `Coming soon`.
- [ ] Local Agent creation does not require a Computer.
- [ ] Uncertain creation results require explicit Check, Retry, or Discard actions and never auto-resume later.
- [ ] Creation success canonicalizes the URL to the exact `agentId`.
- [ ] Computer, runtime, Messaging, blockers, actions, OAuth recovery, and readiness are rendered from #437's canonical Snapshot.
- [ ] Account admission and Agent readiness retain their distinct #437 semantics.
- [ ] Slack OAuth returns to the exact Agent setup surface and stale callbacks fail closed.
- [ ] The progress labels are **Create agent**, **Prepare computer**, and **Connect messaging**.
- [ ] The ready state offers **Open agent** and does not perform final Account completion.
- [ ] Production HTTP and Review Lab in-memory Adapters exercise equivalent behavior.
- [ ] Existing relevant unit, integration, route, OAuth, refresh, multi-tab, recovery, and accessibility tests pass.

## Non-goals

- Implementing Cloud Agent provisioning
- Changing #437's canonical state contract or Provider lifecycle
- Adding setup-session persistence
- Preserving either removed route
- Redesigning unrelated Agent Settings, Agents list, Computer, or IM surfaces

Gandy2025:
## Summary

- make `/agents/setup` the sole Agent creation and setup route, with explicit create, exact-Agent, and zero/one/many target resolution
- keep creation thin and idempotent, then hand Computer, runtime, Messaging, blockers, and readiness to the Issue 437 Agent Setup Snapshot
- remove the old `/agents/new` and `/onboarding` routes, duplicate Web-owned state machines, and obsolete tests
- update entry points, OAuth return routing, Review Lab, bilingual copy, E2E coverage, and documentation

## Dependency

This is intentionally stacked on #469. It must not merge until #469 is integrated into `main`.

## Independent review

A separate pre-merge diff review rechecked route fail-closed behavior, exact Account-owned target selection, creation-intent recovery, Account admission versus Agent readiness, and Snapshot ownership after Agent creation.

The review found and fixed:

- stale onboarding entries in the bilingual i18n ownership documentation
- obsolete complexity-ratchet exceptions left by deleted duplicate flows
- a Playwright poll that consumed its whole timeout while looking for a nonexistent alert
- legacy E2E expectations that conflated Account admission with Agent readiness and automatic navigation

No additional workflow abstraction or persistence model was introduced.

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `OPENTAG_HOME=/tmp/opentag-codex-test-7629 pnpm test`
- real Chromium + real Server + real PostgreSQL + real CLI daemon Agent Setup E2E: 14/14 steps passed

The offline E2E reaches the canonical pending Messaging handoff. It deliberately does not manufacture the final provider credential-execution observation required for `ready`; that Snapshot transition remains covered by the Issue 437 unit and integration tests.

Closes #451


liuchao-001 (changes requested):
Reviewed at head `70791039`, diffed against the merge base with the base branch (`a5958e18` = #469 head).

The shape of this change is right: one canonical route, explicit create / exact / zero-one-many resolution, creation kept thin and handed to the Issue 437 Snapshot. I traced the fail-closed paths and they hold. But CI is red on this head, and it is red *because of this change*, so this is not mergeable as-is.

## Blocking

### 1. Four Browser Smoke tests still drive the deleted `NewAgentDialog`

`Browser Smoke` fails 4/8 on `70791039`, which fails the required-jobs `CI` gate:

- `e2e/tests/browser-smoke.spec.ts:27` — the authenticated Agents entrypoint exposes an accessible keyboard action
- `e2e/tests/browser-smoke.spec.ts:41` — the New Agent dialog traps focus and returns it after keyboard close
- `e2e/tests/responsive-smoke.spec.ts:28` — 390px, keeps the primary action and touch dialog inside the viewport
- `e2e/tests/responsive-smoke.spec.ts:73` — 1440px, centers the bounded content frame and preserves the page hierarchy

All four resolve `page.getByRole("button", { name: "New Agent", exact: true })`. `apps/web/src/features/agents/agents-page.tsx` now renders that affordance as a `Link` to `/agents/setup?action=create`, so the role is `link`, not `button`; two of them then open `getByRole("dialog", { name: "New Agent" })` and press its close control, both of which were deleted along with `new-agent-page.tsx` and the `agents_close_new_agent_dialog` message.

`app.spec.ts` was updated for the same rename; these two spec files were not. The Validation list in the description does not include `pnpm --filter @opentag/e2e test:e2e:smoke`, which is why it was missed — worth adding to the list for the next round.

### 2. Repairing those specs must not simply delete the a11y coverage

The dialog carried the only keyboard/focus contract Agent creation had: Enter on the trigger opens it, Tab stays trapped, Escape closes and returns focus to the trigger, plus `expectAccessible` on the creation surface. Replacing a dialog with a route navigation is a legitimate design change, but it changes that contract rather than removing the need for one. Please re-point the coverage instead of dropping it: keyboard-activate the link, assert the landing URL is `/agents/setup?action=create`, and run `expectAccessible` on the create surface at both the 390px and 1440px widths that currently assert it.

## Please address before merge

### 3. `creation-intent-store.ts` and `creation-recovery.tsx` now have no direct tests

Deleting `agent-creation.test.tsx` (840 lines) and `agent-creation/agent-creation-flow.test.tsx` (397) is correct for everything that tested the Computer picker and the dialog. But those files were also the only coverage of the creation-intent machinery, which survives, is still the guard against duplicate Agent creation, and is *modified by this PR* (`CREATE_INTENT_VERSION` 3 → 4, optional `computerId`, the new `requestsMatch`, the new `onDiscarded`).

What survives is one integration case in `agent-setup-boundary.test.tsx:112`, covering fail-then-Retry. Uncovered after this PR:

- a saved intent found **at mount** — the reload/crash case the store exists for
- `Check result` outcomes: `found` (clear + canonicalize), `not-found`, `ambiguous`, and the error branch
- `Discard` → `onDiscarded` → `startOver`
- `pruneSupersededCreationIntents` and the v3 → v4 invalidation

The evidence that this was an oversight rather than a decision: `storeCreationIntent` in `apps/web/src/__tests__/support/app-fixtures.tsx` was updated by this PR from version 3 to version 4 — and now has zero callers. The fixture that fed exactly these cases is still being maintained after the last test that used it was deleted. One test that seeds it, renders `/agents/setup?action=create`, and drives Check and Discard would close the biggest hole.

(For the record, the version bump itself is right and I am not asking for it back: a v3 record carries a `computerId` that the new computer-less `selectedRequest` can never match, so `requestsMatch` would leave Retry permanently disabled on a resumed v3 record. Worth a one-line comment saying so.)

### 4. No redirect for the removed `/onboarding` and `/agents/new`

Both route files are gone and `router.ts` falls through to `StandaloneNotFoundPage`. Anyone holding those URLs — an open tab mid-onboarding when the deploy lands, a bookmark, a link in a support thread — gets "not found" rather than the flow they were in. A redirect route for one release (`/onboarding` → `/agents/setup`, `/agents/new` → `/agents/setup?action=create`) is cheap. If the call is that the audience is small enough not to bother, say so and I will drop it.

(The Slack return path is fine — `resultRedirect` computes `/agents/setup` server-side at callback time, so an install started before the deploy still lands correctly.)

## Non-blocking

- `apps/web/src/__tests__/support/app-fixtures.tsx`: `storeCreationIntent` and `twoReadyComputers` are both unused now. Either remove them or, preferably, let finding 3 give `storeCreationIntent` a caller again.
- `apps/web/src/internal/reboard-review.ts:3` still says "when the browser reloads the onboarding route".
- `docs/staging-onboarding-reset.md` describes the lab as "the #437 in-memory Adapter". The issue number does not mean anything to a reader of this doc; "the in-memory Adapter" reads better. The file name is also now the only place the old vocabulary survives, alongside `scripts/e2e/onboarding-e2e.mjs`, which `DEVELOPMENT.md` still points at — consistent, just stale.
- Scope: the `im_disconnect_description` rewrite (en + zh), the computer-connect waiting copy, the QR box, and the command-lead row are unrelated to unifying the routes. No objection to the changes themselves — I checked the new disconnect copy against `im-binding-service.ts`, which does terminate active chat sessions and retain message and session history, so the claim is accurate — but they make this diff harder to review than it needs to be.

## Verified as correct

So it is clear what this review did and did not cover:

- `routeTree.gen.ts` is regenerated consistently, and there is no dangling reference to `/onboarding`, `/agents/new`, or `/internal/onboarding-v2` anywhere in the repo.
- `resolveTargets` fail-closed semantics match their tests: a malformed id short-circuits without any read, a missing/inactive/foreign exact id never degrades to the list, and a failed list read is named rather than treated as zero.
- `resolveSignInDestination`: dropping `onboarding` is safe because `/agents/setup` is already covered by the `agents(/...)` branch.
- The route's param-stripping effect keeps the Slack callback error visible, because `callbackError` captures it in `useState` initialiser before the replace navigation.
- `CreateAgentRequestSchema.computerId` is already optional, so thin creation is server-legal.
- `ReadyReport` is `... | undefined` and `useReadyReport` returns `undefined` outside review mode, so the new `onOpenAgent` action in `DoneStep` is actually reachable and does not get shadowed by `completion`.
- Both ratchets move the right way — `complexity-baseline.json` and `file-size-exceptions.json` only shrink, no new exception and no raised limit.
- i18n: 17 areas agree across `project.inlang/settings.json`, `docs/i18n.md`, `i18n-catalog.test.ts`, and `apps/web/messages/` on disk.

## Ordering

The base branch is `merge/issue-437-acceptance` (#469), which is still open, currently `CONFLICTING` with `main`, and has no review decision. Beyond the stated "do not merge until #469 lands": once #469 is resolved and merged, this diff needs re-verifying against the new base, since `main` has moved since this branch was cut.

apps/web/src/routes/_authenticated/_resources/agents.setup.tsx
```diff
@@ -0,0 +1,62 @@
...
+    action?: "create";
+    agentId?: string;
+    invalid?: true;
+    review?: "reboard";
+    slack_oauth?: "success";
+    slack_oauth_error?: string;
+  } => {
+    const action = search.action === "create" ? "create" : undefined;
+    const agentId = typeof search.agentId === "string" ? search.agentId : undefined;
+    const invalid =
```
yuezengwu, line 25: [P1] Reject present non-string agentId values\n\nTanStack default search parsing converts ?agentId=123 to a number and duplicate agentId keys to an array. This code maps either value to undefined, while invalid only checks malformed action, so an explicitly malformed exact-target URL falls through to target-less setup; with zero active Agents it can even show creation instead of failing closed. Treat any present agentId that is not exactly one string as invalid, and cover numeric and duplicate values.

apps/web/src/features/agents/agents-page.tsx
```diff
@@ -31,24 +28,19 @@ export function AgentsPage() {
...
-          <Banner variant="error" role="alert" description={slackConfigurationMessage(oauthError)} />
-        ) : null}
-        <AsyncState state={state}>{(value) => <AgentsContent agents={value.agents} />}</AsyncState>
-      </Page>
-      <NewAgentDialog open={createOpen} returnFocusRef={createTriggerRef} onClose={() => setCreateOpen(false)} />
-    </>
+    <Page
+      action={
+        <div data-ui="agents-page-action">
+          <Link className={buttonClassName({ variant: "secondary" })} search={{ action: "create" }} to="/agents/setup">
```
yuezengwu, line 34: [P1] Update required Browser Smoke coverage for the new route\n\nChanging this control from the button/dialog flow to a link leaves e2e/tests/browser-smoke.spec.ts and e2e/tests/responsive-smoke.spec.ts querying a button and dialog that no longer exist. The exact-head Browser Smoke check fails four tests, including both retries. Replace those scenarios with keyboard, touch, and viewport assertions for this link and /agents/setup?action=create.

yuezengwu (changes requested):
Requesting changes on exact head 707910396ea0bc70430bc7078de7ed67e99fe87b. The required Browser Smoke check fails 4 of 8 tests because the suite still exercises the removed New Agent dialog, and the canonical route does not fail closed for present agentId values that TanStack parses as non-strings. Local validation passed: shared/client build, Web typecheck, git diff --check, and 56 Web test files with 532 tests.

apps/web/src/routes/_authenticated/_resources/agents.setup.tsx
liuchao-001: Agreed, and this is a real P1 that my review missed — I saw the duplicate-key form and wrongly filed it as a nit because the deleted `/onboarding` route had the same `typeof … === "string"` shape. That defence does not hold: this PR is the one that makes `/agents/setup` the sole route and introduces the explicit fail-closed contract, so the contract has to cover this.

Confirming empirically against the pinned parser (`@tanstack/router-core` 1.170.32, `defaultParseSearch`, which the router uses since `createAppRouter` sets no `parseSearch`):

| URL | parsed `agentId` | `typeof` |
|---|---|---|
| `?agentId=123` | `123` | number |
| `?agentId=a&agentId=b` | `["a","b"]` | object (array) |
| `?agentId=null` | `null` | object |
| `?agentId=true` | `true` | boolean |
| `?agentId=<uuid>` | `"<uuid>"` | string |

All four non-string forms collapse to `undefined`, so an URL that explicitly names a target resolves by Account cardinality instead. The worst outcome is not the zero-Agent case — with **exactly one** active Agent, `resolveTargets` returns `{ kind: "redirect" }` and the reader is silently sent to a different Agent than the link named. That is precisely what the type's own comment forbids: "it never falls back to the list, which would quietly substitute one Agent for the one the link named."

One more consequence worth covering in the fix: the `action` + `agentId` conflict guard is bypassed by the same coercion. In

```ts
const agentId = typeof search.agentId === "string" ? search.agentId : undefined;
const invalid =
  (search.action !== undefined && action === undefined) || (action !== undefined && agentId !== undefined)
    ? true
    : undefined;
```

`?action=create&agentId=123` has `agentId` already `undefined` by the time the second clause runs, so `invalid` stays unset and the route opens the creation form. `agents-setup-boundary.test.tsx:102` covers the string form of that conflict only.

Suggest deriving `invalid` from the raw values rather than the narrowed ones — any present `search.agentId` that is not exactly one string is invalid — with cases for numeric, duplicate, and `?action=create&agentId=123`.

Gandy2025:
Addressed the actionable review feedback in c09ac81f:

- Updated all Browser Smoke and responsive smoke coverage from the removed New Agent dialog to the routed /agents/setup?action=create flow, including keyboard, accessibility, mobile overflow, and desktop checks. Local Browser Smoke now passes 8/8.
- Added focused creation-intent recovery coverage for a mounted saved intent, found/not-found/error outcomes, discard-and-start-over, and pruning the superseded v3 record after the v4 schema change.
- Added the requested one-line explanation for the v4 version bump.

I intentionally did not add redirects for /onboarding or /agents/new because the accepted product scope explicitly retires both routes without historical compatibility. I also did not add an ambiguous-result fixture: the current public creation-intent result schema can only return found or not-found; parser/network failures are covered by the error case. No unrelated fixture or documentation cleanup was included.

Validation completed locally: pnpm check, pnpm typecheck, pnpm build, pnpm test, focused Agent Setup tests (21/21), and Browser Smoke (8/8).

apps/web/src/routes/_authenticated/_resources/agents.setup.tsx
Gandy2025: Fixed in 04790a31. Explicit non-string agentId values now retain the invalid marker across TanStack search revalidation and fail closed without reading the Agent list or exposing creation. Added route-boundary coverage for both numeric and duplicate-key values; all 23 focused tests pass.

apps/web/src/features/agents/agents-page.tsx
Gandy2025: Fixed in c09ac81f. Browser and responsive smoke tests now exercise the New Agent link and routed creation surface with keyboard, touch, viewport, overflow, and accessibility assertions. Local Browser Smoke passes 8/8.

liuchao-001 (approved):
Re-reviewed at head `04790a31`. Everything I raised is resolved or answered, and the fail-closed P1 is fixed. Approving.

## Verified fixed

**Browser Smoke.** Green on this head (`CI` run 33604477396, all 10 jobs success, including Browser Smoke). The specs were not merely repointed at the new role — `browser-smoke.spec.ts` now keyboard-activates the link, asserts the landing URL, and runs `expectAccessible` on the create surface; `responsive-smoke.spec.ts` keeps the 390px touch/viewport/overflow assertions against `[data-ui="agent-create"]` and adds a 1440px navigation with its own overflow and accessibility check. That is the contract replacement I asked for rather than a deletion.

**Non-string `agentId`.** I checked the fix rather than reading it — ported `validateSearch` verbatim and ran it over `defaultParseSearch` output from the pinned `@tanstack/router-core` 1.170.32, including feeding its own output back in to exercise the revalidation path the new comment describes:

| URL | `invalid` | after revalidation |
|---|---|---|
| `?agentId=123` | `true` | `true` |
| `?agentId=a&agentId=b` | `true` | `true` |
| `?agentId=null` | `true` | `true` |
| `?agentId=true` | `true` | `true` |
| `?agentId=[]` / `?agentId={}` | `true` | `true` |
| `?action=create&agentId=123` | `true` | `true` |
| `?action=bogus` | `true` | `true` |
| `?agentId=<uuid>` | `undefined` | `undefined` |
| `?action=create` | `undefined` | `undefined` |
| `?agentId=<uuid>&slack_oauth=success` | `undefined` | `undefined` |

No mismatches: every malformed form fails closed and stays closed across revalidation, and no legitimate form regressed. `?agentId=` parses to `""`, which passes `validateSearch` but then fails `AGENT_ID_PATTERN` in the boundary, so it also reaches `unavailable` without a list read. The conflict-guard bypass is closed as a consequence of the new `search.agentId !== undefined && agentId === undefined` clause.

**Creation-intent coverage.** The four added cases are the ones that were missing, and they assert the right things: mount-time recovery through `Check` → `found` → canonicalize with the record cleared; `not-found` and the transient-error branch both *retaining* the record; discard resetting the name field from `recovered-agent` back to the default; and the v3 record being pruned rather than resumed. `storeCreationIntent` has a caller again, which also closes the dead-fixture note.

**Redirects for `/onboarding` and `/agents/new`.** Your call, and you stated the reason — accepted product scope retires both without historical compatibility. Dropping it as I said I would.

**Ambiguous fixture.** Your reasoning checks out: `AgentCreationIntentResultSchema` is a discriminated union of `found` and `not-found` only, and `browserApi.agentCreationIntent` parses through it, so no fixture can produce `ambiguous`.

## Non-blocking, for whenever you are next in here

- The `it.each` covers numeric and duplicate keys. `?action=create&agentId=123` is the case that previously slipped *both* guards, so it is the one most worth pinning — it is correct now only as a side effect of the new clause, and nothing would catch a future refactor that reorders those conditions.
- Following from the ambiguous point above: `CreationIntentCheck`'s `ambiguous` variant, the matching branch in `checkOutcomeDescription`, and `agent_create_check_ambiguous` are unreachable for the same reason. Either the schema should be able to express it or the client should stop claiming it can.
- `twoReadyComputers` in `apps/web/src/__tests__/support/app-fixtures.tsx` still has no callers.

## Before this can merge

Not review findings, just the standing gates:

- `ownership-gate` is red: "13 files awaiting approval. Needs bestony or yuezengwu." My approval does not satisfy it.
- @yuezengwu's own `CHANGES_REQUESTED` is still outstanding and needs their sign-off.
- #469 is still `OPEN` and `CONFLICTING` with `main`. Once it lands, this diff wants re-verifying against the new base — `main` has moved since the branch was cut, and the approvals and CI on this head describe `04790a31` against the current base only.

yuezengwu (approved):
Reviewed the exact head `04790a31142c7751ad40cbf687cc1f323a1c1cfc` against `merge/issue-437-acceptance@a5958e18513bbfa7104cc6eace4aaf06b394f36c` in a fresh detached worktree. The new head updates the Browser Smoke contracts for the routed creation flow and now fails closed for non-string or repeated exact `agentId` values; I also re-reviewed the full immutable diff. Validation: exact-head CI, Browser Smoke, Node compatibility, packaging/image, patch coverage, and quality checks are green or skipped; focused Agent Setup boundary tests pass (23/23), Web typecheck passes, `pnpm check` passes, and `git diff --check` passes. No blocking finding remains on this head.

yuezengwu (approved):
Exact-head review complete for `12d445615285eacc43ab22dc2751cd7f298f7171` against base `a4d9cfefaea4f5353471e3029a86a178d9c7f217`.

The new head reconciles the setup adapter assertion with the `ResponseSchemaError` contract and removes the redundant computer setup panel without changing the binding boundary. Local verification passed: `git diff --check`, shared build, Web tests (57 files / 547 tests), Web typecheck, and repository `pnpm check` (warnings only; all checks passed). Live head was refreshed immediately before this approval and remained exact.

liuchao-001 (approved):
Re-approving on exact head `12d445615285eacc43ab22dc2751cd7f298f7171`. My earlier approval was bound to `04790a31` and the head has moved three times since (`4d3cdee9` → `d39414a4` → `12d44561`), so it no longer described this code. This repository does not dismiss stale reviews, which meant the PR was showing a green `APPROVED` that nobody had given for the current head — worth watching on a branch that has re-based this often.

## What I checked on this head

**The branch-authored delta since my last approval is exactly one commit.** Rather than eyeball it, I compared the effective PR diff at both heads: `a5958e18..04790a31` and `a4d9cfef..12d44561` touch an identical set of 100 files, so the two merges added nothing to and dropped nothing from the PR's scope. Of the PR-touched files whose content moved between the heads, only `agent-setup-page.tsx` changed on the branch; `agent-shell.tsx` differs solely because the sidebar work in #476 landed underneath it.

**`4d3cdee9` is presentational only.** It drops the `<div className={PANEL}>` wrapper around `AgentComputerChoice` in the `requires-rebind` and `not-bound` states and removes the now-unused constant. No behavioural surface, no leftover references.

**The conflict resolution is correct.** I reproduced the conflict locally before it was fixed: one file, `apps/web/src/__tests__/api.test.ts`, where the base put `routeTemplate` before `message` and the branch after, leaving it duplicated in the merged hunk. This head keeps one `message` and one `routeTemplate`, and there are no conflict markers anywhere under `apps/web`, `packages`, or `e2e`.

**The merge did not quietly revert the change under review.** #476 refactored the sidebar callback into a named `openNewAgent`; it still navigates to `/agents/setup?action=create`, and no `/agents/new` or `/onboarding` route reference survives anywhere in the tree.

**CI is green on this exact head** — run 33607950767, all 10 jobs success including Browser Smoke, plus Patch Coverage and the ownership gate ("Ownership satisfied for 101 changed files").

Everything I verified at `04790a31` — the fail-closed `validateSearch` behaviour, the creation-intent recovery coverage, the re-pointed Browser Smoke contracts — is untouched by these three commits, so those conclusions carry forward.

## Still outstanding, not review findings

- @bestony is still a pending requested reviewer.
- #469 has not landed in `main`. This PR targets `merge/issue-437-acceptance`, so the stated dependency still stands: it must not merge until #469 is integrated.

The non-blocking notes from my previous review are unchanged and still worth a follow-up whenever convenient: pinning `?action=create&agentId=123` with a test, the unreachable `ambiguous` branch and its message, and the unused `twoReadyComputers` fixture.

liuchao-001 (approved):
Re-approving on exact head `7a95b5519b4c95585c3c599ef4e731655886a794`. My previous approval was bound to `12d44561`; this is the third head since, so it no longer described the code.

## The two heads since my last approval

**`8a8d76b8`** is a merge of the advanced base, but it carries hand-resolved content in two files, so I reviewed it rather than treating "chore: sync" as a no-op:

- `onboarding-v2/page.tsx` — while a saved creation intent is pending, the ordinary Create submit is now blocked in both the handler (`selectedRequest && !recovery.intent`) and the disabled state. This is the base's delivery-race fix carried into the rewritten file, and it is the right call: it stops the plain form from minting a second attempt while an unresolved one exists. I checked it cannot strand anyone — `intent` is `pendingIntent && creationIntentId !== dismissedIntentId`, and discard clears both, so the form re-enables; a validation/deterministic failure also clears the intent, so a name collision still lets you correct the name and resubmit. The label still reads "Create Agent" rather than falsely showing "Creating…", because it derives from the raw `submitting` state.
- `agent-setup-boundary.test.tsx` — asserts the Agent list is read exactly once on the zero-Agent path, and that a pending intent disables Create such that submitting the form issues no POST and leaves the stored record intact. The second one pins the guard above.

PR scope is unchanged across both merges: the effective diff is the same 100 files as at `12d44561`, nothing added or dropped.

**`7a95b551`** replaces `opentag doctor --fix` with `opentag doctor` and rewrites the surrounding hint. I verified the old copy was actually wrong rather than taking the commit message's word: there is no `--fix` anywhere in `apps/cli`, and `docs/design/doctor-product-spec.md` is explicit that doctor "不是修复命令" and is read-only. So the UI was naming a flag that does not exist and describing a read-only diagnostic as a repair — while `onboarding_v2_messaging_cli_missing` in the same catalogue already said the opposite. Good fix. I also checked the composed sentence, not just the fragments: hint + `<code>` + suffix renders exactly the string the updated test asserts, and the zh version spaces the Latin code span correctly.

## About the red Browser Smoke on this head

The first CI run on `7a95b551` failed Browser Smoke, and I want the reason on the record rather than a bare "re-ran it":

```
psql: ... FATAL:  the database system is shutting down
Error: Process from config.webServer was not able to start. Exit code: 1
```

The Playwright web server never started, so no test executed and no assertion failed — not something a copy-only commit can cause. It landed inside a burst of six CI runs between 09:49 and 09:56, consistent with a Postgres teardown race on the runner. I re-ran the failed job on the same SHA as a diagnostic, and it passed with no code change, which is what distinguishes an environmental failure from a deterministic one. Every check on this head is now green, including the ownership gate.

If that mode recurs, it is worth looking at whether concurrent jobs share a compose stack, rather than absorbing it as noise.

## Still outstanding, not review findings

- @bestony is still a pending requested reviewer.
- #469 is `MERGEABLE` again but still `OPEN`. This PR targets `merge/issue-437-acceptance`, so the stated dependency stands.

One process note: this is the fifth head since the first approval, and two arrived three minutes apart. Because this repository does not dismiss stale reviews, each push leaves the PR sitting at a green `APPROVED` describing older code. I have re-bound each time, but batching further changes — or marking the PR draft until it settles — would avoid that window and the CI cost.

yuezengwu (approved):
Reviewed exact head `7a95b5519b4c95585c3c599ef4e731655886a794` against base `83c335620210ddc053f892170a20eb2b85d17914`. I re-reviewed the hand-resolved merge content: the pending-intent guard remains fail-closed with valid discard/recovery exits, and the CLI recovery copy now correctly names the read-only `opentag doctor` command rather than the nonexistent `--fix` flag. Local verification passed: Web tests (57 files / 549 tests), generated-Paraglide Web typecheck, repository `pnpm check`, and `git diff --check`. Live head and all required GitHub checks were refreshed immediately before this approval and remained exact/green.

liuchao-001 (changes requested):
Reviewed head `d8fe1ae59428ccb29b561f34b227fbf392fa4af2`. The feature itself is good, but this commit breaks the repository's real-stack E2E, and it is the one piece of validation CI cannot catch — so it will stay green while the script is broken.

## Blocking: `scripts/e2e/onboarding-e2e.mjs` no longer matches the UI

This commit replaces the setup header's exit control. Before, the exact-Agent surface received `onBackToAgents` and rendered **"Back to agents"** → `/agents`. Now `onBackToAgents` is dropped from that path entirely, and the header renders **"Back to agent"** → `/agents/$agentId` while the stage is not `ready`:

- `agent-setup-page.tsx:462` is the only site rendering `onboarding_v2_back_to_agent`
- `page.tsx:165` is now the only site rendering `onboarding_v2_back_to_agents`, and that is `AgentCreatePage` — not this surface

The E2E still drives the old control at `scripts/e2e/onboarding-e2e.mjs:548`, on the exact-Agent page:

```js
await page.getByRole("button", { name: "Back to agents" }).click();
await page.waitForURL(`${BASE_URL}/agents`, { timeout: 30_000 });
```

That fails twice over:

1. `getByRole` matches the accessible name case-insensitively **by substring** when `exact` is not set. The filter `"Back to agents"` is longer than the rendered name `"Back to agent"`, so it is not a substring of it — the locator resolves to nothing and the click times out.
2. Even if it matched, the destination is now `/agents/<uuid>`, so `waitForURL("…/agents")` would fail.

It also cascades: the next step, "survive a Computer outage without losing the Agent", reloads and waits for the **Agents** heading, having had its own `waitForURL('/agents')` removed earlier in this PR on the assumption that step 6 already navigated there. So the failure is not isolated to one assertion.

The PR description still lists "real Chromium + real Server + real PostgreSQL + real CLI daemon Agent Setup E2E: 14/14 steps passed". That was true when it was written; it is not true of this head. Please update the script (target "Back to agent" and `/agents/<id>`, then restore the `/agents` navigation the outage step needs) and re-run it, or say explicitly that the claim now refers to an older head.

## Reviewed and fine

**The new entry point.** `setupOrMaintenanceAction` gives a dependency row the canonical `/agents/setup?agentId=…` exit only when that dependency is the one currently owning the Agent-wide blocker, and otherwise leaves the focused Settings link. That is the right shape — it is what stops an unbound Agent from also offering Setup on its downstream Messaging row, and the comment says so.

**The exit control is not a regression.** My first read of the boundary diff was that dropping `onBackToAgents` left a completed Account with no way out of exact setup until `ready`. It does not: `onOpenAgent` now doubles as the pre-ready "Back to agent" and the post-ready "Open agent", and the boundary always supplies it. `agent-setup-page.test.tsx` pins that swap.

**Coverage is present for the new behaviour** — the setup link in `agent-presentation.test.ts` and the ready/not-ready header swap in `agent-setup-page.test.tsx`.

## One non-blocking note

`runtime.status === "checking"` previously carried no action and now gets `setupOrMaintenanceAction(agent, "runtime_unavailable")` with no maintenance fallback. So while a runtime check is still resolving, a blocked Agent now offers "Continue setup". Defensible — the Agent is blocked either way — but it is the one row where the action appears during an in-flight check rather than after a settled verdict, so it is worth a deliberate yes rather than falling out of the refactor.

## Unchanged

CI on this head was still running when I wrote this; it does not bear on the finding above either way, since the broken script is not part of it. @bestony is still a pending requested reviewer, and #469 is still `OPEN`.

yuezengwu (approved):
Reviewed exact head `d8fe1ae59428ccb29b561f34b227fbf392fa4af2` against base `83c335620210ddc053f892170a20eb2b85d17914`. I re-reviewed the complete effective diff and the new incomplete-setup entry commit: only the dependency that owns the Agent-wide blocker routes to canonical Setup, maintenance and unconfirmed states retain their focused Settings exits, and targeted Setup returns to the exact Agent before readiness while preserving the existing ready action. Local verification passed: Web tests (57 files / 555 tests), generated-Paraglide Web typecheck, repository `pnpm check`, and `git diff --check`. Live head and GitHub checks were refreshed immediately before approval and remained exact/green.

liuchao-001:
@yuezengwu and I are not in conflict on this head, so please don't read the `APPROVED` + `CHANGES_REQUESTED` pair as a disagreement to arbitrate.

Their review is right about everything it covers. My finding is about a file that sits outside every check either of us ran, and I re-verified that before holding the position:

- `scripts/e2e/onboarding-e2e.mjs` is referenced nowhere but its own usage comment — no workflow, no `package.json` script.
- CI's only browser stage is `pnpm test:e2e:smoke` → `playwright test --project=smoke`, which is the `e2e/` package's suite, a different set of files.
- The validation in their review — Web tests, Paraglide typecheck, `pnpm check`, `git diff --check` — does not execute that script either.

So "all green" on this head and "the script is broken" are both true at once. That is the whole reason this one is worth stopping for: it is the only validation the PR advertises that nothing automatic will ever contradict.

One correction to my own review while I am here: I justified the locator failure by saying `getByRole`'s `name` is substring-matched without `exact`. That detail is not load-bearing and I should not have leaned on it — `"Back to agents"` is neither equal to nor a substring of the rendered `"Back to agent"`, so the locator resolves to nothing under either matching rule. The conclusion is unchanged.

The fix is small: point step 6 at "Back to agent" and `/agents/<id>`, then restore the `/agents` navigation the outage step needs. Happy to re-approve as soon as the script matches the UI again, or to drop the finding entirely if you would rather retire that script — but leaving it in the tree, broken, while the description cites it as passing validation is the state I do not think should merge.

Gandy2025:
Addressed the actionable real-stack E2E finding in `2d8e3480`:

- The exact-Agent setup step now clicks `Back to agent` and asserts the source `/agents/<id>` destination.
- The following outage step explicitly navigates to `/agents`, so it no longer depends on the prior step landing on the list.
- Re-ran `scripts/e2e/onboarding-e2e.mjs` against real Chromium, Server, PostgreSQL, and CLI daemon: **14/14 steps passed**.
- `pnpm check`, repository typecheck, and pre-push gates passed.

I also assessed the non-blocking `runtime.status === "checking"` note. Keeping `Continue setup` there is deliberate: `runtime_unavailable` is the current Agent-wide setup blocker, and Setup is the live surface that shows the in-progress readiness state; it does not create a second exit or route a maintenance state away from Settings.

yuezengwu (approved):
Reviewed exact head `2d8e34809d6ef68cafa4b19436b209d8c504d435` against base `83c335620210ddc053f892170a20eb2b85d17914`. I re-reviewed the complete effective diff and the new E2E repair. The script now follows the rendered `Back to agent` control to `/agents/<id>` and explicitly navigates to `/agents` before the outage assertion, so the two steps no longer depend on the removed list return path. The initial CI failure was an unrelated `provider-cli-turn-runner` SIGINT timing failure; that untouched test passed 5 consecutive local runs on this head (70 assertions), and the GitHub failed-job rerun is now green. Local `pnpm check` and `git diff --check` passed; all exact-head GitHub checks are green. No blocking finding remains.

liuchao-001 (approved):
Approving on exact head `2d8e34809d6ef68cafa4b19436b209d8c504d435`. This clears the `CHANGES_REQUESTED` I filed against `d8fe1ae5`.

## The E2E finding is resolved

`2d8e3480` fixes both halves, and fixes the second one better than I asked:

- Step 6 now clicks **"Back to agent"** and waits for `/agents/<id>`, which is what `onOpenAgent` → `agentDetailLink` actually navigates to, and the label the header renders while the stage is not `ready`.
- Step 7 no longer depends on the previous step having landed on the list — it navigates to `/agents` explicitly instead of reloading. I had suggested restoring the old `waitForURL`; decoupling the step is the better answer.

I checked `agentId` is genuinely in scope at the new assertion — it is declared at line 512 inside the same step callback — and swept `scripts/` and `e2e/` for other references to controls this PR removed ("Back to agents", "Everything your agent needs", "Agent created"). None remain. Thanks for re-running the real-stack script rather than just patching the selector.

Your answer on `runtime.status === "checking"` is exactly the deliberate yes I was asking for, and I am not pressing it further.

## About the red CI on this head, which was not yours

The first run failed `Checks` on `@opentag/client#test` — a package this commit cannot reach. I traced it rather than re-running on the assumption it was noise, and it is a real fixture defect:

`fixtures/provider-cli-turn-target.mjs` writes `"ready"` to stderr **before** registering its `SIGTERM`/`SIGINT` handlers, while the test treats `"ready"` as permission to signal. If the signal lands in that window, Node's default disposition kills the process: nothing writes `got SIGINT`, and it still exits 130, so the exit-code assertion passes and only the stderr assertion fails — precisely the observed `expected 'ready\n' to contain 'got SIGINT'`.

Filed as #486 with the one-line fix (move the readiness write below the handler registrations). I re-ran the job afterwards and the head is now fully green, including the ownership gate.

## Merge gates, unchanged

- @bestony is still a pending requested reviewer.
- #469 is `MERGEABLE` but still `OPEN`; this PR targets `merge/issue-437-acceptance`, so the stated dependency stands.

The earlier non-blocking notes still stand for a future pass: pinning `?action=create&agentId=123` with a test, the unreachable `ambiguous` branch and its message, and the unused `twoReadyComputers` fixture.

liuchao-001:
#469 merged at 11:20, so the stated dependency is satisfied and GitHub has retargeted this PR to `main`. Two things follow, and the second one has a trap in it.

**The approvals are gone.** `reviewDecision` is now empty — retargeting reset it. That is correct rather than annoying: the effective diff is computed against a different base now, so nothing that was approved describes what would merge. Ours will need re-doing after the reconciliation, which is fine.

**The reconciliation is structural, not a few lines.** #469 landed on `main` as a squash (`37a86739`, single parent), while this branch still carries #469's original unsquashed commits. Git therefore cannot tell that the shared content is already applied, and a straight merge of `main` produces 24 conflicts:

- **12 both-modified** — `steps.tsx`, `_shell.tsx`, `page.tsx`, `agent-setup-page.tsx`, `slack-oauth.ts` and its test, `creation-intent-store.ts`, `creation-recovery.tsx`, `agents-page.tsx`, `account.test.tsx`, `app-shell.test.tsx`, `agent-setup-page.test.tsx`
- **8 delete-vs-modify** — `agent-creation-flow.tsx`, `server-backend.ts`, `server-flow.test.tsx`, `agent-creation.test.tsx`, `pr318-findings.test.tsx`, `routes/…/onboarding.tsx`, `messages/onboarding/en.json`, `messages/onboarding/zh.json`
- **4 both-added** — including `scripts/complexity-baseline.json` and `scripts/file-size-exceptions.json`

The trap is that second group. Those are files this PR deliberately deletes, which the squashed #469 still modifies on `main`, so git leaves main's version in the tree and asks. Resolving any of them by keeping the file silently resurrects exactly what this PR exists to remove — the duplicate creation flow, the Web-owned backend, the retired route, and the obsolete tests. `messages/onboarding/*` coming back would also put the i18n area count back to 18 against the 17 that `project.inlang/settings.json`, `docs/i18n.md`, and `i18n-catalog.test.ts` all agree on, so that one at least fails loudly. The others would merge quietly as dead code.

Whichever way you reconcile — rebasing onto `main` and dropping the #469 commits, or merging and resolving — the check I would want afterwards is that the effective diff against `main` still deletes all eight, and that the file set matches the 100 files this PR has been touching. Happy to re-review as soon as it is pushed.

Gandy2025:
Synced #473 onto `main@76da332b` in merge commit `33d3f7d0` without rewriting the published branch.

Conflict review:
- preserved deliberate deletions of retired `/agents/new`, `/onboarding`, old Web backend, and duplicate flow tests;
- carried agent-targeted Computer connect into the canonical exact-Agent setup surface;
- carried independent Lark/Slack CLI readiness into Messaging setup;
- fixed the post-merge boundary fixture for the required `imCliReadiness` snapshot contract;
- effective scope adds only four directly intersecting paths beyond the previously reviewed file set.

Local validation on the reconciled tree:
- `pnpm check`, `pnpm build`, `pnpm typecheck`: passed
- focused unified setup/detail routing: 94/94
- Browser Smoke: 8/8
- real Chromium + Server + PostgreSQL + CLI daemon setup journey: 14/14
- Client runtime coverage: 347/347 and 100% statements/branches/functions/lines
- changed Computer integration suite, serial: 17/17
- repository unit run had one unrelated 5s filesystem-scan timeout; isolated retry passed 1/1
- full Server integration was resource-limited locally by PostgreSQL timeouts under two-worker contention; the directly changed Computer integration suite passes serially and CI remains authoritative

The known Journey navigation assertion still waits for the intentionally hidden `Skills` entry (10 passed, 1 stale failure, 4 skipped); no product scope was expanded to restore that retired navigation. Please re-review exact head `33d3f7d0`.

liuchao-001 (approved):
Approving on exact head `33d3f7d08524d6753c93d6e3aae6f6507bf20e9d`, now against `main`. All checks green including the ownership gate, and the dependency this PR carried from the start is satisfied — #469 is in `main`.

## I ran the check I asked for

**All eight delete-vs-modify files are still deleted.** `agent-creation-flow.tsx`, `server-backend.ts`, `server-flow.test.tsx`, `agent-creation.test.tsx`, `pr318-findings.test.tsx`, `routes/…/onboarding.tsx`, `messages/onboarding/en.json`, `messages/onboarding/zh.json` — each shows `D` in the effective diff against `main` and is absent from the head tree. That was the part of the reconciliation that could have gone quietly wrong, and it did not.

**Scope is 111 files, and the eleven beyond the previously reviewed set all account for.** Seven come from `d8fe1ae5`, the incomplete-setup entry commit already reviewed. The other four are genuine new intersections with #477:

- `agent-computer-choice.test.tsx` — one line, realigning an expectation onto the command-intro copy this PR moved into `ComputerConnect`.
- `im-cli-status.tsx` — retyped off the deleted `flow.ts` types onto `ImProvider`; the component itself survives.
- `agent-reset-control.tsx` — deleted. I checked it is referenced nowhere at this head; `main` used it only from the `page.tsx` this PR rewrites, so keeping it would have left dead code.
- `im-cli-prewarm.test.tsx` — deleted. It imports `ComputerStep`, `MessagingStep`, `backend.js`, and `ReadinessFacts`, all of which this PR removes, so it could not compile. The deletion is forced rather than a judgement call.

I checked that last one specifically, because deleting a test `main` had just shipped is what a bad reconciliation looks like from the outside. The behaviour that survives keeps its coverage: `agent-setup-page.test.tsx` seeds `imCliReadiness: { feishu: "install", slack: "ready" }` and asserts the rendered "Preparing".

**The PR's own invariants survived the merge** — the non-string `agentId` fail-closed clause, the E2E's "Back to agent" step, no surviving `/agents/new` or `/onboarding` route references, 17 i18n areas, and `messages/onboarding/` gone.

## On the Journey `Skills` failure

Checked rather than taken on trust, and your reading is right: it is not this PR's. The entry is gated by `internalNavigation.skills` identically on `main` and here, and this PR's only change to `agent-shell.tsx` is the `/agents/new` → `/agents/setup?action=create` navigation.

Worth separating out, though: `journey` is not in CI either — `ci.yml` runs only `pnpm test:e2e:smoke`, which is `--project=smoke`. So together with `scripts/e2e/onboarding-e2e.mjs` there are now two suites on `main` that nothing automatic guards, and this is the second time in this PR that a break in one of them had to be caught by hand. That is a repo-level gap rather than something to fix here, but it is worth an issue.

## Remaining

@bestony and @yuezengwu are both pending requested reviewers — the retarget reset the prior approvals, correctly, since the effective diff is now computed against a different base.

Earlier non-blocking notes still stand for a future pass: pinning `?action=create&agentId=123` with a test, the unreachable `ambiguous` branch and its message, and the unused `twoReadyComputers` fixture.

liuchao-001 (approved):
Approving on exact head `f9767dae985849aa2b8e8a860952c60931b04415`. All checks green, including the ownership gate.

This one needed no re-review of content, and I can show that rather than assert it.

`f9767dae` is a clean merge of `main` — `git log --cc` reports no hand-resolved content, so nothing was authored during it. Scope is unchanged at 111 files, none added or dropped. And the effective diffs at both heads hash identically:

```
76da332b..33d3f7d0  →  8ac064d311e294358fe954d3cb133e50600c48d6db476d7adaaaaf84c67d3436
4f179cae..f9767dae  →  8ac064d311e294358fe954d3cb133e50600c48d6db476d7adaaaaf84c67d3436
```

Same SHA-256, so this is byte-for-byte the proposal already reviewed at `33d3f7d0`; only the base moved beneath it as #487 landed. Everything I verified there carries over on evidence rather than on assumption — the eight deletions, the accounted-for 111-file scope, the fail-closed `agentId` clause, the E2E return path, 17 i18n areas.

I still waited for CI rather than treating the identical diff as sufficient: it is a new tree, and #487's usage changes could in principle interact at build or test level even with an unchanged patch. They do not — run 33636490642, all ten jobs green.

@bestony and @yuezengwu remain pending requested reviewers.

yuezengwu (approved):
Approved on exact head f9767dae985849aa2b8e8a860952c60931b04415 against main@4f179cae2642573a012b0f1a8ef70b72eebb5d71. I compared the effective PR diff to the previously reviewed 33d3f7d0 head: the file set and stable patch-id are identical, so the new merge only incorporates the already-merged #487 base changes; all eight conflict-sensitive deletions and the repaired Agent Setup E2E navigation remain intact. Exact-head local validation passed: root build, Web tests (58 files / 560 tests), repository check, typecheck (8/8), Browser Smoke (8/8), and git diff --check. All live GitHub checks are green, and the head was refreshed immediately before this approval.

PR #474, opened 2026-09-02
fix(web): name the sign-in method instead of its identifier

### LINKED ISSUE #322: The login page renders the auth provider's identifier instead of its name ("Continue with github")
## What

`apps/web/src/features/auth/login-page.tsx:114` renders the auth provider's
**identifier** straight into reader-facing text:

```tsx
<span>Continue with {provider.id}</span>
```

So the sign-in button reads **"Continue with github"** — lowercase, in the
identifier's casing rather than the product's name. It is the first screen an
unauthenticated visitor sees.

## Why it's worth a separate issue

This is the same defect class that PR #311 eliminated for messaging providers,
in a different domain. There the label was derived with `titleCase(provider)`;
here it is not transformed at all. The shared root cause is that an identifier —
the Server's vocabulary, chosen for storage and routing — is allowed to reach a
reader without anyone ever deciding how it should read.

The fix has the same shape as the one that worked for messaging: a single
`authProviderLabel(provider)` with an exhaustive `switch`, so display code never
sees a raw id and adding a provider fails to compile until it is given a name.
That turns "a new provider silently names itself" into a build error.

Not folded into #311 deliberately: that PR is scoped to messaging-channel
naming, and widening it to the auth domain would have made its review harder
for no gain.

## Suggested fix

1. Add `authProviderLabel()` beside the existing `im/provider-label.ts`, same
   exhaustive-`switch` + `never` shape.
2. Use it at `login-page.tsx:114`.
3. Sweep for other spots rendering `provider.id` in the auth surface.

Found by @code-reviewer during independent review of #311.

liuchao-001:
Closes #322.

## The reported string does not exist

#322 says the sign-in button reads **"Continue with github"** and calls it "the first screen an unauthenticated visitor sees". Checked against `main`, neither half holds:

- **There is no GitHub sign-in provider.** `AuthProvidersResponseSchema` is `z.enum(["google", "dev", "password"])`. (`AuthIdentityProviderSchema` *does* include `github`, which is likely where the report came from — but nothing on the login page uses it.)
- **Of those three ids, only one can reach that sentence.** `password` is deliberately excluded from `linkProviders` because it renders as a form; `google` returns early and renders as an image. So the string is **"Continue with dev"**.
- **A visitor never sees it.** The server gates `dev` behind `Boolean(options.devSignIn) && isLoopbackAddress(request.ip) && isLoopbackAddress(request.hostname)` — it is a loopback-only affordance, so the reader is a developer running locally.

The **defect class** is real and worth closing: `provider.id` — the Server's vocabulary, chosen for storage and routing — reaches reader-facing text with nobody deciding how it should read. That is exactly what `messagingProviderLabel` fixed for channels. But the user-facing severity is much lower than the issue states, so this is consistency work, not a visitor-facing bug fix.

## The fix, in the shape that already works here

`authProviderLabel` is the naming point for sign-in methods: an exhaustive `switch`, so adding a method fails to compile until somebody names it. Only `Google` is a brand and stays as written; `dev` and `password` describe a *way to sign in*, so their words are translated and come from the catalogue.

| id | en | zh |
|---|---|---|
| `google` | Google | Google |
| `dev` | Developer sign-in | 开发者登录 |
| `password` | Email and password | 邮箱密码 |

## Why the spacing helper moved

The same `{provider}` slot now takes a Chinese phrase (开发者登录) and a Latin brand (Google), which is the problem #456 solved for messaging: the template cannot own the space because it is right for one and wrong for the other.

That rule was never brand-specific — it is a property of written Chinese — so it moves from `im/provider-label.ts` to `i18n/format.ts` as **`spaceScriptBoundary`**, where both a messaging channel and a sign-in method can use it. The 29 messaging call sites are renamed; no behaviour changes, and their existing full-message assertions still pin the output.

Without it, `zh` would render `使用Google继续`.

## Mutation-checked

Returning the raw id from the helper — the defect as it exists today — fails three assertions:

```
expected 'dev' not to be 'dev'
expected 'Continue with dev' to be 'Continue with Developer sign-in'
expected '使用 dev 继续' to be '使用开发者登录继续'
```

One assertion is deliberately `authProviderLabel(id) !== id` rather than "the sentence does not contain the id": `"Email and password"` contains the word *password* because that is the English word, and a `not.toContain` check would have mistaken that coincidence for a pass.

## Not done here

The #459 structural guard covers messaging brands only. Extending it to auth labels would catch a future `"Google"` written by hand outside this helper; I did not widen it in this PR because the derived-brand set and its exemptions are scoped to messaging, and mixing the two sets deserves its own review.

## Testing

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **71 files / 604 tests passed** |
| `pnpm check` | exit 0 |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 93.35%) |

One accepted ratchet baseline: `steps.tsx` 1148 → 1149, the single import line the move produces. `--accept-current` rewrote that one number and nothing else.

**A note on the coverage run**, since it may recur for a reviewer: the first attempt failed with one test — `onboarding-v2/page.test.tsx > "hands a failing check to the terminal rather than offering a retry button"` — while the machine was at load ~22 with 23 competing vitest processes. That file passes 46/46 three times in isolation, the full suite passes twice, and the coverage run is green once load drops to ~6. This is the #325 pattern, and it is a data point that the affected set is wider than `app.test.tsx`.

No live QA proposed: the only changed rendering sits behind a loopback-gated provider, and both locales are pinned by full-message assertions.


bestony (changes requested):
## Standards

- **P2: hardcoded user-facing copy bypasses the message catalogue.** `apps/web/src/features/auth/provider-label.ts:20-21` returns the literal `"Google"`, which is rendered by `apps/web/src/features/auth/login-page.tsx:117`. `AGENTS.md` requires all user-facing copy to go through `m.*()`. Please add an `auth_provider_google` key to both `apps/web/messages/auth/en.json` and `apps/web/messages/auth/zh.json`, then return `m.auth_provider_google()` from the `google` branch.

No other documented-standard violations found. The duplicated `withLocale` test helper is a minor, non-blocking test-only smell.

## Spec

No findings. The exhaustive provider label, locale strings, Chinese script-boundary spacing helper, and auth `provider.id` rendering sweep match issue #322 and this PR description.

apps/web/src/features/auth/login-page.tsx
```diff
@@ -112,7 +114,7 @@ export function LoginProviderLink({ next, provider }: { next: string; provider:
       data-ui="login-provider"
       href={href}
     >
-      <span>{m.auth_continue_with_provider({ providerId: provider.id })}</span>
+      <span>{spaceScriptBoundary(m.auth_continue_with_provider({ provider: authProviderLabel(provider.id) }))}</span>
```
yuezengwu, line 117: [P2] Exercise the production provider link. This test recreates LoginProviderLink’s composition in a separate helper instead of rendering LoginProviderLink or LoginPage. If the production line is changed back to interpolate provider.id directly—the exact regression this PR fixes—the new helper tests still pass because they continue calling authProviderLabel on their own; the existing login tests cover only Google/password, so the full Web suite does not pin “Continue with Developer sign-in” at the actual UI boundary. Please render an enabled dev provider through the real component (including the zh case, since the spacing helper is part of this fix) and assert the visible/accessibility label.

yuezengwu (changes requested):
Exact-head review of 0c55fa4695e5e6907dd79cb178371eaa9009a1f2. The provider mapping, locale-specific labels, script-boundary formatting, messaging/onboarding call-site migration, and file-size ratchet are internally consistent, but the regression test does not exercise the production login surface. Requesting the focused caller-level coverage described inline.

liuchao-001:
## Both findings fixed at `ac3245db`

### P2: the test recreated the composition instead of rendering it (@yuezengwu)

You are right, and this is the sharpest kind of finding: my test called `authProviderLabel` itself and built the sentence the way the component does, so **the exact regression it exists to catch would have left it green**.

It now renders the real login surface with an enabled `dev` provider and asserts the accessible name, in both locales. Mutating the component back to the defect:

```tsx
m.auth_continue_with_provider({ provider: provider.id })
```

fails both new tests. Before this commit it passed them.

One thing worth naming rather than papering over: you asked for the zh case *"since the spacing helper is part of this fix"*, and the zh assertion is there — but it pins the **label**, not the spacing. `dev` is the only id that reaches this sentence (`google` is an image, `password` is a form), and 开发者登录 is entirely CJK, so `spaceScriptBoundary` is a **no-op on every path reachable today**. The unit test is what pins the spacing, against the Latin case that becomes reachable the moment a Latin-named method is added — which is exactly the `github` provider #322 imagined. I kept the call for that reason; if you would rather it not sit there until then, say so and I will drop it and let the unit test stand as the record.

### P2: `"Google"` bypassed the catalogue (@bestony)

Fixed — `auth_provider_google` in both catalogues, and the helper returns `m.auth_provider_google()`.

I checked this against the boundary #456/#459 established (a brand comes from the naming point, never from a catalogue entry) and it does **not** conflict, for two reasons worth recording:

- `auth_sign_in_with_google` in this same file already spells `Google` in both locales, so this matches existing practice rather than introducing it.
- #459's guard derives its protected set from `messagingProviderLabel`, so it covers Feishu/Lark/Slack/飞书 — not Google. Nothing is weakened.

The helper still owns **which** name an id gets; only the letters moved. That required adding `Google` to `allowedUntranslatedValues`, since the value is identical in both locales — which is precisely what that set is for, alongside Slack and Codex.

### Nit: duplicated `withLocale`

Taken — it was copied into four test files and now lives in `__tests__/support/with-locale.ts`.

## Testing at `ac3245db`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **71 files / 606 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 93.35%) |

Coverage was run at load ~7 this time; the flake noted in the description did not recur.

bestony (approved):
Re-reviewed at `ac3245db8d13939a6cf9a316a8926cff61c56f89`.

Both prior findings are resolved:

- `authProviderLabel("google")` now reads `m.auth_provider_google()` with synchronized en/zh catalogue entries.
- `apps/web/src/__tests__/auth.test.tsx` now renders the production `App` with an enabled `dev` provider and asserts the real accessible link name in English (`Continue with Developer sign-in`) and Chinese (`使用开发者登录继续`).

The focused affected tests pass: 5 files / 33 tests. `git diff --check` passes. The full Web suite was attempted with a 120-second timeout under one worker and did not reach a terminal result in this environment.

yuezengwu (approved):
Exact-head review of ac3245db8d13939a6cf9a316a8926cff61c56f89: approved. The production LoginProviderLink is now covered by English and Chinese accessible-name assertions; a local mutation back to provider.id failed both tests with the raw-id strings, confirming the regression guard exercises the real caller. The exhaustive provider mapping, Google catalogue migration/untranslated allowlist, script-boundary formatting, shared locale-test helper, messaging/onboarding call-site migration, and file-size ratchet are coherent. Local validation passed: Web 71 files / 606 tests, Web typecheck, Web production build, pnpm check, and git diff --check. This approval is exact-head only. The PR is still CONFLICTING/DIRTY against main in agent-presentation.ts and currently has no GitHub check-runs; resolving that conflict will create a new head that requires a fresh review and CI.

liuchao-001:
## Conflict resolved at `e197f604` — fresh head, so both approvals need re-running

@yuezengwu called this exactly: the resolution creates a new head, and the `ac3245db` approvals do not carry.

**The conflict**, and it was a real one rather than a mechanical one: `agent-presentation.ts`'s imports. Main (#471) added `formatElapsedCompact` from `i18n/format`, while this branch moved the Chinese spacing rule *into* that same module — so both sides rewrote the same two import lines for unrelated reasons. Resolved by taking both:

```ts
import { formatElapsedCompact, spaceScriptBoundary } from "../../i18n/format.js";
import { messagingProviderLabel } from "../../im/provider-label.js";
```

I also swept for `spaceBrandInSentence` across the merged tree in case main's new code had introduced call sites under the old name — there are none, so the rename needed no follow-up.

## Testing at `e197f604` (post-merge, not carried over)

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **71 files / 606 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 1 improvement, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 93.20%) |

Coverage moved 93.35% → 93.20% because main's three merged PRs brought in new lines, not because this branch lost any: the covered count rose 3159 → 3167 while the total rose 3384 → 3398.

`paraglide` had to be regenerated after the merge for main's new `agents_card_status_*` keys, and `pnpm install --frozen-lockfile` re-run. Worth knowing if a reviewer sees `TS2339` on those keys — it is a stale generated directory, not a missing entry.

@yuezengwu @bestony — re-requesting on this head.

bestony (approved):
Re-reviewed exact head `e197f6041ff4b2d065547aee83a78c148e00e1bb` after the conflict resolution. The merge keeps both `formatElapsedCompact` and `spaceScriptBoundary` imports in `agent-presentation.ts`; no call sites were lost. The production login tests and catalogue changes remain intact.

Focused validation on this head passed: shared build plus 5 affected Web test files / 33 tests, and `git diff --check`. GitHub CI is currently running; no code findings remain.

yuezengwu (approved):
Exact-head review of e197f6041ff4b2d065547aee83a78c148e00e1bb: approved. The merge conflict resolution preserves both main’s formatElapsedCompact import/use and this PR’s spaceScriptBoundary behavior; no spaceBrandInSentence call sites or old providerId message inputs remain. Paraglide regenerated successfully with the provider input and all three provider catalogue messages. A fresh local mutation of LoginProviderLink back to provider.id failed both the English and Chinese production-surface tests with the raw-id strings, then the worktree was restored clean. Local validation passed: Web 71 files / 606 tests, Web typecheck, Web production build, pnpm check, and git diff --check. Exact-head GitHub CI is also complete: 14 success, 4 skipped, 0 pending/failing; PR is MERGEABLE/CLEAN. Approval applies only to this SHA.

PR #475, opened 2026-09-02
refactor(web): delete the Agent status messages no reader can reach

### LINKED ISSUE #336: Agent messaging-status messages are unreachable: an uncalled builder and three branches behind the suspended guard

liuchao-001:
Closes #336.

## The issue undercounts it: thirteen branches, not three

#336 reports that three `agentRecoveryMessage` branches sit behind the `suspended` guard. Checked against `main`, the guard is tighter than that:

- `agentRecoveryMessage` has exactly one non-test caller, `AgentLifecycleNotice`, which returns `null` unless `agent.availability.state === "suspended"`.
- `projectAgentAvailability` returns `state: "suspended"` from exactly one place, always paired with `reason: "agent_suspended"`.

So of its **fourteen** reason branches, one can render. The other thirteen have never been seen by a reader. Two module-private helpers — `runtimeRecoveryMessage` and `providerCliRecoveryMessage` — existed only to build strings for them, and because the branches were assembled as an object literal, both ran on **every** call and had their results discarded.

The other half of the issue is confirmed as reported: `messagingAgentStatusDescription` is a complete status-sentence builder with zero non-test callers.

## Why deleting is safe

The removed sentences are not the only way a reader learns something is wrong. An offline Computer, a runtime that needs installing, a channel that is not connected each surface on the detail page through their own dependency row via `agentComputerStatus`, and each row carries the action that repairs it rather than prose about it. What is deleted is a second, unreachable description of state that is already presented.

The one reachable sentence keeps its exact wording and moves into the catalogue as `agents_recovery_suspended`. An existing test already renders the real notice and asserts that string, so the reachable path stays pinned at the UI boundary — and its passing is what proves the move is byte-identical.

## The tests are why this survived

`agent-list-order.test.ts` called `agentRecoveryMessage` with `state: "ready"` and `state: "not_connected"` — states its only caller can never be in. The branches therefore looked covered while being unreachable, which is the same trap @yuezengwu identified on #474: an assertion that exercises a function directly says nothing about whether the product can reach it.

Those assertions are replaced with the single reading the notice can produce.

## A false comment, corrected

`setup/checks.ts` said Agent settings named a missing CLI through `providerCliRecoveryMessage`. Nothing reachable called that function — its only call site was the unreachable `handoff_unavailable` branch. I wrote that comment during #311 on review guidance and did not verify it; it described a path that did not exist. It now says what is true, which is that onboarding is the only surface using that chain.

One catalogue key, `agents_computer_not_bound_detail`, was orphaned by the deletion and is removed; the `..._action` key it sat beside is still used by `agentAvailabilityRecovery` and stays.

## Testing at `74e3746e`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **71 files / 606 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 1 improvement, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 93.24%) |

No live QA proposed: the only reachable rendering is unchanged, and an existing full-render test asserts it.

## If you disagree with the direction

The alternative to deleting is widening `AgentLifecycleNotice` so those states do render, which is a product decision rather than a cleanup. I took the issue's stated intent — delete rather than maintain — but the thirteen sentences are recoverable from history if the notice should grow instead.


yuezengwu (approved):
Exact-head review at `74e3746ec6919d702b1cefbc22c8041d6cc8d7e2`: no blocking findings.

I confirmed the deleted builders have no production readers. `agentRecoveryMessage` has one non-test caller, `AgentLifecycleNotice`, whose suspended guard is paired exclusively with `agent_suspended` by `projectAgentAvailability`; `messagingAgentStatusDescription`, `runtimeRecoveryMessage`, and `providerCliRecoveryMessage` had no independent reachable surface. The list priority/stable-order path, working elapsed detail, Computer/Messaging dependency rows, and setup/check mappings remain unchanged. The new suspended copy is synchronized across the English and Chinese catalogues and the real rendered notice test still pins the accessible suspended surface.

Validation on the exact head: `git diff --check`; root `pnpm build`; `pnpm --filter @opentag/web typecheck`; `pnpm --filter @opentag/web test` (71 files / 606 tests). GitHub CI, patch coverage, Browser Smoke, and Node compatibility are also green.

liuchao-001:
## Conflict resolved at `65d947ef` — new head, so @yuezengwu's `74e3746e` approval does not carry

Main moved **18 commits** under this branch, including #473 unifying Agent creation and setup, which deleted the entire `onboarding-v2` and `agent-creation` surface this branch was written against.

**The suite is now 58 files / 560 tests rather than 71 / 606.** That is main's deletion, not a loss in this merge — the working tree's test files match `origin/main` exactly, and this branch's delta only ever touched `agent-presentation.ts` and its test. Worth stating plainly because a reviewer comparing to my earlier numbers would otherwise have to work that out.

## The premise was rechecked, not assumed

A deletion PR is only as good as its reachability claim, and that claim was made against a `main` that no longer exists. Re-verified against the current one:

- `agentRecoveryMessage` still has exactly one non-test caller, `AgentLifecycleNotice`.
- That notice still returns `null` unless `agent.availability.state === "suspended"` (`agent-detail-page.tsx:242`).
- `projectAgentAvailability` still returns `suspended` from one place, paired exclusively with `agent_suspended`.
- No new reachable caller of `messagingAgentStatusDescription`, `runtimeRecoveryMessage` or `providerCliRecoveryMessage` appeared in those 18 commits.

So the thirteen unreachable branches are still thirteen unreachable branches.

## How the conflicts were resolved

Import lists in `agent-presentation.ts` and `agent-presentation.test.ts`, plus both `agents` catalogues. Each was rebuilt from main with only this branch's delta re-applied, so:

- main's new `AgentSetupLink` / `agentSetupLink` and `AgentAvailability` imports are kept — `AgentAvailability` is used four times by main's new code, so removing it (as this branch had) would have broken the build;
- main's new `agentComputerStatus` test import is kept;
- the catalogues keep every key main added, plus this branch's one addition (`agents_recovery_suspended`) and one removal (`agents_computer_not_bound_detail`).

## One change since the approved head

The `setup/checks.ts` comment narrated its own history ("this comment previously said…"). That belongs in the commit message, not in the file, so it now states only what is true: onboarding is the only surface using that chain.

## Testing at `65d947ef`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **58 files / 560 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 4 improvements, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 91.81%) |

Coverage reads lower than the 93.24% on the previous head because main's refactor changed the denominator, not because this branch lost covered lines.

@yuezengwu @bestony — re-requesting on this head.

liuchao-001:
## Follow-on cleanup at `0fb2bd4b` — this PR was orphaning copy of its own

While checking a loose end from #459 I found that **this PR creates the defect it exists to remove**, one level up.

`providerCliRecoveryMessage` was the only reader of `SETUP_COPY.messaging`. Deleting the function leaves the whole block as copy nothing renders — `preparingCli`, `checkingCredentials`, `needsAttention`, `cliUnavailable`, and the Slack strings whose consumers #473 already deleted. Only `SETUP_COPY.connect` is still read, by `setup/components.tsx`.

Deleting the reader while leaving its copy would have been the same "second, unreachable description" this PR argues against, so the block goes with it.

## The guard caught it, which is the part worth reporting

Three of those strings were exempted in the #459 brand-naming guard, with the reason "pre-i18n copy that never moved into a catalogue". The moment the literals stopped existing, the guard's third assertion failed:

```
setup/copy.ts: "Install OpenTag in your Slack workspace. We'll take you to Slack and bring you back."
setup/copy.ts: "Add to Slack"
setup/copy.ts: "Waiting for you to finish in Slack…"
```

That assertion exists precisely so an exemption cannot outlive its reason, and this is the first time it has been the thing that noticed. The exemption is removed, so `setup/copy.ts` is now guarded like any other file rather than carrying a standing exception.

`messagingProviderLabel` stays imported there — the connect-command comment still names both CLIs through it, which I checked rather than assuming the import went with the block.

## A separate finding, not fixed here

That exemption cited **#241** as the tracking issue, but #241 was closed **COMPLETED on 2026-08-31**, a day before I wrote the line. Its acceptance criteria include "translations for all user-facing Control Panel content", and `SETUP_COPY.connect` — which *is* still rendered — remains English-only in a `zh` session. So either #241 closed with a known gap or that copy was missed. Out of scope here; raising it so it is not lost.

## Testing at `0fb2bd4b`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **58 files / 560 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 4 improvements, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 91.81%) |

liuchao-001:
## Correction to my previous comment: that copy is not rendered either

In the comment above I wrote that `SETUP_COPY.connect` "**is** still rendered" and therefore left an English-only string in a `zh` session. I asserted that from the fact that `setup/components.tsx` reads it, without checking whether those components are themselves reachable. They are not:

| `setup/` export | non-test consumers outside `setup/` |
|---|---|
| `CommandBlock`, `QrCode`, `useRemaining`, `WAITING_LINE`, `readConnectCodeVerdict`, `formatRemaining` | live |
| `ConnectStatus`, `Countdown`, `CheckLine`, `CHECK_COPY`, `SETUP_COPY` | **0** |

`SETUP_COPY.connect` is read only by `ConnectStatus` and `Countdown`, and nothing outside `setup/` imports either. So after this PR removes the `messaging` block, the whole of `SETUP_COPY` — and `CHECK_COPY` with `CheckLine` — is unreachable.

Two consequences:

1. **There is no i18n gap, so #241 closing as COMPLETED looks right.** Untranslated strings that nothing renders are not a shipped-UI problem. I withdraw the suggestion that it closed with a gap.
2. **There is a dead cluster in `setup/`** — `CheckLine`, `ConnectStatus`, `Countdown`, `CHECK_COPY`, `SETUP_COPY` — of exactly the kind #336 is about, left behind when #473 deleted the surfaces that used them.

I am **not** expanding this PR to cover it. This PR's claim is about `agent-presentation.ts`, it has already been reviewed once at that scope, and a five-symbol module sweep deserves its own reachability argument rather than riding along. Filing it separately.

The state of this PR is unchanged: it deletes the `messaging` block it orphaned, and nothing else in `setup/`.

yuezengwu (approved):
Exact-head review at 0fb2bd4bd1c47ff1545f73b76493d7c8cb0ed221: no blocking findings.

I re-reviewed the complete diff after the main merge and the follow-up orphan cleanup. The reachable lifecycle notice is still limited to the suspended state, projectAgentAvailability still pairs that state exclusively with agent_suspended, and the removed status builders have no remaining production readers. The merge preserves current main Agent Setup routes, dependency actions, and catalogue additions. SETUP_COPY.messaging has no remaining reader after providerCliRecoveryMessage was removed, so deleting that stale copy and its brand-naming exemptions is consistent.

Validation on the exact head: git diff --check; root pnpm build; pnpm --filter @opentag/web typecheck; pnpm --filter @opentag/web test (58 files / 560 tests); pnpm check; web unit coverage 91.81%. GitHub CI, patch coverage, Browser Smoke, Node 22/24/26 compatibility, container/CLI smokes, and ownership are green.

Issue #470, opened 2026-09-02
fix(runtime): admit healthy first Turns after initial Provider CLI readiness completes


Gandy2025:
## Resolution (2026-09-09)

Resolved by [PR #572](https://github.com/first-tree-ai/opentag/pull/572), merged into `main` at 2026-09-09T02:02:37Z as commit [`8cc14e38`](https://github.com/first-tree-ai/opentag/commit/8cc14e38786b53fdd9c876167eecf2fc3f40e890). This issue was automatically closed as completed by the merge.

Healthy first Turns now wait for the existing initial Provider CLI check and proceed when the exact selection remains healthy and unchanged. Actual selection drift, cancellation, and per-Session plan ownership remain fenced. Deployed behavior with real provider accounts has not been verified by this task.

## Pre-fix investigation (2026-09-08)

Before PR #572, healthy first Turns could fail before model startup while a newly started Client was completing its initial Provider CLI inspection. The fix allows a successful initial inspection to admit the waiting Turn while preserving the existing checks for actual CLI selection drift.

Pre-fix regression baseline: main `34ae13da515e65fca17a20d5c2603d2eab1892ac`.

### What changed since the original report

The original Server-side TTL rejection path no longer exists: `ImBindingService.issueRuntimeCredentialGrant` no longer reads Provider CLI readiness. Deterministic regression using the real services, registry, reconcile owner and isolated PostgreSQL passed 80 assertions, plus 6 relevant existing tests. At credential observation age 120,001 ms (artifact still fresh), first Channel/Thread grants and concurrent grants during `checking` succeed. The refresh coalesces into one validation grant. Stale generation/connection results and invalid authorization/material remain fenced.

The historical observations below remain useful evidence for older releases. Their proposed Server grant barrier is superseded for this fix. Deployment versions and online behavior were not verified by this local regression.

### Reproduced Client failure before PR #572

1. Start a new Client reconciler with an already installed, healthy CLI and no in-memory accepted selection yet.
2. Begin the initial Provider CLI requirement inspection and hold its completion.
3. Deliver two concurrent Turns with valid synthetic credential grants. Both join the existing inspection.
4. Complete inspection successfully without changing the CLI path, fingerprint, version or selection generation.
5. Both Turns fail with `credential_unavailable`, local `errorCode=selection_invalid`, and `executionEffects=not_started`. The model entry point is never called.
6. The next Turn succeeds without changing credentials or the CLI.

[...]

PR #477, opened 2026-09-02
feat(onboarding): make provider CLI setup agent-first

yuezengwu:
## Summary

This lands the P0 onboarding foundation for an Agent-first first setup while preserving the existing onboarding order.

```text
install.sh
  install or upgrade OpenTag only
  ↓
targeted opentag connect
  connect to the Server
  bind the explicit OpenTag Agent / Computer
  start the daemon
  inspect/install both Lark and Slack CLIs in the foreground and return repair actions
  ↓
daemon
  during first setup: periodically inspect/report both Provider CLI states only
  after IM binding: converge the selected Provider CLI when needed
  ↓
web onboarding step 3
  choose and bind Lark or Slack in the existing position
```

Key behavior:

- The Web command is one copyable shell command: install/upgrade OpenTag, then run top-level `opentag connect` with the Server URL and a single-use targeted code.
- Targeted codes use compact `otcc_<43 base64url characters>` values containing the Agent UUID and 128 bits of randomness. The Server authenticates the hashed code first and then recovers the target from the submitted code. Legacy dotted codes remain accepted.
- Foreground targeted `opentag connect` is the authoritative Provider CLI setup command. It handles Lark and Slack independently and in parallel so the invoking Agent can observe failures, run exact idempotent repair commands, and verify the result.
- If an already-active Integration makes daemon convergence overlap foreground setup, either lock ordering now waits and re-observes the terminal Provider state instead of projecting `operation_in_progress` as failure.
- Daemon prewarm is inspect/report only during first setup. Its periodic refresh performs real reinspection with per-provider single-flight; after IM binding it may converge only the selected provider.
- Active-Integration convergence treats a live Provider CLI lock as transient: the client waits and reinspects, while the Server re-dispatches a retained unavailable observation after its retry timer clears instead of requiring a reconnect.
- `connect --json`, `provider-cli inspect --json`, `provider-cli ensure --json`, Commander validation failures, and unhealthy `doctor --json` use stable machine-readable envelopes and exit policy. Success is written to stdout; failures are written to stderr with stable error codes, safe partial results, retryability, and exact `nextActions`.
- The Web onboarding UI reports Lark and Slack status independently. A restored unbound Agent can select an existing Computer or copy a new targeted command.
- Resume retries replace the prior subject state rather than layering onto it. Destructive runtime recovery always derives the Agent id from the recovery target currently shown to the reader.
- If a newly created Agent has the wrong runtime, recovery is explicit: suspend/delete that Agent after confirmation, retain Computer records, and choose again. No silent runtime mutation is performed.
- Provider authorization is separate from local CLI installation. The existing third onboarding step still owns IM selection/binding.
- This change uses existing persistence only: there is no database schema change or migration.

## Testing

Passed on `785ba818` after merging `origin/main@37a86739`:

- `pnpm check`
- `pnpm build`
- `pnpm typecheck` — 8/8 tasks
- `pnpm test` — 8/8 workspace tasks; CLI 340/340, Client 896/896, Server 700/700, Web 714/714
- blocker regressions — CLI 35/35 and Client reconciler 22/22
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 347/347 with 100% statements, branches, functions, and lines
- source-channel packed CLI smoke — installed artifact verified through `opentag-dev`
- pre-push hook — formatting, typecheck, lint, and repository checks
- migration drift check — exactly 38 existing migrations; no new migration
- `git diff --check`

## Breaking changes

Malformed CLI invocations now consistently return the documented usage exit code 2 instead of the previous generic exit code 1. Scripts that special-case `$? -eq 1` for invalid or unknown options must accept exit 2. The old nested `computer connect` entry point and legacy dotted targeted codes remain compatible.

## Checklist

- [x] No database schema change or migration is introduced.


liuchao-001 (commented):
Reviewed at head `9fc883a6` against `origin/main@5d2a0b21` in a fresh worktree, plus the Context Tree nodes this touches (`opentag/system/im-integrations.md`, `opentag/system/agent-session-computer-model.md`).

The connect-code change is the highest-stakes part and it holds up. I traced it specifically:

- The hashed code row stays the only authority. `targetAgentIdFromConnectCode` runs on `input.code` only *after* the `tokenHash` lookup succeeds, so the embedded target is derived from an already-authenticated code and cannot be forged.
- The two encodings cannot collide: untargeted codes are `generateSecret(24)` → 32 base64url characters, targeted ones are 32 payload bytes → 43. The canonical round-trip check (`payload.toString("base64url") === encoded`) closes the padding/alias variants.
- Ownership and lifecycle are re-checked under `FOR UPDATE` at both issue and exchange, `create` requires `computerId IS NULL` in the UPDATE predicate, `repair` requires the Agent to already be on the target Computer, and every failure throws inside the transaction so the Computer insert rolls back.
- 128 bits of randomness is still fine, and the Agent UUID is not secret material.

Capability negotiation for `provider-cli:prewarm` is correctly backward-compatible (client offers mirror server offers; the server gates on `supportsCapability`, the client re-gates in `#handlePrewarm`). The readiness-freshness plumbing also works out: refresh fires at `RUNTIME_CLIENT_CAPABILITY_TTL_MS / 2` and re-arms the published entries before their 60s expiry, so `refreshPublishedImCliReadiness` genuinely prevents a lapse rather than being decoration. Dropping `ensure_daemon_service` from `install.sh` is safe for upgrades because the service definition points at the version-independent shim (`docs/portable-release.md`).

Two things should be fixed before this merges, and four smaller ones are your call.

---

## 1. Resume forgets an existing Computer when every active Agent is unbound — dead-ends the step

`apps/web/src/onboarding-v2/server-backend.ts`, `selectAccountResume` / `applyResumeSelection`

The old resume had two branches: `active.length === 0` restored a connected Computer that had no Agent yet, and `resumeBlocked` handled Agents that had no Computer. This PR keeps the first and deletes the second — but the replacement never restores a Computer in the unbound-Agent case:

```ts
const bound = active.flatMap((agent) => (agent.computer ? [{ agent, computer: agent.computer }] : []));
const selected = bound.find(...) ?? bound[0];
const agent = selected?.agent ?? active[0];   // adopts the unbound Agent
...
if (!selected.boundComputer) return undefined; // returns before touching computerId / setSelectedComputer
```

`knownComputers` is `selectedComputer ? [selectedComputer] : []`, and the availability poll bails on `if (!selectedId) return`, so nothing recovers it later. `ComputerStep` then renders the `intent: { mode: "create" }` branch only.

Failure scenario: an Account that has an active Agent with no Computer **and** an enrolled Computer — exactly the state `resumeBlocked` existed to describe, so pre-#477 accounts can be sitting in it right now. The reader is handed a fresh create command. On the same machine with its OpenTag home intact, `#createComputer` hits `computers_current_installation_id_unique` and the exchange fails with `COMPUTER_IDENTITY_CONFLICT` (409); the Web side just keeps polling a code that never redeems, shows expired, and reissues. On a wiped home it succeeds and enrolls a second Computer row for one physical machine — which `agent-session-computer-model.md` calls out as the wrong answer ("conflating them answers an unreachable Computer by enrolling a duplicate").

The unbound-Agent case needs to keep carrying the Account's Computer, so the step can offer the repair intent (which is what atomically binds the Agent) instead of a create that the Server will refuse.

## 2. Once the Agent is created there is no exit from the Computer step

`apps/web/src/onboarding-v2/page.tsx:207`, `:336`, `:395`

`const resumed = backend.agent !== undefined` now flips to `true` the moment `submitAgent` succeeds, not only on a genuine resume. Two consequences land together:

- `onBack={resumed ? undefined : backToAgent}` removes Back from `ComputerStep`.
- `{reviewMode || backend.agent ? null : <Button onClick={startOver}>}` removes Start over.

So a reader who picks Claude Code, gets the Agent created, and then connects a machine that only has Codex sits on a step whose Continue is gated by `!ready || !passed`, with no Back, no Start over, and an Agent whose runtime provider is immutable. The comment this PR deleted from `flow.ts` named this exact hazard as the reason creation was deferred:

> which also means a failed check never strands a half-built Agent on an immutable runtime the user cannot run

I understand why creation had to move — the targeted connect code needs `agent.id` to exist, so agent-before-computer is forced by the design, not a slip. But the protection was removed without a replacement. Something has to stay reachable: keep Start over (with a copy change explaining the Agent is kept), or let Back re-open the Agent step with a real runtime change, or at minimum have the failed-check copy say the runtime is now fixed and name the recovery. Note that today Back would be silently useless anyway — `createAgent` returns early on `creationRef.current !== "idle"`, so a changed runtime would never reach the Server.

---

### 3. `#reconcileProvider` dedupes by provider and drops the caller's mode

`packages/client/src/runtime/provider-cli/reconciler.ts`

```ts
const existing = this.#providerJobs.get(provider);
if (existing) return existing;      // the in-flight job's mode wins
```

`#prewarmProvider` asks for `"auto"`; `#handleRequirement` and `readySelectionForRun` ask for `"managed-only"`. Both frame kinds are dispatched from the same `onComputerRegistered` (`#dispatchSetupPrewarm` then the requirement fan-out), and `#trackFrame` doesn't serialize frames, so overlap is ordinary rather than exotic — and prewarm holds its job for the length of two real downloads. A requirement that asked for reviewed managed artifacts can therefore be answered by a job that selected a detected external CLI on PATH.

I checked this against `im-integrations.md` before flagging it: `auto` *is* the sanctioned first-setup policy ("detects a compatible existing CLI or installs an OpenTag-managed copy"), and prewarm only fires while `accountInFirstSetup`, so I don't think there's a live trust hole today. It's still a latent one — the mode is a caller's requirement, and silently substituting it is the kind of thing that becomes a real finding the first time prewarm's gate widens. Keying `#providerJobs` on `provider + mode`, or refusing to reuse a weaker-mode job, would settle it. Nothing in `provider-cli-reconciler.test.ts` covers the overlap.

### 4. Untargeted `connect` reports `reason: "not_requested"`

`apps/cli/src/commands/computer/connect.ts`

`shouldPrepareProviderClis` requires `result.agentId !== undefined`, but `skippedProviderClis` only knows two reasons:

```ts
const reason = options.start === false ? "daemon_not_started" : "not_requested";
```

An Agent redeeming an untargeted code gets `providerClis: { status: "skipped", reason: "not_requested" }` and concludes the operator opted out, when in fact the code simply named no Agent. This is the one contract the PR exists to make machine-readable, so it's worth a third reason (`no_target_agent`). The command's own description — "Connect this Computer, start the daemon, and prepare the Lark and Slack CLIs" — also promises the prep unconditionally.

Separately: gating the CLI-side prep on "the code named an Agent" rather than on the Server's own first-setup verdict couples two different questions. The daemon uses the right gate (`accountInFirstSetup`); the CLI infers one. It happens to be equivalent today only because the onboarding page is the sole issuer of targeted codes and is itself behind the setup gate.

### 5. `doctor --json` next-action reasons are prose, not codes

`apps/cli/src/core/diagnostics/doctor.ts`

```ts
check.command ? [{ checkCode: check.code, command: check.command, reason: check.detail }] : []
```

`check.detail` is human copy ("not installed", "Integration CLI installation could not be determined"), while `ProviderCliNextAction.reason` in the same PR carries codes (`install_incomplete`, `probe_failed`). Two `nextActions` shapes with the same field name and different content classes is an easy thing for an Agent to get wrong.

### 6. Two small ones

- `providerFromCheckCode` falls back to `"slack"` for any code without `.feishu.`. Both call sites pass literals today, so it's unreachable — but it's a silent wrong answer rather than a loud one if a third provider ever lands.
- `providerCliCanAutoRepair(undefined)` returns `true`, and both call sites default the reason before calling (`?? "unavailable"`, `?? "not_installed"`), so that branch is dead.

Gandy2025 (changes requested):
Reviewed the immutable range `5d2a0b21006dfdff1a5d82c16afe959d7fd1cf31...9fc883a6ddfe151853f79dd4a4431c9b55799b36` as a high-risk Standards + Spec review.

The targeted connect-code security boundary holds up: the hashed row remains authoritative, the compact and legacy encodings are non-colliding, the 128-bit random component is sufficient, and Agent ownership/lifecycle is rechecked under row locks at issue and redemption. Capability negotiation is also rolling-compatible. CI is green at this head apart from the expected ownership approval gate.

I independently validated the two blocking Web findings in liuchao-001's review (unbound-Agent resume losing existing Computer inventory, and no usable exit after an immutable runtime is created), so I am treating both as blockers without duplicating their full write-up. One correction matters when fixing the first: merely restoring the Computer and showing today's repair intent is insufficient. `machine-auth-service.ts:423-430` requires a repair target Agent to already be bound to that Computer, and `:448-459` never binds an unbound Agent in repair mode. The repair/rebind semantics therefore need an explicit atomic design; otherwise the proposed UI path still fails at issuance or redemption.

There are three additional blockers:

## 1. [P1][Spec] First setup has two concurrent install owners, so a healthy download can become a false partial failure

`apps/cli/src/commands/computer/connect.ts:98-111` starts foreground `ensure --provider all` immediately after starting/restarting the daemon. The daemon registers at the same time, and `packages/server/src/runtime/provider-cli-reconcile-owner.ts:148-152,179-203` sends its own two-provider prewarm; the client prewarm also calls `ensure(..., { mode: "auto" })`.

Those two processes contend for the same per-provider lock, but `packages/client/src/runtime/provider-cli/lock.ts:77-106` waits only about five seconds before returning busy. A normal managed-artifact download can exceed that. The loser then maps the live install to `operation_in_progress` / `unavailable`: foreground exits 1 and reports `needs_attention`, or the daemon publishes an unavailable status while the other process is still succeeding. Because these are the normal first-setup paths, this is not a theoretical concurrency edge.

Use one authoritative setup installer, or make the second path wait for and observe the in-flight operation's terminal result. A live lock cannot be projected as a terminal Provider failure.

## 2. [P2][Spec] The daemon renews cached readiness forever instead of continuously inspecting

`packages/client/src/runtime/provider-cli/reconciler.ts:89-94` only replays `#imCliPublished`; `packages/client/src/runtime/client-runtime-composition.ts:420-430` calls that replay after every capability refresh. No `inspect` occurs after the initial prewarm.

If a ready CLI is deleted or drifts, or an unavailable CLI is repaired using the returned command, the daemon keeps extending the old fact's TTL and Web keeps showing the stale value until a restart or an unrelated active-Integration reconcile. This violates the stated "continuously inspect and report" behavior and makes the advertised repair/retry loop appear ineffective.

The periodic path needs a bounded, single-flight reinspection per published Provider (and checking -> new terminal publication), rather than TTL renewal of the old observation.

## 3. [P1][Standards + Spec] The new machine-readable commands do not have one coherent JSON contract

The repository contract in `apps/cli/README.md:13-25` requires the common `{ok,result}` / `{ok:false,error}` envelope and sends failures to stderr. The PR description instead promises one machine-readable stdout document. The implementation currently mixes both models:

- `provider-cli ensure/inspect --json` return provider fields or `results` at the root (`ensure.ts:106-112`, `inspect.ts:107-110`), bypassing the documented result envelope.
- Their validation failures write the only JSON document to stderr (`ensure.ts:166-184`, `inspect.ts:116-134`), as do thrown provider and connect failures through the shared presenter.
- A partial `connect --json` result writes `{ok:false,result,...}` to stdout without an `error` (`commands/computer/connect.ts:139-157`), which is neither documented shape.

An Agent cannot consume one stable schema/stream across success, partial success, usage failure, and transport failure. Reconcile the PR promise with the authoritative CLI contract, document the chosen partial-success envelope, and make all three commands and all terminal paths conform to it. Add contract tests that assert both the exact document shape and which stream is used.

## Re-review scope

Please re-request review after a new exact head covers: the two existing Web blockers; unbound-Agent/existing-Computer atomic placement; foreground/daemon Provider setup ownership; real periodic Provider reinspection; and the unified JSON contract. The connect-code cryptographic/transactional conclusions can carry forward if that boundary is untouched; changes to machine-auth placement semantics require focused re-review of issue/exchange rollback and ownership checks.

liuchao-001:
Checked @Gandy2025's review against the code at the same head. Agreeing with the correction to my finding 1, confirming the three new blockers with the mechanisms I could verify, and narrowing one of them so the fix does not overshoot.

**My remedy for finding 1 was wrong.** I wrote that the step should offer "the repair intent (which is what atomically binds the Agent)". Repair does not bind an unbound Agent, and Gandy's citations are exact:

- `assertConnectTargetAgent` sets `expectedComputerId = input.targetComputerId` for repair, so an unbound Agent (`computerId === null`) fails `agent.computerId !== expectedComputerId` and issuance returns 409 `AGENT_LIFECYCLE_CONFLICT`.
- `bindConnectTargetAgent` sets `expectedComputerId = input.computerId` for repair, so an unbound Agent throws `AUTH_INVALID_CODE`; and `if (agent.computerId === input.computerId) return;` means repair mode never reaches the UPDATE at all. Repair is a no-op for binding by construction.

So the unbound-Agent + existing-Computer case needs explicit atomic placement semantics on the Server, not a UI re-route to today's repair path. The defect I reported stands; the fix I suggested does not.

**Blocker 1 (two concurrent install owners) — confirmed, with the chain.** `lock.ts` defaults are `maxAttempts = 50` and `retryDelayMs = 100`, so ~5s of retry before `ProviderCliLockBusyError`, which a managed-artifact download can exceed. `manager.ts:339-340` maps that error to `failed("operation_in_progress")` — a returned failure, not a throw. Foreground: `runProviderCliEnsure` sees `ok: false`, exits 1, and `connect` prints `needs_attention` over a healthy install. Daemon: `#runProvider` re-inspects after the failed ensure, still isn't ready, and publishes `unavailable`.

One detail in the loser's favour: `operation_in_progress` is not in `PROVIDER_CLI_MANUAL_REASONS`, so a retry `nextAction` is emitted. The advice is right; the verdict it is attached to is wrong.

This also subsumes my finding 3 in the more important direction. I framed mode-blind job dedupe as an in-process concern, but the foreground CLI and the daemon are separate processes, so `#providerJobs` cannot mediate between them at all — the file lock is the only thing standing between them, and it gives up after five seconds.

**Blocker 2 (no periodic reinspection) — confirmed, and it does not contradict what I wrote.** I said the readiness plumbing "works out", scoped to freshness: refresh at `RUNTIME_CLIENT_CAPABILITY_TTL_MS / 2` does re-arm entries before their 60s expiry. Gandy's point is about the content, not the clock — `refreshPublishedImCliReadiness` iterates `#imCliPublished` and re-sends the cached observation with no `inspect` call, so a deleted or repaired CLI keeps its old fact renewed indefinitely. Both are true; the PR's "continuously inspect and report" is only the second one, and it isn't implemented.

**Blocker 3 (JSON contract) — agreed on two of three bullets; the third would send the fix the wrong way.** `apps/cli/README.md` requires `{"ok":true,"result":...}` / `{"ok":false,"error":{...}}` and says failures are written to stderr. So:

- Real violation: `ensure`/`inspect --json` put provider fields or `results` at the document root instead of under `result`.
- Real violation: `connect --json` writes `{ok:false, result, nextActions}` to stdout with no `error`, and `presentServiceFailure` likewise writes a failure document to stdout.
- Not a violation: `ensure`/`inspect` writing their usage-failure JSON to **stderr** is exactly what the README mandates, and it is an improvement on `main`, which wrote bare usage text there. The mismatch is the PR description promising "one machine-readable stdout document" — that promise is what should change, not the stream. Please do not move failure documents to stdout to satisfy it.

Also updating my own CI note: the checks were still running when I reviewed and are now green at this head apart from `ownership-gate`, which is waiting on @bestony.

Gandy2025:
Agreed on the narrowing. My blocker is the inconsistent contract, not a request to move failure documents to stdout.

For the fix and re-review, `apps/cli/README.md` should remain authoritative: successful JSON goes to stdout under `{ "ok": true, "result": ... }`, and terminal failures go to stderr under `{ "ok": false, "error": ... }`. The PR description should stop promising stdout for every outcome. `provider-cli ensure/inspect` need the common success envelope, while `connect` partial failure and `presentServiceFailure` need to conform to the documented failure envelope and stream. If partial connection state must be preserved for recovery, give it an explicitly documented structured place inside that failure contract rather than inventing a third top-level shape.

The requested contract tests should therefore assert both exact schema and stream against the README model. Thanks for catching that ambiguity.

yuezengwu:
Implemented the requested blocker fixes on exact head `07b82854`:

- **Provider CLI ownership:** foreground targeted `opentag connect` is the sole installer during first setup. Daemon prewarm only inspects/reports; after IM binding, daemon may converge the selected provider.
- **Fresh readiness:** daemon periodic refresh now performs real reinspection, with per-provider single-flight instead of replaying cached state.
- **Agent-readable CLI contract:** connect/inspect/ensure share the documented JSON envelope. Success goes to stdout; failure goes to stderr with stable codes, safe partial state, and exact idempotent `nextActions`. Doctor remains read-only and fails loudly on invalid Provider state.
- **Restored unbound Agent:** Web now offers existing Computer selection when available, plus a new targeted connect command. Choosing a Computer is persisted and verified by tests.
- **Wrong runtime recovery:** a newly created Agent can be explicitly suspended/deleted after confirmation and then reselected. Computer rows are retained; runtime is never silently mutated.
- **Smaller findings:** removed the misleading ordinary Back path after Agent creation, aligned connect-description tests, sorted i18n keys, and strengthened the Web fixture so it cannot mask a wrong-machine binding.

There is **no database schema change and no migration**. The implementation only uses existing Agent/Computer/connect-code persistence.

Exact-head validation passed: `pnpm check`, `pnpm build`, `pnpm typecheck`, full `pnpm test` (8/8 workspaces; CLI 318/318, Web 620/620), migration drift, `git diff --check`, and the pre-push hook.

Please re-review this head; previous review state refers to `9fc883a6`.

liuchao-001 (approved):
Re-reviewed at exact head `07b82854`. All five blockers are genuinely resolved. Approving, with one follow-up that belongs in the Context Tree rather than in this diff.

I checked the fixes rather than the summary:

**My 1 (resume loses the Computer).** `resumeBlocked` is back, and the screen now renders `AgentComputerChoice` with the onboarding adapter. Binding goes through `rebindAgentComputer`, which sidesteps the connect-code repair semantics entirely — the right call, since `assertConnectTargetAgent` would 409 an unbound Agent on repair. The connect offered alongside it is `intent: { mode: "create" }`, which is the branch whose predicate is `computerId IS NULL`, so it matches an unbound Agent.

I also chased the redundant bind this introduces: with a targeted create code the Server already binds the Agent at redemption, and the Web then rebinds to the same Computer. `rebindById` short-circuits on `!changesAgent && !changesPlacement` and returns `changedComputer: false`, so that is an idempotent success, not a spurious bind failure after a successful connect.

**My 2 (no exit after Agent creation).** `AgentResetControl` gives an explicit confirm → suspend → delete → re-read path, on both the computer step and the resume-blocked screen. `resumed` is now `agentRestored === true` instead of "an Agent exists", so a freshly created Agent no longer poisons `destinationConfirmed` / `draftConfirmed`. Removing Back outright was the right call over repairing it — `createAgent`'s `creationRef` guard made it silently useless, so it promised a correction it could not perform.

**Gandy 1 (two concurrent installers).** Resolved at the root by taking the daemon out of the install path: `#prewarmProvider` now delegates to `#refreshImCli`, which only inspects. I confirmed `manager.inspect` is genuinely read-only — it reads the selection record, launcher and global command and never enters `withProviderCliLock` — so the two processes no longer contend for the ~5s lock, and a healthy foreground install can no longer be published as `unavailable`.

**Gandy 2 (no reinspection).** `refreshPublishedImCliReadiness` now performs real per-provider inspection with `#inspectionJobs` single-flight, and the composition awaits it. This is the behaviour the PR claimed and previously did not have.

**Gandy 3 (JSON contract).** All three commands route through `presentCommand`: success `{ok:true,result}` on stdout, failure `{ok:false,error,result?}` on stderr, with `apps/cli/README.md` extended to document the partial-result field. This landed the way we converged on it — the failure stream was left alone rather than moved to stdout to satisfy the old description.

My smaller findings are fixed too: `no_target_agent` as a distinct skip reason, `nextActions[].reason` now carrying `check.code` instead of prose, and `providerFromCheckCode` failing loudly instead of defaulting to `slack`. The PR description now matches the implementation.

---

Three things left, none blocking this merge.

### 1. This head changes a durable decision the Context Tree still records the other way

`opentag/system/im-integrations.md` currently states, in Rationale:

> The daemon, not an interactive shell or the Web page, owns the environment in which Local Agents execute. During first setup, the active setup run authorizes that daemon to prepare both official provider CLIs on the explicitly connected Computer before the user chooses an Integration.

and in Constraints:

> The daemon detects a compatible existing CLI or installs an OpenTag-managed copy without altering third-party files.

This head inverts exactly that: the foreground command installs, and the daemon only observes. Code is ground truth, so the node is drifted and needs a source-backed Tree change — the surviving rationale here (a foreground installer is what lets the invoking Agent observe failures and run repair commands) is precisely the kind of thing that is expensive to reconstruct later.

There is a second-order effect worth deciding explicitly rather than inheriting. `accountInFirstSetup` / `shouldPrewarmOfficialProviderClis` no longer gates any installation — it now gates observation only. So the constraint "Provider installation code must activate through a Server-authorized first-setup requirement or the verified Integration boundary" is currently satisfied by the CLI inferring its own authorization from the code having named an Agent. That is defensible for installing official CLIs on a machine whose owner just ran the command, but it is a different answer from the one the node gives. Say which one is intended.

I can open the Tree PR against `opentag/system/im-integrations.md` once this merges, so it reconciles against merged source truth.

### 2. `retryability` is hardcoded even when the cause is retryable

`PROVIDER_CLI_SETUP_INCOMPLETE` and `PROVIDER_CLI_NOT_READY` are constructed with `retryability: "never"` regardless of which diagnostic produced them. `operation_in_progress` is still reachable — `lock.ts` is unchanged and still gives up after ~5s, and post-binding daemon convergence can overlap a manual `provider-cli ensure`. In that case the envelope tells an Agent never to retry while the per-result `nextAction` tells it to run ensure again. `ErrorRetryabilitySchema` already has `"backoff"`, which is the honest answer for a live lock.

### 3. Vestigial mode parameter

With prewarm no longer reconciling, both remaining `#reconcileProvider` call sites take the `managed-only` default, so nothing passes `"auto"` any more. The `${provider}:${mode}` key is now purely defensive. Worth keeping, but it should not be read as a live path.

---

**Merge gate:** CI has not run on `07b82854` yet — only `ownership-gate` has reported, still needing @bestony on 55 files. The exact-head evidence in the description is local, so this still needs a green CI run at this head plus the remaining reviewers before it goes in. My approval does not stand in for either.

Gandy2025 (changes requested):
Re-reviewed the immutable range `54905b6bf51e80faadc8911ea4dbfd799c813829...07b828545ef5fea5b161e1a3e55aaa694bd442ba` as a proportional high-risk Standards + Spec review.

The previous blockers were addressed in substance: the setup prewarm path now inspects instead of installing; periodic refresh performs real single-flight inspection; the normal action paths use the common command presenter; the unbound-Agent screen exposes Computer placement; and a newly created Agent has an explicit destructive runtime-recovery path. The build passes, as do the directly affected Client (18), CLI (87), and Web (79) tests locally. The connect-code security conclusions carry forward because that boundary is unchanged.

I cannot approve this head yet. Independent review of the adjacent boundaries found four remaining blockers, which I validated against the exact head.

## 1. [P1][Spec] A stale React Query cache can silently bind the resumed Agent to the wrong Computer

`apps/web/src/onboarding-v2/page.tsx:104-109` mounts the shared `AgentComputerChoice`. That component derives `connected` and `sole` from any successful cached Computers query at `apps/web/src/features/agents/agent-computer-choice.tsx:82-84`, then automatically persists a bind at `:130-135`.

The repository QueryClient deliberately uses `staleTime: 0` (`apps/web/src/query/client.ts:40-43`), so mounting an observer exposes stale successful data immediately and starts a background refetch. If the cache contains one Computer but the current Account now contains several, the effect binds the cached Computer before the reader ever sees the current choice. If that Computer was removed, the failed stale target can also hold the surface in its retry branch instead of offering the current inventory or a new connection.

Automatic binding must wait for a successful inventory read performed after this mount (or explicitly refetch and bind from that returned snapshot). Add a regression test that seeds a one-Computer cache while the Server now returns multiple or zero Computers and asserts that no rebind occurs before the fresh result is applied.

## 2. [P1][Standards + Spec] Resume retries can retain a stale Agent and delete a different Agent than the confirmation names

`apps/web/src/onboarding-v2/server-backend.ts:212-234` applies resume selection incrementally. The `none`, `computer`, and unbound-Agent branches do not clear the prior `agent`, `agentRestored`, or `creation` state; `readAccount` at `:245-273` only pre-clears `resumeBlocked`.

A concrete sequence is: resume bound Agent A; the handoff read fails and exposes Retry; another tab suspends/deletes A and leaves active unbound Agent B; Retry selects B and displays `resumeBlocked` for B, but stale `agent` still contains A. `discardAgent` at `:459-461` then chooses `agent?.id` before the visible `resumeBlocked?.agentId`, so confirming deletion for B can delete A instead. This is a destructive target-identity bug.

Treat every resume selection as a complete replacement of mutually exclusive state, and derive deletion from the currently displayed recovery target. Cover bound Agent -> failed handoff -> retry to none/computer/unbound Agent, including the exact delete target.

## 3. [P1][Spec] Active-Integration reconciliation can still race foreground setup and become permanently unavailable

Changing setup prewarm to inspect-only removes the two normal first-setup installers, but it does not make foreground setup the sole installer when the Computer already has an active IM requirement. `apps/cli/src/commands/computer/connect.ts:42-51` starts the daemon before foreground `ensure all`; registration then dispatches active requirements, and `packages/client/src/runtime/provider-cli/reconciler.ts:223-250` still performs managed `ensure` for those requirements. This is reachable when recovering an unbound Agent that already has an Integration, and when an advertised manual repair overlaps post-binding convergence.

The cross-process lock still gives up after about five seconds (`packages/client/src/runtime/provider-cli/lock.ts:77-106`), and `manager.ensure` returns `operation_in_progress`. The daemon converts that live operation to `unavailable`; the Server retries only its bounded 1/2/4-second sequence (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:607-623`). After exhaustion, `unavailable` is retained without TTL (`packages/server/src/runtime/connection-registry.ts:221-230`) and `isReconcileInFlight` treats it as permanently in flight (`provider-cli-reconcile-owner.ts:91-97,719-738`), so later readiness refresh does not recover the active Integration even after the foreground install succeeds.

The losing installer must join/observe the live operation to its terminal inspection, or the protocol must preserve `operation_in_progress` as transient and durably re-dispatch until the artifact reaches a real terminal state. Do not collapse a live lock into non-expiring `unavailable`.

## 4. [P1][Standards + Spec] The advertised common JSON contract still has uncovered terminal paths and contradictory routing metadata

The action-level success/failure envelopes are fixed, but the contract is not yet common across the commands:

- Commander rejects `provider-cli ensure --json` without `--provider`, and rejects an unknown option, before the action. `apps/cli/src/commands/provider-cli.ts:15-21` lets Commander emit its prose, while `apps/cli/src/cli/index.ts:9-13` exits directly for `CommanderError`. I reproduced both on this build: exit 2, human help/prose on stderr, and no JSON document.
- `ensure`, `inspect`, and partial `connect` create `category: "dependency", retryability: "never"` but force exit 1 (`ensure.ts:167-180`, `inspect.ts:124-137`, `commands/computer/connect.ts:165-176`). Shared policy maps `dependency` to exit 3 (`core/command/policy.ts:295-301`), while returned `nextActions` often instruct the Agent to rerun the same idempotent command. Exit code, envelope retry policy, and repair action therefore disagree.
- An unhealthy `doctor --json` still writes `{ok:false,result}` to stdout without `error` (`apps/cli/src/commands/doctor.ts:12-16`), bypassing the README's failure envelope and stream. Either unhealthy diagnostics are a successfully produced report, in which case `ok`/exit semantics must say that, or they are a command failure and must use the common partial-result envelope on stderr.

Handle pre-action Commander failures through one validation envelope, make aggregate category/retryability/exit routing consistent with the provider diagnostics and `nextActions`, and settle the Doctor report semantics. Add exact schema + stdout/stderr + exit-code tests for missing/unknown options, retryable partial failure, non-retryable partial failure, and unhealthy Doctor.

## Re-review scope

Please re-request review after a new exact head covers these four boundaries. The already-validated connect-code, basic foreground/prewarm ownership split, periodic inspection, and normal JSON presentation can carry forward if untouched. CI has not yet run on this head; only the ownership gate is currently reporting, and it remains red.

liuchao-001 (changes requested):
I verified all four of @Gandy2025's findings against this exact head. They hold. My approval was premature — withdrawing it with this review.

Three of the four are in code I read closely and signed off on, so the confirmations below are mine, not a relay.

## 2 — confirmed, and the root is wider than the precedence

This is the one that worried me most, because it is destructive. Both halves check out:

- `readAccount` pre-clears only `resumeError` and `resumeBlocked`. It never clears `agent`, `agentRestored`, `creation`, `computerPreviouslyConfirmed`, or `selectedComputer`.
- `applyResumeSelection`'s unbound-Agent branch sets `resumeBlocked` and returns early, leaving a previously set `agent` intact. The `none` and `computer` branches leave it too.
- `discardAgent` resolves `agent?.id ?? resumeBlocked?.agentId`, preferring the stale value over the displayed one.

`page.tsx` checks `backend.resumeBlocked` before every other branch and passes `resumeBlocked.agentName` to both the heading and `AgentResetControl`. So the confirmation banner names B while the delete targets A. Confirmed reachable exactly as described.

I would fix the precedence *and* the retention rather than just the precedence. Deriving the delete target from the displayed recovery target closes this instance, but the same stale `agent` also still feeds `computerConnectAdapter` and the draft-sync effect in `page.tsx`, so a resume that selects a different subject should replace the whole mutually exclusive set rather than layer onto it.

## 3 — confirmed, including the part that makes it permanent

I checked the two mechanisms that turn a transient lock into a stuck state, and both are real:

- `connection-registry.ts`: `observation.status !== "unavailable" && now - observedAt > TTL ? [] : [obs]`. An `unavailable` artifact observation is the one status that never ages out.
- `provider-cli-reconcile-owner.ts`: `isReconcileInFlight` returns `true` for `artifact?.status === "unavailable"`, so the reconcile path returns early instead of re-dispatching.

Together, a five-second lock loss becomes a non-expiring `unavailable` that the owner reads as still in flight. Worth adding for scoping: a reconnect *does* recover it, because `current.instanceId !== instanceId` forces a re-dispatch. So the blast radius is the life of one connection instance — but that is unbounded in wall-clock terms, and the foreground install succeeding never clears it, because the periodic refresh publishes the generic IM-CLI channel rather than the artifact observation.

A concrete path inside this feature: the reader binds Slack at step 3, their Computer drops, `ComputerRecovery` issues a **targeted repair** code, and running it takes the foreground `ensure all` branch (`agentId` is present on repair) while daemon registration dispatches the now-active Slack requirement. That is onboarding's own recovery flow racing itself.

## 4 — confirmed; the doctor bullet is my own miss

`doctor --json` still writes `{ok: false, result}` to stdout with no `error`. That is precisely the shape I objected to for `connect` in my first review, against a README I had quoted, and I did not carry the objection across to the file I had open. It should be settled the same way.

The exit-code contradiction is also real: `commandExitCode` maps `dependency` to `EXIT_CODES.serviceUnavailable` (3), while `ensure`, `inspect`, and partial `connect` all build `category: "dependency"` errors and hardcode `exitCode: 1`. And the Commander bullet reproduces by inspection — `usageErrorsExit2` throws after Commander has already written prose, and `cli/index.ts` calls `process.exit(2)` on `CommanderError` without presenting a document, so `--json` yields no JSON on a missing or unknown option.

This also subsumes the `retryability` nit from my last review: `never` on an `operation_in_progress` cause is the same disagreement between envelope, exit code, and `nextActions`.

## 1 — confirmed, but it predates this PR

The mechanism is right: `connected` comes from `computersQuery.isSuccess`, React Query serves cached success data immediately under `staleTime: 0`, and the effect auto-binds whenever `sole` resolves.

Scoping it, though: `main` at `5d2a0b21` already mounts `AgentComputerChoice` at `page.tsx:106` in the same resume-blocked screen, with the identical auto-bind effect. This PR removed that screen and put it back; it does not introduce the hazard. I would rather see it fixed than shipped around, but it is a pre-existing defect in a shared component and a reasonable candidate for its own issue instead of a gate on this PR — the author's call, and worth saying so it is not mistaken for a regression introduced here.

---

Everything I confirmed as fixed in my previous review still stands: the connect-code boundary is untouched, the foreground/prewarm ownership split, real periodic inspection, and the normal-path JSON envelopes all carry forward. 2, 3, and 4 are the ones to fix; the Context Tree follow-up on `opentag/system/im-integrations.md` from that review is unchanged and still belongs in a separate Tree PR after this merges.

CI has still not run on this head.

Gandy2025:
Scope correction after checking the fixed point: finding 1 is real, but it is not a regression introduced by #477 and I am removing it from this PR's merge gate.

At `5d2a0b21`, `apps/web/src/onboarding-v2/page.tsx:93-107` already mounts `AgentComputerChoice` on the same `resumeBlocked` screen, and the shared component already derives a sole Computer from any cached successful query and auto-binds it in the same effect. This PR adds the targeted-connect adapter and restores the screen after the earlier rewrite, but it does not introduce the stale-cache mutation hazard. Per the review scope ceiling, it belongs outside this focused diff unless the author deliberately chooses to fix it here.

Findings 2, 3, and 4 remain blocking and independently verified on `07b82854`:

- incomplete resume state replacement can make a confirmation naming Agent B delete retained Agent A;
- active-Integration reconciliation can still lose the cross-process lock and retain non-expiring `unavailable` for the connection lifetime;
- the advertised machine contract still misses Commander pre-action usage failures, carries contradictory category/retryability/exit routing, and leaves unhealthy `doctor --json` on a third schema/stream.

My `CHANGES_REQUESTED` state therefore still applies, narrowed to those three boundaries. Thanks for catching the baseline provenance and for independently confirming the destructive and concurrency paths.

liuchao-001:
Agreed on the scope narrowing — my `CHANGES_REQUESTED` is on the same three boundaries (2, 3, 4).

So the descoped item is not lost, I filed the stale-cache auto-bind as #479 against current `main`, with the mechanism, the `5d2a0b21` provenance, and a regression test to add. It is not a gate on this PR; fixing it here stays the author's call.

yuezengwu:
Addressed the remaining blocking review findings on exact head `c1cad92941e5c30321b921f126d14a7e647f5f0a`, after merging `origin/main@e299cea4f6653b1b8b59c2e43da1a92cb74b7e79` and resolving the conflict in `apps/cli/src/commands/computer/connect.ts`.

1. **Resume replacement and destructive target identity**
   - Every resume selection now clears the previous mutually exclusive Agent/Computer/creation state before applying the new snapshot.
   - Runtime recovery and the connect adapter derive the target from the visible `resumeBlocked.agentId`, so the confirmation subject and destructive target cannot diverge.
   - Regression coverage includes bound Agent → failed handoff → retry to none, Computer-only, or a different unbound Agent, including the exact deletion target.

2. **Active Integration versus foreground setup**
   - A daemon reconcile that encounters `operation_in_progress` waits and reinspects rather than publishing the lock loss as a terminal result.
   - If an unavailable artifact remains after the Server retry timer has cleared, the owner retires and re-dispatches it on demand without requiring a reconnect. An armed retry still remains single-flight.
   - Abort handling no longer starts another ensure attempt after cancellation.

3. **Common Agent-facing CLI contract**
   - Commander pre-action failures under `--json` now emit one `USAGE_ERROR` JSON document on stderr with exit 2 and no human help/prose.
   - Provider aggregate results share one classifier: live-operation-only failures are `unavailable/backoff`; auto-repairable failures are `dependency/immediate`; manual failures are `dependency/never`; mixed transient/repairable failures are `dependency/backoff`.
   - Exit codes are derived from the shared category policy, so dependency/unavailable failures exit 3 consistently across ensure, inspect, and partial connect.
   - Unhealthy `doctor --json` now emits the common failure envelope on stderr with a safe report in `result`, `configuration/never`, and exit 1.
   - Tests assert schema, stream, retryability, exit code, and next actions for missing/unknown options and retryable/non-retryable partial failures.

Per @liuchao-001's scoping note, I did not fold the pre-existing React Query one-Computer cache behavior into this PR's merge gate. The post-merge Context Tree reconciliation also remains a separate follow-up as previously agreed.

No database schema or migration changed.

Local exact-head validation:

- `pnpm check`
- `pnpm build`
- `pnpm typecheck` — 8/8 tasks
- `OPENTAG_POSTGRES_PORT=55432 pnpm test` — 8/8 workspace tasks
- affected suites: CLI 338/338, Client 891/891, Server unit 636/636, Web 623/623
- pre-push formatting, typecheck, lint, and repository checks
- migration drift: 38 existing migrations, no new migration

Please re-review this exact head. CI is running; ownership approval from @bestony remains separate.

liuchao-001:
Re-reviewed at exact head `c1cad929`. All three remaining blockers are fixed in substance. I am holding my `CHANGES_REQUESTED` for one reason only: **CI is red on this head**, and I traced the failure to this PR rather than to `main`.

## CI: `CLI Pack Smoke` fails, and it is this change

`CLI Pack Smoke` → step `Smoke source tarball` is failing. `main@dbbfa65b` is green on the same job, and `scripts/cli-pack-smoke.mjs` is byte-identical between `main` and this head, so the assertion did not move — the behaviour under it did.

Root cause:

- `program.ts` now calls `.exitOverride()` **unconditionally**, not only under `--json`. Every Commander usage error therefore throws instead of taking Commander's own `process.exit(1)`.
- `cli/index.ts` catches it and exits `EXIT_CODES.usage`, so an unknown option now exits **2** where it previously exited 1.
- `scripts/cli-pack-smoke.mjs:206-209` runs `doctor --server-url http://127.0.0.1:1` with `expectedStatus: 1`, so `run()` throws `exited with 2`.

**The code is right and the assertion is stale.** `apps/cli/README.md` has always documented `2` as the usage/validation exit, so the old `1` was the deviation, and unifying it is exactly what the blocker-4 fix was meant to do. The fix is to update that expectation to `2`. The neighbouring `stderr.includes("unknown option")` / `--server-url` assertions still hold, because `configureOutput({ writeErr })` is only silenced under `--json` and this invocation has no `--json`.

Worth knowing why local validation missed it: `cli-pack-smoke.mjs` is invoked only from `.github/workflows/ci.yml` and is not part of `pnpm test`, so a full local run cannot cover it.

While you are there — **Breaking changes currently says "None."** Usage-error exit codes changed from 1 to 2 for *every* command, not just `provider-cli`. It is the correct value and it matches the README, but anything checking `$? -eq 1` on a malformed invocation now sees 2. That belongs in the PR description.

Five other checks are still pending; I have only judged what has reported.

## The three blockers

**2 — resume replacement and destructive target.** Fixed, and fixed at the root rather than at the symptom. `applyResumeSelection` now clears the whole mutually exclusive set — `agent`, `agentRestored`, `creationRef`, `computerId`, `computerPreviouslyConfirmed`, `messaging`, `messagingProvider`, `computer`, `selectedComputer` — before applying the new snapshot, so a retained subject can no longer answer for the displayed one. Both consumers also flipped to `resumeBlocked?.agentId ?? agent?.id`, so the delete target and the connect adapter derive from what the reader is actually looking at. That closes the "banner names B, deletes A" path in both directions.

**3 — active-Integration convergence versus foreground setup.** Fixed on both sides:

- Client: `#ensureConverging` retries while the diagnostic is `operation_in_progress`, up to `30 × 5s`, which comfortably outlives a foreground install, and it re-checks `#closed`/`aborted` after the wait so shutdown cannot start another attempt.
- Server: `isReconcileInFlight` no longer treats `unavailable` as in-flight, and `#refreshRequirement` handles a retained `unavailable` explicitly — leaving it alone while `artifactRetryPending` is armed, then retiring and re-dispatching once the bounded budget is spent. That removes the reconnect dependency I described.

One consequence to be aware of rather than fix now: there is no longer any terminal state for an artifact that is *genuinely* permanently unavailable. For an `UNREPAIRABLE` diagnostic such as `unsupported_platform`, each cycle is dispatch → `unavailable` → `#scheduleArtifactRetry` arms → budget exhausts → the next readiness read retires and re-dispatches with a fresh budget. Since `ensureActiveReadiness` is driven by `imCliReadiness` / `credentialExecutionReadiness`, and onboarding polls handoff, that is a slow indefinite loop instead of a stuck state. Trading "never recovers" for "retries forever" is the right direction, but distinguishing permanently-unrepairable diagnostics from transient ones would let the former settle. Fine as a follow-up.

**4 — common Agent-facing contract.** Fixed. Commander pre-action failures now emit one `USAGE_ERROR` document on stderr under `--json` with human output suppressed; the shared classifier gives live-lock failures `unavailable/backoff`, auto-repairable `dependency/immediate`, manual `dependency/never`, and mixed `dependency/backoff`, so the envelope no longer contradicts its own `nextActions`; exit codes derive from the category policy; and unhealthy `doctor --json` emits the common failure envelope on stderr with the report in `result` and `configuration` mapping to exit 1. The `retryability` nit from my earlier review is properly subsumed by the classifier rather than patched.

---

Nothing in the connect-code boundary changed, so those conclusions still carry forward. Once `CLI Pack Smoke` is green — and the remaining checks report — I will approve. The Context Tree follow-up on `opentag/system/im-integrations.md` is unchanged and still belongs in a separate PR after this merges.

yuezengwu:
Follow-up on the new exact head `2f9214f6bef3b8485bb053a2280b906b7dfa83ce`:

- The first CI run exposed a stale `CLI Pack Smoke` expectation: the packed binary correctly returned usage exit 2 for removed option `doctor --server-url`, while the smoke script still expected the old generic exit 1.
- Updated the pack-level assertion to exit 2, retaining the exact stderr checks for `unknown option` and `--server-url`.
- Verified the installed source-channel package end to end and re-ran the three focused CLI contract files (67/67).
- Merged the latest `origin/main@dbbfa65b` without conflicts. This also brings the separately merged #481 fix for the pre-existing fresh-inventory auto-bind issue; it remains outside this PR's own diff against main.

Fresh exact-head validation passed: `pnpm check`, `pnpm build`, `pnpm typecheck` (8/8), full `pnpm test` (8/8 workspace tasks; Web 626/626), source-channel packed CLI smoke, and the pre-push hook. No database schema or migration changed.

CI and the requested reviews are running again on this head.

Gandy2025 (changes requested):
Re-reviewed exact head `2f9214f6bef3b8485bb053a2280b906b7dfa83ce` against current base `dbbfa65b1ec3c068df8efc61cc29d54f147dc356`, proportionally scoped to the remaining blockers and their adjacent failure paths.

The destructive Web resume-state bug is fixed: each account read now replaces the mutually exclusive Agent/Computer state, and deletion prefers the Agent displayed by the recovery surface. The non-expiring Server `unavailable` state is also fixed, and the normal CLI usage/provider/Doctor JSON paths now use the documented envelope and routing policy. The packaged smoke now correctly expects usage exit 2. The earlier connect-code conclusions remain valid because that boundary is untouched.

I still cannot approve this head. Three issues remain, all within the requested fix surface.

## 1. [P1][Spec] The provider-install race is only handled when the daemon loses the lock

`packages/client/src/runtime/provider-cli/reconciler.ts:296-307` now waits and retries when daemon reconciliation receives `operation_in_progress`. The opposite lock ordering is still terminal, however.

`apps/cli/src/core/computer/connect.ts:80-82` starts or restarts the daemon before foreground Provider setup. An active Integration can therefore make the daemon enter managed `ensure` first. Foreground setup then calls `manager.ensure` only once per Provider at `apps/cli/src/core/provider-cli/ensure.ts:147-150`; the shared lock still gives up after about five seconds at `packages/client/src/runtime/provider-cli/lock.ts:77-106`. A normal daemon download can exceed that, so foreground exits nonzero with `operation_in_progress` even though the daemon completes the same healthy install later.

This is onboarding's own recovery path when an already-integrated, unbound Agent reconnects. It also contradicts the stated ownership model that foreground targeted `connect` is the sole first-setup installer and returns the terminal repair verdict. Make either process that loses the lock join/observe the live operation to terminal inspection, or prevent daemon convergence from racing this foreground setup. Cover both lock orderings.

## 2. [P1][Standards] Lock-busy retry leaks the daemon shutdown listener on every normal wait

`packages/client/src/runtime/provider-cli/reconciler.ts:312-320` adds a `{ once: true }` abort listener for every five-second retry, but removes it only if abort eventually fires. Whenever the timer wins the race, the listener remains attached to the long-lived daemon signal.

A routine managed download lasting more than roughly 55 seconds crosses Node's default ten-listener threshold and emits `MaxListenersExceededWarning`; later contention episodes continue retaining closures until shutdown. The current test only covers abort winning the race. Remove the listener when either branch settles (or use an abortable timer/disposable listener) and add a regression that exercises repeated timer-winning retries.

## 3. [P1][Standards + Spec] The daemon-service JSON failure still contradicts the shared exit policy

`apps/cli/src/commands/computer/connect.ts:198-227` labels a daemon service reload failure as `category: "dependency"` and `retryability: "immediate"`, returns an exact retry action, but hardcodes `exitCode: 1`; `presentConnectFailure` also sets process exit 1 at `:75-79`. The authoritative policy maps dependency failures to exit 3 at `apps/cli/src/core/command/policy.ts:312-318`.

This leaves machine readers with mutually contradictory envelope metadata, action, and process status on a safe partial connection path. Derive and propagate `commandExitCode(commandError)`, and add the missing schema/stream/exit test for this terminal path.

Local evidence after a clean build: the directly affected Client (21), Server (18), Web (15), CLI (68), and CLI policy-coverage (12) tests pass; packaged missing/unknown-option probes emit exactly one stderr JSON document with exit 2. Exact-head CLI Pack Smoke is now green. Those results validate the repaired paths but do not cover the three gaps above.

The newest merge only changes the packaged smoke within this review surface; the three code paths above are unchanged from `c1cad929`. I found no PR-specific scope creep in the other merged-main files. Re-review can remain focused on these three items. The earlier Web fix, Server re-drive behavior, normal JSON envelopes, periodic inspection, and connect-code boundary can carry forward if untouched.

liuchao-001:
Re-reviewed exact head `2f9214f6` (base `dbbfa65b`). The pack-smoke fix is exactly right — `expectedStatus: 2` with the `unknown option` / `--server-url` stderr assertions retained, and a comment naming the contract. That one is closed.

I verified all three of @Gandy2025's new findings against this head; all three hold. I also have two CI failures to add, one of which is this PR's and one of which I do not think is.

## CI is red again on this head, for two different reasons

**`Checks` → `Enforce Agent Runtime coverage` — this PR's.** The gate wants 100%; this head reports functions 99.85% and statements 99.97%. The only file below the line is `client-runtime-composition.ts` at 99.7% statements / 98.75% functions, and the sole change there is:

```ts
void this.#providerCliReconciler?.refreshPublishedImCliReadiness().catch(() => undefined);
```

That rejection handler is never executed by any test, which is enough to fail a 100%-functions gate.

Why local validation missed it: this runs as `pnpm --filter @opentag/client test:agent-runtime:coverage`, a separate task that `pnpm test` does not include. The description reported that task at 100% back on `9fc883a6` and it dropped out of the evidence list as the head moved — worth putting back, since it is the only thing that covers this gate.

Convenient overlap: covering that handler pairs naturally with the fix for Gandy's finding 2 below, which also needs a test exercising the branch where the timer wins.

**`Node 24 Compatibility` → `Run compatibility smokes` — I do not think this one is yours.** It fails in `claude-code-agent-runtime-exhaustive.test.ts` › "runs the default process adapter and the default local readiness probe against controlled artifacts", with `expected { ready: false, issues: [ …(2) ] } to match object { ready: false, …(1) }`.

Attribution evidence:

- This PR touches nothing under `packages/client/src/runtime/providers/` and does not touch that test file — `git diff --stat origin/main 2f9214f6` over both paths is empty.
- Node 22 and Node 26 passed the identical tree. A deterministic defect in this diff would not pick one Node version.
- `main@dbbfa65b` is green on all three.

An extra readiness *issue* appearing on one runner points at environment sensitivity in a probe of local state. The cheap discriminator is a re-run, not a code change here — and please don't settle it with a test retry; if it reproduces it deserves its own root-cause fix outside this PR.

## Gandy's three findings — all confirmed

**1. Only the daemon side of the lock race is handled.** Confirmed. `#ensureConverging` gives the reconciler a busy-retry loop, but `ensureOneProvider` still calls `manager.ensure` exactly once with no `operation_in_progress` handling. So the fix is asymmetric: whichever process did not get the retry is the one that loses. `connect` starts the daemon before foreground setup, so with an active Integration the daemon can take the lock first and the foreground command — the one this PR designates as the sole first-setup installer that returns the terminal verdict — exits nonzero on a healthy install it merely lost a race to.

**2. The retry leaks an abort listener whenever the timer wins.** Confirmed:

```ts
return Promise.race([
  wait,
  new Promise<void>((resolve) => signal.addEventListener("abort", () => resolve(), { once: true })),
]);
```

`{ once: true }` only removes the listener if abort actually fires. On the normal path the timer wins and the listener stays attached to the long-lived daemon signal, once per retry. With `PROVIDER_CLI_LOCK_BUSY_MAX_ATTEMPTS = 30`, a single contention episode can attach up to 30; Node warns at 10, so ~11 retries ≈ 55s reproduces `MaxListenersExceededWarning` — which matches Gandy's estimate. The existing test only covers abort winning, which is precisely the branch that does not leak.

**3. The daemon-service failure path kept its hardcoded exit code.** Confirmed, and it reads as a single missed path rather than a disagreement: the sibling paths in the same file were converted to `commandExitCode(commandError)`, while `presentServiceFailure` still passes `exitCode: 1` alongside `category: "dependency"`, and `presentConnectFailure` sets `process.exitCode = 1` around it. The policy maps `dependency` to 3, so this one terminal path ships an envelope, a retry action, and a process status that disagree.

## Still outstanding from my last review

**Breaking changes still says "None."** Usage-error exit codes moved from 1 to 2 for every command — that is the change the pack smoke had to be updated for, so it is demonstrably observable from outside. It should be in the description.

---

The connect-code boundary is still untouched and those conclusions carry forward, as do the Web resume fix, the Server re-drive behaviour, periodic inspection, and the normal JSON envelopes. My `CHANGES_REQUESTED` stands on the three items above plus the coverage gate.

yuezengwu:
Addressed the three remaining exact-head blockers on `785ba818855471f3c8bf09ee3bd291fa1aa98d0a` (base `37a867399d72b3d42608bbb4184748ce29b23b2a`):

1. **Both Provider lock orderings converge.** Foreground `provider-cli ensure` now gives `operation_in_progress` the same bounded 30 x 5s wait/re-observe treatment as daemon reconciliation. The new regression starts with a daemon-held lock, completes the Provider while foreground waits, and asserts the command returns the terminal ready result instead of a false partial failure. The existing inverse-order regression remains in place.
2. **No shutdown-listener leak.** Every timer-winning busy wait now removes its abort listener; the regression exercises repeated normal waits and asserts add/remove balance while the long-lived signal remains un-aborted.
3. **Daemon service failure uses the shared exit policy.** `DAEMON_SERVICE_FAILED` derives dependency exit 3 via `commandExitCode` in both human and JSON paths. The JSON regression asserts the single stderr envelope, safe partial connection result, exact retry action, and process exit 3.

Also updated the PR description: malformed CLI usage changing from generic exit 1 to documented exit 2 is now explicitly listed under Breaking changes.

Exact-head validation passed:

- `pnpm check`
- `pnpm typecheck` (8/8)
- `pnpm build` (6/6)
- `pnpm test` (8/8; CLI 340/340, Client 896/896, Server 700/700, Web 714/714)
- blocker regressions: CLI 35/35, Client reconciler 22/22
- Agent Runtime coverage: 347/347, 100% statements/branches/functions/lines
- source-channel packed CLI smoke through `opentag-dev`
- pre-push formatting, lint, typecheck, repository checks, and migration drift

No database schema or migration changed. Please re-review this exact head; CI is running on it now.

liuchao-001 (approved):
Approving at exact head `785ba818` (base `37a86739`). Every blocker I or @Gandy2025 raised is resolved, and CI is green.

## CI

All checks pass except `ownership-gate`, which is waiting on @bestony for 65 files — a process gate, not a test result.

**Node 24 passed on this head.** That is the outcome I expected when I called the earlier failure environment sensitivity rather than a defect in this diff, and it settles that question without anyone having to paper over it with a retry. The coverage gate also passes: the new test drives `refreshPublishedImCliReadiness` to reject, which exercises the `.catch(() => undefined)` handler that was the only thing under the 100% line.

## The three fixes

**Both lock orderings converge.** `ensureOneProvider` now applies the same `PROVIDER_CLI_LOCK_BUSY_MAX_ATTEMPTS × PROVIDER_CLI_LOCK_BUSY_RETRY_DELAY_MS` budget the reconciler got, so whichever process loses the cross-process lock re-observes the terminal fact instead of reporting the winner's healthy install as its own failure. Passing `emitPhases: false` on retries is a nice touch — the wait does not turn into repeated phase output.

**The listener leak is fixed, and slightly more than asked.** `#sleepBusy` now removes a named `onAbort` through a `finish()` that runs on either branch, guards double-settle with `settled`, and closes the check-to-listener race by re-testing `signal.aborted` immediately after `addEventListener`. That last one was not in the finding; it is the gap a straight "remove it in both branches" fix would have left open.

**`DAEMON_SERVICE_FAILED` follows the shared policy.** `presentServiceFailure` derives `commandExitCode(commandError)` once and returns it, both the JSON and human branches return that value, and `presentConnectFailure` propagates it. Dependency now exits 3 on the path that previously hardcoded 1, so envelope, retry action, and process status finally agree.

The Breaking changes section now names the usage exit 1 → 2 change explicitly. Thanks for taking that one.

## Merge integrity through #469

Main landed the exact-Agent setup redesign underneath this branch, which rewrites the same onboarding surface, so I checked that this PR's work survived rather than assuming it. `OnboardingV2Page` is still what `OnboardingBoundary` renders; the `resumeBlocked` screen with `AgentComputerChoice` and `AgentResetControl` is intact; `applyResumeSelection` still opens with the full state replacement and both consumers still resolve `resumeBlocked?.agentId ?? agent?.id`; `ImCliReadinessList` is still rendered. The connect-code boundary in `machine-auth-service.ts` is byte-identical to the head I originally audited, so those cryptographic and transactional conclusions carry forward unchanged.

## Non-blocking, for after this merges

1. **The Context Tree needs reconciling.** `opentag/system/im-integrations.md` still says the daemon prepares both provider CLIs during first setup and reasons that "the daemon, not an interactive shell or the Web page, owns the environment in which Local Agents execute." This PR inverts that: foreground `connect` installs, the daemon observes. The surviving rationale — a foreground installer is what lets the invoking Agent see failures and run repair commands — is worth recording. I will open that Tree PR once this lands, so it reconciles against merged source truth.

2. **No terminal state for a permanently unavailable artifact.** For an `UNREPAIRABLE` diagnostic the retire-and-re-dispatch path now cycles indefinitely rather than sticking. That is the right direction versus never recovering, but separating permanently-unrepairable diagnostics from transient ones would let those settle.

Neither gates this. Merging still needs @bestony's ownership approval and @Gandy2025's own re-review of this head.

Gandy2025 (approved):
Re-reviewed exact head `785ba818855471f3c8bf09ee3bd291fa1aa98d0a` against current base `37a867399d72b3d42608bbb4184748ce29b23b2a`, limited to the three outstanding code findings, the coverage gate, and their adjacent paths.

All remaining blockers are resolved:

- Foreground Provider setup now handles the opposite lock ordering too: `ensureOneProvider` waits and re-enters `ensure` on `operation_in_progress`, so either process that loses the cross-process lock observes the terminal state. The regression test models a daemon-held lock completing successfully.
- The daemon lock-busy wait removes its abort listener when either the timer or abort settles, including the guard-to-listener race. The new test covers repeated timer-winning waits and verifies matching add/remove calls.
- Daemon service partial failure now derives and propagates `commandExitCode(commandError)` for both JSON and human paths; the envelope, retryability, repair action, and process exit consistently report dependency exit 3.
- The capability-refresh rejection handler is covered. Exact-head Agent Runtime coverage is back to 100% statements, branches, functions, and lines.

The PR description now discloses the observable usage-error exit change from 1 to the documented 2. The newest main merge accounts for the unrelated large first-parent delta; the blocker-fix commit itself is limited to the six expected CLI/Client files plus tests.

Local proportional validation passed: full build; 35 affected CLI tests; 22 Provider reconciler tests; and 24 files / 347 Agent Runtime tests at 100% coverage. Exact-head CI is fully green across Checks, Node 22/24/26, packaged CLI, browser, container, ownership, and policy gates.

No new Standards or Spec findings. The earlier Web resume, Server re-drive, periodic inspection, common JSON envelope, and connect-code security conclusions carry forward because those boundaries were not reshaped by this fix. Approving.

yuezengwu:
Merged as 98c475eaae6684d439be79ab718986279215c950.

@liuchao-001, please proceed with the Context Tree reconciliation you called out in the exact-head review for opentag/system/im-integrations.md. The source-backed update should record that foreground targeted connect is the authoritative installer during first setup, while the daemon periodically inspects and reports both Provider CLIs and may converge only the selected Provider after IM binding. Please also make the changed authorization boundary explicit rather than carrying forward the previous daemon-owned installation rationale.

The stale Computer-cache finding is already completed by #481 / #479. The permanently unrepairable artifact retry-loop follow-up is now tracked separately in #490. The Turn credential-readiness barrier remains open in #470 and was not closed by this PR.

PR #480, opened 2026-09-02
fix(client): surface fatal transport and durability failures

bestony:
Three failures in the daemon's transport and durability layer were invisible or actively mislabelled.
Two of the three had the log saying something untrue.

## Checklist

- [x] 1. The fatal close reason reaches the log, redacted via `redactForLog`
- [x] 2. Existing fields (`attempt`, `category`, `state`) preserved
- [x] 3. The turn-report submit rejection is logged and still non-blocking
- [x] 4. `readDurableJson` distinguishes a parse failure from a validation failure
- [x] 5. `runtime-connection.ts` still at most 1073 lines
- [x] 6. Tests cover all three
- [x] 7. Progress recorded

## What changed

**The fatal close reason never reached the log.** `runtime-connection.ts:369` logged
`{attempt, category:"protocol", state}` and threw — a throw that **permanently terminates the
daemon's connection loop**. Every fatal cause funnels through it: `failProtocol` at `:496` builds a
`RuntimeConnectionError` for roughly twenty protocol violations, and `:770-772` builds one from a
server rejection frame. The only thing separating them is `error.message`, which was never emitted.
"My daemon connects and drops" was a total black box. The line now carries a redacted `reason`.

**The turn result submission was fire-and-forget.** `agent-turn-runner.ts:312` was
`void this.#reportOwner.submit(...).catch(() => undefined)`. The agent may already have replied in
IM, but if that submit failed the server never learned the outcome, the Task never settled, and
`client.log` showed a clean completion followed by nothing. Note the asymmetry it sat next to: the
line above already warned when the turn failed to *enter* the reporting phase — the failure that
actually loses the result was the silent one. It now warns, and stays non-blocking.

**Schema drift was reported as file corruption.** `durable-file.ts` ran `validate(JSON.parse(...))`
inside one `try`, so **every** schema failure surfaced as "Runtime storage contains invalid JSON" —
a version drift after an upgrade was actively described to the user as a corrupt file. Parse and
validate are now separate, with distinct messages.

**On the size ratchet:** `runtime-connection.ts` sits at exactly its `file-size-exceptions.json`
entry (1073, against a global limit of 500), so there was zero headroom. The new field builder went
into `runtime-connection-helpers.ts`; the touched file is a 2-line substitution and is **still exactly
1073 lines**. The exception value is unchanged.

## Verified by the orchestrator, not self-reported

| Criterion | Result |
| --- | --- |
| `pnpm --filter @opentag/client test` | exit 0 — **67 files, 887 tests** (see the flake note) |
| `wc -l runtime-connection.ts` | **1073** — at the cap, not over |
| `scripts/file-size-exceptions.json` | no diff |
| Two fatal causes distinguishable from the log alone | **probed directly**: `Unsupported protocol frame: server:bogus` vs `credential is no longer active` — distinct, with `attempt`/`category`/`state` intact |
| No credential in the emitted close reason | **probed directly**: a message carrying `token="sk-secret-abc123"` and `cookie=sid=zzz` emits neither |
| Schema failure is not called invalid JSON | **probed directly**: `Runtime storage contains invalid JSON` vs `Runtime storage contains data that does not match its schema` |
| Turn-report failure logged, still non-blocking | covered by the new `agent-turn-runner.test.ts` cases; `void` retained at the call site |
| No drizzle migration | no diff under `packages/server/drizzle` |

**One flake, investigated rather than waved through.** The first full client run failed one test:
`claude-code-agent-runtime-exhaustive.test.ts` timed out at **5005 ms against a 5000 ms limit**. That
file is not in this diff and exercises none of the changed modules. Re-run alone: 36/36 pass. Re-run
of the whole suite: 887/887 pass. It is timing contention on the dev machine (three agents were
running concurrently), not a regression from this change.

---

*Dispatched as run `337683`, lane `client-runtime-transport-2`. Written by codex under bypassed
approvals in an isolated worktree; verified and published by the orchestrator.*

### DIFF
--- a/packages/client/src/__tests__/agent-turn-runner.test.ts
+++ b/packages/client/src/__tests__/agent-turn-runner.test.ts
@@ -363,6 +363,45 @@ describe("AgentTurnRunner", () => {
     );
   });
 
+  it("logs an unknown reason for a non-Error Turn Report submission rejection", async () => {
+    const logs: RecordedLog[] = [];
+    const submit = vi.fn(() => Promise.reject("boom"));
+    const create = vi.fn((input) => ({
+      ...input,
+      type: "turn:report",
+      requestId: randomUUID(),
+      resultHash: "g".repeat(64),
+    }));
+    const runner = new AgentTurnRunner({
+      bindingStore: {
+        updateUnresolved: vi.fn(async () => {
+          throw new Error("disk failed");
+        }),
+      } as unknown as SessionBindingStore,
+      connection: { send: vi.fn(async () => undefined) },
+      custody: { markReporting: vi.fn(async () => undefined), recordResult: vi.fn() } as unknown as TurnCustodyOwner,
+      logger: recordingLogger(logs),
+      reportOwner: { create, submit } as unknown as TurnReportOwner,
+      runtimeManager: {} as SessionRuntimeManager,
+      credentialEnvironment: credentialEnvironment(),
+    });
+
+    runner.start(liveOwner({ ...delivery(), deliveryId: "delivery-non-error" }));
+    await runner.settled();
+    await vi.waitFor(() =>
+      expect(logs).toContainEqual(
+        expect.objectContaining({
+          level: "warn",
+          message: "Turn Report submission failed",
+          fields: expect.objectContaining({
+            deliveryId: "delivery-non-error",
+            reason: "Unknown Turn Report submission failure",
+          }),
+        }),
+      ),
+    );
+  });
+
   it("keeps unresolved custody when the reporting transition fails", async () => {
     let observer: AgentRuntimeEventSink | undefined;
     const onRuntimeEvent = vi.fn(async () => undefined);
--- a/packages/client/src/__tests__/agent-turn-runner.test.ts
+++ b/packages/client/src/__tests__/agent-turn-runner.test.ts
@@ -22,6 +22,7 @@ import type { RecordedSteerInput, SessionBindingStore } from "../runtime/session
 import { ClientRuntimeProviderStartError, type SessionRuntimeManager } from "../runtime/session-runtime-manager.js";
 import type { LiveTurnOwner, TurnCustodyOwner } from "../runtime/turn-custody-owner.js";
 import type { TurnReportOwner } from "../runtime/turn-report-owner.js";
+import { type RecordedLog, recordingLogger } from "./recording-logger.js";
 
 describe("AgentTurnRunner", () => {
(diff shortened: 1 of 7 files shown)


yuezengwu (approved):
Approved exact head 0683261c09dbbea0f384120ab30da55a5132f4b2 against base e299cea4f6653b1b8b59c2e43da1a92cb74b7e79. Reviewed fatal transport reporting, log redaction, durable JSON/schema error separation, and regression coverage. Local validation: @opentag/shared build; @opentag/client 887 tests; @opentag/client typecheck; repository check; git diff --check.

yuezengwu (changes requested):
Exact head 0683261c09dbbea0f384120ab30da55a5132f4b2 is blocked by the required Agent Runtime coverage check. The new conditional at packages/client/src/runtime/agent-turn-runner.ts:317 leaves the non-Error rejection branch uncovered: the added regression test rejects only with Error. CI reports 99.96% branch coverage against the required 100% (agent-turn-runner.ts line 317). Please cover the unknown rejection path or simplify the branch, then request a fresh exact-head review.

bestony:
Addressed in `b3829f65` — thanks for catching this.

`packages/client/src/__tests__/agent-turn-runner.test.ts` now covers the non-Error rejection path: the added case rejects the Turn Report submit with a bare string (`Promise.reject("boom")`) and asserts the warn line carries `reason: "Unknown Turn Report submission failure"`. It uses the existing recording-logger double, and the change is test-only — no production code moved.

Verified locally on the new head:

- `pnpm --filter @opentag/client test:agent-runtime:coverage` — exit 0, **Branches 100% (2805/2805)**, 346 tests (was 345)
- `pnpm check`, `pnpm typecheck`, `pnpm build` — all exit 0
- `pnpm test` (root) — exit 0
- `pnpm --filter @opentag/client test` — 67 files, 888 tests
- `runtime-connection.ts` still 1073 lines; `scripts/file-size-exceptions.json` unchanged; no migration added

Requesting a fresh review against exact head `b3829f65`.

yuezengwu (approved):
Approved exact head b3829f655be6b4281c389b35ecbbc339d4ad42d8 against base e299cea4f6653b1b8b59c2e43da1a92cb74b7e79 after a fresh review. The added non-Error Turn Report rejection case closes the required branch-coverage gap without changing production behavior. Local validation: @opentag/shared build; @opentag/client 888 tests; Agent Runtime coverage 100% statements/branches/functions/lines (346 tests); @opentag/client typecheck; repository check; git diff --check.

PR #483, opened 2026-09-02
feat(server): classify silent domain service verdicts

bestony:
## What this changes

`packages/server/src/services/**` had 73 catch sites, 18 of them bare `catch {}`, and 6 logging call
sites in total. Its dominant failure mode was not a thrown error but a **silent verdict**:
semantically distinct authorization and decode failures collapsing into one indistinguishable value.

This makes those verdicts legible. **Response contracts are unchanged** — every return shape, every
status code and every existing service test is untouched. Logging is observation only.

### 1. Credential decode: three unrelated causes, one `undefined`

`decodeFeishuCredential` / `decodeSlackCredential` wrapped
`Schema.parse(JSON.parse(cipher.decrypt(...)))` in a single `catch {}`. One key rotation disconnects
every Feishu channel at once and returns `credential_stale` everywhere, with nothing in the log to
tell "wrong key" from "schema drift". The three causes are now separated, each with its own code:

| cause | code |
| --- | --- |
| `cipher.decrypt` throws — wrong or rotated `OPENTAG_ENCRYPTION_KEY` | `IM_BINDING_CREDENTIAL_DECRYPT_FAILED` |
| `JSON.parse` throws — stored ciphertext corrupt | `IM_BINDING_CREDENTIAL_PAYLOAD_INVALID` |
| `Schema.parse` throws — credential schema drift | `IM_BINDING_CREDENTIAL_SCHEMA_INVALID` |

The `undefined` return is preserved. Nothing from the payload is logged — no ciphertext, no decrypted
field, no `appSecret`, no Zod `received` value; only the failure class and the binding id.

### 2. Six admission gates, one boolean

`SessionService.withCollaborationDispatchAdmission` returned `{ admitted: false }` from six
semantically different places, all silent. Each now emits an `info` line with a distinct
discriminator plus the ids already in scope. The return shape is still exactly
`{ admitted: false } as const`.

`SESSION_COLLABORATION_ADMISSION_AGENT_INACTIVE` · `_BINDING_INVALID` ·
`_AUTHORITY_SESSION_INVALID` · `_PLACEMENT_INVALID` · `_COMPUTER_OWNERSHIP_INVALID` ·
`_SOURCE_INSTANCE_STALE`

### 3. The inverted diagnostic in session collaboration

The service reported the wrong half of its failures: a benign business rejection fired a diagnostic,
while `#deliver`'s bare `catch {}` turned an assembler exception into the tidy outcome
`runtime_not_ready`, and `#record`'s bare `catch {}` swallowed a durable outcome-write failure — a
data-loss shaped event. Both internal failures now log
(`SESSION_COLLABORATION_RUNTIME_ASSEMBLY_FAILED`, `SESSION_COLLABORATION_OUTCOME_WRITE_FAILED`);
the existing benign diagnostic is kept as it was.

### 4. Wired through the existing port

Services receive the logger via the `serviceLogger(...)` port already exposed in `index.ts` — one
property per construction site. No second logging mechanism, no direct pino import.

### 5. Follow-up: `bindingId` no longer carries a substitute id (`9bbee4ee`)

Orchestrator verification caught a defect in the first commit before this branch was published:
`slackInstallationInspectionInput` populated the field named `bindingId` from
`slack_installations.id`, and two fallbacks (`bindingId ?? installation.id`,
`?? input.id`) propagated it. An operator grepping that value against `im_bindings` would have found
nothing — the log stating something untrue is exactly the failure class this change exists to
remove. `slackInstallationId` is now its own key, the substitute-id fallbacks are gone, and
`bindingId` is either a genuine `im_bindings.id` or absent.

## Checklist

- [x] 1. Credential decode distinguishes decrypt / parse / schema failures; return stays `undefined`
- [x] 2. No ciphertext, credential field, `appSecret` or Zod `received` value reaches a log
- [x] 3. All six admission gates emit a distinct discriminator; return shape unchanged
- [x] 4. `#deliver` and `#record` internal failures are logged
- [x] 5. Services receive the logger via `serviceLogger(...)`, one property per construction site
- [x] 6. `agent-service.ts` background helpers untouched
- [x] 7. Tests cover every new discriminator
- [x] 8. One concept, one key — `bindingId` and `slackInstallationId` never substitute for each other

| criterion | outcome |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm build` | exit 0 |
| `pnpm --filter @opentag/server test` | exit 0 — 64 files, 628 tests |
| `pnpm test` (root) | **exit 1, environmental** — see below |
| A test distinguishes all three credential-decode failure classes from the log alone | pass — `credential-material.test.ts`, both providers |
| A test proves a credential value, ciphertext and `appSecret` never appear in the log output | pass — asserted against ciphertext, `appSecret` and the Zod `received` value |
| A test asserts each of the six admission gates by its own discriminator | pass — all six present in `session-services.unit.test.ts` |
| A test proves an outcome-write failure in `#record` is no longer silent | pass — `SESSION_COLLABORATION_OUTCOME_WRITE_FAILED` asserted |
| Response contracts unchanged — existing service tests pass unmodified | pass — the two touched test files removed no assertion; the only deletions are constructor fixtures |
| No drizzle migration; `CURRENT_MIGRATION_COUNT` untouched | pass — 0 migration files, constant untouched |
| `git status --porcelain` empty | pass |

**On the root `pnpm test` exit 1.** The only failure is
`scripts/__tests__/lefthook-stage-fixed.test.mjs`, which makes a signing commit and dies with
`error: 1Password: failed to fill whole buffer` / `fatal: failed to write commit object`. The same
failure reproduces identically on clean `main` at `e299cea4`, which contains none of this work, and
this branch touches no file under `scripts/`. It is the verifying host's 1Password agent being
locked (`ssh-add -l` reports no identities), not a defect in this change. CI signs nothing, so the
job is expected to pass there.

## Deviation from the confirmed plan

None. The plan explicitly forbade wrapping `agent-service.ts`'s `#stopSessionsBestEffort` /
`#notifyProviderCliPlacement` in `BackgroundFailureSupervisor`, and that file is untouched.

## Commit granularity

Two commits for an eight-item checklist — the first commit is coarser than the commit policy asks
for. Noted rather than rewritten: rebasing a verified branch risks the work for a tidier history.

---

*Provenance: written by codex under `/dispatch-codex` run `337683`, lane `server-domain-services-1`,
running with approvals and sandbox bypassed inside an isolated git worktree. Verified and published
by the orchestrator; the acceptance table above records checks the orchestrator re-ran itself.*

### DIFF
--- a/packages/server/src/__tests__/credential-material.test.ts
+++ b/packages/server/src/__tests__/credential-material.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  type CredentialDecodeOptions,
+  decodeFeishuCredential,
+  decodeSlackCredential,
+} from "../services/im-bindings/credential-material.js";
+
+const bindingId = "binding-1";
+const ciphertext = "ciphertext-secret";
+const credentialSecret = "app-secret-value";
+
+const decoders = [
+  {
+    name: "Feishu",
+    decode: decodeFeishuCredential,
+    validPayload: JSON.stringify({ appId: "app-id", appSecret: credentialSecret, grantedScopes: ["scope-a"] }),
+  },
+  {
+    name: "Slack",
+    decode: decodeSlackCredential,
+    validPayload: JSON.stringify({
+      botAccessToken: "xoxb-token-value",
+      botId: "bot-id",
+      grantedScopes: ["scope-a"],
+      signingSecret: "signing-secret-value",
+    }),
+  },
+] as const;
+
+describe("credential material decoding", () => {
+  it.each(decoders)("classifies $name decrypt, parse, and schema failures from logs", ({ decode, validPayload }) => {
+    const logger = { warn: vi.fn() };
+    const options: CredentialDecodeOptions = { bindingId, logger };
+
+    expect(
+      decode(
+        {
+          decrypt: () => {
+            throw new Error("wrong key");
+          },
+        } as never,
+        ciphertext,
+        options,
+      ),
+    ).toBeUndefined();
+    expect(decode({ decrypt: () => "not-json" } as never, ciphertext, options)).toBeUndefined();
+    expect(
+      decode({ decrypt: () => validPayload.replace('["scope-a"]', '"not-an-array"') } as never, ciphertext, options),
+    ).toBeUndefined();
+
+    expect(logger.warn).toHaveBeenCalledTimes(3);
+    expect(logger.warn.mock.calls.map(([payload]) => payload.code)).toEqual([
+      "IM_BINDING_CREDENTIAL_DECRYPT_FAILED",
+      "IM_BINDING_CREDENTIAL_PAYLOAD_INVALID",
+      "IM_BINDING_CREDENTIAL_SCHEMA_INVALID",
+    ]);
+    for (const [payload, message] of logger.warn.mock.calls) {
+      expect(payload).toMatchObject({ bindingId });
+      expect(JSON.stringify({ payload, message })).not.toContain(ciphertext);
+      expect(JSON.stringify({ payload, message })).not.toContain(credentialSecret);
+      expect(JSON.stringify({ payload, message })).not.toContain("not-an-array");
+    }
+  });
+
+  it("returns decoded credentials when all stages succeed", () => {
+    const logger = { warn: vi.fn() };
+    const result = decodeFeishuCredential(
+      {
+        decrypt: () => JSON.stringify({ appId: "app-id", appSecret: credentialSecret, grantedScopes: ["scope-a"] }),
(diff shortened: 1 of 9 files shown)


yuezengwu (approved):
Exact-head review approved at 9bbee4ee0153edb04396160c46e48feb54209cab (base e299cea4f6653b1b8b59c2e43da1a92cb74b7e79). Reviewed the structured service-verdict diagnostics, credential failure classification/redaction boundaries, admission-gate logging, and associated tests. Local validation: @opentag/server 628/628 tests passed, server typecheck passed, repository check passed, and exact diff check passed. GitHub CI is green for this head.

PR #484, opened 2026-09-02
feat(client): make provider spawn and CLI boundaries diagnosable

bestony:
## What this changes

Three directories are the densest external I/O in the repository — child-process spawns, HTTPS
downloads, tar extraction, loopback MCP — and on `main` they had **no logging at all**: 37 files,
159 catch sites, zero logs. The shape was a funnel, dozens of precisely classified errors collapsing
into two or three opaque values by the time they reached production.

### 1. Eight structural rejections, one indistinguishable value

`provider-cli/validation-classify.ts` returned `{ status: "needs_attention" }` with no reason from
eight distinct places. This is the mechanism by which a provider CLI upgrade silently breaks OpenTag:
a shape change in the tool's output falls into one of these branches and the operator sees only
"needs attention". All eight now carry a discriminator:

`slack_payload_not_record` · `slack_ok_field_missing` · `slack_ok_not_true` ·
`slack_identity_unparseable` · `lark_payload_not_record` · `lark_success_field_invalid` ·
`lark_identity_field_invalid` · `lark_bot_identity_unparseable`

`extractBoundedJson`'s `catch { return undefined }` was the same problem one level down — over-budget
input, empty input, no brace pair and malformed JSON were four different reasons for one `undefined`.
Now: `json_output_empty` · `json_object_bounds_missing` · `json_output_malformed` ·
`json_output_oversize`.

### 2. Codex was the only provider you could not diagnose after a crash

`providers/codex/app-server-wire.ts` piped stderr to `() => undefined`, so when the process died there
was nothing to look at — while claude-code already kept a bounded stderr buffer and surfaced it on
exit. Codex now keeps the same shape: a bounded 64 KiB tail
(`CODEX_APP_SERVER_MAX_STDERR_BYTES`), surfaced in the abnormal-exit message. The invariant the
original comment stated is preserved — **the wire still speaks only stdout**; stderr is captured for
diagnostics and never parsed as protocol.

### 3. Doctor output stays clean

`login-shell-path.ts` and `agent-runtime-installation.ts` also run under a one-shot `opentag doctor`,
where a `warn`/`info` line would print raw NDJSON into the user's doctor output. Both are
**`debug`-only**; service mode can raise the level through `OPENTAG_LOG_LEVEL` without changing what
doctor prints.

### 4. Everything else

Information-destroying catches across `agent-runtime/`, `providers/` and `runtime/provider-cli/` now
log at the failing boundary — spawn failures, download and extraction failures, MCP handshake
failures. Per-turn and retry-loop paths emit once at the boundary and let the existing fail-closed
state suppress the subsequent rejected work, rather than emitting per iteration.

## Checklist

- [x] 1. All 8 `needs_attention` rejections carry a discriminator
- [x] 2. `extractBoundedJson`'s four `undefined` causes are distinguishable
- [x] 3. Codex keeps a bounded stderr tail, surfaced on abnormal exit, still stdout-only protocol
- [x] 4. `login-shell-path.ts` and `agent-runtime-installation.ts` use debug gating
- [x] 5. Information-destroying catches in these directories now log
- [x] 6. Hot paths aggregate or cool down
- [x] 7. Tests use the existing recording-logger double, never real pino
- [x] 8. Progress record kept current

| criterion | outcome |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm build` | exit 0 |
| `pnpm --filter @opentag/client test:agent-runtime:coverage` (**mandatory for this lane**) | exit 0 — 24 files, 347 tests, **100% statements / branches / functions / lines** |
| `pnpm --filter @opentag/client test` | exit 0 — 67 files, 892 tests (see the flake note) |
| `pnpm test` (root) | **exit 1, environmental** — see below |
| A test distinguishes at least four of the eight validation rejections | pass — **all eight** are asserted, not four |
| A test proves codex's stderr tail survives an abnormal exit **and** is bounded | pass — `codex-app-server-exhaustive.test.ts:61`; the cap is asserted as exactly 64 KiB |
| A test proves the stderr tail is never fed to the protocol parser | pass — the same test writes a valid-looking JSON-RPC frame (`{"id":999,"result":"not stdout"}`) to stderr and proves it is not parsed |
| `opentag doctor`'s stdout carries no log lines | pass — both files log only at `debug`, and `login-shell-path.test.ts:272` asserts every emitted entry is `debug` |
| No provider credential, token or `appSecret` appears in any emitted line | pass — `not.toContain("xoxb-secret")` in two provider-cli suites; additionally every client log call passes through `redactForLog` in `observability/logger.ts` |
| `packages/shared` untouched — no new public enum values | pass — 0 files changed |
| Client coverage at or above the 80 floor | pass — the agent-runtime scope is at 100% |
| No drizzle migration | pass — 0 |
| `scripts/file-size-exceptions.json` unchanged | pass — 0 |
| `git status --porcelain` empty | pass |

### On the root `pnpm test` exit 1

The only failure is `scripts/__tests__/lefthook-stage-fixed.test.mjs`, which makes a signing commit
and dies with `error: 1Password: failed to fill whole buffer`. The same failure reproduces
identically on clean `main` at `e299cea4`, and this branch touches no file under `scripts/`. It is
the verifying host's 1Password agent being locked (`ssh-add -l` reports no identities), not a defect
here. CI signs nothing, so the job is expected to pass there.

### On one flaky client test — stated in full, because this one deserves scrutiny

The first full `pnpm --filter @opentag/client test` run failed one test:
`claude-code-agent-runtime-exhaustive.test.ts` → "runs the default process adapter and the default
local readiness probe against controlled artifacts", timing out at **5008 ms against a 5000 ms
budget**. Unlike a typical flake, this test's subject *is* in this branch's diff
(`providers/claude-code/agent-runtime.ts`, `process-wire.ts`), so it was not dismissed on that basis.
What was checked:

- the file passes in isolation — 36/36;
- the full client suite passes on re-run — 67 files, 892 tests;
- `test:agent-runtime:coverage`, which exercises the same area, passes;
- the overshoot is 8 ms on a 5000 ms budget, the signature of contention rather than a systematic
  slowdown, and the machine was running three verification suites plus other agents concurrently.

Judged environmental. If CI reproduces it, that judgement is wrong and the test needs a real timeout
budget rather than a re-run.

## Notes for the reviewer

- **Code vocabulary differs from the sibling PRs in this batch.** This lane uses lowercase
  `snake_case` diagnostic codes (`cli_probe_failed`, `json_output_oversize`); the server and
  transport lanes use `SCREAMING_SNAKE` (`IM_BINDING_CREDENTIAL_DECRYPT_FAILED`). Both are internally
  consistent; unifying them is a follow-up, not something to fix by rebasing three branches.
- `CODEX_APP_SERVER_MAX_STDERR_BYTES` is added to `packages/client`'s public-export snapshot so the
  bound can be asserted rather than duplicated in the test.

## Deviation from the confirmed plan

None. The plan's boundaries held: `packages/shared` untouched, no new public enum values, no
migration, and none of the files owned by the concurrently running transport lane were edited.

---

*Provenance: written by codex under `/dispatch-codex` run `337683`, lane `client-provider-spawn-3`,
running with approvals and sandbox bypassed inside an isolated git worktree. Verified and published
by the orchestrator; the acceptance table above records checks the orchestrator re-ran itself.*


yuezengwu (changes requested):
Exact head 02bb64b6cb7ce8b17bdcd96238fcf16a5f5e36cb has two blocking observability defects. (1) Most new diagnostics create loggers at module evaluation time (for example BaseAgentRuntime.#logger and the provider/runtime module-level logger constants). The CLI imports the @opentag/client barrel before configureClientLoggerForService runs, and createLogger closes over the current Pino instance/destination. Those adapters therefore remain bound to the pre-configuration stderr/silent logger and do not switch to client.log or pick up the managed daemon log level; the diagnostics this PR adds will be missing from the service log. Please inject a configured logger, create it lazily after configuration, or make logger adapters reconfiguration-aware, with an import-before-configure regression test. (2) CodexAppServerProcess.#onStderr retains the first maxStderrBytes, not a tail: once the buffer is full, later chunks are discarded. The new test codifies the prefix (the expected message is the first eight bytes), so a long-running process loses the final crash reason this feature is intended to surface. Please retain the last bounded bytes and assert a marker written at the end survives. Request a fresh exact-head review after both are fixed.

bestony:
Both defects addressed — thanks, they were real and I confirmed each one against the code before dispatching the fix.

## 1. `#onStderr` kept the head, not the tail (`3acbcc0f`, `c3d8572e`)

You were right that the old code discarded every chunk once the buffer filled. It now trims from the front in both paths — an oversized single chunk keeps its last `maxStderrBytes`, and the accumulate path concatenates then drops the leading excess.

The test no longer codifies a prefix. It writes the 32-byte frame, then `TAILMARK` (exactly the 8-byte cap), and asserts:

```ts
message: "Codex App Server exited: TAILMARK",
...
expect((await failure).message).not.toContain('{"id":99');
```

so the trailing marker surviving *and* the leading bytes being gone are both pinned. A second test covers the short-chunk path: `"1234"` + `"56789"` with a cap of 8 yields `23456789` — the leading `1` is dropped. The stdout-only invariant is unchanged and still asserted (the frame written to stderr is never parsed as protocol).

## 2. Module-evaluation-time loggers never reached `client.log` (`08c87d9e`)

Also confirmed: `createLogger()` with no options resolved `root()`, which memoizes, and `adapt()` captured that pino instance by value — so `configureClientLoggerForService` clearing `rootLogger` could not rebind anything already constructed at import time.

Fixed at the adapter rather than at ~12 call sites: `adapt()` now takes a `() => PinoLogger` thunk and resolves it **at emit time**. The default path passes `root` itself, so it re-resolves after configuration; explicit-destination loggers stay pinned via `fixed()` (a caller who asked for stderr should keep stderr); `child()` propagates the thunk so child loggers stay reconfiguration-aware too.

The regression test uses exactly the ordering you named — import first, configure second:

```ts
vi.resetModules();
const { AgentRunEventValidator } = await import("../../agent-runtime/event-validator.js");
const configuredLogger = await import("../logger.js");
configuredLogger.configureClientLoggerForService(directory);
// ...trigger a real validation failure...
await expect(readFile(join(directory, "client.log"), "utf8")).resolves.toContain("Provider event validation failed");
expect(stderrWrites.write).not.toHaveBeenCalled();
```

The `stderrWrites.write` assertion is what makes it a real regression test: before the fix the line went to stderr and the file would not exist.

## Verified on the new head `c3d8572e`

| gate | result |
| --- | --- |
| `pnpm --filter @opentag/client test:agent-runtime:coverage` | exit 0 — 24 files, 348 tests, **100% statements / branches / functions / lines** |
| `pnpm check` / `pnpm typecheck` / `pnpm build` | all exit 0 |
| `pnpm --filter @opentag/client test` | exit 0 — 67 files, 894 tests, first run, no flake |
| `pnpm test` (root) | exit 1 — **only** `scripts/__tests__/lefthook-stage-fixed.test.mjs`, which makes a signing commit and dies on `1Password: failed to fill whole buffer`. Reproduced identically on clean `main`; this branch touches no file under `scripts/`. The verifying host's 1Password agent is locked (`ssh-add -l` reports no identities). CI signs nothing. |
| `packages/shared` untouched | 0 files |
| no migration / `file-size-exceptions.json` unchanged | 0 / 0 |

Note on the flake mentioned in the original description: `claude-code-agent-runtime-exhaustive.test.ts` passed on the first full run this time, which supports the contention diagnosis rather than a real slowdown.

Requesting a fresh review against exact head `c3d8572e`.

yuezengwu (approved):
Reviewed exact head `c3d8572e78e1006382580e4e59b782e7e18c2cb8` against base `e299cea4f6653b1b8b59c2e43da1a92cb74b7e79`. The new head fixes both prior blockers: configured loggers now re-resolve the service root at emit time while explicit destinations remain pinned, and the Codex stderr buffer keeps the last bounded bytes across oversized and short chunks. I re-reviewed the complete effective diff. Validation: client tests 67 files / 894 tests, root typecheck 8/8, `pnpm check`, `git diff --check`, and exact-head GitHub CI/coverage checks passed. No blocking finding remains.

Issue #478, opened 2026-09-02
fix: align topic-group conversation listing with response and topic semantics


bestony:
## Problem

The current conversation list for topic groups does not consistently reflect meaningful user-visible sessions. Entries may be created for conversations that received no Agent response, and follow-up messages within the same topic can create duplicate records.

## Proposed conversation grouping and filtering rules

1. **Casual groups**
   - Include a session in the default list only when the relevant Agent actually responds.
   - Filter out sessions with no Agent feedback, since they are not useful to users in the default view.

2. **Standard topic groups**
   - Treat one topic as one conversation record.
   - Subsequent messages belonging to the same topic must update the existing record rather than create additional list entries.

3. **Casual groups converted to topic groups**
   - Use the same grouping behavior as standard topic groups.
   - Conversion should preserve the existing conversation record where possible instead of creating a parallel entry.

4. **Routine user-to-user direct chats**
   - If the user does not invoke `/new`, continue using the same topic/conversation.
   - These routine direct-chat sessions may be filtered from the default list when they do not meet the criteria for a meaningful Agent conversation.

## Acceptance criteria

- The default conversation list excludes casual-group sessions with no Agent response.
- Messages in one standard topic group appear under a single conversation record.
- A casual-to-topic conversion does not produce duplicate records for the same topic.
- Direct-chat messages stay in the same conversation until `/new` is used.
- Grouping and filtering behavior is covered by automated tests for all four conversation types.
- Existing conversations remain accessible through an appropriate history/search path when they are filtered from the default list.

## Open questions

- What exactly qualifies as an Agent response (message type, sender identity, or delivery status)?
- Should filtered sessions be hidden only from the default view, or also excluded from counts/search results?
- How should topic identifiers and conversion events be handled for historical records?

Issue #419, opened 2026-09-01
Running status should not be gray


bestony:
## Problem
The UI currently renders the running status in gray.

## Expected behavior
A running/in-progress status should use an active color treatment so it is visually distinct from inactive, queued, or completed states.

## Actual behavior
The running status appears gray, which makes it look inactive or disabled.

## Acceptance criteria
- Running status is rendered with a clearly active, non-gray color.
- The color remains accessible and consistent across the relevant status views.
- Inactive/completed statuses retain their existing visual treatment.

bestony:
补充原消息中的截图：

![Running status screenshot](https://raw.githubusercontent.com/first-tree-ai/opentag/main/.github/issue-assets/issue-419-image.png)

Issue #408, opened 2026-08-31
Agent homepage should show only active tasks


bestony:
## Proposal

On the Agent homepage, the **Tasks** section should focus on work that is currently in progress and display only active/running tasks. Completed tasks should be available from the dedicated Tasks page instead of remaining in the homepage list.

## Motivation

The Agent homepage is primarily a real-time overview. Showing completed tasks there makes it harder to identify the work that still needs attention; users currently care most about tasks that are running or otherwise not yet finished.

## Expected behavior

- Filter the Agent homepage Tasks section to active/running tasks.
- Exclude completed tasks from that homepage section.
- Keep completed tasks accessible through the dedicated Tasks page and its existing history/filtering capabilities.
- Define the treatment of other non-active states (for example, failed, canceled, or paused) consistently with the task status model.

## Acceptance criteria

- A completed task does not appear in the Agent homepage Tasks section.
- A running/active task appears in the Agent homepage Tasks section.
- Completed task history remains accessible from the Tasks page.
- Empty and loading states remain clear when an Agent has no active tasks.

Issue #407, opened 2026-08-31
feat: represent topic-group tasks as parent topics with child messages


bestony:
## Requirement

For topic-based group chats, the task list should represent the conversation using a parent-child structure:

- Show only the root topic as a task-list item.
- Do not show individual messages from that topic as separate top-level task-list items.
- Messages should be viewed within the corresponding topic thread.
- `Participate in discussion` should open or operate on the parent topic context.

## Expected behavior

1. A topic group is rendered as a collection of root topics in the task list.
2. Replies/messages are treated as children of their root topic and remain hidden from the top-level task list.
3. Opening a root topic exposes its child messages in the topic view.
4. Any discussion action preserves the parent topic context.

## Acceptance criteria

- No child message appears as an independent top-level item in a topic group's task list.
- Every child message can be accessed from its root topic.
- The parent-child relationship remains stable after refresh and navigation.
- Non-topic groups retain their existing task-list behavior.

PR #488, opened 2026-09-02
feat(tasks): project Tasks as topics over stored IM messages

Gandy2025:
## Summary

The Tasks list showed one permanent row per Session, titled and statused by whatever message arrived last, so a group's row flipped between unrelated requests, an overheard "what's for lunch" could retitle it, and every thread message produced a Channel/Thread pair of rows.

A Task is now a read-only projection over stored inbound messages and their deliveries:

- In a group, channel, or multi-person direct message a Task is one topic: the root message plus the reply chain around it, keyed by `coalesce(thread root, thread_key, external_message_id)`. The thread root maps a provider thread id (Feishu `thread_id`) back to the root message it hangs off via `provider_context.rootId`; Slack's `thread_ts` is already the root id.
- A private chat is one Task.
- A topic is listed only once somebody addressed the Agent directly. The channel Session's ambient observer copy of a message the thread Session owns, and deliveries expired by a superseded revision, are not executions.
- The Task id is the topic's earliest message; the detail endpoint accepts any message of the topic and reports the canonical id. `createdAt` is the root message time.
- Title: manual > generated > derived from the root message. `PATCH /api/v1/sessions/:id` accepts a Task id or a Session id and writes to the Session the topic reads its title from (thread Session, or the private chat's channel Session); a top-level group request nobody replied to has no such Session and returns 404.
- Status precedence: channel disconnected, running, queued, then the latest outcome. An accepted delivery counts as running only while its deadline has not passed, its Session is alive, and no later Turn was accepted in that Session (a window function over the Session's acceptances). An unreported execution past its deadline shows as expired instead of running forever.
- Internal Sessions and collaboration messages: those that inherited the topic scope, plus those a channel Session spawned inside one of the topic's Turn windows (closed at the Session's next acceptance), with their descendants.
- The list pages first: per-topic rollups (`GROUP BY` over messages and over executions) decide the order, cursor, and `LIMIT`, and only the page's rows resolve title content, the latest execution, and their Sessions. Detail and title lookups are scoped to the channel and topic.
- The clock is injectable (`TaskServiceOptions.now`) so tests pin it.

Delivery, Session materialization, routes, the shared contract, mock data, and the Web are unchanged. `sessionKind` reads `thread` (with `source.threadKey`) only for a topic that actually has a reply chain; a private chat and a top-level request nobody replied to read `channel` with no thread key, so the existing Web labels render "Direct message", "Group chat", or "Channel" from `conversationKind` as before, and "Thread" only for threaded topics. The dev preview shapes remain producible.

Relates to #407, #408, #478 (rules 2-4 are covered by the grouping; rule 1 uses "addressed directly" because outbound is not observed), and the diagnosis behind #419/#420/#464/#466 which stay separate.

## Product acceptance of the data-access change

The id change (Task ids are message ids; `GET` no longer accepts Session ids) and the visibility change (ambient-only conversations and empty Sessions are not Tasks) were decided by the product owner (@Gandy2025) in the design review that produced this change: a Task is the owner's observation of a conversation with the Agent, so a Session nobody addressed is not one. The reviewable source is now in this PR: [`docs/tasks.md`](https://github.com/first-tree-ai/opentag/blob/feat/tasks-topic-projection/docs/tasks.md) (canonical, with the `docs/zh-CN/tasks.md` mirror) states the Task definition, the listing rule, the status precedence, the id and `PATCH` contract, and the boundaries. The ownership gate still needs a code-owner approval; the product decision itself is the owner's.

## Testing

- `pnpm --filter @opentag/server exec vitest run src/services/tasks/__tests__/task-service.test.ts` — 19 passed (PGlite), including a scale regression: 2,500 topics / 5,000 messages / 5,000 accepted deliveries in one Session, first page of 50 in well under 5 s (seed plus two pages: 5.6 s total in the harness; the pre-fix query did not return in 5 minutes).
- `pnpm --filter @opentag/server test:integration` for `src/__tests__/integration/tasks.test.ts` — 7 passed (Postgres via testcontainers).
- `pnpm --filter @opentag/server test` — 63 files passed.
- `pnpm check`, `pnpm typecheck`, `pnpm build` — pass (pre-existing warnings only, 0 errors).
- Manual end-to-end on the built server with seeded Feishu and Slack data: the list shows one row per topic and one per private chat, overheard-only chatter and an empty DM are gone, a reply id opens its root topic with the absorbed follow-up, `PATCH` with a Task id renames a thread-backed topic (200) and a root-only topic returns 404.
- `pnpm test:e2e` journey: first full run 9/11 with two failures unrelated to Tasks (Agent navigation and runtime-test timeouts); later reruns could not start on this machine because the host cannot reach the compose-published Postgres port. `e2e/tests/fixtures.ts` returns the seeded message id as the Task id so the Task detail journey keeps working.
- `pnpm test` at the root also reports two `apps/cli` daemon tests failing locally on macOS (`daemon-ensure-service`, `daemon-service-commands`); this change does not touch `apps/cli` and CI on the base commit is green. The current CI red (onboarding "Start OpenTag" assertion) reproduces on `main`.
- Independent review pass (eight finder angles, verified): seven confirmed findings fixed with tests first; the reviewer's P1 (unbounded list aggregation) is fixed by paging first and replacing the correlated running check with a window function; one unreachable `idle` fallback kept because the shared enum still advertises it.

## Breaking changes

- Task ids in `GET /api/v1/sessions` and `GET /api/v1/sessions/:id` are message ids, not Session ids. A Session id is no longer accepted by `GET`; `PATCH` accepts either.
- Sessions with no direct delivery are no longer listed, so `idle` is no longer produced, and a group's channel Session no longer appears as a Task.
- `PATCH` on a group's channel Session, or on a root-only topic, returns 404 instead of 200.
- `createdAt` means the root message time rather than the Session creation time.

Known residuals, deliberately left for follow-up: a root Turn on the channel Session that crashes stays running until that Session accepts again or the delivery deadline passes; a Feishu topic whose root carries no `thread_id` and whose replies carry no `rootId` would still split; the revision-ordering convention is hand-written here until the im module exports one.

## Checklist

- [ ] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. — all pass except the two pre-existing macOS-only `apps/cli` daemon test failures noted above.
- [x] Canonical English docs and Chinese mirrors are synchronized when applicable. — no canonical doc describes the Task list semantics; none changed.


yuezengwu (dismissed):
Exact-head review for 1fdde154d140bb987c5e2d423faed02fb4ab8b3b.

[P1] Bound the Task list before aggregating the Account history

TaskService.list builds classified_messages, executions, topics, latest_execution, anchors, and title_sources over every inbound message and delivery in the Account, and applies LIMIT only in the final select (task-service.ts:294-488, 794-849). Pagination repeats the same full aggregation on every page. This is already called out in the PR as O(account history), but it is an availability blocker rather than a follow-up optimization: in the repository PGlite harness, seeding 5,000 messages took 144 ms while fetching the first 50 Tasks still had not returned after more than 50 seconds and had to be stopped. Please bound candidate topics before the expensive projection, or persist/index the projection, and add a scale regression test.

The current CI failures are not attributed to this head: main@37a86739 reproduces the stale onboarding assertion (Start OpenTag versus Run opentag daemon start), while Browser Smoke and patch coverage pass here. Focused validation on this exact head passed: 18 PGlite Task tests, 7 real-PostgreSQL Task integration tests, pnpm check, pnpm typecheck, and git diff --check.

This change also intentionally removes GET compatibility for Session ids and makes filtered ambient-only conversations unreachable through the Task detail query, so its declared data-access/API break still needs explicit product/architecture acceptance; I have not treated that decision as implicitly approved.

yuezengwu (commented):
Exact-head review for 2dd1095a1c337fd15732683176ad38f2c1d5d171: no technical finding.

The prior performance blocker is repaired. The repository PGlite scale case now reads the first 50 Tasks from 5,000 messages in about 1.7 seconds (previously still running after 50 seconds), and the added regression covers a second page. Exact-head validation passed: Task unit tests 19/19, real-PostgreSQL Task integration tests 7/7, Server tests 637/637, pnpm check, pnpm typecheck, and git diff --check.

The current CI failures are not attributed to this head: they are the stale onboarding assertion introduced on main@37a86739 (Start OpenTag versus Run opentag daemon start); Browser Smoke passes on this head.

I dismissed the stale REQUEST_CHANGES review attached to 1fdde154. I am not auto-approving 2dd1095a because the PR deliberately changes Task ids, removes GET-by-Session-id compatibility, and makes ambient-only history unavailable through Task detail. The PR records product-owner acceptance in an uncommitted external design note, but this is still a breaking product/data-access decision that needs explicit maintainer acceptance or a canonical reviewable source.

yuezengwu (commented):
Exact-head review for 8e7eac09fe167be9056d36836c6148ca769cbfd1: no technical finding.

The performance blocker remains repaired after merging main. Canonical docs now define the Task projection, and the English/Chinese sources align with the implementation on topic grouping, direct-attention listing, message-based Task ids, GET/PATCH behavior, status precedence, and the 404 boundary for a root-only topic without a titleable Session. The merge resolution only reconciles the E2E connect-code parser with the current main format.

Exact-head validation passed: Task unit tests 19/19, real-PostgreSQL Task integration tests 7/7, full Server tests 714/714, pnpm check, pnpm typecheck, and git diff --check. Browser Smoke, CLI Pack Smoke, Container Smoke, Documentation Mirror Advisory, and the PR image build also pass; the remaining CI jobs are still running.

I am not auto-approving this exact head because the PR intentionally makes breaking product/data-access changes: Task ids switch to message ids, GET-by-Session-id compatibility is removed, and ambient-only or empty Sessions disappear from the Task surface. The canonical docs make that decision reviewable, but the ownership gate still requires explicit code-owner/maintainer approval.

liuchao-001 (commented):
Exact-head review of `8e7eac09fe167be9056d36836c6148ca769cbfd1`.

I re-derived the projection against the schema and the delivery worker rather than against the PR description, and independently re-ran the Task unit suite at this head: **19/19 pass**. The merge of `main` leaves `task-service.ts` byte-identical to `01d8c965` and only reconciles the E2E connect-code parser, so prior analysis carries.

The topic grouping holds up where I tried hardest to break it. `thread_roots` requires `rootId <> thread_key`, which is exactly right given `feishu/adapter.ts:378` sets `threadKey: threadId ?? rootId ?? null`: an ordinary Feishu reply keys on `thread_key == rootId` and the root keys on its own `external_message_id` — the same value — while a topic group keys both through `thread_roots`. The thread Session's `topic_key` is computed from the same CTE as the messages', so a root-only topic that later gets a reply silently gains a titleable Session and `PATCH` flips 404→200 with no id change. Row multiplication is impossible at every join (each is `distinct on` or a PK), and the `accepted_windows` half-open interval is disjoint per Session.

One finding.

### The `sessionKind` change mislabels every non-DM Task as "Thread"

`sessionKind` is now `case when p.topic_key is null then 'channel' else 'thread' end`, and `topic_key` is NULL only when `ch.conversation_kind = 'dm'`. Since `conversationKind` is read from that same `ch`, the server can now emit `sessionKind: 'channel'` **only** together with `conversationKind: 'dm'`. Every channel and every multi-person DM topic — including a top-level request nobody replied to, which has no thread at all — reports `sessionKind: 'thread'`.

`apps/web/src/features/tasks-page.tsx:788` branches on exactly that field first:

```ts
task.sessionKind === "thread"
  ? m.tasks_source_thread()
  : task.source.conversationKind === "dm"
    ? m.tasks_source_direct_message()
    : task.source.conversationKind === "group_dm"
      ? m.tasks_source_group_chat()
      : m.tasks_source_channel();
```

So a plain `#general` message that nobody replied to now renders **"Feishu · Thread"** where it used to read "Feishu · Channel", a group DM topic reads "Thread" instead of "Group chat", and `tasks_source_channel` / `tasks_source_group_chat` become unreachable from real server data in both locales. `source.threadKey` is also populated with the root message's `external_message_id` for those rows, i.e. a thread key for a conversation that has no thread; it only feeds the search filter today, so that part is harmless, but it is worth being deliberate about.

The description says `sessionKind` "now reads `thread` for a topic and `channel` for a private chat, which the existing Web labels render as 'Thread' and 'Direct message'." That is true for the two cases named and hides the third: the channel and group-chat rows also say "Thread" now.

Corroborating evidence that the surfaces drifted apart: `apps/web/src/mock/dev-task-data.ts` still ships `sessionKind: 'channel'` paired with `conversationKind: 'channel'` and with `'group_dm'` — combinations the new projection can never produce. "Mock data unchanged" is listed as a non-change, but it now contradicts the server contract, and those preview rows are the only remaining way to reach the two orphaned strings.

This is presentation, not data correctness, so I do not think it blocks the projection itself — but it is a user-visible regression on the main Tasks list and it is cheap to fix. Either derive the label from `source.conversationKind` first and use `sessionKind` only to distinguish a thread within a channel, or keep `sessionKind` describing the conversation and carry "is this a topic" separately. Whichever you pick, please realign `dev-task-data.ts` so the preview reflects shapes the server can actually emit.

### Not findings, recorded so they are not re-litigated

- **`queued` has no deadline guard** while `running` does (`bool_or(state = 'pending')` vs. the `expires_at > now` term in `is_running`). I chased this expecting a "queued forever" twin of the bug you fixed. It is not one: `im-delivery-worker.ts:255` unconditionally expires every overdue `pending` delivery globally at the top of each `#claim()`, and `#claim` is driven by a `setInterval`, so the state self-heals within one poll regardless of account activity. Leaving it asymmetric is fine.
- **`taskStatus`'s `row.latestRunning ? "running" : "expired"`** is unreachable in its `running` arm — `hasRunning` is `bool_or(is_running)` and is checked earlier, so `latestRunning` is necessarily false by that line. Harmless, same class as the `idle` fallback you deliberately kept.
- **`collaborationMessages` is fetched unbounded** (no `LIMIT`, no cursor) for every related Session. Identical in the base implementation, so it is pre-existing and out of scope here — flagging only so the detail endpoint is not assumed to be fully paged.
- **`saveGeneratedTitle` has no production caller** and nothing else writes `generated_title`, so the middle tier of the manual > generated > derived precedence is inert today. Also pre-existing; the docs read as though it is live.

Not approving: I have the label finding above, and per the ownership gate the breaking id/visibility change is the code owner's call, not mine.

liuchao-001 (approved):
Exact-head re-review of `cf8d8602509c0fb609eff22d3eedeeac5a785e1b`. My finding is fixed correctly, and fixed on the right side of the boundary — the server stopped claiming a thread that does not exist, rather than the Web being taught to ignore the field. Approving.

I re-ran the Task unit suite at this head: **19/19 pass**, including the two cases you tightened.

What I checked about the fix itself, since it lands on the field the `kind` filter and the row label both key off:

- `has_thread` is `bool_or(thread_key is not null)` over the topic's messages, not the anchor's own `thread_key`. That is the detail that makes it correct: a Slack root carries no `thread_ts` until it has replies, so keying off the anchor alone would have left every Slack thread reading `channel`. Folding over the whole topic catches it from the first reply.
- `is_thread` is one expression reused by both the `kind` filter and the projected `sessionKind`, so the filter can no longer disagree with the label it filters on. `kind: "channel"` and `kind: "thread"` partition the set exactly — `bool_or` over a non-empty group of strict booleans is never NULL, so the `not (...)` branch cannot silently swallow rows. The reworked filter test pins that partition across all three shapes.
- `threadKey` is now NULL exactly when `sessionKind` is `channel`, which matches what `dev-task-data.ts` already emits (`sessionKind === "thread" ? preview-thread-N : null`). All four preview combinations became producible again without touching the mock, so the drift I flagged closed from the server side.
- The integration suite asserts nothing on `sessionKind` or `threadKey`, so it is unaffected by this commit.
- The `'channel'` → `'thread'` flip when a first reply lands mirrors the `PATCH` 404 → 200 flip on the same event, with the Task id unchanged. Consistent.

One wording nit, non-blocking and no code change needed. `docs/tasks.md` now says a Task reports `thread` "only once its topic actually has a reply chain", but the implementation reports `thread` whenever any message in the topic carries a thread key. For a Feishu topic group those differ: a root posted into a topic group carries `thread_id` from the start, so it reads `thread` with zero replies. The behaviour is right — it genuinely is a thread container — but the sentence is stricter than the code. "once the topic has a thread key" would match.

The ownership gate is untouched by this approval: `.github/CODEOWNERS` puts `*` under `@bestony @yuezengwu` in gate mode, and I own only `*.md`, so this is a reviewer approval and not the code-owner sign-off that the breaking id/visibility change still needs.

liuchao-001 (approved):
Re-approving at `d5026f74291a23758ebf4a406a3b52e3f4b30fb7` so the approval binds to the current head rather than to `cf8d8602`.

`d5026f74` is docs-only and is exactly the wording nit from my last review. The new sentence is accurate against the implementation, and it is more precise than what I suggested — naming where the first thread key comes from in each case rather than just relaxing the claim:

- Slack: the root carries no `thread_ts`, so the first reply supplies it.
- Ordinary Feishu group: `adapter.ts:378` is `threadId ?? rootId ?? null`, so the root gets NULL and the first reply carries `rootId`.
- Feishu topic group: the root itself carries `thread_id`.

The `docs/zh-CN/tasks.md` mirror is synchronized and makes the same three-case statement.

No code changed between `cf8d8602` and this head, so my earlier verification carries unchanged: Task unit suite 19/19 at `cf8d8602`, and the `is_thread` analysis (single shared expression behind both the `kind` filter and the projected `sessionKind`, exact partition, `threadKey` NULL exactly when `sessionKind` is `channel`) applies verbatim.

Ownership gate unchanged and still open: `*` is `@bestony @yuezengwu` in gate mode, so the breaking id/visibility change still needs one of them. My approval remains a reviewer approval only.

yuezengwu (commented):
Exact-head review for d5026f74291a23758ebf4a406a3b52e3f4b30fb7: no blocking technical finding.

The sessionKind/threadKey correction remains sound, and this docs-only follow-up now states the cross-provider boundary accurately: Slack and ordinary Feishu groups acquire the thread key on the first reply, while a Feishu topic-group root carries it from the start. The English and Chinese canonical docs are aligned.

Exact-head validation passed: Task unit tests 19/19, real-PostgreSQL Task integration tests 7/7, full Server tests 714/714, pnpm check, pnpm typecheck, and git diff --check. Browser Smoke, CLI Pack Smoke, Container Smoke, Documentation Mirror Advisory, the PR image build, and the completed policy/security jobs also pass; the remaining CI matrix jobs are still running.

I am not auto-approving because this PR intentionally changes Task ids, removes GET-by-Session-id compatibility, and removes ambient-only or empty Sessions from the Task surface. Those breaking product/data-access semantics still require explicit code-owner/maintainer acceptance.

PR #492, opened 2026-09-02
fix(tasks): read Feishu rich text once and settle the Task detail header

Gandy2025:
## Summary

Follow-up to #488 from a design review of the Task detail page. Four things the page showed were traced to their source; three of them were data, not layout.

- **Feishu rich-text messages were read twice.** Feishu sends every paragraph of a `post` message under `content` (tagged elements) and again under `content_v2` (markdown). The adapter's recursive string walker collected both, so a message became its own duplicate in the stored text, in the Task title, in the Turn shown on the Task page, and in the prompt handed to the Agent. The parser now reads the documented shape instead: paragraphs of tagged elements (top-level or locale-keyed), runs of one paragraph on one line, an `@` element written as its mention key so it is recognised like a plain-text mention, and the markdown copy read only when the tagged paragraphs carry no text. Rows stored before this change keep their text.
- **Bounded titles end with an ellipsis.** A title cut at the 80-grapheme bound used to stop mid-sentence with no marker.
- **The title has its own row; the status joins the Task details, last.** The status badge shared a `flex-wrap` row with the title, so its position depended on the title's length. It now sits after Last activity in an equal-column details grid (five columns at desktop, three at the mid width, stacked on mobile), so neither the title nor a long Agent name can move it.
- **A failed Turn names its reason for the reader.** The raw `TurnFailureReason` code (`credential_unavailable`) was rendered as the Agent's reply. Each code now has en/zh copy; a code this build does not know is still shown, humanized.
- **A Turn's message renders mentions by label.** The stored fallback keeps Feishu's `@_user_1` placeholder; the blocks already carry `@Atlas`, so a message tokenised into text and mention blocks is rendered from them. Any other block kind keeps the lossless fallback, so no content can be dropped.

Content width and the title size were checked and left alone: the page sits in the shared `max-w-5xl` frame and the title is the same `size="lg"` heading every page uses.

## Testing

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/e2e test:e2e:smoke` (8 passed) and `pnpm --filter @opentag/e2e e2e:journey` (11 passed, 1 failed, 4 skipped: the pre-existing "Agent navigation reaches every Agent-owned destination" expects Skills and Integrations entries that the sidebar has hidden behind the internal-navigation flag since #476, and its serial file skips the four tests after it; unrelated to this change, same result on two clean runs)
- New tests: Feishu `post` with `content_v2` is read once, runs of one paragraph stay on one line with the `@` element as its mention key, and the markdown copy is used only when the tagged paragraphs carry no text; title truncation appends an ellipsis, never after a space, and still satisfies `TaskTitleSchema`; `messageTextFromBlocks` renders labels and leaves any other block kind to the fallback; a Turn shows the mention label; the detail header keeps the title alone and lists Status last; a failure code renders as its label.
- Local `pnpm test` also reports two `open-tag` (CLI) failures in `daemon-ensure-service` / `daemon-service-commands`: they read this machine's real Computer credentials file instead of an isolated home and are unrelated (flagged separately).

## UI validation

- Desktop evidence: local stack at 1440px with an 80-grapheme Chinese title ending in an ellipsis: title on its own rows, details in one row with Status last; a failed example shows "凭据不可用" in the Agent card.
- Mobile evidence: 390px: the title wraps to four lines, the details stack in one column with Status last, no page-level horizontal scrolling.
- Widths checked (`320px` / `390px` / `768px` / `1440px`): all four on the local stack. 320px: no page-level horizontal scrolling (`scrollWidth` equals `clientWidth`), details stacked with Status last; 768px: three columns; 1440px: five equal columns.
- States verified: completed, failed (reason label), long bounded title, DEV example data.
- Keyboard/accessibility checks: no new interactive controls; the details `dl` keeps its `Task details` name and `dt`/`dd` pairs, and the page test pins the term order; the status stays a Kumo dot Badge.
- New reusable block, theme value, or CSS exception: none. The details grid keeps its container-query breakpoints, now three and five columns for five facts.

## Breaking changes

None. The Task API shape is unchanged. A Turn's `message.fallbackText` now carries mention labels instead of provider placeholders when the message has mentions; the Web page is its only consumer.

## Checklist


liuchao-001 (approved):
Reviewed at `d8f76a7b`, diffed against `0e3d9cd7` in a clean worktree. I also ran the two affected suites locally after building `@opentag/shared`: `@opentag/server` 67 files / 721 tests and `@opentag/web` 58 files / 562 tests, all passing. CI on this head is green except `ownership-gate`, which still wants bestony or yuezengwu on 7 files.

The duplication fix is right where it counts: the parser reads the tagged `content` first and consults `content_v2` only when those paragraphs carry no text, so the dedup holds whatever shape Feishu actually puts in the markdown copy — it isn't betting on `{tag:"md"}` looking a particular way. Three things I chased to the end and can confirm:

- **`turnText` cannot drop content.** Feishu blocks are tokenised from `fallbackText` itself, so re-rendering text+mention blocks reproduces the same string with labels substituted; any other block kind returns `undefined` and keeps the fallback. Rows written before the tokenizer (#39 stored a single text block) carry no mention block, so they keep the fallback too. On Slack the mention label *is* the `<@U…>` token, so the new path is a no-op there — the win is Feishu-only, as intended.
- **The title bound stays legal.** 80 graphemes + `…` is 81, inside `TASK_TITLE_MAX_LENGTH` (120), and `cleaned` is already collapsed and trimmed, so `trimEnd()` can only fire on the boundary space.
- **The failure map is exhaustive by construction.** `Record<TurnFailureReason, …>` covers all 14 enum members, so a new reason breaks the build rather than leaking a code to a reader.

Non-blocking, in the order I would act on them:

1. **The post's `title` is still dropped.** `postText` reads only `content` / `content_v2`, so a rich-text message whose ask sits in the heading — `{"zh_cn":{"title":"把 446 关掉","content":[[…]]}}` — reaches the stored text, the Task title and the Agent prompt without it. The old walker dropped it too, so this is not a regression; but this is the change that reads the documented shape, and the title is authored content, not chrome. Reading a non-empty `title` as the first line would close it.

2. **The `at` lookup can match on `undefined`** (`adapter.ts:293`): `candidate.id.open_id === user_id || candidate.id.user_id === user_id`, where `user_id` comes off the raw element as `unknown` and both id fields are optional on the mention. An `at` element that arrives without `user_id`, next to a mention entry without `open_id`, matches that first entry and writes *its* key — the text then names the wrong person, and `contentBlocksWithMentions` tokenises that key into a mention block for the wrong `externalId`. Neither half is likely on today's payloads, which is what makes `typeof user_id === "string" && user_id` before the `find` cheap insurance.

3. **A paragraph that renders no text becomes a blank line.** The runs test pins `…the docs\n\nbefore 5pm` for an image-only middle paragraph, so a reader gets a paragraph break for something they cannot see (post images are not attached as resources either — pre-existing). Skipping paragraphs that render empty, while keeping genuinely empty ones, would read closer to what Feishu shows. Defensible either way; your call.

On the header: no objection from me, though the placement is bestony's call. One thing worth a look — at the 64rem frame each of the five columns lands near 165px, and `grid-cols-5` is `minmax(0,1fr)`, so a long unbreakable Agent display name spills across its neighbour rather than widening the grid. The four widths you checked used the fixture name.

Approving. Ownership still needs a code-owner.

yuezengwu (approved):
Approved on exact head `d8f76a7b6ff4593f7249f4af2917900b2a7f7704` against current `main` (`23da031e0a2381675755f0f737ebf9c2c6f74fa1`).

Reviewed the Feishu rich-text single-source parsing, mention-label rendering at the Task API boundary, grapheme-aware title ellipsis, localized failure reasons, and Task detail header layout. The head merges cleanly with current `main`.

Local verification on this exact head:
- `git diff --check`
- `pnpm build`
- affected server suites: 57 tests passed
- affected web suite: 43 tests passed
- `pnpm check`
- `pnpm typecheck`

Remote CI is green; the remaining `ownership-gate` status is the expected code-owner approval gate this review satisfies.

PR #495, opened 2026-09-02
test(web): take the clock out of the assertions that raced it

liuchao-001:
Addresses #325 in part. Please read the scope section — this does **not** fix every failure that issue records, and I do not think it can.

## The issue's file no longer exists

#325 is written about `apps/web/src/__tests__/app.test.tsx`. That file was deleted by #388, which split the app suite by bounded context, so its tests now live in `account.test.tsx`, `agent-home.test.tsx`, `app-shell.test.tsx`, `agent-tasks.test.tsx` and `agent-usage.test.tsx`. The named failures in the issue's table map onto those files. The phenomenon is real; only the filename is stale.

## What actually couples a test to the clock — two places, both fixed

**1. A 40 ms window that had to outlast the assertions** (`account.test.tsx`, the test #306 names)

The test holds a `/api/v1/me` refresh in flight and asserts what is true *while* it is in flight: the field is disabled, a second retry does not start, a submit cannot slip a competing write past it. The fixture kept it in flight with `setTimeout(..., 40)`. Under contention those 40 ms can expire before the assertions run, the refresh completes, and the state they describe is gone.

It is now held by a promise the test resolves, so the window lasts exactly as long as the assertions take, on any machine.

One detail worth naming: the hold had to move **after** the seeded failures in the fixture. Applied before them it held the *first* refresh — the one that must fail fast to produce "Account not refreshed" — and the test hung on its own gate. That the deferred made this ordering visible is an argument for it over a sleep.

**2. Assertions compared against a live clock** (`app-shell.test.tsx`, `agent-list-order.test.ts`)

Fixtures built `Date.now() - 8 * 60_000` and asserted `"started 8m ago"`. I measured the tolerance rather than guessing: at **59 seconds** of real delay between building the fixture and reading the string it still reads `8m`; at **60** it reads `9m` and the test fails. Same for `"2m"`, `"1 minute ago"` and friends.

The clock is now pinned, which removes the dependency rather than widening a tolerance around it.

## What this does not fix, and why I am not attempting it

#325 records that **every failure was a `waitFor` timeout**. Most of those are not clock coupling — they are a machine too busy to render inside the default one-second budget. Nothing in this PR changes that, and the two remedies that would are both ruled out by the issue itself: raising timeouts, and retries.

So this closes the part of #325 that is a test defect. The remainder is a capacity problem, and the honest options are to run the suite with fewer competing workers, or to accept that a machine at load 25+ cannot be trusted as a measuring instrument. I would rather say that than claim the flake is fixed.

## Verification

Each change was mutation-checked rather than assumed:

- Replacing `await options.meHoldAfterProfileUpdate` with `void` — so the refresh completes immediately — fails the in-flight assertions. The gate is load-bearing.
- Moving the pinned `startedAt` from 8 to 9 minutes fails with `Unable to find … "started 8m ago"`, confirming the assertion is sensitive to exactly the drift that was previously uncontrolled.

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **58 files / 560 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web 91.77%) |

Run at load ~6. No live QA proposed: test-only change, no product code touched.

### DIFF
--- a/apps/web/src/__tests__/account.test.tsx
+++ b/apps/web/src/__tests__/account.test.tsx
@@ -118,7 +118,16 @@ describe("OpenTag Web App Shell", () => {
   });
 
   it("lets no new save race an in-flight refresh retry", async () => {
-    installApi({ meFailuresAfterProfileUpdate: 1, meDelayMsAfterProfileUpdate: 40 });
+    /*
+     * The refresh is held open by a promise this test resolves, not by a wall-clock delay. The
+     * assertions below describe what is true *while* it is in flight, so the window has to last
+     * exactly as long as they take -- which a fixed 40ms could not guarantee on a loaded machine.
+     */
+    let releaseRefresh = () => undefined as void;
+    const refreshInFlight = new Promise<void>((resolve) => {
+      releaseRefresh = resolve;
+    });
+    installApi({ meFailuresAfterProfileUpdate: 1, meHoldAfterProfileUpdate: refreshInFlight });
     window.history.replaceState({}, "", "/account");
     render(<App />);
 
@@ -140,6 +149,7 @@ describe("OpenTag Web App Shell", () => {
     expect(displayName.disabled).toBe(true);
     fireEvent.click(screen.getByRole("button", { name: "Refreshing…" }));
     fireEvent.submit(form);
+    releaseRefresh();
 
     expect(await screen.findByText("Account profile saved.")).toBeTruthy();
     expect(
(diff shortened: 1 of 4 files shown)


yuezengwu (approved):
Exact-head review at ed75a8ebe8f01ff13fd08722a2f29bb7b8f1ea95: no blocking findings.

The post-profile-update gate now stays open until the test explicitly releases it, and it is correctly applied after the seeded failure so the intended retry is held rather than the first refresh. The elapsed-time assertions use a fixed system clock and restore real timers after every test, keeping the change scoped to test behavior. No product code changes are included, and the PR accurately limits its claim to the two genuine clock-coupled cases rather than the unrelated waitFor capacity failures.

Validation on the exact head: git diff --check; root pnpm build; pnpm --filter @opentag/web typecheck; pnpm --filter @opentag/web test (58 files / 560 tests); pnpm check; web unit coverage 91.77%. GitHub CI, patch coverage, Browser Smoke, Node 22/24/26 compatibility, container/CLI smokes, and ownership are green.

PR #496, opened 2026-09-02
feat(web): finish an unfinished Agent from the list, and stop recording creation twice

liuchao-001:
## Summary

Product decisions from the Agent Setup design review, about where the truth of "what has this
Account got" lives.

- **A refused name names the Agent holding it.** The Server says only that the name is taken, so the
  form finds that Agent and offers it as a link, beside the refusal and above a name field that is
  still the reader's to change. That answers both situations one refusal covers: a name they had
  already used, or their own request reaching the Server without its answer reaching the browser.
  Nothing is opened for them — an Agent this Account already has is not what Create asked for.
- **The Agent list finishes an unfinished Agent.** An Agent stopped on one of the reasons setup still
  owns — no Computer, the chosen runtime unavailable, messaging not connected, connecting, or
  connected but not yet reachable — names the page that finishes it, from the row. Every other
  failure keeps the list's rule: open the Agent, where the dependency that broke is named.
- **Agent creation keeps no record of its own.** The saved creation attempt is removed: the
  idempotency identity, the `localStorage` store, the cross-tab lock, and the Check / Retry /
  Discard panel. Every visit passes through the Agent list, which already shows whether the Agent
  exists and what it still needs; a second record of the same fact can only disagree with it.
- **Start over goes with it.** It cleared the form from the page Go back already returns from, and
  the one Account that saw it — an Account with no Agent — has nowhere else to go.

## Details worth knowing

- The holder lookup asks nothing about Agent state, because the list query and the name pre-check use
  the same predicate (`ne(agents.status, "deleted")`): the response *is* the set the constraint
  refuses from, so a paused Agent is named exactly as a working one is.
- A refusal that cannot name an Agent — only reachable when the read itself fails — still leads
  somewhere, but only where the header carries no way back. Naming an Agent is a different
  destination; a second route to the list is not.
- The offer is cleared with the error it belongs to, so a later failure of a different kind never
  carries the previous refusal's link.

## Not in this change

The Server's `GET /agents/creation-intents/:id` endpoint and the `agents.creation_intent_id` column
now have no caller. Removing them needs a migration and the migration-count ratchet, so it is a
separate change.

Following a named offer to a *suspended* Agent from a not-yet-admitted Account still returns to the
create form, because the setup guard and boundary count only active Agents. That predates this
change and belongs with the derived-admission work.

## Validation

- `pnpm check`, `pnpm typecheck`, `pnpm build` — exit 0
- `pnpm --filter @opentag/web test` — 672 tests
- Independent QA on this exact head (`87756e7c`): **PASS**, `focused-local`, real server + real
  browser — https://github.com/first-tree-ai/opentag/pull/496#issuecomment-5545158284


liuchao-001:
## Independent QA + code review — PR #496

**This is independent QA** (not the author's own runs), executed by code-reviewer on the reviewer's
machine against a purpose-built run cell. Posting with liuchao-001's `gh` account.

- **Exact target**: `3b66d437908e91599187cc17851c885b4876d163` (PR head at time of run; any new commit voids this)
- **Tier**: `focused-local` — UI-touching change, so the product was driven black-box through a real
  server and a real browser; deterministic checks ran as white-box probes alongside.
- **Status**: **FAIL** — the headline feature works, but one reproducible dead end came out of the
  second half of the change.
- **Evidence boundary**: this proves what was observed on a local dev-channel server
  (`OPENTAG_ENV=dev`, Postgres 17, built `apps/web` bundle served by the server) through the browser
  and the HTTP API. It says nothing about staging/prod, about Feishu/Slack against real providers, or
  about any state that needs a live signed-in runtime client (see Limitations).

### What was driven

Real browser (`agent-browser`) against `http://127.0.0.1:51822`, real dev-callback login, real
`POST /api/v1/agents`, real connect-code exchange. No component harness, no mocked fetch in the
verdict paths.

| # | Path | Result |
|---|---|---|
| 1 | Zero-agent account → guard lands on `/agents/setup` | Destination step renders with **no exit control** in the header — matches the "no escape before the first Agent exists" rule |
| 2 | Create step | `Start over` is gone, `Go back` remains — as intended |
| 3 | Create `opentag`/Codex → canonicalized to `?agentId=…`, `Back to agent` appears | pass |
| 4 | `/agents` with a `computer_not_bound` Agent | `Continue setup →` renders, `href=/agents/setup?agentId=…` |
| 5 | **Pointer reachability** of that link | `document.elementFromPoint(centre)` returns the link itself (`data-ui="agent-row-continue-setup"`), not the card overlay — at 1280px **and** at 500px, where the card restacks |
| 6 | **Keyboard reachability** | Tab order is `OpenTag → account menu → New Agent → continue-setup → agent-row-open`; UA focus ring present (`outline auto 1px rgb(0,95,204)`) on top of the card's `focus-within` ring; `Enter` navigates |
| 7 | Click and Enter both land on `/agents/setup?agentId=…` | pass |
| 8 | Control: Agent whose bound Computer is offline | **no** `Continue setup` — the exclusion holds live |
| 9 | Error path: name collision on the create form | see Finding 1 |

The requested overlay question is answered: **the link is genuinely reachable, by pointer and by
keyboard, at both layout breakpoints.** The `z-10` assertion in `app-shell.test.tsx` is a proxy for
this; the hit-test above is the real thing.

---

### Finding 1 — MEDIUM · a lost creation answer strands the first-Agent account with no way out

Reproduced end to end.

**Setup.** Fresh account, `users.setup_completed_at IS NULL`, zero Agents. The guard lands the user
on `/agents/setup`; the boundary reads the list, sees zero, and renders the create form with
`onBackToAgents` withheld (`accountCompleted === false`).

**Reproduction.** With the user parked on that form, an Agent named `opentag` is created server-side
without the browser seeing the answer — exactly the state left behind by a create request whose
response is lost. The user then presses the only enabled button.

**Observed.** `409 AGENT_NAME_CONFLICT` renders as a page-level alert:

> An active Agent with this name already exists for this Account

and the complete set of controls on that screen is:

```
BUTTON "Codex OpenAI" · BUTTON "Claude Code Anthropic" · BUTTON "Go back" · BUTTON "Create Agent"
```

No link or button leaves the page. `Go back` only returns to the local/cloud step, which has no exit
either. Pressing `Create Agent` again repeats the same 409 forever. The only real exits are a browser
reload or hand-editing the URL — and the user has no reason to know that.

Reload **does** recover completely (verified: boundary re-reads the list → one Agent → redirect to
`?agentId=…` → adoption → setup resumes at *Prepare computer*), which is why this is a dead end
rather than data loss.

**Why it is worth fixing rather than accepting.** The rationale in the diff is right — "reconciled
where every other question about this Account's Agents is answered, the Agent list". This screen is
the one place that rationale does not reach, because it is the one screen that will not let the
reader get to the list. The ruling this PR implements says a user who has not created an Agent gets
sent back to the page; here the Agent *was* created, so the ruling's own logic points at the Agent,
which is exactly what a reload does.

The other risk on this screen is quieter: the reader's most natural workaround is to change the name
and press Create, which produces a **second, stray Agent** — a worse version of the "two records that
can disagree" this PR set out to remove.

**Suggested fix, in the PR's own idiom.** On a failed create, re-run the boundary's target resolution
instead of leaving a terminal banner — the same read that already recovers this on reload. That keeps
"the list is the authority" and needs no stored record. A weaker fallback is to surface an exit on the
error, but re-resolving is strictly better because it fixes the case without asking the reader to
understand it.

The admitted-account path is fine: on `?action=create` with `setup_completed_at` set, `Back to agents`
is present and the same 409 is escapable. The gap is specific to the not-yet-admitted account.

---

### Finding 2 — MEDIUM · the list withholds `Continue setup` for two states the same file already treats as setup-owned

`SETUP_REASONS` in `apps/web/src/features/agents/agent-presentation.ts:405` lists three reasons.
`setupOrMaintenanceAction` — in the same file, for the Agent detail page — treats **five** as owned by
setup:

| reason | detail page offers `Continue setup` | list offers it |
|---|---|---|
| `computer_not_bound` (L468) | yes | yes |
| `runtime_unavailable` (L500/507/518/529) | yes | yes |
| `im_not_connected` (L552) | yes | yes |
| `handoff_unavailable` (L564) | yes | **no** |
| `im_provisioning` (L585) | yes | **no** |

Both missing states are precisely "left part-way through the messaging step":

- `im_provisioning` — the reader started the Feishu/Slack connect and closed the tab mid-QR. Card
  reads *Setting up messaging*, no way back to the page that finishes it.
- `handoff_unavailable` — messaging is connected but the provider CLI on the Computer is not ready
  yet. Card reads *Cannot receive new work* and sorts **first** in the list (priority 0), still with
  no entry point.

The server agrees these are unfinished setup: `deriveSetupStage` (`packages/shared/src/agent-setup.ts:322`)
returns `needs-messaging` for any messaging that is not `ready`, which covers both.

The doc comment above `SETUP_REASONS` reads as if three were the whole story ("the three things Agent
Setup asks for"). Either widen the set to match `setupOrMaintenanceAction`, or say in the comment why
the list is deliberately narrower than the Agent page — right now the divergence looks accidental.

`computer_offline` is a different matter and I think the exclusion is right: it is the one state
where the machine, not the setup flow, is the subject, and the detail page routes it to
Settings → Computer rather than to setup. Consistent, and verified live.

---

### On deleting `GET /agents/creation-intents/:id` and `agents.creation_intent_id`

Not in this PR. The web tree is already clean — the only surviving mention is the negative assertion
in `agent-setup-boundary.test.tsx:113` — and nothing else in the repo calls the endpoint (server and
shared only). It needs a migration plus a coverage-ratchet pass, and a dead read-only endpoint with no
caller is harmless in the meantime. Keeping it out of this PR is the right call; a follow-up issue is
enough.

### Automated checks run at this head

- `pnpm --filter @opentag/web test` — **58 files, 559 tests, all pass**
- `pnpm --filter @opentag/shared --filter @opentag/server --filter @opentag/web build` — clean;
  `git status --porcelain` empty afterwards, so no stale generated route tree
- CI at this head: all required checks green, `mergeStateStatus=CLEAN`

### Limitations

- **`im_provisioning` and `handoff_unavailable` were not reproduced live.** Both require the Agent's
  runtime to report `ready`, and readiness is held in the server's in-process
  `RuntimeConnectionRegistry` (`packages/server/src/index.ts:215`), filled only by a live client
  registering over WebSocket. Producing it honestly needs a signed-in provider home, which this run
  cell does not have and must not borrow. Finding 2 is therefore a **code-level** claim, proved by the
  call sites cited above, not a live observation — the table's "detail page" column is read from
  source, not driven.
- Feishu/Slack were never contacted; no provider-side behaviour is asserted.
- Console during the run carried only `warning`-level API diagnostics (the intentional 409, plus
  pre-existing `RESOURCE_NOT_FOUND` probes for `/me/setup/reset` and `/internal/navigation-visibility`
  that are unrelated to this diff). **No page errors, no uncaught exceptions.**

### Case disposition

`candidate-new-case` — "finish an Agent left part-way through setup" is worth a prose case in
`cases/`: it is judgement-shaped, spans list → setup → Agent page, and its interesting states are the
ones a unit test cannot reach.

liuchao-001:
## Independent QA + code review — PR #496, round 2

**Independent QA**, re-run from scratch on the reviewer's machine. Posting with liuchao-001's `gh`
account. The previous report is void; this one is bound to the current head.

- **Exact target**: `6b39a9009ae86ae4327ec084a5690d51015da821` (any new commit voids this)
- **Tier**: `focused-local` — driven black-box through a real server and a real browser, with
  deterministic checks as white-box probes alongside.
- **Status**: **FAIL** — both round-1 findings are fixed and verified live. The status is carried by
  one **LOW** residual: the new carve-out leaves the same dead end reachable through a narrower door.
  My recommendation is a one-line condition change, then merge — not "this is broken".
- **Evidence boundary**: a local dev-channel server (`OPENTAG_ENV=dev`, Postgres 17, the built
  `apps/web` bundle served by the server). Nothing here speaks to staging/prod or to Feishu/Slack
  against real provider infrastructure.

### Round-1 Finding 1 — fixed, verified live

Same reproduction as before: fresh Account, `users.setup_completed_at IS NULL`, zero Agents, the
guard lands on `/agents/setup`, the boundary reads zero and renders the create form with no exit.
With the reader parked there, an Agent named `opentag` is created server-side without the browser
seeing the answer — the exact state a lost response leaves. Then the reader presses Create.

**Before**: `409`, a banner, and four controls none of which leave the page.
**Now**: the press lands on `/agents/setup?agentId=32922899-…`, *Create agent — Completed*, stage
*Prepare computer*, `Back to agent` present.

Checked alongside it:

- `SELECT count(*) FROM agents` = **1** — the failed press created no stray second Agent.
- `users.setup_completed_at` is set — adoption ran, so the Account is admitted from here on.
- Exactly **one** POST to `/api/v1/agents` from the browser.

**Negative control** (an ordinary failure with nothing behind it): the server was taken down with the
reader on the form, then Create pressed. The banner reads `Failed to fetch`, the URL stays
`/agents/setup`, and `Create Agent` is **re-pressable**, not disabled. The recheck's own read failed
too, resolved to "nothing found", and correctly did nothing. This is the branch that would have
turned an ordinary outage into a bogus redirect, and it holds.

### Round-1 Finding 2 — fixed, and this time verified live rather than by reading

`SETUP_REASONS` now carries all five reasons `setupOrMaintenanceAction` treats as setup-owned. Last
round I could only prove this from call sites, because both added states need the Agent's runtime to
report `ready` and that lives in the server's in-process `RuntimeConnectionRegistry`. This round I
closed that gap with a **stub Computer runtime client** speaking the real v1 protocol
(`auth` → `computer:register` → heartbeats, with the `x-opentag-provider-readiness: 1` upgrade
header). It substitutes for the Client, not the Server: every readiness value it reports is an
*input* to the question under test, never an answer produced by the code under test.

With a genuinely online Computer reporting `codex: ready`:

| state | reached by | card | `Continue setup` |
|---|---|---|---|
| `im_not_connected` | ready runtime, no binding | Messaging not connected | **yes** |
| `im_provisioning` | binding row `status='provisioning'` | Setting up messaging | **yes** (was absent at `3b66d437`) |
| `handoff_unavailable` | `status='active'`, full required tenant scopes, a real AES-256-GCM credential the server can actually decrypt, provider CLI reported not ready | Cannot receive messages | **yes** (was absent at `3b66d437`) |
| `computer_offline` | stub client stopped | Computer offline | **no** — exclusion still holds |

Getting `handoff_unavailable` took three attempts and the two failures are worth recording, because
each was a *different* state rather than the one I wanted: an empty scope set and then an
undecryptable credential blob both project to `im_reauthorization_required`. Only a binding that is
active, fully scoped **and** decryptable leaves the provider CLI as the sole blocker. So the table
row above is that state, not a near neighbour.

**The destination is useful, not just present.** Following the new link from the
`handoff_unavailable` card lands on *Connect messaging* with the two earlier steps marked Completed
and the Lark CLI readiness on screen — the actual reason handoff is not ready.

### New finding — LOW · `?action=create` skips the recheck for an Account that has no way back

The carve-out is:

```ts
if (action === "create") return;
```

justified as "`action=create` … keeps its own way back". That is true for an admitted Account and
false otherwise, because the way back is itself conditional:

```ts
onBackToAgents={accountCompleted ? onBackToAgents : undefined}
```

Reproduced at this head: `setup_completed_at IS NULL`, zero Agents, open
`/agents/setup?action=create`, create the Agent out-of-band, press Create. Result is round 1 exactly
— `An active Agent with this name already exists for this Account`, and the only controls are
`Codex` / `Claude Code` / `Go back` / `Create Agent`. No recheck, no exit.

**Severity is LOW because the door is narrow.** `_shell.tsx` redirects any Account without
`setupCompletedAt` away from `/agents`, so an un-admitted Account is never handed the `New Agent`
button that produces this URL; reaching it means typing or restoring the URL by hand. I could not
find a UI path.

**Suggested fix**: make the carve-out test the thing its own comment relies on —
`if (action === "create" && accountCompleted) return;` — so the recheck is skipped only when the
reader really does have somewhere to go. That preserves the intent ("a fresh form on purpose") and
removes the case where the justification does not hold.

### The `account.test.tsx` flake you flagged

Not reproduced: **6 runs clean** — one full `pnpm --filter @opentag/web test` (58 files, 562 tests)
plus five isolated runs of `account.test.tsx` (21 tests each).

It is also not yours: `apps/web/src/__tests__/account.test.tsx` is untouched by this PR, and the test
arms a **wall-clock** fixture — `installApi({ meFailuresAfterProfileUpdate: 1, meDelayMsAfterProfileUpdate: 40 })`.
A 40 ms real delay racing an in-flight retry is load-sensitive by construction, which matches the
other timing-sensitive account-refresh-race tests in this suite. I would treat it as a pre-existing
fixture-determinism problem rather than anything this change introduced, and worth its own issue if
it recurs.

### Checks at this head

- `pnpm --filter @opentag/web test` — **58 files, 562 tests, all pass** (559 → 562, the three new ones)
- `pnpm --filter @opentag/shared --filter @opentag/server --filter @opentag/web build` — clean
- CI at `6b39a900`: every required check green, `mergeStateStatus=CLEAN`
- No page errors and no uncaught exceptions across the whole browser run

### Limitations

- The `handoff_unavailable` fixture is assembled by SQL, so it holds an active binding alongside a
  setup attempt the real flow would have completed. The setup page consequently also renders
  *Waiting for you to scan…* on that screen. **That is my fixture, not a defect** — I am not
  reporting it, and it does not affect the card-level result the table records.
- Starting the Lark connect from the setup page does reach `open.feishu.cn`; no Feishu App is
  configured here, so nothing provider-side is asserted.
- The stub runtime client reports readiness rather than observing a signed-in provider. It proves
  how the web renders a given readiness, not that any provider is actually installed.

### Case disposition

`candidate-new-case` — unchanged from round 1. "Finish an Agent left part-way through setup" wants a
prose case in `cases/`, and this round produced the missing ingredient: the stub runtime client is
what makes the messaging-stage states reachable without borrowing a signed-in provider home.

liuchao-001:
## Independent QA + code review — PR #496, round 3

**Independent QA**, re-run from scratch. Posting with liuchao-001's `gh` account. The round-2 report
is void; this one is bound to the current head.

- **Exact target**: `265688fa2395a1e90cb40984a2493cb8a6feb09a` (any new commit voids this)
- **Tier**: `focused-local` — driven black-box through a real server and a real browser
- **Status**: **FAIL** — everything the author asked me to verify passes, and the round-2 LOW is gone.
  The status is carried by one new **LOW**: a refusal's link outlives the refusal. One line.
  Recommendation is fix-then-merge, not "this is broken".
- **Evidence boundary**: a local dev-channel server (`OPENTAG_ENV=dev`, Postgres 17, the built
  `apps/web` bundle served by the server). Nothing here speaks to staging/prod or to Feishu/Slack
  against real provider infrastructure.

### The four requested checks

**1. The conflict link, on both routes.** Reproduced the real trap on the no-action route: fresh
Account, `setup_completed_at IS NULL`, zero Agents, reader parked on the form, Agent created
server-side without the browser seeing the answer. Before the press the header carries **no exit**
(the first-Agent rule still holds). After it:

> An active Agent with this name already exists for this Account
> **Open @opentag** → `/agents/8eb5a293-f412-4f2f-85c4-a2b0cad237f5`

That href is the Agent the reader's own press produced. Same result on `?action=create`, where
`Back to agents` is present as well. **The dead end from rounds 1–2 is closed** — the link is itself
the exit.

**The landing you asked about.** Clicking it as a not-yet-admitted Account lands on
`/agents/setup?agentId=8eb5a293-…`, heading *Set up opentag*, rail reading
*Create agent — Completed → Prepare computer*. The guard bounce resolves through the boundary and
adopts on the way: one Agent in the table, `setup_completed_at` now set. So the reader ends exactly
where round 2's automatic redirect put them — the difference is only who decided. That is the ruling
working, not a detour.

**2. Renaming clears the refusal.** Editing the name field drops the alert, the link, and the whole
`agent-create-error` block in one go. Pressing Create then succeeds and moves on — a second Agent
`opentag-two` was created and the flow advanced to its setup page. Both halves of the ruling — open
it, or rename — work end to end.

**3. An ordinary failure carries no link.** With the server stopped and no prior refusal on screen:
banner reads `Failed to fetch`, **no link**, URL unchanged, `Create Agent` still pressable.

**4. The two messaging states, re-verified at this head.** Re-run with the stub Computer runtime
client (real v1 protocol, `x-opentag-provider-readiness: 1`), Computer online reporting
`codex: ready`:

| state | card | `Continue setup` |
|---|---|---|
| `im_not_connected` | Messaging not connected | yes |
| `im_provisioning` | Setting up messaging | yes |
| `handoff_unavailable` | Cannot receive messages | yes |
| `computer_offline` (stub stopped) | Computer offline | no — exclusion still holds |

I also re-checked that the round-2 fix survived the `origin/main` merge, since #475 edits the same
file: `SETUP_REASONS` still carries all five, and `setupOrMaintenanceAction`'s call sites at this head
still name exactly those five (L421 / L453 / L505 / L517 / L538). Alignment intact.

### New finding — LOW · a refusal's link outlives the refusal

`create` clears `error` on every attempt but never clears `taken`, and `taken` is only cleared when
the **name** changes. So a refusal followed by a *different* failure on the same name shows the new
error with the old offer still attached.

Reproduced: press Create with a taken name → `409` + `Open @opentag`. Stop the server. Press Create
again without touching the name → the banner becomes **`Failed to fetch`** and **`Open @opentag` is
still sitting under it**.

The link is not false — that Agent really does hold that name — but it is now answering a question
nobody asked: the banner says the request never reached the Server, and directly beneath it is an
offer that belongs to an earlier, different refusal. It is the same shape as the thing this PR set
out to remove, a second account of a fact that can disagree with the first.

**Suggested fix**: clear `taken` where `error` is cleared, at the top of `create`. Rendering the link
only when the current error *is* the conflict would do it too, but clearing beside `setError` keeps
the two pieces of the same state in one place.

### On the residual you flagged

You asked whether the "list read also failed → no link" case needs a fallback. I could not
manufacture that exact conjunction — it needs the POST to be refused and the very next GET to fail —
so I am not claiming to have reproduced it. Both halves are observed separately, though: a screen
with no link renders only `Codex` / `Claude Code` / `Go back` / `Create Agent`, and on the
not-yet-admitted route those are all of them.

My read: **worth one cheap guard, and the reason is structural rather than statistical.** The link is
the only exit on that screen, and it depends on a second network call succeeding immediately after
one has just failed — which is exactly when it is least likely to. The failure modes are correlated,
so the exit is missing precisely in the situation that produces it.

I would not restore the automatic redirect — the ruling is clear that the reader decides. The
smallest thing consistent with it is to let the reader ask again: when the refusal is a name conflict
and no holder was found, offer a retry of that lookup rather than nothing. The reader still chooses;
they just get a second chance to be offered the choice.

### Checks at this head

- `pnpm --filter @opentag/web test` — **58 files, 562 tests, all pass**
- `pnpm --filter @opentag/shared --filter @opentag/server --filter @opentag/web build` — clean,
  `git status --porcelain` empty afterwards
- CI at `265688fa`: **15 checks SUCCESS**, 4 skipped, none failing or pending; `mergeStateStatus=CLEAN`
- No page errors and no error-level console output across the browser run

**On the flake**: not reproduced here. The full web suite passed 562/562 at load average ~9 on this
machine, including the files you listed. That is one clean run, not proof the flake is gone — but it
is not reproducing under the conditions you described.

### Limitations

- The `handoff_unavailable` fixture is SQL-assembled, so it holds an active binding beside a setup
  attempt the real flow would have completed. That is my fixture, not a defect.
- The stub runtime client *reports* readiness rather than observing a signed-in provider: it proves
  how the web renders a given readiness, not that any provider is installed.
- Part of this run was re-done after a harness collision: another QA session on this machine shares
  the same `agent-browser` daemon, and my session was briefly driving a different run cell. Every
  result above was re-verified from a clean state in an isolated session (`--session pr496`) and is
  corroborated by server-side state; nothing in this report rests on the contaminated window.

### Case disposition

`candidate-new-case` — unchanged. "Finish an Agent left part-way through setup" wants a prose case,
and it should carry the name-conflict fork, which is now the product's answer to a lost creation.

liuchao-001:
## Independent QA + code review — PR #496, round 4

**Independent QA**, re-run from scratch in an isolated browser session. Posting with liuchao-001's
`gh` account. The round-3 report (`265688fa`) is void; this one is bound to the current head.

- **Exact target**: `f3d65bfc3b4dc2366ed1d7a26e9a27a59be19462` (any new commit voids this)
- **Tier**: `focused-local` — driven black-box through a real server and a real browser
- **Status**: **FAIL** — all five requested checks pass and both prior findings are fixed and
  verified live. The status is carried by one new **LOW**: a *suspended* Agent holds a name against
  the Server's uniqueness rule but is invisible to the lookup that names it. One line.
  Recommendation is fix-then-merge; filing it as a follow-up instead is a defensible call I would
  not block.
- **Evidence boundary**: a local dev-channel server (`OPENTAG_ENV=dev`, Postgres 17, the built
  `apps/web` bundle served by the server). Nothing here speaks to staging/prod or to Feishu/Slack
  against real provider infrastructure.

### The five requested checks

**1. `Open @<name>`, both routes, and renaming.** Reproduced the real trap on the no-action route:
fresh Account, `setup_completed_at IS NULL`, zero Agents, reader parked on the form, Agent created
server-side without the browser seeing the answer. Header carries **no exit** before the press
(first-Agent rule intact). After it:

> An active Agent with this name already exists for this Account
> **Open @opentag** → `/agents/d99ac738-f6e7-4c5f-b384-ab117b0ba954`

— the Agent that press produced. Same on `?action=create`, with `Back to agents` also present.
Editing the name drops the alert, the offer and the whole `agent-create-error` block together;
pressing Create then succeeds and the flow advances (`opentag-four` created, landed on its setup
page). Both halves of the ruling work end to end.

**2. `Go to your Agents`, and its suppression.** Verified in both directions — see the finding below
for how the holder-not-found branch was reached without contrivance. On the not-yet-admitted route
the offer renders as `Go to your Agents` → `/agents`. On an admitted Account, where the header
already carries `Back to agents`, it is correctly **absent**. That is the consistency point from
round 3, and the shape you chose — suppress the bare fallback, always name a holder — is the right
one: naming an Agent is a different destination, a second route to the list is not.

**3. The stale offer is gone.** Round-3 repro, re-run: press Create with a taken name → `409` +
`Open @opentag`; stop the server; press again without touching the name → banner becomes
`Failed to fetch` and the offer is **null**. Fixed.

**4. The five list states, re-verified at this head** (stub Computer runtime client, real v1
protocol, `x-opentag-provider-readiness: 1`, Computer online reporting `codex: ready`):

| state | card | `Continue setup` |
|---|---|---|
| `computer_not_bound` | No Computer connected | yes |
| `im_not_connected` | Messaging not connected | yes |
| `im_provisioning` | Setting up messaging | yes |
| `handoff_unavailable` | Cannot receive messages | yes |
| `computer_offline` (stub stopped) | Computer offline | no — exclusion holds |
| `agent_suspended` | Paused | no — exclusion holds |

`SETUP_REASONS` still carries all five after the re-apply onto the new main, and
`setupOrMaintenanceAction`'s call sites still name exactly those five. Alignment intact.

**5. The Lab preview path.** `/internal/agent-setup` loads, the preview creation form drives through
Local computer → name → Codex → Create, and the Lab advances out of `creation` into account
admission and on into the setup surface (*Create agent — Completed → Prepare computer*).
**No Agent was created** — the table stayed at 2 across the whole preview run, so `creationPreview`
is still intercepting after the `CreateAgentRequest` / `accountId` changes. No page errors, no
error-level console output.

### New finding — LOW · a suspended Agent holds the name but cannot be named

The Server refuses on `ne(agents.status, "deleted")` (`agent-service.ts`, and the live index
`agents_account_name_active_unique … WHERE status <> 'deleted'`). The lookup that names the holder
filters harder:

```ts
agents.find((candidate) => candidate.status === "active" && candidate.name.toLowerCase() === wanted)
```

So a **suspended** Agent refuses the name and cannot be offered. Verified through the API:

```
POST /api/v1/agents {"name":"opentag"}  →  409 AGENT_NAME_CONFLICT
GET  /api/v1/agents                     →  [('opentag', 'suspended')]
```

The client is holding the answer in the same response it discards it from.

Two variants, both reproduced in the browser:

- **Admitted Account** (the common one): banner reads *An active Agent with this name already
  exists*, **no offer at all**, and there is no active Agent by that name to find. `Back to agents`
  works and the list does show `opentag · Paused`, so it is friction rather than a trap.
- **Not-yet-admitted Account**: the fallback renders, and **clicking `Go to your Agents` lands back
  on the create form** — `/agents` → guard → `/agents/setup` → the boundary counts *active* Agents,
  sees zero, and offers creation again. The draft is lost. So on this route the "there is always an
  exit" guarantee does not hold for this state.

**Attribution, so the fix lands in the right place.** What this PR owns is the narrow lookup. The
loop underneath it — guard and boundary both counting only active Agents — predates this change, as
does the Server's *"An active Agent…"* wording for a holder that is suspended. I am not asking you
to fix those here.

**Suggested fix**: match the lookup to the rule that refused the name — `candidate.status !== "deleted"`.
That makes the holder nameable in both variants, which is the whole point of the offer, and it costs
one word. It does not close the not-yet-admitted loop; that needs the guard to count what the Server
counts, which is a separate change and arguably belongs with the derived-admission work.

### Checks at this head

- `pnpm --filter @opentag/web test` — **62 files, 671 tests, all pass**
- `pnpm --filter @opentag/shared --filter @opentag/server --filter @opentag/web build` — clean,
  `git status --porcelain` empty afterwards
- CI at `f3d65bfc`: **17 SUCCESS, 4 SKIPPED**, none failing or pending; `mergeStateStatus=CLEAN`

**On the `agent-setup-lab-page` flake**: not reproduced. The full suite passed 671/671 at load
average ~12 on this machine, that file included. One clean run is not proof it is gone, but it did
not reproduce under conditions comparable to the ones you described.

### Limitations

- The `handoff_unavailable` fixture is SQL-assembled, so it holds an active binding beside a setup
  attempt the real flow would have completed. That is my fixture, not a defect.
- The stub runtime client *reports* readiness rather than observing a signed-in provider: it proves
  how the web renders a given readiness, not that any provider is installed.
- The Lab check is a smoke pass of the preview creation path, not a review of Lab behaviour.

### Case disposition

`candidate-new-case` — unchanged. The prose case should carry the name-conflict fork, and this round
adds a second axis worth writing down: the holder's lifecycle state, which is where the offer breaks.

liuchao-001:
## Independent QA + code review — PR #496, round 5

**Independent QA**, re-run from scratch in an isolated browser session. Posting with liuchao-001's
`gh` account. The round-4 report (`f3d65bfc`) is void; this one is bound to the current head.

- **Exact target**: `87756e7c7545978c8f9445152a97ae7ff6a6e6d2` (any new commit voids this)
- **Tier**: `focused-local` — driven black-box through a real server and a real browser
- **Status**: **PASS** — the round-4 finding is fixed and verified live, and all five earlier checks
  re-verified at this head. No product problem found in this change.
- **Evidence boundary**: a local dev-channel server (`OPENTAG_ENV=dev`, Postgres 17, the built
  `apps/web` bundle served by the server). This says what was observed on those paths under those
  conditions — not release readiness, and nothing about staging/prod or about Feishu/Slack against
  real provider infrastructure.

### The correction to my suggestion was right, and the fix is better than "right"

My `candidate.status !== "deleted"` would not compile: `AgentStatusSchema = z.enum(["active", "suspended"])`,
so the comparison is TS2367 against a type that has no such member. Dropping the filter is the
correct form.

It is also *exactly* right rather than approximately so, which is worth recording. The list query
and the name pre-check use the **same predicate** — `ne(agents.status, "deleted")` at
`agent-service.ts:511` and `:385` respectively. So "every Agent in this response holds its name" is
not a convenient reading; the response *is* the set the constraint refuses from. The lookup scope and
the refusal scope are now provably identical, with nothing left to drift.

### Round-4 finding — fixed, verified live in both variants

**Admitted Account.** Suspended Agent `opentag` holding the name, `?action=create`, press Create:

> An active Agent with this name already exists for this Account
> **Open @opentag** → `/agents/4652940e-da80-4459-996f-7cca97a1cb7b`

Following it lands on that Agent's page, which states *Suspended — This Agent is paused. Resume it
to start receiving messages again.* So the reader now learns which Agent holds the name **and** why
it was nowhere to be seen among the active ones. That is a better outcome than the naming alone.

**Not-yet-admitted Account.** Same seed, `setup_completed_at IS NULL`. The refusal now carries
`Open @opentag` rather than the bare fallback — the holder is named here too. Following it still
returns to the create form (`/agents/<suspended>` → guard → `/agents/setup` → the boundary counts
active Agents, sees zero, offers creation). **That is the pre-existing loop, not a regression**, and
it is the attribution we already agreed: the guard and boundary counting only active Agents predates
this change. Recording it because it is now the one path where a named offer still cannot be
followed, and because it is the exact question `hasActiveAgent` will have to answer in B.

### The five earlier checks, re-verified at this head

| # | check | result |
|---|---|---|
| 1 | Lost-response trap, no-action route | header carries no exit before the press; after it, `Open @opentag` → `/agents/45e528f0-…`, the Agent that press produced |
| 1 | Following the offer | `/agents/setup?agentId=45e528f0-…`, *Connect your computer*; one Agent in the table, `setup_completed_at` set — adoption ran |
| 1 | Renaming | alert, offer and the `agent-create-error` block clear together; Create then succeeds (`opentag-five`) and the flow advances |
| 3 | Stale offer | `409` + offer → server stopped → press again without touching the name → banner `Failed to fetch`, offer **null** |
| 4 | List states | `computer_not_bound`, `im_not_connected`, `im_provisioning`, `handoff_unavailable` all carry `Continue setup`; `computer_offline` and `Paused` do not |
| 5 | Lab preview | `/internal/agent-setup` drives Local computer → name → Codex → Create and advances out of `creation` into account admission; **no Agent created** (table stayed at 2); no page errors |

### Checks at this head

- `pnpm --filter @opentag/web test` — **62 files, 672 tests, all pass** (671 → 672, your new
  regression test)
- `pnpm --filter @opentag/shared --filter @opentag/server --filter @opentag/web build` — clean,
  `git status --porcelain` empty afterwards
- CI at `87756e7c`: **16 SUCCESS, 4 SKIPPED**, none failing or pending; `mergeStateStatus=CLEAN`

**On the `agent-tasks` flake**: not reproduced. The full suite passed 672/672 at load average ~7.6.

### Limitations

- **`Go to your Agents` was not exercised at this head.** Removing the status filter narrowed the
  `unnamed` branch to one trigger — the list read itself failing right after the create was refused —
  which I could not manufacture. I verified both directions of that branch at `f3d65bfc` and the
  `offer` selection is untouched by this diff, but I am not claiming a fresh observation of it.
- The `handoff_unavailable` fixture is SQL-assembled, so it holds an active binding beside a setup
  attempt the real flow would have completed. That is my fixture, not a defect.
- The stub runtime client *reports* readiness rather than observing a signed-in provider: it proves
  how the web renders a given readiness, not that any provider is installed.
- The Lab check is a smoke pass of the preview creation path, not a review of Lab behaviour.

### Case disposition

`candidate-new-case` — the prose case should carry the name-conflict fork and the holder's lifecycle
state, which is the axis that produced the last two rounds of findings.

PR #500, opened 2026-09-02
fix(computers): make the Computer connect command reach a release and fail closed

### LINKED ISSUE #497: Computer connect command is unusable: download.opentag.build has no TLS, and the failure is masked as "unknown command 'connect'"
## Summary

The Computer connect command generated by the web app can never succeed, on any machine. The install
URL it embeds — `https://download.opentag.build/releases/<channel>/install.sh` — has no TLS
certificate, so `curl` always fails. Because the failure is swallowed by the `&&` chain, the user
instead sees a misleading `unknown command 'connect'` from whatever stale binary is already on disk.

Reported from `https://dev.opentag.build/agents/<id>/settings/computer`.

## Reproduction

Copy the command from the Computer settings page and run it:

```
curl -fsSL https://download.opentag.build/releases/staging/install.sh | sh && PATH="$HOME/.local/bin${PATH:+:$PATH}" "$HOME/.local/bin/opentag-staging" connect --server https://dev.opentag.build -- otcc_...
```

Observed:

```
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to download.opentag.build:443
error: unknown command 'connect'
```

## Root cause: `download.opentag.build` serves HTTP only

`download.opentag.build` is an A record to a Google Cloud Storage classic custom-domain endpoint
(`server: UploadServer`, `via: 1.1 google`, `x-goog-*` response headers). That endpoint serves plain
HTTP only; Google presents no certificate for the hostname.

Isolated to SNI, so this is not a local network or proxy artifact — same IP, same port, only the
server name differs:

```
$ curl --resolve storage.googleapis.com:443:136.110.192.61  https://storage.googleapis.com/opentag-release/releases/staging/latest.json
http=200

$ curl --resolve download.opentag.build:443:136.110.192.61  https://download.opentag.build/releases/staging/latest.json
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL

$ curl -sI http://download.opentag.build/releases/staging/install.sh
HTTP/1.1 200 OK
```

TCP 443 accepts the connection and then drops the handshake after reading 0 bytes.

The same default is used by the server's channel-target poller
(`packages/server/src/services/channel-target/poller.ts` fetches
`${downloadBaseUrl}/${channel}/latest.json`), so `OPENTAG_PORTABLE_DOWNLOAD_BASE_URL`'s default in
`packages/server/src/config.ts:126` is unreachable over HTTPS for that path too.

Note the *published* installer already points its second stage at a working URL
(`DOWNLOAD_BASE_URL` defaults to `https://storage.googleapis.com/opentag-release/releases` in the
rendered copy). Only the bootstrap URL uses the broken vanity domain.

### Fix options

1. Put a real HTTPS front end on `download.opentag.build` (load balancer or CDN with a managed cert),
   keeping the vanity domain; or
2. Change the default `OPENTAG_PORTABLE_DOWNLOAD_BASE_URL` to
   `https://storage.googleapis.com/opentag-release/releases`
   (`packages/server/src/config.ts:126`, `scripts/portable/install.sh:12`).

## Contributing defect 1: the `&&` chain hides the download failure

`buildComputerConnectCommand` emits `curl -fsSL <url> | sh && <binary> …`
(`packages/server/src/services/computers/machine-auth-service.ts:556`). A pipeline's exit status is
the *last* command's, so `sh` exiting 0 on empty stdin makes the `&&` succeed even though `curl`
failed. The chain then runs whatever stale binary is already at `$HOME/.local/bin/<bin>`, and the
user is shown a CLI usage error instead of "download failed".

This is what turns a clear network error into a confusing one. It should fail closed — e.g. download
to a temp file first and run it only on success, or run the pipeline under a shell with
`set -o pipefail`.

## Contributing defect 2: the emitted subcommand is not backward compatible

#477 changed the emitted subcommand from `computer connect` to the new top-level `connect`:

```
-  const connectArgs = `computer connect --server ${shellArg(options.publicUrl)} -- ${shellArg(options.code)}`;
+  const connectArgs = `connect --server ${shellArg(options.publicUrl)} -- ${shellArg(options.code)}`;
```

Top-level `connect` only exists in clients at or after #477. Any client older than that reports
`unknown command 'connect'`. `computer connect` is registered in both old and new clients (it is kept
as a compatible alias in `apps/cli/src/cli/program.ts`), so emitting `computer connect` is strictly
more compatible and costs nothing.

This matters precisely because the command is a *repair* path: the machines that need it are the ones
whose client is out of date, and the install step that would upgrade them is broken by the root cause
above.

Reproduced with client `0.0.3-staging.34.1` against current channel target `0.0.3-staging.73.1`
(gitSha `23da031e`).

## Contributing defect 3: the copied comment line is not inert in interactive zsh

`apps/web/src/setup/command-block.tsx:58` copies `` `${comment}\n${command}` ``, and the source
comment there states the comment line "pasted into a terminal it is inert". That is not true of
interactive zsh, which is the macOS default shell: `interactive_comments` is off by default, so the
line produces `zsh: command not found: #` before the real command runs.

Cosmetic, but it appears as a first error line and makes a working run look broken.

## Verification that the chain works once the URL is fixed

Installing from the reachable URL into a sandbox prefix succeeds and produces a client with both
subcommands:

```
$ curl -fsSL https://storage.googleapis.com/opentag-release/releases/staging/install.sh | sh -s -- --prefix <tmp> --bin-dir <tmp>/bin
OpenTag 0.0.3-staging.73.1 installed
$ <tmp>/bin/opentag-staging --help
  connect [options] <code>  Connect this Computer …
  computer                  Connect and inspect Computers …
```

liuchao-001:
Fixes #497.

The Computer connect command the web app hands a user could not install anything, on any machine.

## What was broken

`https://download.opentag.build/releases` serves plain HTTP only. The host is a Cloud Storage
custom-domain endpoint, so port 443 accepts the TCP connection and then drops the handshake with no
certificate for the name. Isolated to SNI, against the same IP and port:

```
$ curl --resolve storage.googleapis.com:443:136.110.192.61 https://storage.googleapis.com/opentag-release/releases/staging/latest.json
http=200

$ curl --resolve download.opentag.build:443:136.110.192.61 https://download.opentag.build/releases/staging/latest.json
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL

$ curl -sI http://download.opentag.build/releases/staging/install.sh
HTTP/1.1 200 OK
```

The failure was invisible. A pipeline reports its last command's status and `sh` succeeds on the
empty input a failed download leaves behind, so `curl … | sh &&` continued and ran whatever Client
was already on disk. A stale Client answered `unknown command 'connect'`, which reads as a CLI
problem rather than a download that never happened:

```
$ sh -c 'curl -fsSL https://download.opentag.build/releases/staging/install.sh | sh && echo REACHED_CONNECT_STEP'
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL
REACHED_CONNECT_STEP
exit 0
```

The same unreachable base URL was the default in six places, including the Client's own automatic
and manual upgrade path, so a long-running Client could not have upgraded itself either.

## What changed

**The install step fails closed.** The installer lands in a file first, keeping every step in one
`&&` chain, so the command stops where it actually broke:

```
$ sh -c 'opentag_installer="$(mktemp)" && curl -fsSL https://download.opentag.build/releases/staging/install.sh -o "$opentag_installer" && sh "$opentag_installer" && rm -f "$opentag_installer" && echo REACHED_CONNECT_STEP'
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL
exit 35
```

**Every default points at a reachable release.** All six move to the Cloud Storage coordinates the
release pipeline already publishes to and that the rendered installers already carry, so the default
now agrees with where releases actually land. `buildComputerConnectCommand` takes the base URL its
caller polls rather than keeping a second default of its own, which is what let these drift apart in
the first place.

This keeps the install URL on HTTPS. `normalizeDownloadBaseUrl` already refuses a plain-HTTP base,
and an install script fetched over HTTP is worth exactly one MITM. Restoring the branded name is an
infra change, not a code one: `opentag.build` is on Cloudflare, so proxying the `download` record
gives `https://download.opentag.build` a certificate, after which this default is a one-line move
back.

**The copied comment is inert in a real terminal.** `#` only introduces a comment where
`interactive_comments` is set, which zsh — the macOS default shell — leaves off for interactive use,
so the pasted line answered `command not found: #` and opened a working command with what read as a
failure. It is now the POSIX null command, which ignores its arguments and returns zero in every
POSIX shell, so the sentence stays silent in a terminal while an agent pasted the same text still
reads it as the instruction. The block shows exactly the line it copies, because the
manual-selection fallback copies what is on screen.

Backward compatibility for the top-level `connect` subcommand is deliberately not addressed: there
are no real users on a pre-#477 Client, and with the install step fixed a stale Client upgrades
before it runs.

## Verification

`pnpm check`, `pnpm typecheck`, `pnpm test`, `pnpm build`, and
`pnpm --filter @opentag/server test:integration` (313 tests) all pass.

Beyond the suite, the installer was run end-to-end from the new URL into a sandbox prefix and
produced `0.0.3-staging.73.1`, the current channel target. The two shell transcripts above are the
before and after of the masking bug.

Needs `focused-local` QA on the Computer settings page: issue a connect command, confirm the copy
button yields a first line a terminal ignores, and confirm the command installs and connects.


Gandy2025 (approved):
Reviewed `23da031e0a2381675755f0f737ebf9c2c6f74fa1...38d4110a1003b8b065a9fc9d5de8dd2229f92ce9` as a high-risk release/install and cross-surface change.

**Standards:** No blocking findings. The implementation preserves explicit error handling, keeps the Server poller and generated command on one configured release base, quotes the installer URL and temporary path, and updates the canonical English and Chinese documentation together.

**Spec:** No blocking findings. The generated Computer command now stops on a failed download, all source defaults use the published Cloud Storage coordinates, and the displayed/copied instruction is inert in interactive zsh while remaining readable to an agent. Keeping the top-level `connect` command is acceptable here because a successful install upgrades the stale Client before invocation.

Verification on this exact head:

- `pnpm check`
- Web Computer-connect tests: 18 passed
- affected Server unit/API/config tests: 86 passed
- portable release script tests: 20 passed
- staging `latest.json` and rendered `install.sh` both return 200; the rendered installer pins the same GCS base
- the inert line exits 0 under interactive zsh, bash, and sh
- required CI is green apart from the expected ownership gate

Non-blocking Standards note: `docs/zh-CN/deploying.md` and `docs/zh-CN/portable-release.md` mirror the English edits, but their `Last synced with` dates still read `2026-09-01`. Please refresh those dates with the next edit to these files, per `CONTRIBUTING.md`.

This approval covers the code review only. The PR's requested focused-local Computer settings → copied command → install → connect QA remains a merge prerequisite, and the ownership gate still requires bestony or yuezengwu.

Gandy2025 (approved):
Re-reviewed the exact-head delta `38d4110a1003b8b065a9fc9d5de8dd2229f92ce9..c5f8f813d4edfe3bf7ded1921b02b3ec7e4d6ca8`.

The new commit changes only the two Chinese mirror `Last synced with` dates from `2026-09-01` to `2026-09-02`, resolving the prior non-blocking Standards note. `git diff --check` passes.

**Standards:** Passed on the new head; the mirror metadata now satisfies `CONTRIBUTING.md`.

**Spec:** Unaffected. The previous install/connect review and verification evidence carries forward because no product, shell, Server, Client, or test behavior changed.

The focused-local Computer settings → copied command → install → connect QA remains required before merge. The ownership gate still requires bestony or yuezengwu.

yuezengwu (approved):
Reviewed exact head c5f8f813d4edfe3bf7ded1921b02b3ec7e4d6ca8 after the synchronize; I did not carry the earlier-head verdict or test result forward.

Standards and Spec: no blocking findings. The Server now has one configured release authority, the generated command stops at a failed download, the displayed first line is a POSIX null command, and the current staging installer upgrades to a Client that exposes both connect entry points.

Current-head verification:

- git diff --check, pnpm check, pnpm build, pnpm typecheck, and pnpm test
- Server integration: 313 passed
- Agent Runtime coverage: 351 passed, 100 percent statements, branches, functions, and lines
- isolated local Browser Smoke: 8 passed
- live staging install from the new GCS URL into a sandbox prefix produced 0.0.3-staging.73.1
- an invalid download returned curl exit 56 and never reached the connect marker
- interactive zsh and sh both accepted the copied null-command line silently
- exact-head Web flow created an Agent, displayed the inert line, bound the Computer through top-level connect, started the daemon, and advanced Prepare Computer to completed

CI triage: the first Browser Smoke attempt failed before any test because PostgreSQL reported that it was shutting down during disposable database creation. The completed rerun is green, consistent with the isolated local 8/8 pass.

Non-blocking deployment note: the staging GCS install.sh and latest.json return 200, while the production equivalents currently return 404 because no production portable target is published. This does not block the staging failure reported in #497, but the production install command in the docs will remain unusable until the first prod portable release is published.

liuchao-001:
## Independent QA — `PASS`

**This is independent QA** run by `code-reviewer` (not the author). Posting under liuchao-001's `gh` account.

| | |
|---|---|
| **Status** | `PASS` |
| **Tier** | `focused-local` |
| **Exact head** | `c5f8f813d4edfe3bf7ded1921b02b3ec7e4d6ca8` |
| **Scope** | Computer settings → issued connect command → pasted into a real interactive zsh → install → connect |

**Evidence boundary.** Everything below was observed against a local staging-mode Server built from this head. It does **not** cover the release pipeline, prod, or the launchd daemon install (see Limitations).

---

### 1. The premise was verified independently

The PR rests on `download.opentag.build` having no TLS certificate. I re-derived this myself rather than taking the description's word for it, collapsing every variable except SNI — same IP, same port, same request:

```
$ dig +short download.opentag.build @8.8.8.8     →  136.110.192.61

SNI = storage.googleapis.com   @ 136.110.192.61:443  →  200
SNI = download.opentag.build   @ 136.110.192.61:443  →  curl: (35) SSL_ERROR_SYSCALL
http://download.opentag.build  @ 136.110.192.61:80   →  200  (server: UploadServer, x-goog-*)

$ openssl s_client -connect download.opentag.build:443 -servername download.opentag.build
no peer certificate available
SSL handshake has read 0 bytes and written 1560 bytes
```

My own resolver returns a proxy fake-IP (`198.18.5.209`), so I pinned the real address with `--resolve` to rule the proxy out. **Premise confirmed: the name terminates TLS with no certificate, and the direction of the PR is correct.**

### 2. (a) The copied first line is silent in a terminal — confirmed

Driven through a real ZLE prompt under a pty, not `zsh -c` (which parses `#` differently and would have hidden the bug). `interactive_comments` is confirmed **off** in this shell.

Old form, with the reporter's exact hostname:

```
% # Run this command in the terminal on superMacBook-M4-Pro.local
zsh: command not found: #
```

Old form, when the Computer name contains an apostrophe (`Ada's Mac`) — **worse than the reported symptom**: the unterminated quote swallows the real command and strands the terminal:

```
% # Run this command in the terminal on Ada's Mac
quote> echo REACHED_COMMAND      ← never runs
quote> exit                      ← never runs
```

New form at this head — returns cleanly, next command runs:

```
% : 'Run this command in the terminal on Ada'\''s Mac'
% echo REACHED_COMMAND
REACHED_COMMAND
```

**Adversarial probe on the escaping.** `displayName` is user-controlled (it is the reported hostname), so I ran six hostile names through `inertComment()` into a real interactive zsh: `Ada's Mac`, `hi!bang`, `` back`tick` ``, `dollar$(touch …)`, `semi; touch …`, and a deliberate quote-break `a'; touch …; '`. **All six were silent, none errored, and no injection fired** (no sentinel file created). History expansion on `!` does not fire inside the single quotes.

### 3. (b) The command really installs and connects — confirmed

I took the command **the running product actually rendered** (read verbatim out of the DOM, not reconstructed) and pasted it into a real interactive zsh:

```
: 'Run this command in the terminal on the Computer you want to connect'
opentag_installer="$(mktemp)" && curl -fsSL https://storage.googleapis.com/opentag-release/releases/staging/install.sh -o "$opentag_installer" && sh "$opentag_installer" && rm -f "$opentag_installer" && PATH="$HOME/.local/bin${PATH:+:$PATH}" "$HOME/.local/bin/opentag-staging" connect --server https://localhost:52951 -- otcc_…
```

Result:

```
Downloading OpenTag portable metadata: https://storage.googleapis.com/opentag-release/releases/staging/latest.json
Downloading OpenTag 0.0.3-staging.73.1 for darwin-arm64
OpenTag 0.0.3-staging.73.1 installed at …/.local/share/opentag/staging/versions/0.0.3-staging.73.1
Connected Computer ea1c6157-c4fe-486b-99cf-764097a77f9b and bound Agent d926e7f4-54c7-467a-b151-faf117ba6e68
```

Server-side confirmation (not just CLI output):

```json
{"computerId":"ea1c6157-c4fe-486b-99cf-764097a77f9b",
 "displayName":"superMacBook-M4-Pro.local","platform":"darwin",
 "connectionStatus":"offline"}
```

The Server also logged `Advertised channel target changed → 0.0.3-staging.73.1` at startup, so the channel-target poller reaches the release from the new default too — that path was equally dead before.

`connectionStatus: offline` is the correct result for the `--no-start` run described under Limitations.

### 4. The masking bug is genuinely fixed, not just described

Both shapes against the same still-broken URL:

```
OLD:  curl -fsSL <broken> | sh && echo REACHED_CONNECT_STEP
      curl: (35) SSL_ERROR_SYSCALL
      REACHED_CONNECT_STEP          ← fell through
      exit=0

NEW:  opentag_installer="$(mktemp)" && curl -fsSL <broken> -o "$opentag_installer" && sh … && echo REACHED_CONNECT_STEP
      curl: (35) SSL_ERROR_SYSCALL
      exit=35                       ← stopped where it broke
```

### 5. Positive control on the single-source-of-truth claim

The PR argues the drift was caused by `buildComputerConnectCommand` holding a second default. I verified the value is genuinely wired through config rather than baked in, by starting a control Server with `OPENTAG_PORTABLE_DOWNLOAD_BASE_URL=https://mirror.example/releases`:

- poller followed it → `Channel target poll failed … url: https://mirror.example/releases/staging/latest.json`
- issued command followed it → `curl -fsSL https://mirror.example/releases/staging/install.sh …`

Both consume the one value. Confirmed at `packages/server/src/index.ts:403` → `config.channelTarget.downloadBaseUrl`.

### 6. Diff audit

- No `download.opentag.build` reference survives anywhere in the repo at this head.
- All six defaults moved, including `apps/cli/src/core/update/portable-installer.ts` — so the Client self-upgrade path is fixed too.
- The **already-published** `staging/install.sh` at GCS carries `DOWNLOAD_BASE_URL=…storage.googleapis.com…`, so the emitted command works today, before this PR ships. No re-release is needed to unblock users.
- Top-level `connect` is registered at `apps/cli/src/cli/program.ts:32` and `computer connect` still exists; the published `0.0.3-staging.73.1` is built from `23da031e`, which is current `main`, so a freshly installed Client accepts the emitted `connect`. The deliberate no-backward-compat decision holds.
- `c5f8f813` vs `38d4110a` is docs-only (two `Last synced with` dates). I rebuilt and re-ran everything at `c5f8f813`; nothing in the earlier head's evidence was carried forward.

---

### 🐛 Findings — none blocking

- **[LOW]** `machine-auth-service.ts:562` — `rm -f "$opentag_installer"` sits inside the `&&` chain, so the temp file is left behind on exactly the failure this PR exists to surface (installer runs and fails). A failing `rm` would also suppress the connect step. Consider `sh "$f"; status=$?; rm -f "$f"; [ $status -eq 0 ] && …`, or simply accept the leak — `mktemp` files are in `$TMPDIR`. Not worth blocking on.
- **[NIT]** The command leaves `opentag_installer` set in the user's interactive shell afterwards. Cosmetic.
- **[NIT]** `docs/portable-release.md:21` still documents the standalone install as `curl -fsSL … | sh`. It has no `&&` chain so it cannot mask a later step, but it is the same shape this PR argues against — worth aligning next time these files are touched.

### ⚠️ Note for liuchao-001 on the "elegance" ask

You asked that the pasted line stay elegant. It does: the block renders `: '…'` in the muted comment style, so it still reads as an instruction line rather than as code (screenshot in evidence). The trade is that the copied text now carries `: '` and `'` around your sentence. That is forced — when the clipboard is unavailable the UI falls back to selecting what is on screen, so displayed and copied text must be identical. I think the trade is right, but it is your call.

### ✅ Verification performed

- Live product path: onboarding → Create agent → Prepare computer → issued command → real interactive zsh paste → install → connect. Black-box via `agent-browser` against a real Server.
- Interactive-zsh pty tests: old vs new comment form, plus six hostile `displayName` values.
- Fail-closed shell comparison against the real broken URL.
- Config override positive control (poller + issued command).
- Network premise re-derivation with SNI isolated.
- Static audit of all 24 changed files against `origin/main`.
- Suites re-run against my own build of this head: `@opentag/web` 561 passed (58 files), `@opentag/server` 714 passed (67 files), `scripts/__tests__/portable-build.test.mjs` 20/20 — all exit 0.
- PR CI at this head: everything green except `ownership-gate`, which is the expected approval gate.

### 🚧 Limitations — what this run does **not** prove

1. **The daemon service install was not exercised.** I ran connect with `--no-start`. Reason: `apps/cli/src/core/daemon/service/launchd.ts:254` does `bootout` on label `opentag-staging` before bootstrapping, and this host has the operator's **real** `opentag-staging` service registered — a faithful run would have booted it out and hijacked the label. This PR changes no daemon-service code. I verified afterwards that the operator's plist, binary and `.zshrc` are all untouched.
2. **The copy button's clipboard payload was not read black-box** — `readText` is permission-denied in this browser profile. I verified the *displayed* line instead, which is what the manual-selection fallback copies, plus the new unit test covering `writeText`. The displayed and copied strings come from the same `commentLine` value.
3. Local Server only, behind a self-signed TLS front (staging mode requires HTTPS), with `NODE_EXTRA_CA_CERTS` and a sandbox `HOME` for the install. Not a release-pipeline or prod verification.
4. The onboarding wait page did not flip to "connected" on its own — the browser logged a transient `AUTH_INVALID_TOKEN` on `/api/v1/me` during my ~7-minute install idle, and a reload showed the correct connected state. Out of this PR's scope and confounded by my headless harness; flagging it only so it is not mistaken for something this PR caused.

### 🌳 Context Tree

No tree change needed. The durable decision that *could* be recorded — "release artifacts are addressed by their Cloud Storage coordinates until `download.opentag.build` gets an HTTPS front" — is captured in the PR description and issue #497, and the PR frames it as reversible in one line. If liuchao-001 decides the Cloudflare-proxy route is *not* happening, that becomes a durable constraint worth a node.

### QA case disposition

`candidate-new-case` — "issued Computer connect command installs and connects from a clean machine" is a recurring, judgement-heavy path with no existing case. The shell-level fail-closed comparison and the `inertComment` hostile-name matrix are deterministic and belong in the product test suite instead, not in a QA case.

### Residual state (retained warm environment, not uncleaned)

Kept: `otqa-pg` Postgres container, pnpm store, the `opentag-pr500` worktree. Removed: my two Servers and TLS fronts, the sandbox `HOME`, and the `pr500qa` / `pr500ovr` databases. The operator's real `opentag-staging` service, binary and shell profile were verified untouched.

### Next steps

1. Optional: decide on the `rm -f` nit — I would merge without it.
2. Merge is still gated on the ownership gate (`bestony` or `yuezengwu`); I am not going around it.
3. **Any new commit invalidates this report** — it is bound to `c5f8f813` only.

liuchao-001:
Triage of the three non-blocking findings from the independent QA. **None applied**, and the reasoning is the same for all three: this head carries a `PASS` QA report and two approvals, and every reviewer stated that a new commit invalidates that evidence. Trading a verified head for three cosmetic improvements is a bad exchange. Recording the decisions here so they are not silently dropped.

### [LOW] `rm -f "$opentag_installer"` sits inside the `&&` chain — **not changing**

Two distinct sub-claims, and I agree with the diagnosis of both:

1. *The temp file leaks on exactly the failure this PR exists to surface.* True. It is a `mktemp` file in `$TMPDIR`, which the OS reclaims; the leak costs nothing a user will notice, and cleaning it up on the failure path means either a `trap` or splitting the chain, both of which make the copied command harder to read. The command is read by humans and pasted into agents, so its legibility is a real constraint, not a nicety.

2. *A failing `rm` would suppress the connect step.* Also true, and the more interesting of the two — `rm` should not be able to gate the thing the command exists to do. But the file was created by `mktemp` moments earlier, in a directory the user owns, and `-f` already suppresses the missing-file case. I could not construct a realistic path where that `rm` fails and the rest of the chain would otherwise have succeeded.

So the reviewer's suggested `sh "$f"; status=$?; rm -f "$f"; [ $status -eq 0 ] && …` is strictly more correct, and I would take it in a change that is touching this line anyway. It is not worth a new head on its own.

### [NIT] `opentag_installer` stays set in the user's shell — **not changing**

Correct, and it is the deliberate cost of the shape. Scoping it away needs a subshell, which then cannot hand the exit status to the connect step without extra plumbing — the readability trade above, for a stray variable in an interactive shell.

### [NIT] `docs/portable-release.md:21` still documents `curl -fsSL … | sh` — **not changing here**

Agreed on the principle, and the reviewer's own qualifier is the reason it is a nit rather than a bug: that one-liner has no `&&` chain, so there is no later step for a failed download to fall through to. The failure is visible where it happens. Aligning it is worth doing the next time these files are edited, exactly as suggested.

---

One thing from the QA worth surfacing, because it makes the comment fix more load-bearing than the original report suggested: with a Computer name containing an apostrophe, the **old** `#` form did not merely print `command not found: #` — it left the shell stranded at a `quote>` continuation prompt and swallowed the real command entirely. The reported symptom was the mild version of that bug.

PR #505, opened 2026-09-02
refactor(web): delete the setup pieces nothing renders

### LINKED ISSUE #494: Dead cluster in apps/web/src/setup: five exports with no reachable reader
## The cluster

Five exports in `apps/web/src/setup/` have no reachable reader. Counting non-test consumers **outside** `setup/` on `main`:

| Symbol | Consumers outside `setup/` |
|---|---|
| `CommandBlock` | 3 |
| `QrCode` | 2 |
| `useRemaining` | 2 |
| `WAITING_LINE` | 5 |
| `readConnectCodeVerdict` | 2 |
| `formatRemaining` | 2 |
| **`CheckLine`** | **0** |
| **`ConnectStatus`** | **0** |
| **`Countdown`** | **0** |
| **`CHECK_COPY`** | **0** |
| **`SETUP_COPY`** | **0** |

`SETUP_COPY.connect` is read only by `ConnectStatus` and `Countdown`; `CHECK_COPY` is read only by `CheckLine`. Nothing outside `setup/` imports any of the three components, so the copy and the components hold each other alive without a reader.

## How it got here

#473 unified Agent creation and setup and deleted the `onboarding-v2` surfaces that rendered them. The exports survived because they are re-exported from `setup/index.ts`, which makes them look like a public surface rather than orphans — a module boundary hides the fact that nobody is on the other side of it.

`SETUP_COPY.messaging` was the same cluster and is being removed in #475, because that PR deleted its last reader directly and could not leave it behind. `connect` was not reachable from that PR's diff, so it stayed.

## Why it is worth doing

This is the #336 class: code that reads as live and is not. It carries a specific cost here — `SETUP_COPY` and `CHECK_COPY` are English-only literals, so anyone auditing i18n coverage finds untranslated user-facing copy and cannot tell without a reachability trace that no user can reach it. That is a false signal for exactly the audit most likely to look.

## Suggested direction

Delete `CheckLine`, `ConnectStatus`, `Countdown`, `CHECK_COPY` and `SETUP_COPY`, along with their `setup/index.ts` re-exports and any tests that exercise them directly.

Before deleting, confirm the reachability claim against `main` at that time rather than against this issue — #475 is currently removing `SETUP_COPY.messaging`, and `main` moves quickly enough that the counts above should be re-derived, not trusted.

If any of the three components is meant to be reused by the unified setup flow rather than deleted, that is a product call and this issue should become "wire them up" instead.

## Not a translation gap

Filed after I briefly concluded the opposite on #475. Since nothing renders these strings, #241 ("Add i18n support for the Control Panel") closing as COMPLETED is not contradicted by them.

liuchao-001:
Closes #494.

## The cluster is ten, not five

#494 counted five dead exports. Re-derived against current `main` — which has moved twice since it was filed, so the issue explicitly said to re-derive rather than trust it — the real count is ten. Non-test consumers **outside** `setup/`:

| Symbol | Consumers | |
|---|---|---|
| `CommandBlock` | 3 | keep |
| `QrCode` | 2 | keep |
| `useRemaining` | 2 | keep |
| `WAITING_LINE` | 5 | keep |
| `readConnectCodeVerdict` | 2 | keep |
| `formatRemaining` | 2 | keep |
| `MessagingCliStatus` | 4 | keep |
| `CheckLine`, `ConnectStatus`, `Countdown` | 0 | delete |
| `CHECK_COPY`, `SETUP_COPY` | 0 | delete |
| `deriveChecks`, `messagingCliCheck`, `CheckRow`, `CheckState`, `RuntimeStatus` | 0 | delete |

The five the issue did not name were kept alive by the ones it did: `CHECK_COPY` existed for `CheckLine`, `SETUP_COPY` for `ConnectStatus` and `Countdown`, and the check-row types for `CHECK_COPY`. `messagingCliCheck`'s last reader was the onboarding step #473 deleted.

## Why a whole module went unnoticed

`setup/index.ts` re-exported all of it. A barrel makes a symbol look like a public surface, so "is anything importing this?" stops being a question anyone asks — the module boundary hides that nobody is on the other side of it. #473 deleted the onboarding surfaces that rendered these, and none of it went along.

## The data went with the code

This is the mistake #475 made and had to correct mid-review, so it is done here up front:

- `setup/copy.ts` is deleted outright — both its exports were dead.
- `setup.css` loses `.ots-check*` and `.ots-slot--status`, which styled only the deleted components.
- `layout.test.ts` loses the assertions that measured those two heights, and the media-query test whose only subject was `.ots-check`.
- `@keyframes ots-fade` **stays**: it is shared with `.ots-pulse`, which three live surfaces still render. I checked rather than assuming it went with its neighbours.
- `checks.test.ts` keeps its `formatRemaining` block and loses the two describing deleted functions.

## Verification

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **58 files / 552 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 4 improvements, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web **92.42%**, up from 91.77%) |

Coverage rises because what is removed was largely untested — which is itself a symptom of the problem, not an argument that the deletion is risky.

Test count drops 560 → 552: the eight removed all exercised deleted functions directly. None asserted anything a reader can see, which is why they stayed green while their subjects became unreachable.

No live QA proposed: nothing rendered changes, and the surviving exports are untouched.


yuezengwu (changes requested):
Reviewed exact head 3344d9af60eeadf8d6662c7637a4e5ac29320bb1. Two blockers remain:

1. This deletes the `.ots-slot--status` min-height rule and its layout regression assertion, but the live `ComputerConnect` status row still renders that class in `apps/web/src/features/computer-connect/computer-connect.tsx`. The deleted rule intentionally reserves the tallest status state; removing it makes the connect-status transition reflow the page, so this is not a behavior-neutral dead-code deletion. Please preserve the rule/test or replace them with equivalent layout protection and coverage.
2. The exact head is currently conflicting with `main` (`apps/web/src/setup/copy.ts` is deleted here but modified on main). Please synchronize and resolve the conflict without discarding the current copy changes.

There is also a stale explanatory comment in `apps/web/src/setup/checks.ts` that still names the deleted `messagingCliCheck`/`CheckState` path; please remove or update it while repairing this change.

Any synchronized head will require a fresh exact-head review.

liuchao-001:
## All three fixed at `18edca7f`

### Blocker 1: `.ots-slot--status` — you are right, and my method was wrong

I checked whether each exported *symbol* had a reader. A CSS class can have one without any symbol doing, and this one does: `computer-connect.tsx:482` writes it directly on the live status row. Deleting its reserved height would have let that row reflow when the status lands — a behaviour change dressed as a dead-code deletion.

The rule and its layout assertion are restored, and the comment now names the surface that actually renders it, so the next person deleting `ConnectStatus`'s neighbours does not repeat this.

`.ots-check` stays deleted — I re-checked it the same way, and its only writer was the `CheckLine` this removes.

### Blocker 2: the conflict with `main`

Resolved. `main`'s change to `copy.ts` was #500 editing `repairCommandComment` inside `SETUP_COPY`; I confirmed it added no reader, so the modify/delete resolves to delete. **#500's actual fix is untouched** — it lives in `messages/computer-connect/{en,zh}.json`, which is what `CommandBlock` renders. Only the dead duplicate goes.

Worth noting as evidence for this PR: that string had **no reader even inside `setup/`**, so a PR merged today spent effort correcting copy nothing displays.

### The stale comment

There were two. The one you found above `formatRemaining` described `messagingCliCheck`/`CheckState`; the file header described the check-row derivation. Both are rewritten to what the file holds — the readiness vocabulary and the connect-code countdown.

## `main` is red, independently of this PR

While rebasing I found `agent-computers.test.tsx` failing on **pristine `origin/main`** (`5c8db374`): #500 replaced the `#` command comment with the POSIX null command `: '…'` and updated `computer-connect.test.tsx`, but that surface's test asserts the same rendered text and was missed.

Fix is up as **#507** (one assertion). This branch merged `main`, so it inherits that failure:

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | 552 passed, **1 failed — the inherited one** |
| same, excluding `agent-computers.test.tsx` | **57 files / 537 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 4 improvements, no new baselines) |

I did not fold #507's one-liner into this branch — it would collide with the standalone fix and hide a red `main` behind an unrelated refactor. This goes green on its own once #507 lands.

yuezengwu (commented):
Exact-head re-review completed for 18edca7f302815fdcaabf0c7c4793a38bb44fcd2. The prior findings are resolved: the still-live `.ots-slot--status` rule and regression assertion are preserved, the stale deleted-symbol commentary is gone, and this branch merges cleanly into current `main`. I found no new blocking defect in this PR diff.

Local validation on this exact head passed: `pnpm check`, `pnpm build`, `pnpm typecheck`, `git diff --check`, and 59 focused Web tests covering the retained layout and remaining setup components.

I am not approving yet because required CI is red on the already-reproduced `apps/web/src/__tests__/agent-computers.test.tsx:78` mismatch in current `main`, which #507 repairs. This PR does not modify that test or behavior. Once the base CI repair is merged and required checks rerun, this exact head must be refreshed before approval.

liuchao-001:
## Head refreshed at `d8cc6170` — the base repair is merged and this is green on it

@yuezengwu — #507 merged as `33fbf6ec`, so the red you were waiting on is gone. I verified that independently before refreshing: a detached checkout of pristine `origin/main` now runs **58 files / 561 tests** clean.

This branch has merged that `main`. The only conflict-free change it brought in was the repaired assertion in `agent-computers.test.tsx`, which this PR does not touch.

## Testing at `d8cc6170`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **58 files / 553 tests passed** |
| `pnpm check` | exit 0 (ratchet passed, 4 improvements, no new baselines) |
| `node scripts/unit-coverage.mjs --project web` | exit 0 (web **92.70%**) |

553 rather than main's 561: the eight removed tests all exercised deleted functions directly, none of them asserting anything a reader can see — which is why they stayed green while their subjects became unreachable.

Nothing else changed since the head you re-reviewed, so the `.ots-slot--status` restoration and the rewritten `checks.ts` comments are as you saw them.

yuezengwu (approved):
Exact-head re-review completed for d8cc6170dbe3abcbe279e3bebc17d9849fee993b against base 33fbf6ecff1004a8b0d1fd90e72dbaad4493f649. This head was reviewed from zero after synchronization. The deleted setup exports have no remaining readers, the live ComputerConnect status rule and its layout assertion remain intact, and the change is consistent with the verified canonical Context Tree. I found no blocking defect. Local validation passed: git diff --check, pnpm check, pnpm build, pnpm typecheck, and the complete Web suite with 58 files and 553 tests. All current GitHub checks are passing or skipped. Approving this exact head.

PR #507, opened 2026-09-02
test(web): follow the inert command comment into the Agent computers test

liuchao-001:
## `main` is red, and has been since #500 merged

`apps/web/src/__tests__/agent-computers.test.tsx` fails on pristine `origin/main` (`5c8db374`):

```
FAIL src/__tests__/agent-computers.test.tsx > generates a command naming the assigned Computer
Unable to find an element with the text: # Run this command in the terminal on Ada's Mac
```

I confirmed this on a detached checkout of `origin/main` with nothing of my own in it, after hitting it while rebasing an unrelated branch.

## Why

#500 deliberately replaced the `# …` comment in the connect command with the POSIX null command `: '…'` — and for a good reason, quoted from its own comment: zsh leaves `interactive_comments` off for interactive use, so a pasted `#` line runs and answers `command not found: #`, which reads as a failure on a command that actually works.

It updated the catalogues and `computer-connect.test.tsx`, which now asserts `: 'Run this command in the terminal on Ada'\''s Mac'`. It missed `agent-computers.test.tsx`, which renders the same block through the Agent computers surface and still expected the `#` form.

## The fix

One assertion, updated to the form the block renders. Nothing in the product changes.

I am **not** reverting or re-litigating #500's decision — the inert comment is right, and this is only the assertion it left behind.

## Testing at `ef51f075`

| Command | Result |
|---|---|
| `pnpm --filter @opentag/web test` | **58 files / 561 tests passed** |
| target file alone | 16/16 passed |

Prioritising this over my own #505, which is blocked behind the same red.

### DIFF
--- a/apps/web/src/__tests__/agent-computers.test.tsx
+++ b/apps/web/src/__tests__/agent-computers.test.tsx
@@ -75,7 +75,9 @@ describe("OpenTag Web App Shell", () => {
     expect(screen.getByRole("region", { name: "Reconnect Ada's Mac" })).toBeTruthy();
     expect(screen.queryByRole("heading", { name: "Reconnect Ada's Mac" })).toBeNull();
     expect(await screen.findByRole("button", { name: "Copy command" })).toBeTruthy();
-    expect(screen.getByText("# Run this command in the terminal on Ada's Mac")).toBeTruthy();
+    // The comment is the POSIX null command, not a `#` line: zsh runs a pasted `#` and answers
+    // "command not found", so #500 made it inert. This is the rendered form.
+    expect(screen.getByText(": 'Run this command in the terminal on Ada'\\''s Mac'")).toBeTruthy();
     const repairRequests = vi
       .mocked(fetch)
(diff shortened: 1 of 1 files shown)


yuezengwu (approved):
Reviewed exact head ef51f075b9f8da675cb97553fdec0765e44a7553. This is the correctly scoped repair for the post-#500 main regression: it updates the remaining rendered-surface assertion to the POSIX null-command form without changing product behavior. The branch is cleanly mergeable; `git diff --check`, root build, and the focused 16-test Web file passed locally, and all GitHub checks are pass/skipping.

liuchao-001:
## Correction: the SHA in the testing heading was wrong

The heading said `Testing at 4bf3b28`. **That commit does not exist** — `git cat-file` rejects it. The real head is `ef51f075`, which is what @yuezengwu reviewed, and I have corrected the body.

The results under it were genuine and re-run on `ef51f075`: 58 files / 561 tests, and 16/16 for `agent-computers.test.tsx` alone. What was wrong was the label, not the evidence — I wrote the SHA without reading it back after the push.

Flagging it rather than editing silently, because a testing heading is a claim a reviewer is entitled to check against a real commit.

PR #498, opened 2026-09-02
feat: lead with Slack wherever a messaging channel is chosen

liuchao-001:
## What

Nothing in the product should read as though Feishu were the recommended messaging channel. Three places did.

**1. Settings → Messaging, unconnected state** (`im-tab.tsx`)

Lark was an emphasis (filled) button and Slack a secondary one, in that order, with no marks. Both connect actions now use the same neutral `secondary` variant, Slack leads, and each carries its provider's official mark via the existing `ProviderIcon`, passed as kumo's `icon` prop so the loader replaces the mark while a connect is in flight.

**2. The sentence above those buttons**

It named Lark first while the buttons led with Slack. The call site now names Slack first.

That sentence is the one place both brands appear in a single line, which is exactly where a hard-coded space cannot be right: whichever brand the zh catalogue spaced for, the other one is wrong. So the zh template drops its space and `spaceScriptBoundary` places it around whichever brand turns out to be Latin — the same composition the disconnect copy in this file already uses. English is unaffected.

**3. Onboarding's channel choice** (`agent-setup-service.ts`)

The setup flow still offered Feishu → Slack, contradicting the settings page. The order lives in the **server's** action derivation, not only in the web memory adapter that mirrors it, so both moved. Slack leads where it is offered; Feishu remains the entry that survives a deployment without Slack OAuth (`slackOAuthAvailable === false` still yields Feishu alone).

## Why

Which messaging app a team already lives in decides this choice, not us. Emphasis styling, sentence order, and card order are all ways of implying a default, so they have to agree — and they now agree on presenting the two channels as equals with Slack first.

## Scope

`apps/web` plus one `packages/server` action-ordering change. No API shape change, no migration, no behavioural change to binding lifecycle. Accessible names are unchanged (`Connect Slack` / `Connect Lark`) because `ProviderIcon` is decorative and `aria-hidden`.

## Tests

- `agent-messaging.test.tsx` locks the equal-choice contract: button order, class parity between the two buttons, `secondary`'s own surface and text tokens, and — read out of each inlined SVG's `<title>` — that each button wears *its own* mark rather than merely *a* mark.
- `brand-in-message.test.ts` pins the finished sentence in both locales, including the Chinese case where one Latin and one CJK brand share a line (`连接 Slack 或飞书，…`).
- `agent-setup-page.test.tsx` pins the rendered onboarding card order; the server and adapter suites pin both branches of the action derivation.
- `pnpm check`, `pnpm typecheck`, `pnpm test` green locally.

[...]

PR #508, opened 2026-09-02
test(web): follow the repair command's comment to its shell-safe form

liuchao-001:
## What

`main` is red. `apps/web/src/__tests__/agent-computers.test.tsx` still expects the Computer repair block to open with:

```
# Run this command in the terminal on Ada's Mac
```

That `#` form is precisely what #500 removed. An interactive shell rejects a pasted comment line, so the block now opens with the POSIX null command, closing and reopening the apostrophe so the sentence cannot escape its quotes:

```
: 'Run this command in the terminal on Ada'\''s Mac'
```

`computer-connect.test.tsx` was updated to match — it carries the same string as a named constant with that reasoning. This one assertion was missed.

## Evidence

The failure is on `main` itself, not on any branch: `agent-computers.test.tsx > generates a command naming the assigned Computer without leaving the Agent` fails identically on a clean `origin/main` checkout at `5c8db374`, and main's own CI run for that SHA fails on **Node 22, Node 24, Node 26 and Checks** with this test.

With this change, on the same checkout: `pnpm check`, `pnpm typecheck` and `pnpm test` all exit `0`.

## Scope

One assertion in one test file. No product change — the rendered form is already correct and deliberate; only the expectation was stale.

## How it was found

It surfaced while merging current `main` into #498, which is unrelated to Computer connect. Filed separately rather than folded in, so the fix for a red `main` is not gated on that PR's review.

### DIFF
--- a/apps/web/src/__tests__/agent-computers.test.tsx
+++ b/apps/web/src/__tests__/agent-computers.test.tsx
@@ -75,7 +75,9 @@ describe("OpenTag Web App Shell", () => {
     expect(screen.getByRole("region", { name: "Reconnect Ada's Mac" })).toBeTruthy();
     expect(screen.queryByRole("heading", { name: "Reconnect Ada's Mac" })).toBeNull();
     expect(await screen.findByRole("button", { name: "Copy command" })).toBeTruthy();
-    expect(screen.getByText("# Run this command in the terminal on Ada's Mac")).toBeTruthy();
+    // The POSIX null command, with the apostrophe closed and reopened so the sentence cannot escape
+    // its quotes: a `#` comment is what an interactive shell rejects when the block is pasted.
+    expect(screen.getByText(": 'Run this command in the terminal on Ada'\\''s Mac'")).toBeTruthy();
     const repairRequests = vi
       .mocked(fetch)
       .mock.calls.filter(([path, init]) => path === "/api/v1/computer-connect-codes" && init?.method === "POST");


liuchao-001:
## Independent QA — PR #508 (bound to head `52c5e759`)

**Status: PASS** · **Tier: `test-only`** · run by code-reviewer (independent of the author)

- **Exact target**: `52c5e759cada5bd65f1f29ea340a31c2bab55bb9`, base `main` `5c8db374`, confirmed with
  `gh pr view --json headRefOid`; worktree from `refs/pull/508/head`.
- **Tier**: `test-only` is right — zero product code changes, the subject *is* one assertion, and
  the deterministic check is the assertion itself. No run cell, no live tier claimed.
- **Evidence boundary**: this proves the named commands passed and that the repaired assertion
  matches what the component actually renders under jsdom. It says nothing about whether the
  emitted command behaves correctly when pasted into a real terminal — that was #500's question and
  I did not re-qualify it here.

### I did not take either assertion on trust

Rather than compare old string to new string, I replaced the assertion with a probe that forces the
rendered node into the failure output:

```
expected [ Array(1) ] to deeply equal [ '__QA_PROBE__' ]
+   ": 'Run this command in the terminal on Ada'\''s Mac'
+ opentag computer connect --server https://opentag.example.com -- example",
```

So the page really does render the POSIX null-command form, and the new expectation matches it. The
old `# Run this command in the terminal on Ada's Mac` matches nothing on the page — a **stale
assertion, not a product defect**, exactly as described.

The producing code agrees: `apps/web/src/setup/command-block.tsx:145`

```ts
function inertComment(comment: string): string {
  return `: '${comment.replace(/'/g, `'\\''`)}'`;
}
```

`computer_connect_repair_command_comment` is unchanged in both catalogues; only the wrapper changed
in #500. And the escaping is semantically right, not just string-equal: `: 'Run … Ada'\''s Mac'` is
the null command with one argument, and closing/reopening the quote around the apostrophe is
precisely what stops an interactive paste from dropping into a `quote>` continuation.

### Sweep and sibling agreement

- No `# Run this command` expectation survives anywhere under `apps/`, `packages/`, `docs/`, `e2e/`.
  This was the last one.
- `computer-connect.test.tsx:18` already carries the same string as `REPAIR_COMMENT`. I compared the
  two **source literals** programmatically, not by eye: byte-identical. The two suites now agree.

### Controls

| Run | Result |
|---|---|
| PR head `52c5e759`, `agent-computers.test.tsx` alone | **16 passed** |
| Same worktree, that one file reverted to `origin/main` (`git checkout origin/main -- …`), nothing else changed | **1 failed / 15 passed**, `Unable to find an element with the text: # Run this command in the terminal on Ada's Mac` |
| PR head, `pnpm check` | exit `0` |
| PR head, `pnpm --filter @opentag/web test` | **58 files / 561 tests**, exit `0` |

A controlled negative and positive on the same tree, so the repair is what changes the outcome.

CI on this PR is now green across the board — including **Node 22.22.2 / 24 / 26 Compatibility**
and `Checks`, which are the exact jobs failing on `main` at `5c8db374`.

### One note, deliberately not a change request

Do **not** tidy this by computing the expectation in the test, e.g.
`inertComment(m.computer_connect_repair_command_comment({ computerName }))`. That would assert the
code against itself and stop catching the drift this PR is repairing — the same failure mode we hit
on #498, where a test re-implemented the production composition and therefore never observed the
call site. Hard-coding the finished string is the correct choice here, and the duplication between
the two test files is the price of it. If it ever becomes three places, share a fixture constant;
do not derive it from the helper.

### Recommendation

Merge-ready from my side: correct, minimal, controlled, and it unblocks `main`. Landing this first
and then merging `main` into #498 is the right order — that produces a new head for #498 and I will
re-run its QA against it.

liuchao-001:
Closing as a duplicate of #507, which fixed the same stale assertion and merged to `main` at `33fbf6ec`.

The two repairs are byte-identical in the assertion itself — only the explanatory comment differs — so `main` is green and nothing here is still needed.

This was a race, not a disagreement: #507 was opened at 23:42:51Z and this at 23:46:43Z, after a search of open PRs that predated #507's existence. Recording it so the duplicate is legible rather than mysterious.

Thanks to whoever picked it up first. The QA round already run against this branch (`52c5e759`) applies to an equivalent change, and its finding — do not compute the expectation from `inertComment(...)`, because a test that re-implements the production composition stops observing the call site — is worth keeping in mind for whoever touches that assertion next. #507 hard-codes the finished string too, so it is already on the right side of that.

PR #509, opened 2026-09-03
fix(feishu): read a rich-text title and settle @ elements and empty paragraphs

Gandy2025:
## Summary

Follow-ups from the review of #492, which read a Feishu rich-text (`post`) message from its documented shape. Three of the four were raised there as non-blocking; the fourth is the layout case the reviewer asked to be checked.

- **A rich-text title is read as the first line.** The old walker and #492 both dropped `title`, yet it is authored content: a request whose ask sits in the heading reached the stored text, the Task title, and the Agent prompt without it.
- **An `@` element without a usable id is written by name.** The lookup could match a mention entry that had no id either (`undefined === undefined`) and write that entry's key, naming the wrong person and tokenising the wrong mention. The id must now be a non-empty string before any mention is consulted.
- **A paragraph that renders nothing is left out.** An image on its own line used to become a blank line in the text; it is skipped, while an authored empty paragraph still stays.
- **A long unbreakable Agent name wraps inside its details column** on the Task detail page instead of spilling across its neighbour in the five-column grid.

## Testing

- `pnpm check`, `pnpm typecheck`, `@opentag/server` and `@opentag/web` test suites on the rebased branch; the full gate including `pnpm build` and the e2e smoke suite passed on the same diff before the rebase.
- Adapter tests: the locale-wrapped `post` fixture now expects its title as the first line; runs of one paragraph stay on one line with the image-only paragraph dropped; an `@` element without an id next to a mention without an id renders by name.

## UI validation

- Desktop evidence: local stack at 1440px with a 60-character unbreakable Agent display name in the DEV example Task: the name wraps inside its 179px column, no details cell overflows (`scrollWidth` equals `clientWidth` for all five), Status stays in the fifth column.
- Mobile evidence: unchanged from #492 (stacked single column below 36rem).
- Widths checked (`320px` / `390px` / `768px` / `1440px`): 1440px for the new case; the other widths are unchanged from #492, where all four were checked.
- States verified: failed example with the long name.
- Keyboard/accessibility checks: no interactive change.
- New reusable block, theme value, or CSS exception: none.

## Breaking changes

None.

## Checklist

### DIFF
--- a/apps/web/src/features/tasks-page.tsx
+++ b/apps/web/src/features/tasks-page.tsx
@@ -386,14 +386,14 @@ export function TaskDetailPage({
           aria-label={m.tasks_details()}
         >
           <TaskDetailFact label={m.tasks_agent_label()}>
-            <span className="inline-flex items-center gap-2">
+            <span className="flex min-w-0 items-center gap-2">
               <span
-                className="grid size-7 place-items-center rounded-full bg-kumo-brand text-xs font-medium text-kumo-inverse"
+                className="grid size-7 shrink-0 place-items-center rounded-full bg-kumo-brand text-xs font-medium text-kumo-inverse"
                 aria-hidden="true"
               >
                 {task.agent.displayName.charAt(0)}
               </span>
-              <strong>{task.agent.displayName}</strong>
+              <strong className="min-w-0 break-words">{task.agent.displayName}</strong>
             </span>
           </TaskDetailFact>
           <TaskDetailFact label={m.tasks_source_label()}>
--- a/packages/server/src/__tests__/feishu-adapter.test.ts
+++ b/packages/server/src/__tests__/feishu-adapter.test.ts
@@ -172,7 +172,7 @@ describe("Feishu adapter", () => {
       "[unsupported:text]",
       "[image]",
       "[file]",
-      "first\nsecond",
+      "Title\nfirst\nsecond",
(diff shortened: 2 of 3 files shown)


yuezengwu (approved):
Approved on exact head `28ff6848c5d0495bd0479eb683d7f641e850c672` (PR base `5a7137867b2af1d1dc738178e293745d872aa8cc`). A read-only merge-tree check against current `main` (`fa28253fd8abdf4b8cdc64d16959e299dcf2fea2`) is clean.

The title, mention-id, and rendered-empty-paragraph changes preserve one normalized plaintext source and address the three follow-ups from #492 without changing Integration or Session authority. The layout change gives the Agent flex row and its name a shrinkable width while keeping the avatar fixed. Local Browser QA at 1440px confirmed all five detail cells remain within their 179px columns (`scrollWidth === clientWidth`), and the Agent name resolves to `min-width: 0` plus `overflow-wrap: break-word`.

Local verification on this exact head:
- `git diff --check`
- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/server test:integration` (18 files / 313 tests)

Remote code checks are green; `ownership-gate` is the expected code-owner approval gate. Canonical context checked at `first-tree-ai/opentag-context-tree@062af96ce5917dae60c796f4a23e0175a70de5c3`.

PR #511, opened 2026-09-03
docs: rewrite the README for launch

serenakeyitan:
Refs #510

## Summary

Rewrites `README.md` and `README.zh-CN.md` from an engineering changelog into a product introduction, and merges `main` in.

**Structure.** Centered hero (logo, tagline, description, badges, nav, EN ↔ 中文) · an SVG walkthrough of the four-step story · What is OpenTag? · three capability sections (context, tools on that machine, ownership) · a Quickstart that runs · Runtimes and chat-platform tables · architecture diagram · task-indexed documentation table · Project status · Contributing · License.

## Review findings addressed

Merged rather than rebased, deliberately: this branch is under active review with line-anchored findings, and a force-push would orphan them. `README.md` and `README.zh-CN.md` conflicted and were resolved onto the rewritten files.

| Finding | Fix |
| --- | --- |
| Retired `enrollment` vocabulary, 5 hits | Replaced with connection wording. `legacy-persistence-allowlist.test.ts` passes locally. |
| Quickstart could not produce a login code | `bootstrap:admin` now runs *before* the foreground server, with the required `OPENTAG_BOOTSTRAP_EMAIL` and `OPENTAG_BOOTSTRAP_DISPLAY_NAME`. |
| Missing `export PATH` | Restored in the install block, with a note to persist it in the shell profile. |
| Node floor wrong | Corrected to `main`'s engines: 22.22.2 / 24.15.0 / 26, in badge and prose. |
| Project status and demo-data disclosure deleted | Both restored. The disclosure is narrowed to **Skills and Integrations**; Tasks ships for real after #488. |
| `TRADEMARKS.md` / `docs/tasks.md` links and trademark notice absent | All three added, plus an explicit Anthropic trademark and non-affiliation notice under License. |
| "Cloud coming soon" | Removed. Nothing in the repo stands behind it. |
| Context-repo-on-GitHub claims | Removed. The repo has no GitHub client and no context-tree concept. The section is back to the three doc-backed capabilities it had displaced: bounded Thread Session history, `${OPENTAG_HOME}` files, internal Session collaboration. |
| "every connector you do, no setup needed" / "your whole toolbox" | Replaced with the bounded capability: a tool that is installed, signed in, and visible to the provider runtime can be used from the thread, subject to that tool's own permissions. OpenTag does not discover or expose tools. |
| "no token handed to a third party" | Narrowed to no credential custody. The CLIs still authenticate to their own providers. |
[...]

PR #513, opened 2026-09-03
docs: use the website's own walkthrough animation

serenakeyitan:
## Summary

Replaces the hand-drawn SVG walkthrough with a recording of the real one on [opentag.build](https://opentag.build): the site's own markup, CSS, and JavaScript driving the four steps, captured from a local checkout of `first-tree-ai/opentag-site` and cropped to `#steps`. Recorded with the site's own `--speed` knob at 1.5, so one cycle is 15.9s rather than 29.75s. Nothing about the animation is reimplemented.

Only three files change: the two READMEs (image path and alt text) and the asset itself.

## On the earlier review of this asset

The GIF was replaced by the SVG in #511 because reviewers read its provider picker and connector grid as product claims. **That is a deliberate owner decision reversed here:** this is the marketing walkthrough, it is the same one visitors already see on the website, and it is meant to run slightly ahead of what has shipped. The README prose around it is unchanged and stays scoped to what exists, including the bounded connector statement and the Project status section.

Merging without waiting on the ownership gate, as the owner's call.

## Testing

Documentation only.

- Both READMEs still resolve every relative link, anchor, and image.
- The asset is the byte-identical recording previously carried at `e5e44c46`, 1,791,172 bytes.

## Breaking changes

None.


liuchao-001 (commented):
Post-merge review — review was requested at 03:53:24Z and the merge landed at 03:53:33Z, so this is a record of findings rather than a gate. Two of them are worth a follow-up regardless of the owner decision about what the animation is allowed to claim.

## 1. The recording stops after step 2 of 4

`docs/assets/opentag-walkthrough.gif` is 1000×580, 190 frames, 15.83s, `NETSCAPE2.0` loop count 0 — it repeats forever. Sampling each of the four step badges in the left rail across all 190 frames: `01` and `02` turn green (`#81bf2a`), `03` and `04` stay grey (`#dddfdd`) in every single frame. The pane only ever renders `1/3 Connect your workspace`, `2/3 Choose your model`, `3/3 You're set up`, and the Product Eng thread. **Shared knowledge** and **Connect your stack** never play.

So the README hero loops with half of its own left rail permanently inert, and both alt texts describe four steps that a reader never sees.

The PR body's own arithmetic points the same way: at `--speed 1.5`, a 29.75s cycle is 19.83s, not 15.9s. The capture is about four seconds short of one full cycle.

## 2. The alt text widened the claim that the #511 review had narrowed

Two of the three changed lines are prose, and both moved:

- EN: "thread context read from the root message, and reaching the tools already set up on that machine" → "shared knowledge kept on your own machine, and connecting the rest of your stack"
- zh-CN: "……以及连接你的其余工具。"

`README.md:54` still bounds this to "a tool that is installed, signed in, and visible to the provider runtime on that Computer … OpenTag does not discover or expose tools." "Connecting the rest of your stack" is the wider claim #511 took out, and it now reaches screen-reader users only. Cheap to fix, and independent of the animation question.

## 3. For the record on the asset

The blob is `19ff78ca`, 1,791,172 bytes, first committed at `e5e44c46` (Sep 2, 19:03 PDT) and dropped from the #511 branch before that PR merged at 20:45. This restores those exact bytes rather than being a new capture — the body says as much; noting it only so the history is explicit in one place.

## 4. Where running ahead costs something

Marketing running ahead of what has shipped is the owner's call and I am not relitigating it. Two specific costs:

- The provider picker (Claude Code, Codex, **Kimi**, **DeepSeek**, **Z.ai**, "Any provider") sits about 110 lines above the README's own Runtimes table, which lists exactly two: `codex` and `claude`. Same page, two answers.
- The recording carries third-party marks — Anthropic, OpenAI, Slack, GitHub, Google Analytics, plus Kimi / DeepSeek / Z.ai — and shows the bot pulling Analytics numbers and opening GitHub issue #482, while Project status says the Skills and Integrations areas are interface previews backed by demo data. #511 moved to lettered marks specifically to avoid this, and the trademark / non-affiliation notice under License now sits below a hero full of logos.

## 5. Process

Review was requested from four accounts and the merge landed twelve seconds later with zero reviews and CI unfinished; the Ownership Gate's own "Evaluate ownership" check reported at 03:53:39, six seconds after the merge. Unblocking yourself as owner is one thing — but requesting review you do not intend to wait for makes the request decorative. (The PR run's Browser Smoke failure was a flake: the push run on `main` is green on the same tree.)

Happy to send the alt-text correction as a one-line-per-README follow-up. The truncated capture needs someone with the `opentag-site` checkout to re-record a full cycle.

Gandy2025 (commented):
Post-merge review of `56fc50117027d8db8137595e169d7ba440b4336a...31e4c8bbf65b89f551dcfa001de2ff56e82d1d16` (low-risk documentation change).

I independently validated the substantive finding already recorded by @liuchao-001: the GIF is 190 frames / 15.83 seconds and loops indefinitely, but steps 03 and 04 never become active in any decoded frame. The capture therefore does not deliver the four-step walkthrough described by the PR and both alt texts. A full-cycle re-recording is the smallest correction.

One factual correction to the existing review record: the PR body's timing arithmetic is sound. The site's four base durations total 23.8 seconds, the deployed baseline uses `--speed: .8` (`23.8 / .8 = 29.75`), and a local value of `1.5` gives `23.8 / 1.5 = 15.87` seconds. The truncation is proven by the frame content, not by dividing 29.75 directly by 1.5. The observed first two steps running at roughly their unaccelerated durations suggests the capture did not actually apply the intended speed to the rail clock.

On the Spec axis, I treat the stated owner decision that the marketing animation may run ahead of shipped capability as authoritative and do not repeat the provider/connector claim objections. On the Standards axis, the incomplete capture is the only content blocker I validated beyond the already-recorded post-merge process concern.

liuchao-001:
@Gandy2025 is right and my arithmetic in point 1 was wrong — withdrawing it. The rail's base durations are `const DUR = [6.5, 7.5, 4.8, 5]` (`opentag-site`, `public/index.html`), totalling 23.8s, and the deployed `--speed:.8` is what makes the cycle 29.75s. Dividing 29.75 by 1.5 double-counted that .8; `23.8 / 1.5 = 15.87s` is exactly the length of this capture, so the PR body's number was self-consistent all along. The truncation stands on the frame content, which is how it was actually established.

What the frames say about *why* it truncated, since it points at the fix:

Sampling the step-01 and step-02 badges across all 190 frames (12.003 fps): **01** is green from 0.00s to 7.42s, **02** from 7.42s to the final frame — 8.41s in and still running when the recording stops. At `--speed: 1.5` step 01 would run `6.5 / 1.5 = 4.33s`. The observed pacing is the deployed `.8` clock (step 01's remaining 7.42s of its 8.125s, with the capture starting about 0.7s in — frame 1 already has the cursor mid-scene over the Feishu card). So the knob never reached the clock, while the window length was chosen as if it had, and a 29.75s cycle got cut a little past its midpoint.

The mechanism is in the rail script: `SPEED` is read once inside the IIFE via `getComputedStyle(...).getPropertyValue('--speed')`, `DUR` is derived from it there, and every panel scene animates on `var(--dur)`, which JS sets per step from that array. A `--speed` applied *after* that script has run — devtools, an injected stylesheet, `style.setProperty` on `<html>` — restyles only the few `calc(x / var(--speed))` entrance animations and leaves the rail and the scenes on the old clock. Changing the `:root` declaration itself (`public/index.html:240`) before the page loads is the version that takes.

So the re-record has two acceptance checks rather than one:

1. All four badges go green in the capture — 03 and 04 currently never do.
2. Step 01 lasts about 4.3s if `1.5` genuinely applied. If it lasts ~8s, the knob did not land and the capture needs the full 29.75s at the deployed `.8`.

liuchao-001:
Follow-up for the alt-text finding is open as #530 — the one clause per README, against current `main`, with the animation itself untouched.

Two notes from opening it. The recording still needs someone with an `opentag-site` checkout to re-record a full cycle; the acceptance checks are in the comment above. And #518, currently draft, rewrites both READMEs and deletes the bulleted statement that bounds the connector claim, so the two changes meet on adjacent lines — flagged there.

PR #516, opened 2026-09-03
fix: complete targeted Computer preparation (F3)

yuezengwu:
## Summary

- Return the exact target Agent Runtime only to Clients advertising support in existing `clientVersion` SemVer build metadata. Unmarked Clients retain the legacy strict response shape; new Clients accept older Servers but fail closed on missing Runtime evidence. No version floor or database schema changes.
- Reuse the canonical selected Runtime factory's complete capability/credential probe, with a bounded deadline. Never install or guess a Runtime. Add the idempotent, explicitly selected `computer runtime-inspect` verification command. Its transient failures use `backoff`; manual install/auth/version/configuration failures use `never`.
- Project human and JSON output from one schema-validated `LocalComputerPreparationResult`: Computer (connection/daemon checks), selected Runtime, Lark, Slack. Both normal and fail-closed paths derive their counts from required rows. Prepare both Providers even after Runtime or daemon failure. `--no-start` skips only daemon; `--no-prepare-provider-clis` explicitly skips both Providers.
- Preserve the committed Server exchange through local persistence failures, report `codeConsumed: true` / `localPersistenceReady: false`, and never replay the code or start a daemon with incomplete local files. Save credentials first, then identity; add exact-installation `computer repair-local` recovery without another exchange. If the new credential was not saved, request a new code after fixing storage. No two-file atomicity or rollback to potentially revoked credentials is claimed.
- Supply independent repair/verify actions, including fallback Provider rows. Preserve custom Home in recovery/daemon commands and keep warnings non-blocking. Local READY does not assert Server/Web readiness; the Runtime probe is point-in-time local evidence, not daemon parity.

Non-goals: no Web, setup snapshot/service, prewarm-owner, migration, IM authorization, or release workflow changes. This is F3 of the September 3 onboarding repair plan, independently deployable from F1/F2/F4.

## Testing

Validated head: `c5640f62d4d9252b673029cf3f16bd73e3126749`, with main `761fc12042a34b43451b42c1091f19fbb9dbeda2` incorporated by merge `a4d8659f7d37a51656cccf58636725a0ec2003ab`. The published branch was not rebased or force-pushed. The export-snapshot conflict retained both sides.

- `pnpm install --frozen-lockfile` — passed; no F3 dependency manifest/lock changes.
- `pnpm check` — passed, including ownership-policy/migration/complexity checks (existing warnings retained; PR approval is a separate gate).
- `pnpm build` — passed.
- `pnpm typecheck` — passed.
- `pnpm test` — passed: Shared 213, Client 950, Server 738, CLI 411, Web 584, plus 311 root script tests.
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 357 tests passed; statements/branches/functions/lines all 100%.
- `pnpm --filter @opentag/server test:integration` — 18 files / 314 tests passed against owned PostgreSQL containers.
- Built CLI `connect --help`, `computer runtime-inspect --help`, and `computer repair-local --help` — passed; explicit provider/installation inputs and skip scopes visible.
- `git diff --check` — passed; dedicated worktree clean.

GitHub validation on this same head passed: [required CI](https://github.com/first-tree-ai/opentag/actions/runs/33724730037)
(including Node 22/24/26 compatibility, CLI packaging, browser/container smoke, integration and Runtime coverage), and
[patch coverage](https://github.com/first-tree-ai/opentag/actions/runs/33724730085): 773/773 changed executable lines (100%; threshold 80%). PR image build, documentation mirror advisory, and quality scoreboard also passed.

All local gates were rerun after merging main and implementing the review fixes. No validation was transferred blindly from the previous head.

Coverage includes selected Codex/Claude-only probes; Runtime missing/credential/version/capability/configuration failures; bounded cancellation; old/new exchange compatibility including a frozen published strict parser against the real exchange route; Provider success/single/double/manual/lock/missing-result/warning cases; both skip flags; daemon failures; custom Home shell quoting; one JSON document; real phase forwarding; connection-preserved/no-code-reuse outcomes; schema-validated fallback and human/JSON parity. Persistence tests cover both writes failing before/after committing, fresh/existing homes, refusal of old/missing/unreadable credentials, repeated exact-installation recovery, daemon exclusion, and secret redaction.

The tests use owned local fixtures and Testcontainers PostgreSQL. No claim of live full-onboarding QA or a recording is made here.

## Breaking changes

Targeted connect JSON now reports `result.preparation` and `result.guidance` instead of the legacy Provider-only projection. After a successful exchange, incomplete local preparation returns exit 3 with `LOCAL_COMPUTER_PREPARATION_INCOMPLETE` and retryability `never`, because connect codes are one-time. Post-exchange local write failure instead returns `COMPUTER_LOCAL_PERSISTENCE_FAILED`, `connected: true` (Server-side only), `codeConsumed: true`, and `localPersistenceReady: false`, with local repair/new-code guidance. Standalone Provider CLI repair semantics are unchanged. Explicit required-check skips remain blocking rather than reporting local success.

Legacy Client exchange response shapes and machine-credential file formats remain unchanged; credential/identity write ordering and failure recovery are corrected.

## Checklist

- [x] Canonical English docs and Chinese mirrors are synchronized when applicable (the updated CLI README has no Chinese mirror).

## Review and merge

This CLI/shared/server change is gated. A non-author CODEOWNER approval and green required CI are required; no admin bypass or self-approval. Refresh the live PR head and latest main immediately before squash merge. Do not transfer review evidence to a different head.

Gandy2025 and liuchao-001 both approved code at `c5640f62`. [Review disposition and validation](https://github.com/first-tree-ai/opentag/pull/516#issuecomment-5521706322) cover all original findings. Their non-blocking explicit-skip summary suggestion is deferred: skip rows already render `skipped (blocking)` with repair/verify actions, but have no diagnostic-code summary. The identity format remains version 2; consolidating that existing version literal is also not part of this fix.

Current handoff: implementation, review corrections, and validation complete; **not merged**. The ownership gate still reports 26 gated files requiring bestony approval. That review remains requested; neither the author nor these two code approvals substitute for the required non-author CODEOWNER. This PR has not been squash-merged.


Gandy2025 (changes requested):
Reviewed immutable range `a203c2dd4d8484758a38814cbc1657e309cc9bf0...e0327a6dfc53a88dab336af1eca22ecede87a521`.

Risk: high — this crosses the one-time credential exchange, compatibility negotiation, local persistence, daemon startup, and provider/runtime readiness boundaries.

Standards: no blocking findings against the repository instructions, package dependency direction, public-export conventions, or adjacent CLI/runtime patterns.

Spec: changes requested for three validated blockers.

1. **P1 — Preserve the consumed-code outcome when local persistence fails.** In `apps/cli/src/core/computer/connect.ts:65-74`, the Server exchange succeeds before `writeComputerIdentityAtomically` and `storeBoundAccountComputer`. Either write can then throw into the generic result-less handler at `apps/cli/src/commands/computer/connect.ts:85-95`. At that point the code has already been consumed, the Agent may be bound, and the machine token rotated, but the command omits `connected: true` and the explicit do-not-reuse guidance. Identity-first ordering can also leave the local identity/credential pair inconsistent. Please distinguish post-exchange persistence failures from pre-exchange failures, return a non-retryable consumed-code result with safe recovery guidance, avoid or recover the partially advanced local pair, and test failure of each write.

2. **P2 — Keep fallback Provider rows independently verifiable.** `failedClosedProviderComponent` in `apps/cli/src/core/computer/preparation.ts:293-311` supplies only `nextAction`, although `apps/cli/README.md:29-31` promises individual repair/verify actions. Both the unexpected ensure catch (`:345-350`) and projection fallback (`:378-383`) use these rows. Add `verifyAction: { command: providerInspectCommand(provider) }` and cover both fallback paths.

3. **P2 — Do not mark transient standalone inspection as never retryable.** `apps/cli/src/commands/computer/runtime-inspect.ts:19-24` assigns `retryability: "never"` to every non-ready result. However, `temporarily_unavailable`, deadline, and unexpected probe failures explicitly tell the user to wait and run the idempotent verifier again (`apps/cli/src/core/computer/runtime-probe.ts:176-185,230-260`). Derive standalone retryability from the diagnostic (for example, `backoff` for transient/time-bounded failures and `never` for install/auth/version/configuration repair). Keep the targeted `connect` envelope itself at `never` because its code is one-time.

Validation on this head: `pnpm build`, `pnpm check`, `git diff --check`, 48 focused CLI tests, 20 focused shared-contract tests, and 37 focused Server service tests passed. Required CI is green except the expected ownership gate and the still-running patch-coverage job at review time.

liuchao-001 (commented):
Reviewed `a203c2dd…e0327a6d` in a dedicated worktree at that exact head.

The compatibility design is the strongest part of this PR. Smuggling the marker into `clientVersion` build metadata is the only option that keeps *both* strict parsers intact — a new optional request field would be rejected by an older Server's `.strict()`, and an unconditional `runtimeProvider` would be rejected by an older Client's. `parseSemVer` splits `build` into identifiers, so `build.includes(...)` is an exact-element match rather than a substring match, and `isSupportedClientVersion` ignores build metadata, so the marker cannot move the release floor. The integration test that freezes a hand-written legacy parser and asserts it *rejects* the marked response (`packages/server/src/__tests__/integration/computers.test.ts`) proves the claim in both directions against the real route rather than against an `omit()` of the evolving schema. I also confirmed `readComputerEntry` normalizes to exactly four fields, so `runtimeProvider` never reaches `computer-credentials.json` — the "persistence unchanged" claim holds.

Failing closed on `runtime:unconfirmed` rather than defaulting to a provider is the right call and matches the Context Tree constraint that a verdict admits only the subject it was made about.

I independently confirm @Gandy2025's three findings; they are real and I have nothing to add to them. Below are findings I did not see covered there.

---

### 1. The fallback projection is the one that escapes its own invariant check (should fix)

`failedLocalComputerPreparation` (`apps/cli/src/core/computer/preparation.ts:364-385`) hand-builds the result: `requiredCount: 4` is a literal and `readyCount: computer.status === "ready" ? 1 : 0` is a hand-rolled special case, and the object is returned **without** `LocalComputerPreparationResultSchema.parse`. Its sibling `runLocalComputerPreparation:352-360` derives both counts from `components` and parses.

`assessLocalPreparation` is the mechanism the whole design leans on to keep `status`/`localReady`/`readyCount` from contradicting the rows beneath them — and the only path that skips it is the one taken *after something already went wrong*. It is not wrong today (a `ready` Computer component is always unblocked with two ready children, so `1` is correct), but nothing enforces that. Add a fifth component, or make `computerComponent` able to return `ready` alongside a blocking child, and this path silently emits a document whose `localReady` its own `components` do not support. The CLI's own test re-parses that fallback document through the schema, which is a good instinct — the production path should do the same.

Suggestion: extract one `projectPreparation(components)` that derives the counts and parses, and call it from both places.

### 2. `blocking` means two different things in two files (low)

`packages/shared/src/computer.ts:321` documents the field as "A blocking Check in any non-ready status (including `skipped`) keeps its Component non-ready" — i.e. status-sensitive. But `concealedGateCheckIds:370-374` flags *any* child with `blocking: true` regardless of status, so a `ready` + `blocking` child makes its parent permanently un-`ready`. Meanwhile `formatPreparationCheckLine` (`apps/cli/src/core/computer/formatting.ts:34`) guards with `check.status !== "ready" && check.blocking`, i.e. it assumes `ready` + `blocking` is a representable row worth rendering without the marker.

Pick one meaning. If `blocking` means "currently blocking", the Check schema should reject `status: "ready"` with `blocking: true` the same way `readyComponentIssue` already does for Components, and the formatter's status guard becomes dead. If it means "would block if not ready", `concealedGateCheckIds` needs the status test. Right now the doc, the validator, and the renderer each describe a different rule.

### 3. `preparationBlockerCodes` is exported but never called in production (low)

`preparation.ts:402` is reachable only from `apps/cli/src/__tests__/computer-preparation.test.ts:209`. Related: `formatPreparationCheckLine` renders id/status/version/message but never `diagnosticCode`, so codes like `install_incomplete` and `DAEMON_SERVICE_FAILED` exist only in `--json`. A one-line "Blocked by: …" in the human summary is exactly what this function computes; otherwise it should be dropped rather than kept alive by a test.

### 4. The parity claim is stronger than the code (low, doc accuracy)

`runtime-probe.ts:18-22` says reusing the canonical factories means "a connect-time verdict and the daemon's readiness cannot disagree." Two deliberate differences survive: the daemon `mkdir`s both Runtime homes before `realpath` (`packages/client/src/runtime/client-runtime-composition.ts:455-458`) while `canonicalizeProviderHome` intentionally does not, and the daemon composes the factories with a shared `createLoginShellDiscovery()` while the probe passes no `discovery` at all and re-derives login-shell PATH per run. "Runs the same probe against the same resolved environment" is accurate; "cannot disagree" is not.

### 5. Nits

- `preparation.ts:242` calls `warningsFromProvider(result)` twice to build one optional key.
- `preparation.ts:281` re-specifies `verifyAction`, which `base:254` already carries.

---

### Observations, not change requests

- The marked version is persisted into `computers.client_version` by both `#createComputer:304` and `#repairComputer:346`, so between exchange and the daemon's first registration frame (`computer-service.ts:125` overwrites it) that column holds `0.0.2+opentag-connect-runtime-v1` rather than a release version. Nothing renders the column today, so this is harmless — but it is the one place the capability marker leaks past the wire, and it is worth knowing before any surface starts displaying it.
- On a targeted connect, `--no-start` and `--no-prepare-provider-clis` now both guarantee a non-zero exit, because a skipped required Check stays blocking. That is stated in the README and I think it is the right default; flagging it because it is a silent contract change for any script that passes those flags and checks the status. The server-authored bootstrap command passes neither, and the `issue #239` container scenario connects untargeted, so neither in-repo caller breaks.

### Verification I ran on this head

`pnpm test` for `@opentag/shared` (203 passed) and `@opentag/server` (725 passed) both green. CLI: 366/367 — the single failure is `src/__tests__/auto-update-defaults.test.ts > refreshes the daemon service through the stable shim after an injected install` (expected `'installed'`, got `'installing'`), which passes on re-run in isolation, is untouched by this PR, and is green in CI. I am recording it as a local timing flake in the full `--maxWorkers=1` run, not as a finding against this PR, but it looks worth a separate look.

Not approving while @Gandy2025's blockers are open.

yuezengwu:
Addressed the review feedback in `c5640f62` after merging current main `761fc120` (merge `a4d8659f`; the only conflict was the export snapshot, retaining both sides).

@Gandy2025 — all three requested fixes are implemented:

1. **Consumed-code persistence failure:** both local writes now remain inside the post-exchange result boundary. Failures return `COMPUTER_LOCAL_PERSISTENCE_FAILED`, exit 3, `retryability: never`, `connected: true` explicitly meaning the committed Server exchange, `codeConsumed: true`, and `localPersistenceReady: false`. No daemon mutation or preparation runs on this path. Credentials are written first; identity is reconstructable from them. The emitted, idempotent `computer repair-local --installation-id … --home …` rewrites the saved credential durably and reconstructs identity without another exchange. It rejects missing/unreadable/wrong-installation credentials, including the old binding, and directs the operator to fix storage and obtain a **new** code if the new credential was never saved. No rollback to potentially revoked tokens or two-file atomicity is claimed. Tests cover each write failing before/after committing, with fresh/existing homes, repeated recovery, daemon exclusion, exact-Home guidance, and secret redaction.
2. **Provider fallback verify:** both thrown-ensure and projection-fallback rows now include the exact read-only `provider-cli inspect` action; both paths are tested.
3. **Standalone Runtime retryability:** `temporarily_unavailable` and `runtime_probe_failed` (including timeout/unexpected probe failures) report `backoff`; install/auth/version/configuration failures remain manual/`never`. Targeted connect remains `never`. The diagnostic matrix is tested.

@liuchao-001 — the additional findings are also addressed:

- Both normal and fallback results use one `projectPreparation(components)` to derive required/ready counts and parse the shared invariant. Tests exercise invalid-row fallback, optional rows, and blank daemon errors.
- `blocking` now means currently blocking: ready Checks and Components are rejected when blocking; the renderer no longer hides that flag based on status.
- `preparationBlockerCodes` now supplies a `Blocked by: …` guidance line in both human and JSON output, with parity assertions.
- Probe documentation now describes point-in-time local evidence and explicitly records the home-creation/discovery-cache differences from the daemon.
- Removed duplicate warning projection and redundant verify action.

Validation on the final source state: frozen install, `pnpm check`, build, typecheck, and full `pnpm test` passed (311 script tests; Shared 213, Client 950, Server 738, CLI 411, Web 584). Runtime coverage: 357 tests, all four metrics 100%. PostgreSQL integration: 18 files / 314 tests passed. `git diff --check` passed. No unrelated timing-flake changes, migrations, Runtime installation, or Server/Web readiness claims were added.

Fresh-head verification is now complete: [CI](https://github.com/first-tree-ai/opentag/actions/runs/33724730037) passed and [patch coverage](https://github.com/first-tree-ai/opentag/actions/runs/33724730085) is 773/773 changed executable lines (100%; threshold 80%). Gandy2025 and liuchao-001 both approved this exact head. The PR body has been refreshed to the same head, test counts, and CI evidence.

The new non-blocking explicit-skip summary suggestion is acknowledged and deferred as the reviews permit: skipped rows still show `skipped (blocking)` and their actions; no new code head is introduced just for that summary enhancement. Consolidating the existing identity version literal is also deferred. Required bestony approval remains outstanding (26 gated files). No review has been dismissed by the author, and this PR has not been merged.

liuchao-001 (approved):
Re-reviewed at head `c5640f62` in a fresh worktree. Confirming the merge first, since this head merged current main rather than rebasing: the net diff against `761fc120` is still confined to the same 27 files, and the one conflicted file (`packages/client/src/__tests__/public-exports.snapshot.json`) kept both sides — I diffed main's export list against this head's and nothing from main was dropped, including `runContextTreeCli` from #455. No unrelated work rode in on the merge.

All five of my findings are addressed, and I checked each rather than taking the summary:

- `projectPreparation` now derives `requiredCount`/`readyCount` from the rows and parses, and both `runLocalComputerPreparation` and `failedLocalComputerPreparation` go through it. The invariant test is better than what I asked for: feeding a `ready` + `blocking` Runtime row makes the normal projection throw and degrade into the fail-closed projection with `runtime_probe_failed`, rather than crashing the command. Filtering on `component.required` instead of array length is also the right generalization.
- `blocking` now means one thing. The Check schema rejects `ready` + `blocking` the same way the Component schema does, `readyComponentIssue` drops its now-redundant branch, and the renderer's status guard is gone. Doc, validator, and renderer agree.
- `Blocked by: …` is emitted from `preparationBlockerCodes` into both projections with parity assertions.
- The probe comment now states what it actually guarantees, and names both divergences (home creation, discovery cache) explicitly.
- Duplicate `warningsFromProvider` call and redundant `verifyAction` both gone — and the `verifyAction` removal is correct, since the `manual` branch already spreads it from `base`.

I also read @Gandy2025's #1 fix closely, because it is the substantial one, and it holds up. `allocateComputerIdentity` mints a fresh UUID and writes nothing, so the identity-stage failure is a real state and not a redundant rewrite; ordering credentials first is right because identity is fully derivable from the saved credential, and `repair-local` reconstructs it from `credential.installationId`, which is the same value the exchange used as `installationId`. The refusal when `credential.installationId` does not match is the load-bearing part — it stops a repair from reviving an older binding whose credential a repair-mode rotation may already have revoked. The credentials-uncommitted case honestly routes to a new code instead of pretending recovery is possible. I confirmed no daemon mutation on this path: the early return precedes both `restart()` and `installAndStart()`, and the `preflight()` that ran earlier is a read-only `systemctl --user show-environment` probe. The test that pins both sides of the rename boundary for each of the two writes, across fresh and existing homes, is the right shape.

On #3, the current `AgentRuntimeProbeIssue` union is closed and the switch handles four codes explicitly, so the `default:` arm is reachable only by `temporarily_unavailable` — which `runtimeInspectionError` classifies as `backoff`. Coverage is complete for today's union, not just for the two named codes.

---

### One new non-blocking nit

The `Blocked by: …` line goes silent exactly when the blockers are the explicit skips. `providerComponentFromEnsureResult`'s `skipped` branch and `daemonChildCheck`'s `noStart` branch both omit `diagnosticCode`, and `preparationBlockerCodes` only collects rows that have one. I ran the combination through `projectPreparation` directly rather than inferring it:

```
--no-start --no-prepare-provider-clis
status:   needs_attention
codes:    []
guidance: ["Connection preserved. Do not reuse the one-time connect code.",
           "Server/Web confirmation still requires fresh daemon observations."]
```

Three blocking rows, no `Blocked by:` line. The rows themselves still say `skipped (blocking)`, so nothing is hidden — but the summary line added for exactly this purpose is absent in the one case where the answer is simplest. Giving the skip branches a code (`daemon_skipped` / `provider_skipped`), or falling back to blocking row ids when no codes exist, would close it. Fine to leave for a follow-up.

### Two small notes

- The PR body's **Validated head** still reads `e0327a6d` with the old counts, while the head is `c5640f62`. Your comment carries the current numbers, but the body is the artifact the merge checklist points at, and it currently contradicts its own "Do not transfer review evidence to a different head" rule.
- `repair-local` hardcodes `version: 2` for the identity record, which is now the third place that literal lives (`allocateComputerIdentity`, `validateIdentity`, here). Harmless today since `validateIdentity` normalizes to 2 anyway.

### Verification I ran on this head

`@opentag/shared` 213, CLI 411, `@opentag/server` 738 — all green, matching your reported counts exactly. The `auto-update-defaults.test.ts` failure I saw last round did not recur in the full `--maxWorkers=1` run, which confirms it was a local timing flake and not something to chase.

CI on this head: green except `ownership-gate` (26 files, needs bestony). `Checks`, `Enforce Patch Coverage`, `Generate Quality Scoreboard`, and `Node 24 Compatibility` were still running when I looked — I am not asserting anything about those, and patch coverage in particular still has to land on its own.

Approving on the code. The non-author CODEOWNER approval and green required CI remain separate gates I am not satisfying here.

Gandy2025 (approved):
Re-reviewed immutable range `761fc12042a34b43451b42c1091f19fbb9dbeda2...c5640f62d4d9252b673029cf3f16bd73e3126749`.

Risk remains high because this crosses one-time credential exchange, local persistence, daemon startup, and Runtime/Provider readiness boundaries. The current correction holds up:

- The post-exchange persistence path now preserves the committed Server outcome, reports `connected: true` / `codeConsumed: true` with non-retryable recovery guidance, writes the non-reconstructable credential first, and exits before daemon or preparation work. `repair-local` validates the exact installation, reconstructs identity without another exchange, and fails closed when the new credential was not saved. Tests cover both writes on both sides of the durable-rename boundary, fresh and existing homes, idempotent recovery, no daemon mutation, and secret redaction.
- Both Provider fallback paths now retain independent read-only verify actions.
- Standalone Runtime inspection uses `backoff` for transient/unexpected probe failures and `never` for manual install/auth/version/configuration repair, while the one-time connect envelope remains `never`.
- Normal and fail-closed preparation now share `projectPreparation`, derive counts from required rows, and pass through the shared schema. The Check/Component `blocking` invariant is consistent across schema and formatting.

Standards: no blocking findings. The final 27-file PR diff remains scoped, respects package boundaries and public exports, and the merge conflict resolution preserves both main's Context Tree exports and this PR's preparation exports.

Spec: no remaining blockers or adjacent correctness findings.

Validation on this head: `pnpm build`, `pnpm check`, `pnpm typecheck`, `git diff --check`, 57 focused CLI tests, 20 focused shared-contract tests, and the relevant public-export tests passed. The two daemon command tests also pass 15/15 with an isolated `OPENTAG_HOME`; their initial local failure came from this machine's incompatible default-Home credential fixture, not this diff.

Non-blocking merge checklist: refresh the PR body's `Validated head` and test counts from `e0327a6d` to this head before merge, so the written evidence does not claim transfer across heads. Explicit skip rows also remain readable but currently produce no `Blocked by:` summary code; that can be tightened separately.

Approving the code at `c5640f62`. Fresh-head required CI remains a separate gate.

PR #518, opened 2026-09-03
feat(docs): update readme

ggordonhall:


liuchao-001:
Heads-up on one line this touches, not a review of the rewrite.

This drops the **"It works with the tools already on that machine"** bullets from both READMEs — including *"a tool that is installed, signed in, and visible to the provider runtime on that Computer … OpenTag does not discover or expose tools"* — while leaving the hero `` alt text, which ends "connecting the rest of your stack" / "连接你的其余工具", unchanged. That sentence was added in #511 specifically to bound this claim, so after this merges the wider version is the only statement of tool reach left on the page, and for screen-reader users it is the only one they ever get.

#530 narrows that clause against current `main` (two lines, one per README). It touches the `` line your first hunk carries as context, so whichever of us merges second gets a small conflict there — happy to rebase mine onto yours if this lands first, or you can take the clause into this rewrite and I will close mine.

PR #521, opened 2026-09-03
feat(web): refine agent setup computer preparation

Gandy2025:
## Summary
- redesign Agent Setup Step 2 around a stable Computer summary and unified command surface
- present the selected Runtime and aggregate Messaging support as one row each while retaining provider-specific checks underneath
- keep initial, repair, expired, and retry command states in the same fixed-size component; remove repeated readiness from Step 3

## Validation
- `pnpm check`
- `pnpm typecheck`
- `pnpm build`
- focused web tests: 164 passing
- full web suite: 654 passing
- setup/creation Journey E2E: 2 passing
- manual desktop and 390px mobile UI walkthrough; no horizontal overflow

## Test environment notes
- full repository tests passed for all affected web code; two unrelated CLI daemon tests read a malformed local OpenTag credential unless `OPENTAG_HOME` is isolated, and both pass in an isolated home
- the full Journey suite reaches the unrelated Agent navigation test, where latest main can exhaust its 90s timeout waiting for the Skills nav item; the two Agent Setup/creation journeys pass independently

PR #524, opened 2026-09-03
fix(release): include Context Tree runtime in portable artifacts

yuezengwu:
## Summary

Fix the portable packaging failure in [npm Publish run 33740499085](https://github.com/first-tree-ai/opentag/actions/runs/33740499085/job/100601025424), which rejects the CLI's declared `@first-tree-ai/context-tree@0.1.8` dependency under the old no-node_modules contract.

- Materialize the frozen production dependency closure into a physical `app/node_modules` tree, preserving the package manifest, CLI, skills, templates, and licenses. Record exact package names/versions in `app/dependency-closure.json`; never fetch dependencies or run lifecycle scripts during assembly.
- Share strict graph/asset verification between the builder and artifact verifier. Reject missing or unreferenced packages, inconsistent pins/records, unsupported dependency shapes, escaping source resolution, nested node_modules, symlinks, and native addons.
- Add staging and production portable-app checks to CLI Pack Smoke. Run the relocated Context Tree CLI and OpenTag's real Context Tree resolution path in a clean temporary environment, without operator credentials or configuration writes.
- Synchronize the English/Chinese portable release documentation and add focused regressions.

## Testing

Head: `99f7c2fc13a9a4820272d3a2110c55f8bae49826`, based on main `21ac2485ffb72673812557292d3fafee94922ca9` at the start of validation.

During validation, main advanced to `217697e19a42052d4a8ef7f4b14b2594669b6385` with separate Web onboarding changes (#521). There is no changed-file overlap with this packaging fix. This PR retains the exact locally validated head; PR CI must verify its merge result against main.

Repository validation after updating to main:

- [x] `pnpm check`
- [x] `pnpm build`
- [x] `pnpm typecheck`
- [x] `pnpm test` — 330 script tests; Shared 218, Server 773, Client 950, CLI 431, Web 651.
- [x] `pnpm --filter @opentag/server test:integration` — 314 tests.
- [x] `pnpm --filter @opentag/client test:agent-runtime:coverage` — 357 tests; statements, branches, functions, and lines all 100%.

All six gates ran serially with `CI=1 TURBO_FORCE=true`. The source fingerprint was unchanged throughout validation.

Packaging acceptance already performed on the unchanged packaging/CLI source before the main update:

- 39 focused tests passed, including the actual frozen dependency closure.
- Staging and production portable-app smoke passed.
- Real four-platform staging artifacts built with official embedded Node v24.19.0 and passed structural/dependency verification.
- macOS ARM64 and a read-only, network-none Linux ARM64 Docker consumer passed actual OpenTag/Context Tree invocation outside the checkout.
- Repeated four-platform builds produced identical tarball hashes and manifest bytes.
- Five corrupted payloads were rejected: missing Context Tree CLI, missing skill, incorrect package version, external symlink, and native addon.

The main update changes only unrelated Server/Web code; packaging/CLI inputs are unchanged. The local artifacts use synthetic release coordinates and are not published onboarding evidence. x64 artifacts received structural verification only; x64 runtime execution was not available locally.

## Breaking changes

None. Existing CLI behavior, dependency pin, lockfile, installer, publication workflow, and release safety gates are preserved. This PR does not publish a release or update a channel pointer. W3 onboarding still requires the normally published staging installer after merge and successful main CI/release.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -280,6 +280,12 @@ jobs:
           --version "${{ steps.version.outputs.version }}"
           --binary "${{ steps.version.outputs.binary_name }}"
 
+      - name: Smoke staging portable app
+        run: >-
+          node scripts/portable/smoke-portable-app.mjs
+          --channel staging
+          --version "${{ steps.version.outputs.version }}"
+
       - name: Restore and prepare production identity
         run: |
           git checkout -- apps/cli/package.json apps/cli/src/build-info.ts
@@ -302,6 +308,12 @@ jobs:
             --name open-tag \
             --version 0.0.2 \
             --binary opentag
+
+      - name: Smoke production portable app
+        run: >-
+          node scripts/portable/smoke-portable-app.mjs
+          --channel prod
+          --version 0.0.2
 
   container-smoke:
     name: Container Smoke
--- a/docs/portable-release.md
+++ b/docs/portable-release.md
@@ -61,12 +61,15 @@ Each payload contains:
 VERSION                 Plain-text version
 INSTALL.json            Release metadata, platform, install mode, and app entry
 node/bin/node           Embedded Node.js runtime, checked against the official SHASUMS256.txt
-app/                    Bundled CLI, its package manifest, LICENSE, README, THIRD_PARTY_NOTICES
+app/                    Bundled CLI, its package manifest, LICENSE, README, THIRD_PARTY_NOTICES,
+                        dependency-closure.json, and the embedded node_modules/ dependency tree
 bin/<binName>           Artifact-local shim, used before the payload is activated
 ~~~
 
-The bundled CLI has no `node_modules`: `apps/cli` declares no runtime dependencies, and the build fails closed if that
-ever changes rather than shipping an artifact that only breaks when a user runs it.
+The bundled CLI's Context Tree integration resolves the installed `@first-tree-ai/context-tree` package at runtime —
+its `package.json` through `createRequire`, its bundled CLI, and its packaged skills and templates — so every payload
+embeds that package and its full production dependency closure as a physical `app/node_modules` tree.
+`app/dependency-closure.json` records the exact direct pins and every embedded package name and version.
 
 ## Automatic upgrades
 
@@ -145,8 +148,31 @@ node scripts/portable/verify-portable-artifact.mjs \
   --tarball .portable-release/staging/0.0.2-staging.1.1/open-tag-staging-0.0.2-staging.1.1-darwin-arm64.tar.gz
 ~~~
 
-The verifier checks the published checksum, the extracted layout, and the metadata, and additionally runs the artifact
-when it targets the host platform.
+The verifier checks the published checksum, the extracted layout, the metadata, and the embedded dependency graph
+(exact direct pins, the closure record, packaged assets, and the no-symlink/no-native layout). When the artifact
+targets the host platform it additionally executes it: the OpenTag CLI through its shim, and the embedded Context
+Tree CLI (`--version`, `--help`, `list`) against an isolated home. It also runs OpenTag's
+`context-tree connect` against a nonexistent managed name, verifying real runtime resolution without connecting a
+tree or persisting configuration.
+
+The portable builder ships `@first-tree-ai/context-tree` and every production dependency the frozen install resolves
+for it as a deterministic flat `app/node_modules` tree: one real directory per package, no symlinks, no nested
+`node_modules`, and no native addons. Packages are copied from the installed store only — nothing is fetched and no
+lifecycle script runs; Context Tree's own `postinstall` ships inside its published files but stays inert because
+OpenTag invokes the packaged CLI directly. The app manifest keeps only the truthful direct pins, and both the builder
+(before its identity smoke) and the artifact verifier (for every platform) run the same graph checks.
+
+**Fail-closed dependency shapes.** `apps/cli` may declare only `@first-tree-ai/context-tree`, pinned to an exact
+`x.y.z` version. Unknown or non-exact direct pins, non-empty optional or peer dependencies, `os`/`cpu`-constrained
+packages, lifecycle scripts on embedded packages other than Context Tree's own, symlinks inside published files,
+native addons, and same-name packages resolving to different installed roots all fail the build. These dependency
+shapes need explicit packaging support before they can ship.
+
+**CI gate.** Every `CLI Pack Smoke` run for staging and production assembles the real portable app from the freshly
+built release-channel CLI, verifies its embedded graph, and runs both the relocated Context Tree CLI and OpenTag's
+runtime resolution probe with a temporary `HOME` and a minimal environment (no `NODE_PATH`, `NODE_OPTIONS`, or host
+credentials; `scripts/portable/smoke-portable-app.mjs`). The gate needs no downloaded Node.js runtime,
+so a dependency regression fails ordinary CI rather than the release.
 
 The embedded Node.js version is pinned in `scripts/portable/node-version.txt` and must be an exact `vX.Y.Z`. Its tarball
 is checked against the official `SHASUMS256.txt` for that release before it becomes part of an artifact.
--- a/docs/zh-CN/portable-release.md
+++ b/docs/zh-CN/portable-release.md
@@ -1,7 +1,7 @@
 # OpenTag Portable 发布指南
 
 > Canonical source: [../portable-release.md](../portable-release.md)
-> Last synced with: 2026-09-02
+> Last synced with: 2026-09-03
 
 Portable release 是一份自包含的 OpenTag 安装包：每个平台一个 tarball，同时携带 bundle 后的 CLI **和它自己的
 Node.js runtime**，因此没有 Node.js、没有 npm、没有任何 package manager 的机器也能安装并运行 OpenTag。它发布到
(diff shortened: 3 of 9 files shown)


liuchao-001 (approved):
Reviewed at head `99f7c2fc13a9a4820272d3a2110c55f8bae49826` in a fresh worktree. **Approving** — the packaging contract change is correct, fails closed in the right places, and the runtime probe actually proves what it claims. No blocking findings.

## What I verified independently

- **The probe is a real end-to-end check, not a tautology.** `runContextTreeRuntimeProbe(probeOpenTag: true)` asserts the `No managed Context Tree named "..." exists` message, and that message is only reachable in `apps/cli/src/core/context-tree/connect.ts` after `resolveContextTreeAssets()` resolved the embedded package *and* `runContextTreeCli(assets, ["list"])` spawned the relocated CLI and returned a parseable payload. It also cannot write config: `writeContextTreeConfig` runs only on the success path, so the isolated-`HOME` leftovers assertion holds by construction rather than by luck.
- **Resolution shape matches the consumer.** `packages/client/src/runtime/context-tree.ts` needs exactly `package.json` (via `createRequire`), `dist/cli/index.mjs`, and `skills/`; `assertContextTreeAssets` covers all three plus templates. The CLI is always spawned as `node <cliPath>`, so the flat `app/node_modules` layout and the 0644 normalization are sufficient.
- **The new tests pass and are not hollow:** 19/19 in ~1.4 s here, including the real frozen-install closure test that assembles, verifies, and runs the embedded CLI.
- **Real closure on this checkout:** 38 packages, 13 MB unpacked, ~1.8 MB gzipped added to each of the four tarballs.
- **Redistribution hygiene:** every one of the 38 embedded packages ships its own `LICENSE*` file, so attribution travels with the payload.
- **Full-tree extraction downstream:** both `scripts/portable/install.sh` and `apps/cli/src/core/update/portable-installer.ts` extract the whole tarball with no allowlist, so `app/node_modules` survives first install and self-update. `cpSync(..., {recursive:true})` per platform copies it into every artifact, and the template has no symlinks, so the copy is faithful.
- Stale-contract sweep is clean: no remaining reference to `assertBundledCliHasNoRuntimeDependencies` or to the "ships without node_modules" claim outside the one accurate comment in `runtime-dependencies.mjs`.

## Non-blocking notes

1. **The closure is manifest-derived, not load-derived.** `@first-tree-ai/context-tree` builds with `tsdown --deps.alwaysBundle '/.*/'`, so its four declared deps and their transitive tree are never loaded. I checked empirically: with *only* `@first-tree-ai/context-tree` present and no transitive packages, `--version`, `list`, `verify`, `read`, and `install` all run normally. So 37 of 38 packages — `zod` alone is 8.2 MB unpacked, `yaml` 1.2 MB, plus the whole `micromark*` tree — are dead weight in every artifact (~+1.8 MB gz per platform).

   I am **not** asking to trim it: a future context-tree version could add a lazy `import()`, and a manifest-derived closure is the rule that stays verifiable without trusting someone else's bundler config. Worth recording the trade-off explicitly, though — right now the doc reads as if the closure is what the runtime needs, when it is what the manifest declares.

2. **`app/THIRD_PARTY_NOTICES` no longer describes everything the payload redistributes.** It is generated from the *bundled* package list (`scripts/third-party-notices.mjs`), and `pnpm check` still passes because that list did not change — but the artifact now physically ships 37 more third-party packages. Per-package `LICENSE` files are present, which is the usual npm-style answer, so this is a question rather than a defect: should the notices contract say that the portable payload also redistributes the embedded closure under its own in-tree licenses?

3. **`chmodSync(destination, 0o644)` also strips executable bits.** Correct today (nothing is exec'd directly, and pnpm store files are often read-only, so normalizing is what makes the build deterministic), but the graph verifier would not flag a future embedded package that genuinely needs an executable bin — it would just fail at runtime. A one-line comment saying the mode normalization assumes "everything is loaded through `node`, nothing is exec'd" would pin the assumption where the next reader hits it.

4. **The artifact verifier asserts an exact user-facing sentence.** Rewording the `No managed Context Tree named "..." exists` error becomes a release-verifier failure. It is fail-closed and CLI Pack Smoke catches it on the PR that reworded it rather than at release time, so the blast radius is fine — just flagging the coupling for whoever next edits that string.

## CI

Green on the head I reviewed, including the two new `CLI Pack Smoke` portable steps. `ownership-gate` still wants bestony on 7 files; that gate is independent of this review.

bestony (changes requested):
Reviewed at exact head `99f7c2fc13a9a4820272d3a2110c55f8bae49826`.

## Standards

No documented-standard violations found. The diff follows `AGENTS.md` and `CONTRIBUTING.md`: code and canonical docs are English, the Chinese mirror is synchronized, failures are explicit, scope is focused, and the commit/branch conventions are correct.

Non-blocking judgment call: `scripts/portable/smoke-portable-app.mjs:24` duplicates the argument-parser shape already present in `scripts/portable/verify-portable-artifact.mjs:41`. A shared helper would reduce validation drift, but this does not block the fix.

## Spec

Blocking: the new portable smoke can report success without running either required Context Tree path.

`verifyPortableDependencyGraph()` intentionally accepts an app with no runtime dependencies (`scripts/portable/runtime-dependencies.mjs:432-439`). `runContextTreeRuntimeProbe()` then silently returns when `dependency-closure.json` is absent or when the direct Context Tree pin is absent (`scripts/portable/runtime-dependencies.mjs:545-549`). Both the staging/production smoke and artifact verifier use that sequence.

I reproduced the combined behavior with a dependency-free app manifest: graph verification returned `{ packageCount: 0 }`, and `runContextTreeRuntimeProbe({ probeOpenTag: true })` returned successfully without spawning the relocated Context Tree CLI or OpenTag's real resolution path.

That means removing the direct `@first-tree-ai/context-tree` pin together with `node_modules` and the closure record passes this new gate, contrary to requirements 1 and 8 and allowing the original packaging failure class to recur. Please make the Context-Tree-specific smoke/verifier call sites require the exact direct pin and closure (or make the probe fail when it is required), and add a regression covering the wholly missing dependency.

Summary: Standards has 0 hard findings and 1 non-blocking smell; Spec has 1 blocking finding, the silent no-op in the required runtime probe.

liuchao-001:
Following up on @bestony's Spec finding, since it lands on a path my own approval did not cover. I reproduced it, and I think the mechanism is real but the impact statement needs one correction.

**Confirmed — the mechanism is exactly as described.** Against a hand-built app whose manifest declares no dependencies:

~~~
verify: {"packageCount":0}
probe returned: undefined
~~~

`verifyPortableDependencyGraph` takes the dependency-free branch (`runtime-dependencies.mjs:432-439`) and `runContextTreeRuntimeProbe` returns at the `dependency-closure.json` check (`runtime-dependencies.mjs:546`) without spawning anything. My approval verified that the probe is a genuine end-to-end check *when the pin is present*; I did not test its absence, and on that path it is a silent no-op.

Worth adding to the finding: `smoke-portable-app.mjs:57-59` then prints `0 dependency package(s) embedded; Context Tree runs from the relocated app` — a gate that can print a false success sentence is a bad primitive regardless of whether anything reaches it today.

**Correction to the impact.** "Removing the direct pin together with `node_modules` and the closure record passes this new gate ... allowing the original packaging failure class to recur" overstates what CI would let through. Two guards fire on that exact edit:

1. `scripts/__tests__/portable-runtime-dependencies.test.mjs:179` — added *in this PR* — asserts the checked-in `apps/cli/package.json` yields exactly `[{ "@first-tree-ai/context-tree", "0.1.8" }]`. Drop the pin and `readPortableDirectDependencyPins` returns `[]`, so the `assert.deepEqual` fails and the `node --test scripts/__tests__/*.test.mjs` leg goes red. A source-level regression for the wholly missing dependency therefore already exists; what is missing is one at the smoke/probe level.
2. `scripts/cli-pack-smoke.mjs:147` fails the prod channel when the packed CLI does not carry `@first-tree-ai/context-tree/scripts/postinstall.mjs`.

And with the pin present, nothing is silent: a missing `node_modules`, a missing closure record, or record/manifest drift all fail closed in `verifyPortableDependencyGraph`. The silent path requires the pin's absence, which is precisely the case the two guards above already catch.

**Where that leaves me.** I still read this as defense-in-depth rather than a shippable-defect blocker — but I agree it should be fixed before merge, because a gate whose success message does not depend on anything having run is worth about as much as no gate, and the fix is small: have the two OpenTag-specific call sites (`smokePortableApp`, `verifyHostRuntime`) require the Context Tree pin and a non-zero closure instead of inferring it, e.g. a `requireContextTree: true` option on the probe that turns both early returns into failures, plus a regression asserting the dependency-free app is rejected *there* while `verifyPortableDependencyGraph` keeps accepting the historic shape.

I am not re-reviewing to block — @bestony's review already holds the merge and I have no separate objection. My approval stands for the packaging change itself; it was scoped to the dependency-present path and does not extend to this one.

bestony:
Follow-up correction to my review: the mechanism is confirmed, but the impact statement in the original review was too broad. Removing the checked-in direct pin is already caught by `scripts/__tests__/portable-runtime-dependencies.test.mjs:179`, and the existing CLI pack smoke also checks the packed Context Tree payload.

The remaining finding is the runtime-gate contract itself: `smokePortableApp` and `verifyHostRuntime` can print a successful result while `runContextTreeRuntimeProbe` returns without spawning anything when the app is dependency-free. This is still worth fixing before merge because the required probe's success does not currently imply that the required Context Tree and OpenTag paths ran. A `requireContextTree` option for those two call sites plus a regression for the no-dependency app would close it.

I am not adding a separate nested-`node_modules` finding: the source-side skip is intentional for pnpm install droppings that are replaced by the flat closure, while the shipped-artifact verifier rejects nested directories.

PR #525, opened 2026-09-03
feat(web): refine agent setup readiness lab

Gandy2025:
## Summary

- restore the compact two-step readiness presentation and remove unnecessary separators around the computer summary
- reorganize Mock Control around core Computer, Runtime, and Messaging combinations, with clearer copy and reliable customization/reset behavior
- cover the state model, accessibility contract, responsive layout, and real Mock Control interactions with unit and Playwright tests

## Review

- independent code review completed: PASS
- review findings around recoverable pending state, customization baselines, accessible naming, focus behavior, and reconnect state were fixed before this PR

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `OPENTAG_HOME=<isolated-temp-home> pnpm test`
- `pnpm --filter @opentag/server exec vitest run src/__tests__/integration --maxWorkers=1` (18 files, 314 tests)
- `OPENTAG_POSTGRES_PORT=55438 OPENTAG_E2E_ADMIN_DATABASE_URL=postgresql://opentag:opentag@127.0.0.1:55438/postgres pnpm --filter @opentag/e2e e2e:journey` (17 passed)


bestony (changes requested):
## Standards

- **P2 - Restore the unrelated E2E coverage removed by this PR.** `CONTRIBUTING.md` requires the smallest coherent change and excludes unrelated refactors. In `e2e/tests/app.spec.ts:149-155`, this setup/readiness PR removes Skills and Integrations from the “every Agent-owned destination” journey. At `e2e/tests/app.spec.ts:217-267`, it also replaces the populated Usage two-column assertion with an empty-state assertion although neither the navigation surfaces nor Usage implementation/fixtures change here. These edits weaken regression coverage outside the stated scope; restore them or move a justified change to a separate PR.

- **P3 - Make the scenario copy mapping exhaustive.** `agent-setup-lab-controls.tsx:34-75` repeats the same `LabScenario` dispatch for labels and descriptions, while both functions silently fall through to the “everything ready” copy. Adding a scenario already requires edits in the scenario list, preset map, and both dispatches. A typed `Record<LabScenario, ...>` would prevent an omitted branch from displaying the wrong state.

## Spec

- **P1 - Do not let Mock Control mutate an Agent's Runtime after creation.** The product contract fixes Runtime Provider at Agent creation, but the Runtime override is rendered for every non-fixture scenario (`agent-setup-lab-controls.tsx:522-556`) and `changeRuntime` rebuilds the current Agent with the new Provider (`agent-setup-lab-page.tsx:220-225`). The added test at `agent-setup-lab-page.test.tsx:204-223` explicitly creates with Claude Code, then changes the already-created Agent to Codex. This makes the lab's “reliable customization” model an impossible production state. Restrict this override to direct-setup presets, as its own hint says.

- **P1 - Give the rebind checkpoint distinct old and target Computer identities.** `agent-setup-lab-model.ts:230-247` gives both the cross-account “Previous Mac” and the owned “Review Mac” the same `LAB_COMPUTER_ID`. Binding the selected owned Computer therefore only clears `requiresComputerRebind`; it does not exercise a placement move to another Computer. The current test (`agent-setup-lab-model.test.ts:47-51`) checks only the initial state and misses this. Use different IDs and assert the post-rebind ID.

- **P2 - Exercise the promised responsive readiness surface in a real browser.** The new Playwright case leaves Runtime-missing Step 2 at desktop size, switches to an everything-ready state, and only then sets the 390px viewport (`e2e/tests/app.spec.ts:57-83`). It never checks the compact two-row readiness surface or failure copy under mobile CSS, so declaration-level unit tests cannot detect clipping or overlap there.

- **P3 - The E2E removals above are also outside all three PR summary bullets.** They should not be coupled to the readiness-lab change.

Summary: Standards 2 findings (worst P2: unrelated regression coverage removed); Spec 4 findings (worst P1: the lab permits impossible Runtime mutation and does not actually model Computer replacement).

Gandy2025:
Addressed every requested change in 23b7fc3e:

- Restored Skills/Integrations navigation coverage and the populated Usage two-column assertion.
- Replaced fallback scenario-copy dispatches with an exhaustive typed Record.
- Restricted Runtime override to direct-setup presets; full creation journeys preserve the Runtime selected at creation.
- Modeled rebind with distinct previous/target Computer IDs and asserted the post-rebind target.
- Added a real 390×844 readiness failure check covering both compact rows, clear failure copy, viewport containment, overflow, and accessibility.

Validation after the fixes:
- pnpm check
- pnpm typecheck
- pnpm --filter @opentag/web test (62 files / 665 tests)
- journey E2E (17/17)
- pre-commit and pre-push hooks

Please re-review when convenient.

bestony (changes requested):
## Standards

- **P2 - Restore the original end-to-end paths instead of replacing them with mocks.** The destination and populated Usage assertions are back, but `e2e/tests/app.spec.ts:163-165` now intercepts `navigation-visibility`, and `e2e/tests/app.spec.ts:281-286` replaces the real Usage response with a visual fixture. The base test exercised both through the real journey backend, and this PR does not change either implementation or its fixtures. These mocks therefore still weaken unrelated integration coverage and remain outside the readiness-lab scope, contrary to `CONTRIBUTING.md`'s smallest-coherent-change rule. Restore the base behavior exactly, or move a separately justified E2E change to another PR.

- **P3 - Make the newly added readiness-fixture copy mapping exhaustive too.** `apps/web/src/onboarding-v2/readiness-lab-fixtures.ts:31-56` now maintains `ReadinessScenario` labels in both a typed record and a separate conditional dispatch, but the dispatch silently falls through to `readiness-warning`. Adding a fixture can therefore compile while showing the wrong localized label. Use one exhaustive `Record<ReadinessScenario, ...>` for the localized functions as well, matching the fix already made for `LabScenario`.

## Spec

- **P2 - Close Mock Control before asserting the mobile readiness surface.** In `e2e/tests/app.spec.ts:60-79`, the test opens Mock Control, resizes to 390x844, selects the failure scenario, and then checks the readiness rows without ever clicking `closeLab`. The fixed panel spans nearly the full mobile viewport (`agent-setup-lab-controls.tsx:508-525`) and sits above the page at `z-30`, so the surface under test is occluded. `expectWithinViewport` only checks visibility and bounding boxes; it does not detect another element covering those boxes. This test can therefore pass while the compact rows are unusable and still does not exercise the promised clipping/overlap behavior. Close the panel before the row assertions (and reopen it afterward for the remaining control interactions).

Summary: Standards 2 findings (worst P2: unrelated integration paths remain mocked); Spec 1 finding (P2: the mobile surface is asserted while occluded).

yuezengwu (approved):
Approved for exact head `23b7fc3e3e86e1ba6392c6e8d27b3990d0a05933`.

The successor addresses the prior blockers: unrelated Skills/Integrations and populated Usage E2E coverage is restored; scenario copy is exhaustive; Runtime override is limited to direct-setup presets; rebind uses distinct previous/target Computer IDs and verifies the target; and the 390×844 browser path now covers the compact readiness failure surface, containment, overflow, and accessibility.

Exact-head validation passed: full diff check; `pnpm check`; uncached build (6/6); uncached typecheck (8/8); Web tests (62 files / 665 tests); Journey E2E (17/17 on isolated DB/ports); and GitHub CI including Browser Smoke, Checks, and Enforce Patch Coverage.

This approval applies only to the exact head above and does not transfer to any successor SHA. It does not dismiss or rewrite the predecessor-head CHANGES_REQUESTED review; merge remains external.

Gandy2025:
Addressed the second review in 89ae120f:

- Restored the two pre-existing E2E paths exactly: no navigation-visibility interception and no Usage response fixture. The PR diff now only adds the setup-lab browser scenario to app.spec.ts.
- Replaced the readiness-fixture conditional label dispatch with an exhaustive Record<ReadinessScenario, () => string>.
- The mobile test now closes Mock Control before checking the compact readiness rows, then reopens it for flow actions.

Validation:
- pnpm check: pass
- pnpm typecheck: pass
- focused Web test: 15/15 pass
- focused real-browser setup-lab E2E at 390×844: pass
- pre-commit and pre-push hooks: pass

I also ran the complete Journey suite after restoring the base tests exactly. The new setup-lab case passes, but the unchanged baseline navigation case times out on Skills because the local dev E2E server returns 404 for the staging-only navigation-visibility endpoint and therefore hides Skills/Integrations. I left that unrelated baseline code untouched rather than reintroducing a mock or widening this PR.

bestony (approved):
## Standards

PASS. The pre-existing navigation and populated Usage E2E paths are restored exactly, with no route or response mocks remaining in the PR diff. The new readiness-fixture label dispatch is exhaustive through `Record<ReadinessScenario, () => string>`. No documented-standard violations or actionable baseline smells found.

## Spec

PASS. The mobile browser case now closes Mock Control before asserting the 390x844 compact failure rows, viewport containment, and overflow, then reopens it for the remaining real-control interactions. The earlier Runtime immutability, distinct Computer rebind, state-model, and reset/customization findings remain correctly fixed.

Approved for exact head `89ae120f45a397a42cf1092ef358b72e5947dca5`.

Independent validation: full build; repository typecheck (8/8 tasks); Web tests (62 files, 665 tests); and the focused real-browser setup-lab journey at 390x844. The known local Skills navigation timeout is in the unchanged base journey and was not masked or altered by this PR.

Summary: Standards 0 findings; Spec 0 findings.

yuezengwu (approved):
Approved for exact head `89ae120f45a397a42cf1092ef358b72e5947dca5`.

The successor closes the remaining predecessor-head findings: the original navigation and populated Usage E2E paths are restored without mocks, localized readiness-scenario copy is exhaustive, and Mock Control is closed before the 390×844 readiness-surface assertions.

Exact-head validation passed: full diff check; `pnpm check`; uncached build (6/6); uncached typecheck (8/8); Web tests (62 files / 665 tests); focused 390×844 Journey E2E (1/1); and all GitHub checks. The full Journey run had one Skills-navigation timeout with 12 passing and 4 serial-dependent skips; an equivalent creation → settings → navigation probe fails identically on base `b92175fb5e02bdf491022184148e03da3347be0e` because the local `navigation-visibility` endpoint returns 404, confirming this is baseline behavior rather than a PR regression.

This approval applies only to the exact head above and does not transfer to any successor SHA. Merge remains external.

PR #528, opened 2026-09-04
fix(onboarding): prepare reused computers before messaging

yuezengwu:
## Summary

- keep the same three-step Agent setup experience while restoring four explicit preparation rows for Computer, the selected Runtime, Lark CLI, and Slack CLI
- let foreground `connect` own first-connection repair while existing-Computer placement, reactivation, rebind, and **Check again** start a Server-owned idempotent Runtime/CLI preparation
- add a fenced preparation result protocol so stale heartbeats or older results cannot overwrite the active checking operation
- make the Web refresh action start the real operation before fetching the latest setup snapshot

The selected Runtime is inspected for the operator-provided CLI, capabilities, and credentials. OpenTag does not install Runtime CLIs. Server-owned preparation may install or repair the supported IM CLIs.

Formal W3 acceptance remains release-gated: after this reaches staging, rerun the published staging installer and capture the continuous Web recording before considering the onboarding release accepted.

## Testing

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test` (3376 tests passed across the root and workspace suites)
- `pnpm --filter @opentag/client test:agent-runtime:coverage` (358 tests; 100% statements, branches, functions, and lines)
- `pnpm --filter @opentag/server test:integration` (314 tests)
- `git diff --check`
- local browser validation at `/internal/agent-setup`

## UI validation

- Desktop evidence: local Agent Setup Lab verified the mixed state `Computer ready / Codex ready / Lark CLI needs attention / Slack CLI ready`; no screenshot or recording attached
- Mobile evidence: not separately captured; this change reuses the existing responsive `ReadinessList` and adds no CSS
- Widths checked (`320px` / `390px` / `768px` / `1440px`; explain any N/A): default desktop viewport only; explicit breakpoint capture is deferred to the release-gated W3 Web recording
- States verified: waiting for Computer, Runtime checking/ready/attention, each IM CLI checking/install/ready/attention, stale-result fencing, and refresh failure
- Keyboard/accessibility checks: automated tests verify the ordered readiness list, stable four-row component identities, labels, and status semantics; no separate manual keyboard pass
- New reusable block, theme value, or CSS exception: none; restores the existing `ReadinessList`

## Breaking changes

None. The prewarm request additions are optional for rolling compatibility; the new result frame and HTTP refresh route are additive.

## Checklist


bestony (changes requested):
## Changes requested

> Reviewed head **2ca42b3daa1e4fb5d14d594c55233d74ae2fd663**

I found two blocking regressions in the placement/preparation flow:

- **`packages/server/src/runtime/provider-cli-reconcile-owner.ts:233-248`**: `beginPreparation()` fences the Computer as `checking` before sending a `provider-cli:prewarm` frame containing the new `mode` and `runtimeProvider` fields. The capability negotiation is still the existing `runtime.providerCliPrewarm` v1 (`packages/shared/src/runtime-protocol.ts:41-55`), and the older client schema is strict and rejects those unknown fields. During a rolling deployment, an older daemon therefore drops the frame and never emits `provider-cli:prewarm:result`; the server keeps the preparation request fence forever, so setup remains stuck in checking until reconnect. Gate `ensure`/the result protocol behind a newly negotiated capability (or retain a compatible legacy path and a bounded failure fallback) before setting the fence.

- **`packages/server/src/runtime/provider-cli-reconcile-owner.ts:560-575`**: `onAgentPlacementChanged()` awaits `prepareComputer()` before calling `onActiveBindingChanged()`. `prepareComputer()` throws whenever prewarm is not eligible or cannot start (`:220-225`, `:535-547`). The production eligibility predicate is explicitly false for an active Agent that already has a current messaging binding (`packages/server/src/services/computers/computer-service.ts:44-61`). Consequently, rebind/reactivate of an Agent with an existing Feishu/Slack binding exits before dispatching its active integration requirement, leaving CLI credential reconciliation stale. Make preparation non-fatal to placement notification and always run `onActiveBindingChanged()` (while preserving preparation errors for diagnostics), or otherwise condition the call on eligibility.

Focused tests passed for the changed shared/client/server suites (175 tests); TypeScript checks and `git diff --check` also pass. The two scenarios above are not covered by the added tests, and the current `ownership-gate` check is still failing because 21 files await approval.

bestony (changes requested):
## Changes requested

> Reviewed head **e25de611485034c6836304f24a04a71b7efef639**

The previous two findings are addressed, but one blocking state-consistency issue remains:

- **`packages/server/src/runtime/connection-registry.ts:171-186,201-215,421-438`**: a preparation fence sets `providerReadinessObservedAt`/`imCliReadinessObservedAt` once, then deliberately ignores all readiness values from heartbeats while the fence is active. Those sources expire after `RUNTIME_CLIENT_CAPABILITY_TTL_MS` (60 seconds), while the client can spend up to 150 seconds retrying the provider lock (`packages/client/src/runtime/provider-cli/reconciler.ts:354-361`) and the new server fallback remains armed for five minutes (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:915-921`). A legitimate long preparation therefore projects as `waiting` before it finishes. More importantly, the fallback calls `completePreparation()` and clears the fence (`:934-950`); an older client that dropped the extended ensure frame then continues sending its pre-operation `ready` heartbeat, and `touch()` accepts it once the fence is gone, immediately replacing the fallback's `unavailable` result with stale ready state. Keep an explicit preparation/quarantine marker until a matching terminal result or fresh compatible observation is received, and/or refresh the active-operation TTL without accepting heartbeat readiness. Add a regression test that sends a stale ready heartbeat after the fallback and verifies it cannot restore ready.

The focused provider protocol/reconciler tests (41 tests) and the server TypeScript check pass, but this stale-heartbeat-after-fallback case is not covered.

bestony (changes requested):
## Changes requested

> Reviewed head **ad9d371887ff0908fea6ea2b243baf3bca9eac7c**

The TTL/quarantine fix adds the requested stale-ready regression, but its release condition is still too permissive:

- **`packages/server/src/runtime/connection-registry.ts:527-557,565-589`** treats any preparation-related heartbeat whose status is merely not `ready` as a fresh compatible observation, clears `preparationQuarantine`, and then accepts all fields from that same heartbeat. A strict legacy Client that dropped the extended ensure frame can continue reporting its pre-operation `checking` state (for example, Runtime `ready` plus one CLI `checking`, or the inverse). That stale `checking` satisfies `status !== "ready"`, disarms the quarantine, and allows a subsequent stale all-`ready` heartbeat to restore readiness without the ensure ever running. `checking` is not terminal evidence that preparation completed; it is exactly the in-flight state this fence is meant to protect. Keep the quarantine until a matching prewarm result, or only release it for a clearly terminal, generation-bound observation; add a regression where the legacy heartbeat remains `checking` and then sends stale `ready`.

The focused connection-registry/provider-reconcile tests pass on this head, including the current stale-ready case, but they do not cover stale `checking` followed by stale `ready`.

bestony (approved):
Reviewed head fbc47dc6a5d3bbb390378be0d5573133b4c5dde5. The preparation fence now remains sealed across both in-flight and fallback states; heartbeats refresh liveness/TTL only and cannot overwrite readiness or release the fence. The added regression covers stale checking followed by stale ready, and the focused plus full server unit suites pass. No further blocking findings.

PR #529, opened 2026-09-04
feat(slack): expand agent interaction capabilities

yuezengwu:
## Summary

- Expand the fixed Slack bot capability contract from 7 to 18 scopes so one Agent can discover its Team, users, channels, DMs, and existing MPIMs; join task-relevant public channels; open one-to-one DMs; react to messages; and upload files.
- Project the inbound human Slack user ID into the runtime provider reference and give the Agent bounded native-CLI guidance for pagination, targeted history and thread reads, cross-conversation work, rate limits, message lifecycle operations, scheduled-message cancellation, reactions, one-to-one DMs, task-relevant public-channel joins, and Slack's external file-upload flow.
- Make the public-channel recovery contract explicit: do not roam or bulk-join; on `not_in_channel`, confirm the task-relevant target is public, join once, and retry the original action once. Private, MPIM, unknown, or `channel_not_found` targets require a human invite or clarification instead of guessing.
- Explain the first Slack event required to finish connection observation and add synchronized English/Chinese live-acceptance checklists for the single-Team/single-Agent path, including post-join persistence and both receive modes.

Non-goals: Agent View, multi-Agent routing, slash commands, security redesign, message-retention changes, and creating new MPIMs.

## Testing

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm --filter @opentag/client test:agent-runtime:coverage` (358 tests; 100% statements, branches, functions, and lines)
- `pnpm --filter @opentag/server test:integration` (18 files, 314 tests)
- Exact-head live local acceptance at `7f262e4c` against the dedicated `Local-team` workspace and `OpenTag Local QA` app: 18/18 installed scopes, active installation and binding, credential generation advanced on reauthorization, and identity observation closed.
- Verified public/private mentions, ambient channel delivery, DM and thread replies, Team/user/channel discovery, pagination, history/replies, send/update/delete, scheduled-message cancellation, reactions, external file upload, proactive one-to-one DM, and expected failure for an unjoined private channel.
- Verified `im:read` / `mpim:read` discovery calls complete without `missing_scope` (the isolated test workspace currently has no returned DM/MPIM conversations for the bot token); no MPIM creation is promised.
- Verified the real public-channel recovery path on a newly created channel: the first send returned `not_in_channel`; the Agent confirmed the named target was public, joined exactly once, retried exactly once, and sent successfully without joining unrelated channels.
- Verified post-join behavior under both modes: `all_message` persisted and delivered an unmentioned message as `ambient`; `mention_only` persisted an unmentioned message without creating a delivery, then accepted an explicit mention as `direct` and replied. The installation was restored to `all_message` afterward.
- Verified inbound persistence for one Slack message across `created -> edited -> deleted`, with every revision accepted for delivery.
- The dedicated Slack Team/App, installation, and QA channels are intentionally retained as personal test assets; no credentials or live installation IDs are included in this PR.

## UI validation

- Desktop evidence: Local OpenTag setup reached `opentag is ready`; Slack desktop/web interaction completed the live acceptance above.
- Mobile evidence: N/A; this change adds copy to the existing responsive handoff layout and introduces no layout or style changes.
- Widths checked (`320px` / `390px` / `768px` / `1440px`; explain any N/A): `1440px` live; smaller widths N/A because the new content uses the existing hint paragraph without structural or CSS changes.
- States verified: Slack `waiting-handoff` without CLI progress (automated), ready state (live), public/private/DM message flows (live), reauthorization-required to connected (live), and both channel receive modes (live).
- Keyboard/accessibility checks: no new control; the existing `role=status` remains intact and the instruction is ordinary semantic text.
- New reusable block, theme value, or CSS exception: None.

## Breaking changes

Existing Slack installations that do not have all 18 required bot scopes will require reauthorization before OpenTag reports them healthy.

## Checklist


liuchao-001 (changes requested):
Reviewed at head `0ec5e145` in a fresh worktree off `origin/main`.

The mechanical half of this PR is clean: the `runtimeProviderMessageRef` extraction is a faithful move, the `authorUserId` projection correctly excludes bot and system authors (and a bot ID mislabelled `human`), it is genuinely wired through to the Agent, and the scope lists are now drift-guarded against all four docs plus the manifest. My concerns are about the capability contract, not the code that implements it.

## Decide before merge

**1. `channels:join` plus the self-join instruction silently widens what OpenTag ingests and retains.**

`provider-outbox-instructions.ts:38` now tells the Agent "Join a public channel with `conversations.join` `{channel}`", and `channels:join` is in the fixed contract (`packages/shared/src/im-binding.ts:86`). Downstream:

- inbound admission is per-binding with no channel allowlist (`packages/server/src/services/im/im-message-inbox.ts:445-476`);
- `agents.receiveMode` defaults to `all_message` (`packages/server/src/db/schema/agents.ts:24`), and agent creation never sets it explicitly (`agent-service.ts:391-403`), so that default is what new Agents actually run with.

So an Agent can now, on its own initiative, add the bot to any public channel in the Workspace — and from that moment every message in that channel is persisted as an inbound row and, in the default receive mode, delivered as an `ambient` Turn that materializes Sessions and Tasks. Before this PR the only way into a channel was a human invite, which was itself the consent act.

The PR lists "message-retention changes" as a non-goal, but this is a retention change with the Agent as the actor. `docs/slack-live-acceptance.md` covers only that the join succeeds ("A public-channel `conversations.join` from the Agent CLI may add the bot to a public channel"); nothing checks what happens to that channel's traffic afterwards.

If unprompted self-join is not intended, the guidance should say so ("join a public channel only when the user's task asks you to"), and the acceptance checklist should gain a step that verifies post-join ingest under both `mention_only` and `all_message`.

**2. The `not_in_channel` recovery contradicts the scope this PR just paid for.**

The same bullet at `provider-outbox-instructions.ts:38` says "Join a public channel with `conversations.join`" and then "If Slack returns `not_in_channel`, `channel_not_found`, or a membership error, ask the user to invite the bot instead of retrying."

For a public channel, `not_in_channel` is precisely the case `channels:join` was added to solve. As written, an Agent that hits it while posting to a public channel will ask a human to invite it, and `channels:join` goes unused in its most common trigger. The recovery needs to split by conversation kind.

These two readings can't both be intended — which is the reason to settle #1 first and then make this sentence follow from it.

## Should fix while the contract is still cheap to change

**3. The guidance promises DM/MPIM discovery the sixteen scopes don't grant.**

`provider-outbox-instructions.ts:34-35` tells the Agent it may operate on a "public channel, private channel, DM, or MPIM" by ID and lists `conversations.info` / `conversations.list` for discovery. The contract has `channels:read` and `groups:read` but not `im:read` or `mpim:read`, so `conversations.info` on a DM or MPIM ID and `conversations.list` with `types=im,mpim` both fail `missing_scope`. Line 38's "Open a DM with `conversations.open` `{users}`" likewise works for a single user (`im:write`) but fails for a multi-person DM, which needs `mpim:write`.

This matters more than an ordinary doc nit because every change to `SLACK_REQUIRED_BOT_SCOPES` flips existing installs to `reauthorization_required` (`im-binding-provider-cli.ts:158`) — as this PR's own breaking-change note says. Adding `im:read` / `mpim:read` later costs a second forced reauthorization for everyone. Either add them now (and decide on `mpim:write`), or narrow the guidance to channels and state that a DM or MPIM is reachable only by the ID that arrived on an inbound event.

## Nits

**4.** `packages/server/src/__tests__/im-delivery-worker.test.ts:835-838` — `toMatchObject({ provider: "slack", authorUserId: "human" })`. `"human"` is the base fixture's `authorExternalId`, but it reads like `authorKind` leaked into the ID field. A fixture value such as `U_FIXTURE` would make the assertion self-evident.

**5.** Same test, the following line — `not.toMatchObject({ authorUserId: botUserId })` is redundant after the exact assertion above it.

## What I verified

- Read the full diff. The `runtimeProviderMessageRef` move is byte-for-byte faithful apart from the new author projection; `RuntimeProviderMessageRef` is still used in `im-delivery-worker.ts`, so no dead import.
- `authorUserId` really reaches the Agent: `agent-turn-runner.ts:357` serializes `providerRef` into the prompt target, and history items at `:379` carry per-message author IDs. The history byte budget counts the serialized `providerRef` (`im-delivery-worker.ts:1302-1319`), so the extra field trims how many history items fit rather than overflowing a bound.
- The new 4 KiB guard has room: the guidance is 2,921 bytes across 11 entries.
- Scope lists agree in all five places the new drift guard checks, and no stale "seven" / "七项" wording is left anywhere in the repo.
- `pnpm --filter @opentag/shared test` — 22 files, 219 tests, pass.
- `pnpm --filter @opentag/client test` — 70 files, 955 tests, pass (after building `@opentag/shared` first).
- Context Tree `opentag/system/im-integrations.md` and `messaging-and-routing.md` at `origin/master`: no constraint this PR contradicts.

One process note: the live-acceptance evidence in the description is author-run and isn't bound to head `0ec5e145`. The checklist itself is genuinely good — the reauthorization and expected-error sections especially — so it would be worth having an independent pass re-run it against the exact head once the questions above are settled.

CI on this head is green apart from `ownership-gate`, which is the approval gate (needs bestony), not a code failure.

yuezengwu:
Addressed the requested changes on exact head `7f262e4c`.

The review did not conflict with the product decisions discussed here: cross-conversation access and post-join persistence/delivery are intentional, while public-channel self-join is limited to a concrete task target rather than arbitrary roaming.

Changes made:

- Bounded cross-conversation guidance to task-relevant targets and explicitly prohibited arbitrary discovery-driven roaming or bulk joining.
- Split membership recovery by conversation kind. For a task-relevant public channel, the Agent may join once and retry the original action once after `not_in_channel`; private, MPIM, unknown, and `channel_not_found` targets require an invite or clarification.
- Added `im:read` and `mpim:read`, taking the fixed contract and public manifest to 18 scopes.
- Narrowed `conversations.open` guidance to exactly one user ID / a 1:1 DM. Existing MPIMs may be used when already reachable; this PR does not promise MPIM creation and does not add `mpim:write`.
- Added synchronized EN/ZH acceptance coverage for post-join behavior in both `mention_only` and `all_message` modes.
- Renamed the human test fixture to `U_FIXTURE` and removed the redundant negative assertion.

Verification:

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- Agent Runtime coverage: 358 tests, 100% statements/branches/functions/lines
- Server integration: 18 files, 314 tests

Exact-head live acceptance against the retained personal `Local-team` / `OpenTag Local QA` assets also passed:

- Reauthorization completed with 18/18 scopes; installation and binding are active, credential generation advanced, and identity observation is closed.
- `im:read` / `mpim:read` discovery calls complete without `missing_scope`.
- On a newly created public channel, the first send returned `not_in_channel`; the Agent confirmed the named target was public, joined exactly once, retried exactly once, and sent successfully without joining unrelated channels.
- After that join, an unmentioned `all_message` event was persisted and delivered as `ambient` and received an Agent response.
- In `mention_only`, an unmentioned event was persisted without a delivery; an explicit mention was accepted as `direct` and received an Agent response. The test Agent was restored to `all_message` afterward.

Please re-review this head.

liuchao-001 (approved):
Re-reviewed at head `7f262e4c`. All five findings are resolved, and the two that were questions rather than defects got real answers rather than hedges.

**1 — self-join and ingest.** Answered directly: cross-conversation access and post-join ingest are intentional, and the trigger is now "relevant to the current task" with explicit anti-roaming limits rather than being left implicit. What makes this land is that the consequence is now stated where each audience will actually meet it — in the Agent's own guidance ("Joining enrolls future messages in normal OpenTag ingress: they are persisted, then `mention_only` or `all_message` controls delivery. Do not join merely to explore."), in `slack-app-setup.md` for the operator, and as a live acceptance step in both language mirrors. Worth naming plainly for the record: the trigger moved *wider* than the head I first reviewed — it was "when the user's task names another conversation", it is now "even if the user did not name it". That is a deliberate product decision, it is documented as one in the setup doc and the PR's own summary, and the guardrails that came with it are the right ones. My concern was that the behavior was unstated, not that it was wrong.

**2 — `not_in_channel` contradiction.** Gone. The recovery now splits by conversation kind: confirm the target is public and task-relevant, join once, retry once; private channels, MPIMs, `channel_not_found`, and unknown types go to a human invite. `conversations.info` on an unjoined public channel is reachable with `channels:read`, so the confirm step the guidance asks for is actually executable.

**3 — scopes vs. promised capability.** `im:read` and `mpim:read` added, taking the contract to 18. `mpim:write` was correctly *not* added; instead `conversations.open` is narrowed to exactly one user ID and MPIM creation is declared a non-goal, so the guidance and the scope set now describe the same capability. That is the better trade — a scope you don't need is a forced reauthorization you don't have to spend.

**4, 5 — nits.** `U_FIXTURE` and the redundant negative assertion, both done.

## Verified on this head

- 18 scopes, in sorted contract order, identical across `im-binding.ts`, the manifest (bumped to `minor_version: 3`), both `slack-app-setup.md` mirrors, and both `slack-live-acceptance.md` mirrors — the drift guard covers all five and passes.
- No stale "sixteen" / "十六" anywhere; every count reference reads eighteen / 十八. The only `sixteen` hits left in the repo are an unrelated `sixteen-character-code` fixture in `account-api.test.ts`.
- The EN and ZH mirrors are genuinely synchronized on the new prose, not just on the scope lists — the cross-channel-read, both-receive-modes, DM/MPIM, and expected-error sections all match.
- Guidance is now 3,411 bytes across 12 entries, still 685 bytes under the 4 KiB bound.
- `pnpm --filter @opentag/shared test` — 22 files, 219 tests, pass.
- `pnpm --filter @opentag/client test` — 70 files, 955 tests, pass.

## One note, not blocking

The acceptance checklist now says "Discover existing DMs and MPIMs with `conversations.list`; verify that both conversation types are visible when the bot is a member" — but the exact-head run couldn't satisfy that precondition, since the isolated workspace returned no DM or MPIM conversations for the bot token. Thanks for saying so plainly instead of calling it a pass; the honest read is that `im:read` / `mpim:read` are confirmed present (no `missing_scope`) but the positive visibility case is still unobserved. An empty result and a working result look the same from here. Worth picking up opportunistically the next time that workspace has a DM or MPIM the bot belongs to — no need to hold this PR for it.

Approving. `ownership-gate` still needs bestony, so this doesn't unblock the merge on its own.

PR #531, opened 2026-09-04
fix(e2e): gate the e2e stack on the real Postgres server

### LINKED ISSUE #502: Flaky CI: Browser Smoke bootstrap races Postgres's init-server handover and fails to create the e2e database
`Browser Smoke` failed on a docs-only commit whose entire diff is two Markdown date lines, on a head
whose parent had passed the same job. The failure is in the e2e stack bootstrap, before any test runs:

```
[WebServer] Error: Command failed: docker compose -f /home/runner/work/opentag/opentag/docker-compose.yml exec -T -e PGPASSWORD=opentag postgres psql -U opentag -d postgres -v ON_ERROR_STOP=1 -Atc create database "opentag_e2e"
[WebServer] psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  the database system is shutting down
Error: Process from config.webServer was not able to start. Exit code: 1
```

Run: https://github.com/first-tree-ai/opentag/actions/runs/33692118198/job/100452925224

## Root cause

`e2e/scripts/stack-server.mjs` starts Postgres and then gates on `waitForDatabase()`, which polls
`select 1` until one succeeds:

```js
await run("docker", ["compose", "-f", composeFile, "up", "-d", "postgres"]);
await waitForDatabase();
await resetDatabase();   // drop database … ; create database "opentag_e2e"
```

The `postgres` image runs a **temporary internal server** during initialization so it can apply
init scripts, then shuts that server down and starts the real one. The temporary server accepts
connections on the Unix socket, which is exactly what both `waitForDatabase()`'s probe and
`resetDatabase()`'s `docker compose exec … psql` use. So a single successful `select 1` does not
mean the database is ready — it can be the init server, and the very next statement lands in the
window where that server is shutting down.

That is what the `FATAL: the database system is shutting down` message reports: not a crash, but the
expected handover between the init server and the real one.

The compose `healthcheck` does not close this either, because `pg_isready` is satisfied by the
temporary server for the same reason.

## Suggested fix

Gate on a signal that distinguishes the two servers rather than on one successful query. Options,
roughly in order of directness:

- require N consecutive successful probes separated by the handover window, rather than the first one;
- probe for the real server's readiness explicitly (the init server is not reachable over TCP, only
  the socket), so a TCP probe from the host distinguishes them;
- retry `resetDatabase()` on the specific `57P03`/"shutting down" class instead of failing the whole
  web server start.

The first two remove the race; the third only narrows it, so it is the weakest of the three.

## Impact

This blocks unrelated pull requests: the job is required, so a PR with no product change at all goes
red and needs a re-run. It is a bootstrap race rather than a test assertion, so it can strike any PR.

liuchao-001:
Closes #502.

## Summary

`Browser Smoke` was going red at the e2e bootstrap, before any test ran, on changes that could not have
caused it — most recently on 2026-09-04, and at least five times in the two days after the issue was filed:

```
[WebServer] psql: error: ... FATAL:  the database system is shutting down
Error: Process from config.webServer was not able to start. Exit code: 1
```

The `postgres` image runs a temporary server during first-time initialization so it can apply the init
scripts, then shuts it down and starts the real one. That temporary server listens on the Unix socket
only. `stack-server.mjs` gated on a single `select 1` over that socket, so the gate could be satisfied by
the temporary server, and the `drop database` / `create database` that follow then landed in the handover.

Measured on `postgres:17` with initialization slowed to 8s so the window is observable, probing both
transports from inside the container:

```
t=4s   socket=1     tcp=connection refused      <- temporary init server
t=11s  socket=1     tcp=connection refused
t=12s  socket=FATAL tcp=1                       <- handover: this is the CI failure
t=13s  socket=1     tcp=1                       <- real server
```

So the fix is to gate on the transport only the real server answers: every `psql` in the bootstrap path
(`stack-server.mjs`) now connects over TCP inside the container, and `waitForDatabase()` cannot return
while initialization is still going. No retry and no sleep — the signal itself now distinguishes the two
servers. `e2e/tests/fixtures.ts` still uses the socket and is deliberately left alone here: it runs only
after Playwright's `webServer` gate, so it cannot race initialization. It is tracked in #533 with the
mechanical check that would keep the rule from being re-broken.

The same socket-only probe was written in three more places with the same blind spot, so they get the
same one-flag correction: the compose `healthcheck`, the Container Smoke wait loop in CI, and the Docker
doctor's `--health-cmd`.

## Testing

- Replayed the old and new gate against a `postgres:17` container with initialization slowed to 8s, asking
  after the gate returns which server accepted it (`show listen_addresses` is empty on the init server and
  `*` on the real one):
  - old gate: `(empty => INIT SERVER)` on both runs, and one run reproduced the handover failure;
  - new gate: `*` on both runs, with 71 and 97 consecutive `drop`/`create` cycles across the whole
    initialization window and no failure.
- `pg_isready` over the socket reports ready at t=1s (init server) and over TCP at t≈9s (real server),
  which is the signal the other three call sites relied on.
- `pnpm check`, `pnpm build`, `pnpm typecheck` pass. `pnpm test` reported two failures in
  `packages/client` (`claude-code-agent-runtime-exhaustive`, `pi-agent-runtime-exhaustive`); both are
  process-spawning tests hitting the 5s per-test budget on a machine at load average ~10, and both pass
  on a re-run with headroom. Nothing in this change is reachable from `packages/client`.
- `pnpm test:e2e:smoke` locally with the compose volume removed first, so the run went through the
  first-time initialization path this changes: 8 passed.

## Breaking changes

None. Local development is unaffected: a container with an existing data volume never runs the
initialization server, so the TCP probe succeeds immediately.

## Checklist

- [ ] Tests and documentation cover the changed behavior — the change *is* the e2e/CI readiness gate, so
      it has no unit-test surface; it is covered by the `pnpm test:e2e:smoke` run and the controlled
      old-vs-new replay under Testing.
- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass — see the note on two
      load-sensitive `packages/client` timeouts under Testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -353,7 +353,9 @@ jobs:
             --env POSTGRES_PASSWORD=opentag \
             postgres:17-alpine
           for _ in $(seq 1 30); do
-            if docker exec opentag-postgres-ci pg_isready --username opentag --dbname opentag; then
+            # Probe over TCP: the image's temporary initialization server listens on the Unix socket
+            # only, so a socket probe reports ready before the real server exists.
+            if docker exec opentag-postgres-ci pg_isready --host 127.0.0.1 --username opentag --dbname opentag; then
               exit 0
             fi
             sleep 1
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,9 @@ services:
     volumes:
       - opentag-postgres-data:/var/lib/postgresql/data
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U opentag -d opentag"]
+      # Probe over TCP: the image's temporary initialization server listens on the Unix socket only,
+      # so a socket probe reports ready before the real server exists.
+      test: ["CMD-SHELL", "pg_isready -h 127.0.0.1 -U opentag -d opentag"]
       interval: 5s
       timeout: 5s
       retries: 10
--- a/e2e/scripts/stack-server.mjs
+++ b/e2e/scripts/stack-server.mjs
@@ -36,6 +36,11 @@ function connectionTarget(url) {
   };
 }
 
+// The postgres image runs a temporary server during first-time initialization so it can apply the
+// init scripts, then shuts it down and starts the real one. That temporary server listens on the Unix
+// socket only, so every statement here goes over TCP inside the container, which only the real server
+// answers. A socket connection cannot tell the two apart and lands in the handover as
+// `FATAL: the database system is shutting down`.
 async function psql(url, sql) {
   const { database, password, username } = connectionTarget(url);
   const { stdout } = await execFileAsync(
@@ -50,6 +55,8 @@ async function psql(url, sql) {
       `PGPASSWORD=${password}`,
       "postgres",
       "psql",
+      "-h",
+      "127.0.0.1",
       "-U",
       username,
       "-d",
@@ -68,6 +75,7 @@ async function run(command, args, options = {}) {
   return execFileAsync(command, args, { cwd: repositoryRoot, ...options });
 }
 
+// Gates on the real server: `select 1` cannot succeed over TCP while initialization is still running.
 async function waitForDatabase() {
   const deadline = Date.now() + 60_000;
   let lastError;
--- a/scripts/e2e/doctor-docker.mjs
+++ b/scripts/e2e/doctor-docker.mjs
@@ -152,7 +152,9 @@ async function main() {
     "--env",
     "POSTGRES_PASSWORD=opentag",
     "--health-cmd",
-    "pg_isready -U opentag -d opentag",
+    // Probe over TCP: the image's temporary initialization server listens on the Unix socket only,
+    // so a socket probe reports ready before the real server exists.
+    "pg_isready -h 127.0.0.1 -U opentag -d opentag",
     "--health-interval",
     "1s",
     "--health-timeout",


bestony (approved):
## Standards

No findings. The change is focused, comments are in English, and the same TCP-readiness rule is applied consistently across the heterogeneous Compose, Actions, e2e, and Docker doctor surfaces. Extracting those declarations into a shared abstraction would add coupling without reducing meaningful risk.

## Spec

The Issue #502 requirements are implemented correctly: the e2e stack and Compose healthcheck now probe the real server over TCP, so the temporary Unix-socket-only initialization server cannot satisfy readiness. One non-blocking scope note: Container Smoke and Docker doctor were not explicitly named in the issue, but the PR body identifies them as the same blind spot, and the preventive sweep is coherent with the fix.

## QA

PASS, focused-local, exact head 94e613e63410aaf339a7f48fa5c0aca8bb5e326c.

- pnpm check passed.
- pnpm test:e2e:smoke passed from a fresh task-specific PostgreSQL volume: build succeeded and 8/8 browser smoke tests passed.
- GitHub CI, Browser Smoke, Container Smoke, node compatibility, patch coverage, and image build checks are green.
- Case disposition: no-change. Existing Browser Smoke covers the affected bootstrap path; no separate reusable QA case is needed for this review.

Summary: Standards 0 findings; Spec 1 non-blocking scope note. No blocking issues found.

liuchao-001:
## Independent QA + code review — `PASS`

Posted by **code-reviewer** (First Tree QA/review agent), running through @liuchao-001's `gh`
credentials. @liuchao-001 authored this PR, so this comment is **not** an approval and does not
satisfy the ownership gate — that still needs @bestony or @yuezengwu.

| | |
|---|---|
| **Status** | `PASS` |
| **Tier** | `focused-local` (per `packages/qa`-style tiering) |
| **Bound head** | `94e613e63410aaf339a7f48fa5c0aca8bb5e326c` |
| **Scope** | the readiness criterion at the four changed call sites, and the e2e stack's first-time-initialization bootstrap path |
| **Environment** | macOS 15 / Docker Desktop 29.1.5, `postgres:17` + `postgres:17-alpine`, Node 22.23.2, Playwright 1.62.1 |

**Evidence boundary.** Everything below was measured on my machine, not on GitHub
`ubuntu-latest`. This run can show the criterion is transport-discriminating, reproduce the
reported failure on the base commit, and show the same harness clean at this head. It **cannot**
measure the CI flake rate, and a green `Browser Smoke` on this head is not by itself evidence —
that was the whole problem.

---

### 1. The diagnosis holds, and not just empirically

The image's entrypoint disables TCP on the initialization server explicitly, and appends the flag
last so it wins:

```
# /usr/local/bin/docker-entrypoint.sh:290-302 (postgres:17)
docker_temp_server_start() {
  ...
  # does not listen on external TCP/IP and waits until start finishes
  set -- "$@" -c listen_addresses='' -p "${PGPORT:-5432}"
```

Per-tick from inside the container, initialization widened to 8s:

| | `postgres:17` | `postgres:17-alpine` |
|---|---|---|
| socket accepts, `show listen_addresses` = `` (init server) | t=1.4s … 8.6s | t=1.4s … 8.4s |
| TCP over that same window | `connection refused` throughout | `connection refused` throughout |
| both accept, `show listen_addresses` = `*` | t=9.0s | t=9.4s |

The alpine run also caught the handover directly: at t=9.4s the **socket** probe returned no
response while TCP had already flipped to the real server.

### 2. Your question 1 — "TCP accepts ⇒ ready" — checks out in both directions

- **No false ready:** TCP is refused for the entire initialization window, and the first tick where
  TCP answers is already the real server (`listen_addresses = '*'`).
- **No false not-ready that matters:** if the real server ever opened the port before it accepted
  queries, every consumer still treats that as not-ready — `psql` gets
  `FATAL: the database system is starting up` and `waitForDatabase()` retries, `pg_isready` returns
  non-zero (`PQPING_REJECT`) and both wait loops keep looping. There is no state where the new
  criterion reports ready on a server that cannot serve.

### 3. The old criterion is satisfied by the wrong server — deterministically

I ran the gate functions **extracted verbatim** from each revision of
`e2e/scripts/stack-server.mjs` (the only functional delta between the two harnesses is
`"-h", "127.0.0.1"`; harness diff is in the artifacts), against a compose stack that differs from
the repo's only by a 2s init script and an isolated port/project name:

- **base gate: 15/15** iterations returned with `show listen_addresses` = `` → satisfied by the
  temporary initialization server *every time*.
- **head gate: 15/15** returned `*` → the real server every time.

### 4. Issue #502 reproduces on base, and does not on head

Same harness, `resetDatabase()` looped immediately after the gate returns so the loop spans the
handover:

- **base: 6/6 runs failed** — 7 failures across 177 `drop`/`create` cycles, always on
  `drop database if exists "opentag_e2e" with (force)` or `create database "opentag_e2e"`: the exact
  statements and the exact position from the issue. Four faces of one ~150 ms window:
  - `FATAL:  the database system is starting up`
  - `FATAL:  terminating connection due to administrator command` → `connection to server was lost`
  - `connection to server on socket "…/.s.PGSQL.5432" failed: No such file or directory`
  - and CI's own `FATAL:  the database system is shutting down`
- **head: 6/6 runs clean** — 0 failures across 169 cycles, gate answered by `*` in every run.

Worth stating precisely: the string I reproduce is usually *not* CI's literal one. These are
adjacent instants of the same handover (connect during shutdown vs. after the socket file is
unlinked vs. before the real server accepts), so it is the same defect, not a lookalike.

### 5. Your question 2 — the three sibling `-h 127.0.0.1` sites: no side effects found

- **Auth / `pg_hba.conf`.** `initdb` writes `host all all 127.0.0.1/32 trust` on **both**
  `postgres:17` and `postgres:17-alpine`, so loopback TCP needs no credential;
  `POSTGRES_HOST_AUTH_METHOD` only governs the appended `host all all all` rule. The `PGPASSWORD`
  the e2e `psql` already passes stays harmless, and `pg_isready` never needed credentials.
- **CI Container Smoke on the alpine image.** `pg_isready --host 127.0.0.1` is present and behaves
  identically to the Debian image (table in §1); the long-form `--host` is valid.
- **`doctor-docker`'s health gate.** This is the one that needed checking, because
  `waitForContainerHealth()` returns on `running` as well as `healthy` — if `.State.Health` were
  absent the gate would be a no-op. It is not: for a container that declares a health command, the
  exact format string that function uses reports `starting`. Measured with an 8s init:

  | `--health-cmd` | container reported | real server exists at |
  |---|---|---|
  | `pg_isready -U … -d …` (before) | **healthy at t=1.8s** | t=9.2s |
  | `pg_isready -h 127.0.0.1 …` (after) | `starting` until t=9.4s, healthy at t=10.2s | t=9.4s |

  So before this change `doctor-docker` started the real Server against a database that did not
  exist yet. Same story for the compose `healthcheck`. Both are genuinely fixed, not cosmetically.

### 6. Your question 3 — the real product path, at this head, through first-time init

`pnpm test:e2e:smoke` in a worktree checked out at `94e613e6`, with
`docker compose down -v` **before every run** so each one takes the first-time-initialization path
this change is about:

```
run 1: 8 passed (11.1s)
run 2: 8 passed (9.5s)
run 3: 8 passed (9.2s)
```

`git status --porcelain` clean after `pnpm build`. All repo CI is green at this head; the only red
check is `ownership-gate` ("4 files awaiting approval").

### 7. Completeness of the sweep

The four corrected sites are the only Postgres-container readiness probes in the repo.
`packages/server`'s integration tests use embedded PGlite rather than a container;
`e2e/global-teardown.ts` and `scripts/e2e/onboarding-e2e.mjs` use the host `psql` binary over the
published TCP port, which the initialization server cannot answer either. No workflow other than
`ci.yml` provisions Postgres.

---

### Non-blocking

1. **`e2e/tests/fixtures.ts:74` still runs `psql` over the Unix socket inside the container**, so
   "every `psql` the e2e stack runs now connects over TCP inside the container" in the description
   is not literally true. It is *not* a live race — that helper only runs after Playwright's
   `webServer` gate, i.e. long after `bootstrapAndSeed()`. But it is a fifth copy of the shape that
   produced this bug in four places, and it is the one the next person will copy.
2. **Consider a mechanical guard.** This blind spot was hand-replicated four times and nothing stops
   a fifth. A grep-style assertion in `pnpm check` — any `psql`/`pg_isready` executed *inside* the
   postgres container must pass `-h`/`--host` — would hold the fix in place. The comments are good,
   but they only work on a reader who reads them.
3. **The wait budgets are no longer effectively instant.** The CI Container Smoke loop (30 × 1s) and
   the `doctor-docker` health gate (`--health-retries 30` at 1s) used to pass at ~1.5s because the
   socket answered during init; they now have to outlast initialization (~1.5s natural here, 9s in
   my slowed runs). Comfortable, but the margin is real now.
4. **The gate now depends on `listen_addresses = '*'`**, which the official image sets via
   `postgresql.conf.sample`. If the repo ever mounts a custom `postgresql.conf` or moves off the
   official image, all four probes need revisiting — worth a clause in the `stack-server.mjs`
   comment.
5. Unrelated observation from the smoke runs: two of three logged
   `IM_DELIVERY_WORKER_SCHEDULING_FAILED / ECONNREFUSED` during teardown, after the tests passed —
   the background IM delivery worker outliving `docker compose down`. Pre-existing and independent
   of this change; flagging only because I saw it.

### Limitations

- macOS + Docker Desktop, not `ubuntu-latest`. On this host `docker compose exec` costs ~1.5s, which
  by itself pushes the base gate's first probe *past* the natural ~1.5s initialization window — so
  with an unmodified init the base gate looked fine here (25/25 answered by the real server). That
  is a host artifact, not evidence for the base commit: on a Linux runner exec is ~10× cheaper and
  the 1s retry cadence lands the probe inside the window. I widened initialization to 2s to place
  the probe where a runner puts it, which is what produced §3 and §4.
- The reproduction is of the **mechanism** under a controlled window, not of CI's timing
  distribution. I cannot prove absence of flake on GitHub runners; what I can say is that the
  criterion is now transport-discriminating by construction, and that the four faces of the failure
  all disappear at this head under a window wide enough to hit them 6/6 times on base.
- No `full-isolated` claim: this says nothing about release readiness or any surface beyond the e2e
  bootstrap.

### Case disposition

`candidate-new-case` — "e2e stack bootstraps through first-time Postgres initialization" is worth a
prose QA case, and the base-vs-head gate replay is a reusable shape. Separate task, not this PR.

liuchao-001:
Thanks — the base-vs-head gate replay (base 15/15 satisfied by the initialization server, head 15/15 by
the real one) and the 6/6 reproduction are stronger evidence than what I brought, and the
`waitForContainerHealth()` check in particular went somewhere I had not: I had assumed the doctor's
health gate was already meaningful, and it was not.

Triage of the five non-blocking items:

1. **`e2e/tests/fixtures.ts` still uses the socket** — correct, and the description overclaimed. Filed as
   #533 rather than pushed here: it is not a race (that helper runs after the `webServer` gate), and a
   new head on this PR costs another full review + exact-head QA round on a fix that is currently green
   and holding back a flake that hits unrelated PRs today. I have reworded the PR description so it no
   longer claims more than the diff does.
2. **Mechanical guard in `pnpm check`** — agreed, and it is the item with real value, since the shape was
   hand-copied four times. Also #533; it needs its own small design (how the check recognizes an
   in-container invocation) and does not belong in a flake fix.
3. **Wait budgets are no longer effectively instant** — acknowledged, no change. The CI loop (30 × 1s) and
   the doctor's `--health-retries 30` now have to outlast initialization instead of being answered during
   it. On the measurements here that is ~1.5s against a 30s budget. The margin is real now, which is the
   point: the old "instant" pass was the bug.
4. **Document the `listen_addresses = '*'` dependency** — agreed, folded into #533 as a comment clause in
   `stack-server.mjs`. Worth stating precisely: a custom `postgresql.conf` or a non-official image would
   not silently reintroduce the race, it would make the probes hang until their deadline — loud, not flaky.
5. **`IM_DELIVERY_WORKER_SCHEDULING_FAILED` / `ECONNREFUSED` during teardown** — I saw it too, in my own
   smoke run, after the tests passed. Pre-existing and independent of this change. Not filing it: it is a
   shutdown-ordering log artifact with no failing assertion behind it yet, and I would rather not open a
   half-diagnosed report against the delivery worker.

One correction to the status in your comment: `ownership-gate` is green now. It had evaluated at 18:13,
before the approval at 18:25; the review relay re-ran it and it reports "Ownership satisfied for 4 changed
files." Every check on this head passes and `mergeStateStatus` is `CLEAN`. The PR is waiting only on
@yuezengwu, who is still a requested reviewer.

yuezengwu (approved):
Reviewed exact head `94e613e63410aaf339a7f48fa5c0aca8bb5e326c`. No blocking findings in the four changed readiness probes.

I independently replayed the unmodified `psql`, `waitForDatabase`, and `resetDatabase` functions from base `03d1c4a2a913835a3b85d4bcbfe3b2d377aaf66a` and this head, using isolated Compose projects with fresh tmpfs data and an 8-second SQL initialization delay:

| Image | Base gate | Base reset failures across handover | Head gate | Head resets |
| --- | --- | --- | --- | --- |
| `postgres:17` | Temporary server at 1.40s; TCP probe rejected | 2 | Real server at 9.28s | 16 passed, 0 failed |
| `postgres:17-alpine` | Temporary server at 1.38s; TCP probe rejected | 1 | Real server at 9.15s | 18 passed, 0 failed |

The observed base failures were missing Unix sockets during handover, consistent with the initialization race; I am not claiming reproduction of the exact CI error string. The TCP healthcheck also reached healthy after initialization on both images.

Independent validation at this head: `pnpm check`, `pnpm build`, `pnpm typecheck`, and `git diff --check` passed; Browser Smoke passed 8/8 against another fresh, isolated database. No author-branch changes were made, and the temporary Docker resources were removed.

The remaining socket helper in `e2e/tests/fixtures.ts` runs after the webServer readiness gate and is not a blocker for this fix. Browser teardown still emitted an IM delivery worker diagnostic while PostgreSQL was shutting down, after all assertions passed; this approval does not claim that shutdown behavior is resolved. Validation is focused local Docker coverage, not a measurement of GitHub runner flake rates or a full unit/integration rerun. Approval applies only to the SHA above.

Issue #486, opened 2026-09-02
Flaky: provider CLI turn target announces "ready" before installing its signal handlers


liuchao-001:
## What fails

`packages/client/src/__tests__/provider-cli-turn-runner.test.ts` › *private Turn launcher subprocess* › **forwards SIGTERM and SIGINT to the target**

```
AssertionError: expected 'ready\n' to contain 'got SIGINT'
```

Seen on CI run 33619890738 (job `Checks`, `@opentag/client#test`). It reproduces only under load, which is why it shows up on a busy runner and not locally.

## Root cause

This is not timing noise that a retry should paper over — the fixture publishes its readiness signal *before* it can honour a signal.

`packages/client/src/__tests__/fixtures/provider-cli-turn-target.mjs`, `sleep` mode:

```js
process.stderr.write("ready\n");                       // ← announced here
const finish = (signal) => { ... };
process.on("SIGTERM", () => finish("SIGTERM"));
process.on("SIGINT", () => finish("SIGINT"));          // ← installed only here
setInterval(() => undefined, 60_000);
```

The test treats `"ready"` as permission to signal:

```js
await expect.poll(() => stderr.includes("ready"), { timeout: 5_000 }).toBe(true);
child.kill(signal);
```

If the signal lands in the window between the write and the two `process.on` registrations, Node's default disposition terminates the process immediately. Nothing writes `got SIGINT`, and the process still exits 130 — so `expect([130, 143]).toContain(code)` passes and only the stderr assertion fails. That is exactly the observed failure: `stderr` was `'ready\n'` and nothing else.

The same window exists for `SIGTERM`; it is simply narrower in practice because it is the first of the two `signalRun` calls.

## Fix

Move the readiness write below the handler registrations, so `"ready"` means what the test assumes it means:

```js
const finish = (signal) => { ... };
process.on("SIGTERM", () => finish("SIGTERM"));
process.on("SIGINT", () => finish("SIGINT"));
setInterval(() => undefined, 60_000);
process.stderr.write("ready\n");   // only now is the target actually ready
```

No change to the test itself is needed, and no retry or timeout increase is involved — the ordering makes the poll-then-kill sequence race-free rather than merely less likely to lose.

## Context

Found while reviewing #473, whose head only touched `scripts/e2e/onboarding-e2e.mjs` and therefore cannot reach this package. Filing separately so the diagnosis is not lost in that PR's thread.

liuchao-001:
[...]

PR #532, opened 2026-09-04
refactor(web): let the Computer panel say each fact once

liuchao-001:
## Why

The Agent Computer settings page said the same thing several times over and spaced every line alike, so a reader got one undifferentiated column of sentences and had to read all of it to find the step that fixes anything.

## What changed

**Grouping by spacing.** The panel holds two groups: what this machine is and why it cannot work, then the step that reconnects it. 12px inside a group, 24px between them. Status and last seen are one fact about one machine, so the badge moved under the name it describes and carries last seen as its `StatusIndicator` detail — smaller and dimmer, since it is reference rather than the headline. The recovery sentence explains the badge directly above it, so it joins that group.

**The reconnect step gets a heading** naming it, reusing `computer_connect_repair_title` (already this region's accessible name) minus the machine name the identity header carries two rows above.

**The connect surface stops reserving rows it has nothing to put in.** Idle repair has neither an intro nor a countdown, and holding 20px open for them stranded the command block roughly 50px below the heading, with the equally empty status row underneath. Both are now reserved from the first issued command onward, where a real announcement can arrive and move the block. Nothing can move in idle: the first click swaps the whole surface for the pending skeleton. The status region stays mounted throughout so its first aria-live announcement is heard.

**Copy.** The connect step's idle intro goes — "Start OpenTag on X; this page will continue when it reconnects" repeated the offline sentence directly above it word for word. `Generate a repair command` → `Generate an install command`, which is what the command does. Every recovery sentence stays, including the Offline and Unable-to-confirm ones.

No new component: the black command surface here is the same `ComputerConnect` the new-Agent flow uses, in `repair` mode. Every change is a copy slot on that shared component or layout on the settings page, so the onboarding repair step inherits the same fixes.

## Testing

- `pnpm check` — pass
- `pnpm --filter @opentag/web typecheck` — pass
- `pnpm --filter @opentag/web test` — 62 files, 666 tests, pass
- Rendered the panel in Chrome across Offline / Provider-not-ready / Online and the issued-command state to measure the spacing; the reserved-row problem above was found that way, not by reading the code.

Live QA should confirm the same three states plus the onboarding repair step in a real browser at mobile width, where `.ots-command-lead` takes its 68px stacked height.


liuchao-001:
Rebased onto current `main` (37 commits) and re-verified at `f176eebb`.

The merge was textually clean, but running the merged tree surfaced a semantic conflict worth naming: the activation-funnel test `main` added while this branch sat asks for the connect button by its old name, `Generate a repair command`, which this branch renames to `Generate an install command`. Nothing overlapped in the diff — `main` appended a new test block, this branch renamed the action in a different file — so only executing the merged tree catches it. Fixed in `f176eebb`.

Re-verified on the merged tree: `pnpm check` clean, `apps/web` typecheck clean, `pnpm --filter @opentag/web test` 73 files / 810 tests passing. Also re-rendered the panel in Chrome across Offline / Provider-not-ready / Online to confirm the spacing rework survived the merge unchanged.

Independent QA against this exact head is requested and pending; that evidence is not here yet.

liuchao-001:
## Independent QA — `FAIL` · tier `focused-local` · head `f176eebb4c8355f913524dff69e12accb40a4b2a`

Run by code-reviewer as the independent QA gate (posted with @liuchao-001's GitHub account, per our
standing arrangement). Author-run checks do not count toward this; everything below is my own run
against the exact head.

**Evidence boundary.** A real `@opentag/server` serving the real `apps/web` production bundle built
from `f176eebb`, driven through a browser. Connection and provider readiness were supplied by a stub
client speaking the real v1 WS protocol — that stands in for the Client, which is what *reports*
readiness, not for the Server that answers. Covered: Agent → Computer settings in Offline /
Provider-not-ready / Online / command-issued / issue-failed, the same `ComputerConnect` repair step
reached from onboarding-v2 `agents/setup`, and `create` mode on `/agents/computers`; 1280px and
375px, `en` and `zh`. No claim beyond those paths.

### The settings page is right

The panel this PR set out to fix reads the way it was asked to. Measured, not eyeballed: 12px inside
each group, 24px at the group boundary, and 12px from the "Reconnect computer" heading to the
command block — the ~50px detachment is gone. Online shows badge only; Provider-not-ready correctly
withholds the reinstall affordance; `zh` matches all four decisions. Last-seen contrast is 5.32:1 at
12px, so the dimming stays AA. No horizontal overflow at 375px.

On the two things you asked me to check specifically:

- **The `aria-live` region survives the transition.** I tagged the `role="status"` node while idle
  and re-read it after the command was issued: same DOM node, `aria-live="polite"`, now carrying
  "Waiting for Ada's Mac to reconnect…". The first announcement lands in an already-mounted live
  region. Checked on the real DOM, not jsdom.
- **`create` mode cannot be affected.** It initialises to `issuing` and never reaches `idle`, so
  removing the idle lead row cannot touch "Connect a Computer". Confirmed live.

### F1 — MEDIUM — onboarding's repair step loses its only instruction

This is the blocker. Deleting `computer_connect_repair_intro` is justified as removing a line that
repeated the offline sentence above it word for word. That duplication only exists on the settings
page, which renders `computerRecoveryMessage(agent)`. Onboarding's `BoundComputerSection`
(`apps/web/src/onboarding-v2/agent-setup-page.tsx:1245`) renders no such sentence, so the deleted
line was that step's only statement of the remedy.

Same run cell, same seeded Agent, same offline Computer, base built and observed the same way:

| | what the step says |
|---|---|
| **base** (merge-base) | `Ada's Mac · macOS · Offline` → **"Start OpenTag on Ada's Mac; this page will continue when it reconnects."** → `Need to reinstall? / Generate a repair command` |
| **head** (`f176eebb`) | `Ada's Mac · macOS · Offline` → `Need to reinstall? / Generate an install command` |

The step footer still reads "Complete the action above, then check again", but the only copy above
it is now a reinstall escape hatch. A reader whose OpenTag is simply not running is pointed at
reinstalling, and nothing tells them the page resumes on its own.

Three ways out, your call: give the settings caller a way to suppress the intro instead of dropping
it for every caller; add the equivalent sentence to `BoundComputerSection`; or keep the intro and
let the settings page drop its own recovery sentence in the offline case.

### F2 — LOW — "Last seen" claims staleness for a Computer that is online

`computerStatusLine` carries the comment "a Computer that is online now has nothing to add", but
gates on the composite `ready` (machine reachable **and** provider ready) rather than
`computerState.state === "ready"`. In the four Provider branches the machine is online and
heartbeating, so the badge row renders `● Not ready   Last seen just now · Sep 9, 2026, 9:24 PM`.
Pre-existing logic — but this PR moved the line onto the badge row, so it now reads as one
self-contradicting statement. `agent-computer-settings.tsx:69`.

### F3 — LOW, pre-existing — the "no command, no intro" rule is applied to `idle` only

`issue-failed` and `expired` still render "Paste this command into the coding agent on Ada's Mac."
directly above a block reading "Failed to fetch / Try again". The PR's own test comment says
"Nothing introduces a command that has not been issued"; the guard is `state.kind === "idle"`. In
that same state `.ots-slot--status` holds 28px of empty height because the error moved into the
command block — the dead space this PR set out to remove. Not introduced here; flagging it because
the rule you wrote is broader than the guard you shipped.

### F4 — observation, not a defect

Removing the idle lead reservation means the command block shifts down when the command arrives:
+32px at 1280px, +80px at 375px. Base reserved the row in both states, so there was no shift. It
happens at a deliberate click behind a full skeleton swap, so it's a fair trade for the 50px — worth
knowing that on mobile the control just pressed moves out from under the reader's finger.

### Automated checks I ran myself at this head

- `pnpm --filter @opentag/web test` — 73 files / 810 tests pass
- `pnpm --filter @opentag/web typecheck` — pass
- CI on `f176eebb` — 15 success, 4 skipped, 0 failing

### Limitations

Connection/provider state came from a protocol-level stub, not a real OpenTag install. No
screen-reader run — the live-region claim rests on DOM identity across the transition. Only 1280px
and 375px measured. To build the base control I temporarily checked out the merge-base of the
changed `apps/web` files, then restored to `f176eebb`; the tree was verified clean and every head
observation above was re-captured after the restore.

**Case disposition**: `candidate-new-case` — "a shared connect surface is driven from both onboarding
and settings; copy removed for duplication on one caller must be checked on the other" is worth a
committed case.

Recommend holding merge until F1 is answered. F2/F3 are fine as follow-ups.

liuchao-001:
## F1 fixed, plus F2 — new head `2fb0671b`

Thanks — F1 is a real defect and I verified it independently before fixing: `BoundComputerSection` renders `ComputerStepHeader` + `ComputerSummary` + `ComputerConnect` and no recovery sentence, and the header only says "Connect your computer / {name} runs on your own computer." Nothing there states the remedy. My original justification checked that onboarding *had* a status badge, not that it had the *instruction* — that was the error.

**F1 — fixed, option 2.** `computer_connect_repair_intro` is restored and rendered by `BoundComputerSection` between the summary and the command surface. Not option 1: a suppress-intro flag would put caller-specific policy inside the shared component, and this sentence is caller framing ("why you are here, what happens next") rather than command-lifecycle copy, which is what `ComputerConnect` should own. Not option 3: dropping the Settings offline recovery sentence would reverse an explicit decision by @liuchao-001 to keep all seven.

The offline onboarding test now asserts the remedy is stated. I confirmed the guard is not vacuous — with the fix reverted it fails with "Unable to find an element with the text: Start OpenTag on Review Mac…", and passes with it restored.

**F2 — fixed.** `staleFor` now gates on `computerState.state === "ready"` rather than the composite `ready`. The Provider branches leave a reachable machine, so they no longer render a staleness claim; `agent-computers.test.tsx` asserts no `Last seen` in the sign-in case. You were right that this is pre-existing logic — but this branch is what moved it onto the badge row, so it is this branch's contradiction to remove.

**F3 — not fixed, deliberately.** Correct that my comment claimed a broader rule than the guard implements. Rather than widen the guard in passing, I narrowed the claims: both the component comment and the test comment now say idle only, and name `issue-failed`/`expired` as still introducing an absent command. Those two states need a copy decision (what should sit above a block showing "Failed to fetch"), which is a separate change.

**F4 — acknowledged, no change.** The +80px mobile shift is real and worth the 50px it buys.

Verified at `2fb0671b`: `pnpm check` clean, typecheck clean, 73 files / 810 tests passing. Re-rendered the onboarding repair step through `/internal/agent-setup` (`computer-reconnect` scenario) to confirm the sentence is back, and the Settings Provider-not-ready state to confirm the badge now stands alone. To force the lab into that scenario I temporarily changed its default constant and restored it; `git diff` on that file is empty.

Re-running QA against this head is on you — the previous run is void.

liuchao-001:
## Independent QA (round 2) — `PASS` · tier `focused-local` · head `2fb0671b75e0650ec42c125a97ba2cc1de26ffd1`

Run by code-reviewer as the independent QA gate (posted with @liuchao-001's GitHub account). The
`f176eebb` run is void; everything below is a fresh run against this head.

**Both fixes hold, and I verified the guards are not vacuous rather than taking the claim on trust.**
One thing is not in your control and needs an answer before merge — the red check at the bottom.

### F1 — fixed, verified live and by mutation

The onboarding repair step states the remedy again. Same run cell, same seeded offline Computer,
`/agents/setup`:

`Ada's Mac · macOS · Offline` → **"Start OpenTag on Ada's Mac; this page will continue when it
reconnects."** → `Need to reinstall? / Generate an install command`

Confirmed at 1280px and 375px and in `zh` ("在 Ada's Mac 上启动 OpenTag；重新连接后此页面会自动继续。").
No overflow at 375px.

**Mutation check.** I deleted the new `<p>` from `BoundComputerSection` and re-ran the suite: 1 failed
/ 809 passed, `Unable to find an element with the text: Start OpenTag on Review Mac; this page will
continue when it reconnects.` Restored, 810/810. The guard bites.

Agreed on the reasoning for option 2 over option 1 — a suppress flag would have put caller policy in
the shared component, and this sentence is caller framing, not command-lifecycle copy.

### F2 — fixed, verified live and by mutation

Online-but-Provider-not-ready now renders `● Not ready` alone with no staleness claim; the recovery
sentence still names the Provider and the problem. Offline still carries
`Last seen 3 days ago · Sep 6, 2026, 9:45 PM`, so the fix did not overreach.

**Mutation check.** Reverting `staleFor` to the composite `ready` fails
`agent-computers.test.tsx > explains an unready Provider on the Computer page` with
`Last seen 21 days ago · Aug 19, 2026, 5:00 PM`. Restored, green.

I also walked the other `computerState.state` branches: `unconfirmed` and `not_bound` carry a null
`lastConfirmedAt`, and a reachable machine with no Provider readiness at all still reports `Online`.
The change touches only the four Provider branches, which is what it claimed.

### F3 — narrowing the claim instead of the guard is the right call

Verified both comments now say idle only and name `issue-failed`/`expired` as still introducing an
absent command. Agreed that "what belongs above a block showing *Failed to fetch*" is a copy decision
and not something to settle inside a spacing PR.

### Regression sweep at this head

Settings Offline / Online / issued / `zh` all unchanged from what you had approved: 12px inside a
group, 24px at the boundary, 12px heading→command in idle, 44px once the lead row arrives. No console
or page errors in any state.

**One observation, no action needed.** The onboarding step now shows both "Start OpenTag on Ada's
Mac…" and, once a command is issued, "Paste this command into the coding agent on Ada's Mac." Base
showed only one at a time, but only because they shared a slot by accident. Both are true and the
order reads fine; noting it so nobody meets it later and calls it a duplicate.

### Automated checks I ran myself at this head

- `pnpm --filter @opentag/web test` — 73 files / 810 tests pass
- `pnpm --filter @opentag/web typecheck` — pass

### ⚠️ Merge is blocked by a red check that is not yours

`Node 22.22.2 Compatibility` **fails** at this head:

```
packages/client/src/__tests__/provider-cli-turn-runner.test.ts:427
  private Turn launcher subprocess > forwards SIGTERM and SIGINT to the target
  AssertionError: expected 'ready\n' to contain 'got SIGTERM'
```

Not attributable to this PR, and I checked rather than assumed:

- the PR touches `apps/web` only; that file is **byte-identical to `origin/main`**
- `Node 24` and `Node 26` pass at this same head on the same code
- it passes 5/5 locally on Node 22.23.2

It is a genuine latent race in the test, not noise to wave through: `signalRun` polls until the
target prints `ready` on stderr and then immediately signals the **launcher**. Printing `ready` does
not prove the trap is armed, so the process can die before anything writes `got SIGTERM` — which is
exactly the observed `'ready\n'`. Worth its own issue against `packages/client`; re-running the job
should clear this PR.

**Case disposition**: `candidate-new-case` — unchanged from round 1: a shared connect surface driven
from both onboarding and settings, where copy removed as "duplicated" on one caller must be checked
on the other.

The `apps/web` change is good to merge once that job is green.

liuchao-001:
Following up on the red check in my round-2 QA: `Node 22.22.2 Compatibility` is green after a re-run.
All checks at `2fb0671b` are now 15 success / 4 skipped / 0 failing, and `mergeStateStatus` is `CLEAN`.

The re-run produced no new commit, so the head is still `2fb0671b75e0650ec42c125a97ba2cc1de26ffd1`
and the round-2 `PASS` above remains bound to it. Nothing to re-run on my side.

The flake has an existing issue — #486, same fixture and same root cause, with this occurrence's
SIGTERM branch now recorded there. My earlier suggestion to open a new issue is superseded by that.

PR #539, opened 2026-09-05
fix(im): fail closed on external call host admission and bound the waiter queue

bestony:
Makes `ExternalCallPolicy` fail closed. An empty host allowlist now denies every host instead of allowing all of them, provider-supplied URLs are admitted before a bearer token is ever attached, the concurrency waiter queue is bounded, and a call that misses its deadline stops holding capacity it no longer uses.

Covers three P1 items from `todo.zh-CN.md`: *让外部 URL 策略默认拒绝*, *限制外部调用等待队列*, and the `external-call-policy.ts` half of *让每个 Provider 操作都传递取消信号*. The `im-delivery-worker.ts` half of that last item belongs to a separate branch and is deliberately untouched here.

## Checklist

- [x] Empty allowlist denies; five constructor sites pass real hosts
- [x] URL admission requires HTTPS, rejects IP literals and non-default ports, rejects encoded authority, compares IDN hosts after punycode normalization
- [x] Slack `fetchResource` admits the provider URL before constructing the bearer header
- [x] `maxQueueDepth` rejects immediately with `IM_PROVIDER_OVERLOADED` and emits queue depth/rejection metrics
- [x] A timed-out action keeps its permit until the underlying action settles
- [x] Focused tests cover all eight required behaviours, including the Slack HTTP credential boundary

## The actual exposure

Production wiring was already allowlisted — `packages/server/src/index.ts` builds both IM policies with a real `allowedHosts` list. The hole was `hostAllowed()` treating an **empty** list as "allow everything", combined with five sites that constructed a policy with no hosts at all:

- `services/im/im-resource-service.ts`
- `services/im-bindings/feishu/connection-manager.ts`
- `services/im-bindings/feishu/adapter.ts`
- `services/im-bindings/feishu/registration.ts`
- `services/im-bindings/slack/default-api-client.ts`

Each now passes only the hosts it actually wraps: Slack gets `slack.com` and `files.slack.com`; Feishu/Lark gets `open.feishu.cn` and `open.larksuite.com`; the provider-agnostic resource service gets the union.

The sharpest edge was `DefaultSlackApiClient.fetchResource`, which sent `authorization: Bearer <token>` to whatever `url_private_download` / `url_private` Slack returned. A compromised or spoofed provider response could therefore have pointed the bot token at an arbitrary host, over plain HTTP. The URL is now admitted first, and the header is only built after it passes.

`allowHttpLoopbackForTests` exists for tests only: no production wiring sets it, and even with it set the loopback host must still be allowlisted.

Each criterion was re-run against this branch's checkout, not taken from the agent's claim.

| # | Criterion | Result |
| --- | --- | --- |
| 1 | `pnpm --filter @opentag/server test` exits 0 with a test per required behaviour | **PASS** — 68 test files, 797 tests |
| 2 | `new ExternalCallPolicy({})` rejects every host | **PASS** — test asserts `IM_PROVIDER_HOST_NOT_ALLOWED` *and* that the transport was never called |
| 3 | No `authorization` header reaches an unadmitted URL | **PASS** — see below |
| 4 | At `maxQueueDepth`, a further call rejects immediately | **PASS** — see below |
| 5 | `pnpm check` exits 0 | **PASS** |
| 6 | `pnpm typecheck` exits 0 | **PASS** |
| 7 | No change to `packages/server/src/runtime/im-delivery-worker.ts` | **PASS** — absent from the diff |

**On criterion 3**, the proof is stronger than "an error was thrown": the Slack adapter test feeds `http://files.slack.com/private` as the provider URL, asserts the rejection carries `IM_PROVIDER_HOST_NOT_ALLOWED`, and then asserts the global `fetch` call count is *unchanged* — so no request was constructed at all, and no header with it. The branch also records a grep of every `authorization` construction site in `packages/server/src`; both live sites sit directly behind a `policy.admitUrl(...)` call.

**On criterion 4**, the test does not merely assert the overload error. It stamps the clock, issues the overflowing call with `timeoutMs: 1_000`, and asserts the rejection arrives in under 100 ms — distinguishing "rejected immediately" from "waited out the deadline", which is the behaviour the item actually asks for.

Note on test fixtures: `slack-adapter.test.ts` previously used `https://files.example/download`, a host that default-deny now correctly rejects. The fixture was moved to `files.slack.com` — a fixture correction, not a weakened assertion; the surrounding assertions are unchanged and one was added.

## Deviations

None. The branch records "No plan deviation".

## Note for the reviewer

This branch carries a single commit for a six-item checklist. The dispatch policy asks for continuous, per-unit commits; this lane produced one. Flagged rather than rewritten — rebasing verified work costs more than the untidy history.

---

Produced by dispatch-codex run `d76ae8`, lane `im-call-policy-hardening-2`. Written by codex running under bypassed approvals in an isolated git worktree; every acceptance criterion above was re-run and checked by the orchestrator before this pull request was opened.

### DIFF
--- a/packages/server/src/__tests__/external-call-policy.test.ts
+++ b/packages/server/src/__tests__/external-call-policy.test.ts
@@ -324,7 +324,12 @@ describe("ExternalCallPolicy", () => {
         resolve();
       };
     });
-    const policy = new ExternalCallPolicy({ maxConcurrency: 1, maxAttempts: 1, defaultTimeoutMs: 10 });
+    const policy = new ExternalCallPolicy({
+      maxConcurrency: 1,
+      maxAttempts: 1,
+      defaultTimeoutMs: 10,
+      abandonmentWindowMs: 1_000,
+    });
     const timedOut = policy.run("deadline.holds", async () => actionGate, { timeoutMs: 10 });
     await expect(timedOut).rejects.toMatchObject({ code: "IM_PROVIDER_CALL_DEADLINE_EXCEEDED" });
     expect(actionSettled).toBe(false);
@@ -340,6 +345,39 @@ describe("ExternalCallPolicy", () => {
     settleAction();
     await expect(queued).resolves.toBe("released");
     expect(actionSettled).toBe(true);
+  });
+
+  it("releases a permit after the bounded abandonment window", async () => {
+    const metrics: Array<{ type: string; operation: string }> = [];
+    const policy = new ExternalCallPolicy({
+      maxConcurrency: 1,
+      maxAttempts: 1,
+      defaultTimeoutMs: 10,
+      abandonmentWindowMs: 10,
+      onMetric: (metric) => metrics.push(metric),
+    });
+    const abandoned = policy.run("abandoned", async () => new Promise<void>(() => undefined));
+    await expect(abandoned).rejects.toMatchObject({ code: "IM_PROVIDER_CALL_DEADLINE_EXCEEDED" });
+
+    await expect(policy.run("after-abandonment", async () => "released", { timeoutMs: 100 })).resolves.toBe("released");
+    expect(metrics).toEqual(
+      expect.arrayContaining([expect.objectContaining({ type: "abandoned", operation: "abandoned" })]),
+    );
+  });
+
+  it("does not retry a timed-out action while its provider call is still pending", async () => {
+    let dispatches = 0;
+    const policy = new ExternalCallPolicy({
+      defaultTimeoutMs: 10,
+      abandonmentWindowMs: 10,
+    });
+    const call = policy.run("no-overlap", async () => {
+      dispatches += 1;
+      return new Promise<void>(() => undefined);
+    });
+
+    await expect(call).rejects.toMatchObject({ code: "IM_PROVIDER_CALL_DEADLINE_EXCEEDED" });
+    expect(dispatches).toBe(1);
   });
 
   it("backs off and transitions the circuit", async () => {
--- a/packages/server/src/__tests__/external-call-policy.test.ts
+++ b/packages/server/src/__tests__/external-call-policy.test.ts
@@ -94,6 +94,55 @@ describe("ExternalCallPolicy", () => {
     await expect(policy.fetch("https://api.example.test/file")).resolves.toMatchObject({ status: 200 });
   });
 
+  it("denies every host when the allowlist is empty", async () => {
+    const transport = vi.fn();
+    const policy = new ExternalCallPolicy({ transport });
+    await expect(policy.fetch("https://example.test/resource")).rejects.toMatchObject({
+      code: "IM_PROVIDER_HOST_NOT_ALLOWED",
+    });
+    expect(transport).not.toHaveBeenCalled();
+  });
+
+  it("rejects private, loopback, and encoded IP literal hosts", async () => {
+    const policy = new ExternalCallPolicy({
+      allowedHosts: ["127.0.0.1", "10.0.0.1", "::1"],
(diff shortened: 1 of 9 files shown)


yuezengwu (approved):
Reviewed exact head `cbb19948a658d80a87fb502364440bfcd4ce7200`. No blocking findings in the URL admission, bounded wait queue, cancellation/permit handling, or updated provider construction sites.

Independent focused validation in a clean detached worktree: 94 tests passed across ExternalCallPolicy, Slack adapter, Feishu adapter/registration/setup failure, and IM resource API suites. `pnpm check`, Server typecheck, and `git diff --check` passed. The transport tests verify denial before request dispatch, explicit host/HTTPS restrictions, prompt queue rejection, and retaining capacity until a timed-out underlying action actually settles.

This is approval of the code at the SHA above, with focused local validation; it does not claim a new live-provider deployment or a full repository test rerun. No author-branch changes were made.

bestony:
Pushed `0588986d` and `0d9cf5af`. **This supersedes the head you approved (`cbb19948`), so the PR needs a fresh review** — I am sorry to spend your approval, but an independent audit found that one change on the approved head traded a bug for a worse one, and shipping it would have been the wrong call.

**The problem with the approved head.** Holding the concurrency permit until the underlying action settles was unbounded, and the actions this policy wraps could not be aborted: the Slack `WebClient` was built with no `timeout`, and `auth.test()` / `files.info()` ignored the `AbortSignal` the policy passed. A hung socket therefore never settled and the permit was never returned. Production shares one policy with `maxConcurrency: 16`, so sixteen hung calls would have exhausted it permanently — and because the same PR added the bounded waiter queue, every later call would then have failed `IM_PROVIDER_OVERLOADED` forever with no recovery path. Your review note reads that the tests verify capacity is retained until a timed-out action settles, which is exactly right; what the tests did not cover is that the settle might never come.

**Fixes.** The permit is now released when the action settles **or** when a bounded abandonment window elapses, whichever comes first, with an `abandoned` metric emitted through `#onMetric`. `SLACK_WEB_CLIENT_OPTIONS` now sets an explicit `timeout` so the SDK calls actually settle, and the policy signal is forwarded into the Slack and Feishu transports at each call site that accepts one.

**Second defect, also fixed.** `#runAttempts` was starting the next retry while the previous attempt's promise was still in flight, so a timed-out call could reach the provider twice concurrently — the *重试重叠、重复 Provider dispatch* harm the originating todo item names explicitly. The next attempt is now gated on the previous action settling, and a test asserts a never-settling timed-out action is dispatched **exactly once** under the default `maxAttempts: 2`.

Three tests added: `releases a permit after the bounded abandonment window`, `does not retry a timed-out action while its provider call is still pending`, and `passes the policy signal into Slack WebClient transport and configures a timeout`.

Verified independently before pushing: `pnpm --filter @opentag/server test` 68 files / 800 tests (up from 797, no regressions), `pnpm check` and `pnpm typecheck` all exit 0. No migration added; `im-delivery-worker.ts` and `packages/server/src/index.ts` remain untouched as other branches own them.

yuezengwu (approved):
Independently re-reviewed the full change at `0d9cf5af4bcbf4d92301bb365ef2e7aed09134a2`; the approval of the previous head was not reused.

No blocking finding in the URL admission, credential boundary, bounded waiter queue, deadline/cancellation handling, or SDK transport changes. On this head, 93 focused tests across six files passed, as did the shared build, `pnpm check`, server typecheck, and `git diff --check`.

Validation boundary: bounded abandonment releases the policy permit after the grace period; it is not proof that every uncooperative SDK operation has terminated. No live Slack/Feishu account was used for these tests.

PR #540, opened 2026-09-05
fix(im): move pending overflow cleanup out of the ingest transaction

bestony:
> **Migration collision resolved — this branch now lands at `0040`.** Four sibling PRs from this batch merged ahead of it, two carrying migrations, so `main` owns both `0038_melted_katie_power.sql` and `0039_clean_squirrel_girl.sql`. This branch's original `0038_vengeful_justin_hammer.sql` was **regenerated, not renumbered**, and is now `0040_watery_electro.sql`. Renumbering would have been unsafe: a renamed migration keeps its snapshot's `prevId`, and two snapshots sharing a `prevId` make `drizzle-kit generate` abort permanently *with exit code 0*, so no gate catches it. An earlier version of this note prescribed renumbering; that instruction was wrong and was never followed.

Takes the ordered `OFFSET` scan out of the inbound-message ingest transaction. Previously every inserted delivery triggered its own overflow expiry pass, so a busy Session repeated an expensive ordered query while holding the ingest transaction open.

Covers the P1 item *将 pending overflow 清理移出 ingest transaction* from `todo.zh-CN.md`.

## Checklist

- [x] Per-delivery `#expireOverflow` call replaced by one gated consideration per session+attention
- [x] Cheap pending count gate; no ordered scan while under capacity
- [x] Overflow pass runs after commit, bounded and de-duplicated, failures isolated from ingest
- [x] Covering index added, guided by EXPLAIN
- [x] Covering index migration, regenerated as `0040_watery_electro.sql` after the `main` merge; `CURRENT_MIGRATION_COUNT` → 41
- [x] Tests: no in-transaction scan under capacity, boundary semantics preserved, async failure isolated

## Rebasing onto the new `main` (`5a41a9b6`)

Three commits, in the order they had to happen:

- **`5d7030c9`** merges `main` in. A merge, not a rebase — this branch must stay append-only so it can be pushed fast-forward. The conflicts were confined to four migration-machinery files; no production code conflicted.
- **`ef7653fa`** regenerates the migration. `db:generate` diffed the schema against `main`'s `0039` snapshot and produced exactly one new file, `0040_watery_electro.sql`, whose entire content is the same index this PR always added:

  ```sql
  CREATE INDEX "im_message_deliveries_overflow_idx" ON "im_message_deliveries" USING btree ("session_id","attention","state","reason","message_id");
  ```

  `CURRENT_MIGRATION_COUNT` moves to **41**, matching the regenerated journal.
- **`24005696`** restores `packages/server/drizzle/meta/0038_snapshot.json`.

That last commit is worth a sentence, because the bug it fixes is easy to reproduce and hard to see. `meta/0038_snapshot.json` is a **shared path**: this branch's old 0038 snapshot and `main`'s 0038 snapshot have the same filename, which is why the merge raised it as an add/add conflict. Deleting "this branch's" copy therefore deleted `main`'s, and the branch briefly removed a file `main` has, breaking the `prevId` chain that `0039_snapshot.json` depends on.

**Nothing caught it automatically.** `pnpm check`'s migration drift check passes without validating the snapshot chain, and — the part that made it genuinely invisible — `git diff --diff-filter=D` reports *nothing*, because rename detection pairs the deleted `0038_snapshot.json` with the newly added `0040_snapshot.json` and scores it a rename. Only `git diff --no-renames --diff-filter=D` exposes it. Reviewers auditing a regenerated migration elsewhere in this repo should use `--no-renames`.

### Acceptance criteria — verified independently by the orchestrator

| # | Criterion | Result |
| --- | --- | --- |
| 1 | No file deleted relative to `main` under `packages/server/drizzle/`, checked with `--no-renames` | **PASS** — empty |
| 2 | `meta/0038_snapshot.json` byte-identical to `main`'s copy | **PASS** |
| 3 | Snapshot chain has no gap: `0037`, `0038`, `0039`, `0040` all present | **PASS** |
| 4 | `0040_watery_electro.sql` creates `im_message_deliveries_overflow_idx` on the same five columns in the same order | **PASS** — read from the generated SQL |
| 5 | `main`'s `0038`/`0039` SQL and `0039_snapshot.json` byte-identical; journal 41 entries == `CURRENT_MIGRATION_COUNT` 41 | **PASS** — neither number moved for the snapshot-only restore, which is correct |
| 6 | `pnpm --filter @opentag/server test` exits 0 | **PASS** — 70 files, 853 tests |
| 7 | `pnpm --filter @opentag/server test:integration` exits 0 | **PASS** — 18 files, 324 tests |
| 8 | `pnpm check` exits 0 | **PASS** |
| 9 | `pnpm typecheck` exits 0 | **PASS** |
| 10 | History append-only; merges into `main` with no conflicts | **PASS** — `c0f24863` still an ancestor, `git merge-tree` reports 0 conflicts |

The overflow-cleanup behaviour @yuezengwu approved at `c0f24863` is unchanged by all three commits; this round touched migration machinery only.

## What changed

`im-message-inbox.ts` previously called `#expireOverflow` **inside the insert loop**, once per delivery, inside the ingest transaction — and `#expireOverflow` is a CTE that joins `im_message_deliveries` to `im_messages`, orders by `occurred_at desc, id desc`, and applies `offset <capacity>` (100 direct / 500 ambient).

Now:

- the in-transaction work is a cheap `pending` / `reason is null` **count gate**; a session under capacity pays no ordered scan at all;
- over-capacity candidates are collected into a map keyed by `(sessionId, attention)`, so one ingest schedules at most one pass per key;
- the pass runs **after commit**. A queued pass does not spawn a second one; a failure is logged, clears the key, and a later over-capacity ingest retries it — the already-committed ingest is never failed by it;
- the expiry `UPDATE` carries a second `pending` / `reason is null` guard so an async pass cannot stomp a delivery that was accepted concurrently;
- `im_message_deliveries_overflow_idx` on `(session_id, attention, state, reason, message_id)` backs the gate.

`packages/server/src/index.ts` gets one line: `ImMessageInbox` is now constructed with a `serviceLogger("im-inbox")`, which is what lets the post-commit pass report its own failures without failing ingest.

Each criterion was re-run against this branch's checkout, not taken from the agent's claim.

| # | Criterion | Result |
| --- | --- | --- |
| 1 | `pnpm --filter @opentag/server test` exits 0 | **PASS** — 68 test files, 791 tests |
| 2 | `pnpm --filter @opentag/server test:integration` exits 0 with the three required scenarios | **PASS** — 18 test files, 318 tests; four new scenarios, listed below |
| 3 | Boundary semantics unchanged: over-capacity pending deliveries still become `expired` / `capacity`, in-capacity ones untouched | **PASS** — asserted via `toMatchObject({ state: "expired", reason: "capacity" })` for both the 100-direct and 500-ambient boundaries |
| 4 | Evidence the ingest transaction no longer runs the ordered `OFFSET` query under capacity | **PASS** — an integration test drives 20 repeated under-capacity ingests with a query-log assertion that no `OFFSET` query executes |
| 5 | `CURRENT_MIGRATION_COUNT` equals 39 and exactly one new file under `packages/server/drizzle/` | **PASS** |
| 6 | No change to `packages/server/src/runtime/im-delivery-worker.ts` | **PASS** — absent from the diff |
| 7 | `pnpm check` exits 0 | **PASS** — migration drift, ownership, workspace contract and complexity ratchet all pass |
| 8 | `pnpm typecheck` exits 0 | **PASS** — all 6 workspaces |

New integration scenarios:

- `does not run the ordered overflow pass inside an under-capacity ingest`
- `commits ingests while one overflow pass is queued for the session and attention`
- `expires ambient deliveries beyond capacity and preserves the 500 retained entries`
- `keeps a committed ingest when async overflow expiry fails and retries it later`

EXPLAIN evidence recorded by the branch: against a migrated database with the two older overlapping indexes disabled, the gate plans as `Index Only Scan using im_message_deliveries_overflow_idx` with all four gate predicates in `Index Cond`. The ordered expiry plan is now confined to the post-commit pass.

## Deviations

None from the confirmed plan.

## Cross-branch note for the reviewer

- The one-line `packages/server/src/index.ts` edit sits at the `ImMessageInbox` construction site (~line 274). A sibling branch in this batch edits the server startup/config region of the same file (~lines 150–200). The regions are disjoint and expected to merge textually, but both branches touching one file is worth knowing before you sequence them.
- `packages/server/src/db/schema/im-messages.ts` is also edited by a sibling branch. This branch deliberately adds **only** its one index there, to keep the two diffs mergeable.

---

Produced by dispatch-codex run `d76ae8`, lane `inbox-overflow-async-7`. Written by codex running under bypassed approvals in an isolated git worktree; every acceptance criterion above was re-run and checked by the orchestrator before this pull request was opened.


packages/server/src/services/im/im-message-inbox.ts
```diff
@@ -624,10 +645,66 @@ export class ImMessageInbox {
...
+  ): Promise<void> {
+    for (const candidate of candidates.values()) {
+      if (await this.#hasOverflow(transaction, candidate.sessionId, candidate.attention)) continue;
+      candidates.delete(this.#overflowKey(candidate.sessionId, candidate.attention));
+    }
+  }
+
+  #queueOverflowExpiry(candidate: OverflowCandidate): void {
+    const key = this.#overflowKey(candidate.sessionId, candidate.attention);
+    if (this.#overflowPasses.has(key)) return;
```
yuezengwu: [P2] Preserve overflow arriving during an in-flight pass

The early return discards a new cleanup request even when the running transaction has already taken its snapshot or committed. If it has reduced pending deliveries to 100, another ingest can commit delivery 101 before the pass promise settles; this branch drops that request, and `finally` deletes the key without a follow-up. With no later ingest, capacity remains exceeded indefinitely. I reproduced pending=101 / passes=1 using the real inbox and PGlite while delaying the cleanup transaction's completion acknowledgement. Retain/coalesce the new request and reconsider it after the current pass completes; the existing test only pauses before the expiry query and misses this window.

yuezengwu (changes requested):
Reviewed exact head `3c366b56d8b5366ccd72f5ea5568fd37a9a1290b`. Please preserve overflow work that arrives while the current pass is still registered.

I reproduced the lost-work window using the repository's PGlite fixture and the actual ImMessageInbox: delay resolution of the cleanup transaction after it commits, ingest one additional delivery while its pass remains in the map, then release the transaction completion. Pending count remains 101 with only one pass, instead of returning to the direct capacity of 100. The two existing scheduling tests pass; the additional regression fails. This is a delayed transaction-completion acknowledgement, not an injected cleanup failure.

The async path must remember that new overflow arrived during a pass and arrange another bounded consideration after that pass settles. Please add this interleaving to the regression coverage. I made no author-branch changes; this finding is independent of the currently reported CI failure.

bestony:
Pushed `c0f24863`. You were right, and the existing scheduling tests could not have caught it.

**The defect.** `#queueOverflowExpiry` kept the `(sessionId, attention)` key in `#overflowPasses` until the pass settled — correct for preventing two concurrent passes, but it also meant overflow arriving *during* a registered pass was silently dropped: the ingest saw the key, skipped scheduling, and nothing re-checked the session afterwards.

**The fix.** A `#overflowRechecks` map now records that late overflow arrived while a pass was registered. When the running pass settles, the code re-reads `hasOverflow` and schedules exactly **one** follow-up pass if it is still over capacity. It is bounded — a single follow-up per settle, never a chain — and the follow-up goes back through `#queueOverflowExpiry`, so the guarantee that two passes never run concurrently for the same key is preserved.

**Regression coverage mirrors your sequence**, parameterised over both attentions (`direct` 100 / `ambient` 500): delay the cleanup transaction's completion acknowledgement after it commits, ingest additional deliveries while the pass is still in the map, assert the count sits at capacity + 1 and that `beforeOverflowExpiry` has fired only once, then release and assert with `expect.poll` that the pending count returns to **capacity** and that `beforeOverflowExpiry` fired exactly **twice** — proving the recheck happened and that it did not become an unbounded chain. Final state is asserted exactly: three rows `expired`/`capacity`, the rest `pending` with a null reason. This is a delayed transaction-completion acknowledgement, matching your reproduction, not an injected cleanup failure.

Verified independently before pushing: `pnpm --filter @opentag/server test` 69 files / 795 tests, `pnpm --filter @opentag/server test:integration` 18 files / 318 tests, `pnpm check` and `pnpm typecheck` all exit 0. The changed-line coverage gate was reproduced locally with the repository's own `evaluatePatchCoverage`: **58/58 lines, 100%**.

This new commit supersedes the head you reviewed, so the PR needs a fresh look.

yuezengwu (approved):
Re-reviewed the complete change at `c0f248634f22466273c70f2c0ec2c5aa7dcdfeb9`; the previous head's finding and test results were not carried forward.

The delayed-commit-acknowledgement regression is resolved on this head. I reran the original PGlite reproduction: overflow arriving while the previous pass remains registered is eventually reduced back to capacity. The new direct/ambient tests also verify the follow-up pass. Seven unit/regression tests and 103 PostgreSQL integration tests across the IM binding and setup-backfill suites passed. Shared build, `pnpm check`, server typecheck, and `git diff --check` passed; the migration snapshot adds the expected delivery index and the migration count is 39.

No blocking finding remains in this review. This exact-head approval does not waive repository checks or authorize merging the branch.

yuezengwu (approved):
Fresh technical review of `240056966bd9c874408180ba79b636fdef621376` against main `5a41a9b66020ecea817bca338f713d3d2568fee5`. The current main commit is an ancestor of the tested head; GitHub reports MERGEABLE and CLEAN.

I re-read the full production diff and independently audited the regenerated migration. All existing migration SQL files and snapshots are byte-identical to main, including `0038_snapshot.json`. The snapshot IDs form a valid predecessor chain without duplicate IDs; `0040` follows `0039` and adds only `im_message_deliveries_overflow_idx` besides snapshot identity fields. The journal has 41 entries, matching `CURRENT_MIGRATION_COUNT`, and all recorded SQL hashes match their files. No existing migration file is deleted, checked without rename detection.

Fresh validation: 359 tests passed (35 affected unit tests and the complete PostgreSQL integration suite: 324 tests across 18 files). Shared/Client builds, repository checks, Server typecheck, and `git diff --check` passed; the detached worktree remained clean.

The exercised cases include repeated under-capacity ingests without ordered OFFSET queries, post-commit cleanup while new ingests continue, the 100-direct/500-ambient retention boundaries, failure isolation with later retry, and coalesced rechecks after a delayed cleanup commit acknowledgement. The cleanup UPDATE retains its pending/reason guards for concurrent delivery state changes.

All reported CI checks currently pass or are skipped. This is technical approval of the current head; rollout of the schema migration remains a maintainer decision.

PR #541, opened 2026-09-05
perf(server): bound Agent activity reads and aggregate usage in SQL

bestony:
Bounds the Agent list's activity read. A crashed runtime could leave `accepted` deliveries with no report forever, and every one of those rows stayed eligible for **every** `/agents` query — with usage then aggregated row-by-row in JavaScript.

Covers the P1 item *限制 Agent activity 读取* from `todo.zh-CN.md`.

## Checklist

- [x] Bounded recovery window; stale unreported `accepted` rows project to `unknown`
- [x] Usage aggregation pushed into SQL (with the residual JS step justified in writing)
- [x] Explicit limit and deterministic order on the activity read
- [x] Tests: stale-row projection, old-vs-new aggregation parity, bounded row count
- [x] No migration added; index recommendation recorded with EXPLAIN evidence

## What changed

The activity query's `where` clause used to be `or(accepted AND reportedAt IS NULL, acceptedAt >= usageStartedAt)`. The second branch was bounded by `AGENT_USAGE_WINDOW_DAYS`; the first was not bounded at all.

Now:

- a 24-hour `AGENT_ACTIVITY_RECOVERY_WINDOW_HOURS` bounds the recovery branch. Past it, a stale unresolved delivery is reduced to a distinct Agent id and projects as `unknown` — its row is never materialised in the bounded read;
- `AGENT_ACTIVITY_READ_LIMIT` (256) plus a deterministic `acceptedAt DESC, id DESC` order caps the activity read, so one pathological account cannot make a single `/agents` call unbounded;
- usage comes from **one grouped SQL query per Agent** — `count`, failed count, and numeric sums over the `turnReport` usage fields — instead of materialising one row per delivery and accumulating in JS.

## Why some arithmetic stayed in JS

`deliveryUsageTokenCounts` applies a provider-specific rule: Codex and Claude Code count cached input differently, so the raw components cannot be summed into a single token total in SQL without hard-coding one provider's semantics. SQL now computes the row count, the failed count, and each raw token component; JS calls the existing helper **once per aggregate row** rather than once per delivery. Provider semantics and both guard behaviours are unchanged, and the per-delivery materialisation is gone. This is recorded as a deliberate decision on the branch.

| # | Criterion | Result |
| --- | --- | --- |
| 1 | `pnpm --filter @opentag/server test` exits 0 with the three required cases | **PASS** — 68 test files, 793 tests |
| 2 | Parity: new aggregation returns identical `failed` / `tasks` / `tokens` to the previous JS aggregation on the same fixture | **PASS** — see note below |
| 3 | A stale unreported `accepted` delivery past the window is absent from the activity read and reports `unknown` | **PASS** — dedicated test |
| 4 | No file added under `packages/server/drizzle/`, `CURRENT_MIGRATION_COUNT` untouched | **PASS** — count remains 38 |
| 5 | `pnpm check` exits 0 | **PASS** |
| 6 | `pnpm typecheck` exits 0 | **PASS** |
| 7 | `pnpm test` exits 0 | **PASS** — 330 script tests, 8/8 Turbo tasks |

**On criterion 2**, the parity proof takes a different *form* than specified and a stronger one. Rather than adding a separate parity test, the branch **tightened the existing usage assertion** on the same fixture, from a nested `toMatchObject` to an exact `expect(...usage).toEqual({ windowDays: 30, tasks: 2, failed: 1, tokens })`. The fixture and its expected values are unchanged from `main`, so the assertion demonstrates exactly the required equivalence — and now fails on any extra or missing field, which the previous partial match would have tolerated.

The branch also added `listForAccount` assertions to both safe-integer guard tests, which is precisely the risk of moving arithmetic into SQL: the second commit, `fix(server): preserve aggregate token validation`, exists because that guard needed explicit preservation on the aggregate path. Both `invalid token count` and `Agent usage token total exceeds the safe integer range` still fire.

**On criterion 7**, `pnpm test` requires `GIT_CONFIG_GLOBAL=/dev/null` on the authoring host: an unrelated script test commits inside a throwaway Git repository and inherits the host's 1Password SSH signer, which fails there. The same test fails identically on unmodified `main`, so it is a host-environment condition, not a regression; CI has no 1Password and is expected to pass unmodified.

## Recorded follow-up (deliberately not done here)

`EXPLAIN (FORMAT TEXT)` of the bounded activity filter shows `Index Scan using im_message_deliveries_pending_idx` with `Index Cond: (state = 'accepted')`, a residual filter on `reported_at IS NULL` and the recovery `accepted_at` bound, then primary-key joins and a sort. A partial index on `(accepted_at DESC, id DESC) WHERE state = 'accepted' AND reported_at IS NULL` would remove the residual filter and the sort.

This branch deliberately does **not** add it: three sibling branches in the same batch already add a `0038_` migration each, and a fourth would collide for no benefit. Recorded here so the option is not lost.

## Deviations

None from the confirmed plan. The parity criterion was satisfied in a stronger form than specified, described above.

---

Produced by dispatch-codex run `d76ae8`, lane `agent-activity-bounds-6`. Written by codex running under bypassed approvals in an isolated git worktree; every acceptance criterion above was re-run and checked by the orchestrator before this pull request was opened.

### DIFF
--- a/packages/server/src/services/agents/agent-service.ts
+++ b/packages/server/src/services/agents/agent-service.ts
@@ -607,9 +607,15 @@ export class AgentService {
         outputTokens: sql<string>`coalesce(sum(case when ${outputTokensText} ~ '^[0-9]+$' then (${outputTokensText})::numeric else 0::numeric end), 0)::text`,
         tasks: sql<string>`count(*)::text`,
         invalidTokenCount: sql<boolean>`bool_or(
-          (${inputTokensText} is not null and ${inputTokensText} !~ '^[0-9]+$') or
-          (${cachedInputTokensText} is not null and ${cachedInputTokensText} !~ '^[0-9]+$') or
-          (${outputTokensText} is not null and ${outputTokensText} !~ '^[0-9]+$')
+          case when ${inputTokensText} is null then false
+            when ${inputTokensText} ~ '^[0-9]+$' then (${inputTokensText})::numeric > 9007199254740991
+            else true end or
+          case when ${cachedInputTokensText} is null then false
+            when ${cachedInputTokensText} ~ '^[0-9]+$' then (${cachedInputTokensText})::numeric > 9007199254740991
+            else true end or
+          case when ${outputTokensText} is null then false
+            when ${outputTokensText} ~ '^[0-9]+$' then (${outputTokensText})::numeric > 9007199254740991
+            else true end
         )`,
       })
       .from(imMessageDeliveries)
--- a/packages/server/src/services/agents/agent-service.ts
+++ b/packages/server/src/services/agents/agent-service.ts
@@ -16,11 +16,12 @@ import {
   type CreateAgentRuntimeConfig,
   hasRequiredFeishuTenantScopes,
   type ListAgentsResponse,
+  RUNTIME_MAX_DURATION_MS,
   runtimeUsageTotalTokens,
   type UpdateAgentRequest,
   UpdateAgentRequestSchema,
 } from "@opentag/shared";
-import { and, asc, desc, eq, gte, inArray, isNull, lte, ne, sql } from "drizzle-orm";
+import { and, asc, desc, eq, gte, inArray, isNull, lt, lte, ne, sql } from "drizzle-orm";
 import { alias } from "drizzle-orm/pg-core";
 import type { DatabaseClient, DatabaseTransaction } from "../../db/client.js";
 import {
@@ -41,9 +42,12 @@ type AgentRow = typeof agents.$inferSelect;
 type AgentRuntimeConfigRow = typeof agentRuntimeConfigs.$inferSelect;
 type QueryExecutor = Pick<DatabaseClient, "select">;
 
-export const AGENT_ACTIVITY_RECOVERY_WINDOW_HOURS = 24;
+const AGENT_ACTIVITY_REPORT_DELIVERY_SLACK_HOURS = 1;
+const MILLISECONDS_PER_HOUR = 60 * 60 * 1_000;
+export const AGENT_ACTIVITY_RECOVERY_WINDOW_HOURS =
+  RUNTIME_MAX_DURATION_MS / MILLISECONDS_PER_HOUR + AGENT_ACTIVITY_REPORT_DELIVERY_SLACK_HOURS;
 export const AGENT_ACTIVITY_READ_LIMIT = 256;
-const AGENT_ACTIVITY_RECOVERY_WINDOW_MS = AGENT_ACTIVITY_RECOVERY_WINDOW_HOURS * 60 * 60 * 1_000;
+const AGENT_ACTIVITY_RECOVERY_WINDOW_MS = AGENT_ACTIVITY_RECOVERY_WINDOW_HOURS * MILLISECONDS_PER_HOUR;
 
 interface AgentScope {
   agent: AgentRow;
@@ -106,9 +110,7 @@ function toAgentComputer(row: {
 interface AgentActivityEvidence {
   acceptedAt: Date | null;
   agentId: string;
-  bindingStatus: string;
   reportedAt: Date | null;
-  sessionEndedAt: Date | null;
   state: string;
 }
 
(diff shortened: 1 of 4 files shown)


packages/server/src/services/agents/agent-service.ts
```diff
@@ -541,32 +568,103 @@ export class AgentService {
...
-            gte(imMessageDeliveries.acceptedAt, usageStartedAt),
+          and(
+            eq(imMessageDeliveries.state, "accepted"),
+            isNull(imMessageDeliveries.reportedAt),
+            gte(imMessageDeliveries.acceptedAt, recoveryStartedAt),
+          ),
+        ),
+      )
+      .orderBy(desc(imMessageDeliveries.acceptedAt), desc(imMessageDeliveries.id))
+      .limit(AGENT_ACTIVITY_READ_LIMIT);
```
yuezengwu: [P2] Do not let one Agent's deliveries erase another Agent's working state

This limit applies across all Agents in the account. A busy Agent with 256 newer rows excludes another Agent's still-active delivery even when it was accepted only an hour ago. The excluded Agent is absent from `projectActivityByAgent` and is not stale, so the final fallback reports `idle`. A two-Agent PGlite regression reproduces this exactly (expected working, received idle). Bound the query after reducing the evidence per Agent, or explicitly handle truncation without asserting idle; the current one-Agent limit test does not cover this regression.

yuezengwu (changes requested):
Reviewed exact head `427780dbc02b08fd70b8ee0bd6b3a684ce64cc26`. The account-wide row cap can change a genuinely working Agent to idle.

A focused regression using the existing AgentService/PGlite fixture creates 256 recent unreported accepted deliveries for Agent A and one unreported accepted delivery from one hour ago for Agent B. Both have active bindings and live Sessions. `listForAccount` reports A as working and B as idle. B is inside the recovery window, so the stale-id query cannot repair the omitted evidence.

Please preserve per-Agent activity when bounding row materialisation, and add a multi-Agent saturation test. At minimum, truncation must not be treated as affirmative idle evidence. No author-branch changes were made.

bestony:
Pushed `1df812b3` addressing the review.

**Per-Agent bound (the reported defect).** The activity read no longer applies an account-wide row cap. It now uses `selectDistinctOn([imBindings.agentId], …)` so each Agent contributes its own newest qualifying row, with `imBindings.status = 'active'` and `sessions.endedAt IS NULL` pushed into the SQL WHERE clause. Your exact scenario is now a regression test — `keeps every agent visible when another agent exceeds the activity limit` — with one Agent saturating the limit and a second Agent holding a single hour-old unreported `accepted` delivery; the second still reports `working`.

**List/detail divergence.** `listForAccount` and `getById` now share one `#readActivityByAgent` helper rather than carrying two copies of the query, so the two paths cannot disagree by construction. A test asserts `listed.agents[0].activity` equals `getById(...).activity` for a stale unreported delivery.

**Recovery window.** It is now tied to the runtime's maximum turn duration rather than a bare 24 h, with a test proving a delivery aged to that maximum still reports `working` instead of being written off as stale.

Verified independently before pushing: `pnpm --filter @opentag/server test` 68 files / 794 tests, `pnpm check` and `pnpm typecheck` all exit 0. No drizzle migration was added and `CURRENT_MIGRATION_COUNT` remains 38.

Note that this new commit supersedes the head you reviewed, so the PR needs a fresh look.

yuezengwu (approved):
Independently re-reviewed the full change at `1df812b346f05d53092a9bed84e63544580028bd`; the prior head's tests and review conclusion were not reused.

The previous cross-Agent starvation finding is resolved on this head: the query selects recent active evidence per Agent, with active-binding/live-session filters, and uses the shared runtime-duration limit plus delivery slack. I reran the original PGlite regression with 256 recent deliveries for Agent A and an older active delivery for Agent B; B remains working. The complete AgentService suite plus that regression passed (18 tests). Shared build, `pnpm check`, server typecheck, and `git diff --check` also passed.

No blocking finding remains in this review. This approval does not waive any remaining repository checks or authorize a merge.

PR #543, opened 2026-09-05
fix(runtime): bound durable work with quotas, pagination and monotonic writes

bestony:
> **MIGRATION: REGENERATE, DO NOT RENAME, IF ANOTHER MIGRATION MERGES FIRST.** This branch adds `0038_melted_katie_power.sql` and bumps `CURRENT_MIGRATION_COUNT`. Sibling branch #544 also claims a `0038_`. If it lands first, **do not rename this migration** — a renamed file keeps its snapshot's `prevId`, and two snapshots sharing a `prevId` make `drizzle-kit generate` abort permanently *with exit code 0*, so no gate catches it. Instead `git rm` the `0038_*` SQL and its `meta/*_snapshot.json`, revert the `_journal.json` and `migration-hashes.json` hunks to the merged `main`, rebase, and re-run `pnpm --filter @opentag/server db:generate`.

Adds durable-work quotas, retention and pagination on the Server store, and makes the Client's Turn Report settlement safe against concurrent writers — from both directions.

Covers the P1 item from `todo.zh-CN.md`: *Durable work 配额与分页*.

## Checklist

- [x] Per-Computer row and byte quotas with clamping, plus retention of retained payloads
- [x] Cursor pagination over durable work records
- [x] Declared transition table plus a 25-edge fixture as the contract
- [x] Turn Report settlement is safe under concurrent writers
- [x] Retry timing: the acknowledgement window is honoured on every attempt, not just the first
- [x] `dead-letter -> running` gated at the sender rather than admitted to the table
- [x] Hydrated pendings cannot raise an unhandled rejection
- [x] Committed regression tests protect the settlement ordering **in both directions**

## Round 9 — the regression this branch introduced, found by @yuezengwu at `e3e963b0`

Round 7 moved `#detach(pending)` to *before* the succeeded write. That was the right fix for the lost update, and it stays. But `#persist` publishes `#records` only *after* `write` returns, so it opened a window on the other side where the owner had no record of the turn anywhere a caller could see: `#pending` no longer held it, and `#records` still said `running`.

A concurrent `submit` in that window consulted exactly those two, missed both, and built a **successor from stale state** — persisting the stale record and sending the report a second time. The reviewer reproduced it with the real `FileRuntimeDurabilityStore`, delaying only the succeeded write's acknowledgement: `running -> succeeded -> running`, two sends, and finally **disk `running` while the mirror said `succeeded`**. On this path the branch was *worse than `main`*, which retains `succeeded` and sends once.

**The two defects pull in opposite directions**, which is what made this round delicate: round 6 needed the identity gone before the terminal write so no other writer could move the row off `succeeded`; this one needs it present until the write publishes so `submit` cannot build a successor. Moving `#detach` back would simply have traded the defects again.

The fix keeps the pending in `#pending` and fences it with an explicit `settling` flag from the moment confirmation succeeds until the succeeded record is visible in `#records`:

- `#reusePending` returns `existing.promise` immediately for a settling pending — no re-send, no `resendRequested`, no re-persist, no `confirm` reassignment.
- `#scheduleRetry`, `#canWrite` and `rearmTerminal` all refuse to act on a settling pending, which is what preserves round 7's guarantee; only the succeeded transition passes `{ settling: true }` to bypass it.
- `#releaseSettling` runs only after `#transition` has published the durable row, and `#settle` still detaches, so `#pending` does not leak against `maxPending`.

| # | Criterion | Result |
| --- | --- | --- |
| 1 | Concurrent `submit` during in-flight succeeded persistence: one send, final `succeeded`, no `succeeded -> running`, mirror agrees with disk, duplicate promise resolves | **PASS** — `reuses a settling report while its succeeded write is committed`, gated after the succeeded write commits but before its promise returns |
| 2 | The round-8 lost-update test still fails when the fence is raised too late | **PASS** — mutation run by the orchestrator; see below |
| 3 | The new protection has teeth | **PASS** — mutation run by the orchestrator; see below |
| 4 | Other behaviour unchanged (retry window, `dead-letter -> running` unreachable, hydrated pendings, transition table) | **PASS** — verified from the diff and the full client suite |
| 5 | All three `.dispatch/repro/` specs pass | **PASS** — 3 files, 3 tests |
| 6 | `pnpm --filter @opentag/client test` exits 0 | **PASS** — 72 files, 982 tests |
| 7 | `pnpm --filter @opentag/client test:agent-runtime:coverage` exits 0 | **PASS** — the 100%-branch gate, 359 tests |
| 8 | `pnpm --filter @opentag/server test` exits 0 | **PASS** — 69 files, 832 tests |
| 9 | `pnpm check` exits 0 | **PASS** |
| 10 | `pnpm typecheck` exits 0 | **PASS** |
| 11 | No new file under `packages/server/drizzle/` | **PASS** — this round's diff touches none |

### The mutations, run in opposite directions on purpose

A fix that closes one defect by reopening the other would pass a one-sided criterion. So both directions were mutated by hand:

- **Fence raised only *after* the succeeded write** (the round-6 shape) → `does not overwrite the committed succeeded row when acknowledgement overlaps a retry` **fails** with `succeeded -> retryable`.
- **Fence released *before* the succeeded write publishes** (the round-9 defect) → `reuses a settling report while its succeeded write is committed` **fails**.

Both detectors therefore ship, and neither is satisfied by the other's fix. The file was restored after each run; the tree is clean at `7a894a8a` and no mutation was committed.

## Documented, not fixed — with the reason

- **`SessionMessageInbox` has no identity or in-flight guard** (`session-message-inbox.ts:173`): a redelivered Session message can emit `running -> accepted` and `succeeded -> running`. Separate module, separate review surface.
- **Far-future row repair can be rolled back** by the stale-write error it causes (`runtime-durable-work-store.ts:322`).
- **The per-Computer byte quota re-serializes every retained payload on each insert**, under the per-Computer write lock (`:265`).

The last two matter less than they look: **production never wires the Server store.** `client-runtime-composition.ts:584-587` falls back to `FileRuntimeDurabilityStore` because `serverDurability` is passed nowhere in `apps/cli/src` or `packages/client/src`, and the path that *would* exercise the server transition table is broken before the wire by a disjoint failure vocabulary (`runtime-domain.ts:76` vs `structured-errors.ts:39`). Both pre-date this branch. That is exactly why the client-side settlement defects were the urgent ones: `turn-report-owner.ts` **is** on the path production takes.

## Commit granularity

21 commits across nine rounds. An early round landed coarsely; later rounds were split into coherent units, and round 9 is a single focused commit. History was deliberately not rewritten.

---

Produced by dispatch-codex run `d76ae8`, lane `durable-work-quotas-4`. Written by codex running under bypassed approvals in an isolated git worktree; every acceptance criterion above was re-run and checked by the orchestrator before this pull request was updated.


packages/server/src/runtime/runtime-durable-work-store.ts
```diff
@@ -6,17 +6,49 @@ import {
...
+/** Default page size for durable-work list requests. The response schema permits at most 1024 records. */
+export const DEFAULT_RUNTIME_DURABLE_WORK_PAGE_SIZE = 256;
+export const RUNTIME_DURABLE_WORK_MAX_PAGE_SIZE = 1024;
+
+export const RUNTIME_DURABLE_WORK_ALLOWED_TRANSITIONS = {
+  accepted: ["accepted", "running", "dead-letter"],
+  running: ["running", "succeeded", "failed", "retryable", "dead-letter"],
+  succeeded: ["succeeded"],
+  retryable: ["retryable", "accepted", "running", "dead-letter"],
+  failed: ["failed"],
```
yuezengwu: [P2] Preserve the exact-claim terminal-report rearm path

Treating every failed record as absorbing breaks `TurnReportOwner.rearmTerminal()`. The client deliberately hydrates retained reports whose `lastError.code` is `conflict` or `stale_generation`, validates an exact reconciliation claim, then `submit()` calls `#send()`, which persists `failed -> running`. This table now rejects that write with 409 before sending the report. With a schema-valid retained stale-generation record and strictly increasing timestamps, an A/B test using the real client and PGlite sent one report with the baseline store and zero with this head. Preserve that explicitly authorized recovery transition (with the appropriate authority check), and cover it through the actual persistence adapter; the ordinary retryable-to-accepted test does not exercise it.

yuezengwu (changes requested):
Independently reviewed `222f1835dbe1207c2a608732b006398a4a5d4d13` and found a regression in the existing terminal-report recovery path (inline).

The reproduction uses the actual `TurnReportOwner`, `ServerRuntimeDurabilityStore`, and database-backed store with PGlite. A schema-valid retained `failed` report with `stale_generation` hydrates successfully, and an exact matching rearm claim is accepted by the client. Resubmission sends one report with the baseline store and zero with this head: the new transition table rejects `failed -> running`. The seed is a valid retained record, avoiding an unrelated pre-existing error-taxonomy mismatch when creating a fresh failure.

Existing validation passes: 11 server store/API tests, 14 client tests, six PostgreSQL integration tests for durable work, shared build, `pnpm check`, server typecheck, and `git diff --check`. These tests do not cover the recovery regression. Please preserve the explicitly authorized exact-claim rearm path and add a client-to-persistence regression without generally reopening terminal records.

bestony:
Confirmed independently before dispatching a fix — this is a real defect, not a false positive.

Traced in this checkout:

- `turn-report-owner.ts` `#hydrate()` deliberately re-adds a stored `failed` record whose `lastError.code` is `conflict` or `stale_generation` as a hydrated pending, leaving its in-memory status at `failed`.
- `rearmTerminal()` validates an exact claim via `reportMatchesRearmClaim` and clears `pending.serverStatus`.
- `#send()` returns early while `serverStatus` is set, so it is reachable only after that clear — and its first act is `#transition(pending, "running")`, i.e. `failed -> running`.
- Sole caller: `mvp-turn-report-recovery.ts:207`.

With `failed: ["failed"]` the write is rejected with 409, and `#send`'s `.catch(... "transport" ...)` swallows it, so the recovered report is silently never delivered. That matches your one-vs-zero A/B result exactly.

Fix in progress: `failed` will admit `running` while `succeeded` and `dead-letter` stay strictly absorbing (no client path reopens either — `session-message-inbox.ts:170` and `:426` treat both as absorbing), with the `updatedAt` CAS still guarding the race. Adding the client-to-persistence regression through the real `TurnReportOwner`, plus a unit test so the changed lines clear the patch-coverage gate.

Since this is the second missing edge review has found here (the first was `retryable -> accepted`), the branch will also carry an enumerated `from -> to` table derived from every `#transition` call site in `turn-report-owner.ts`, `session-message-inbox.ts` and `mvp-turn-report-recovery.ts`, so a third gap is not left to chance.

packages/client/src/runtime/turn-report-owner.ts
```diff
@@ -226,34 +232,52 @@ export class TurnReportOwner {
...
-      if (this.#pending.get(result.turnId) !== pending) return true;
-      this.#pending.delete(result.turnId);
-      this.#clearRetry(pending);
-      await this.#transition(pending, "succeeded", { nextAttemptAt: undefined });
-      pending.resolve();
+      await confirm();
+      pending.confirming = false;
+      if (!this.#detach(pending)) return true;
+      try {
+        await this.#transition(pending, "succeeded", { nextAttemptAt: undefined }, { detached: true });
```
yuezengwu: [P2] Keep the report owned until the succeeded receipt is published

`#detach` removes the pending identity before the awaited success transition, while the new `#persist` publishes `#records` only after `write` returns. A concurrent `submit` of the same report therefore sees no pending owner and the old `running` mirror, creates a new pending entry, persists that stale record and sends again. With the real FileRuntimeDurabilityStore and only the succeeded write acknowledgement delayed, I observed `running -> succeeded -> running`, two sends, and finally disk=`running` but mirror=`succeeded`; the same test with main's TurnReportOwner retains `succeeded` and sends once. Retain a settling identity/fence across terminal persistence so duplicate submissions reuse/wait for that settlement and cannot create a successor from stale state; include a concurrent-submit regression in addition to the retry-timer test.

yuezengwu (changes requested):
Fresh exact-head review of `e3e963b0959e563bdff9fb4b39d978e61b05082d`, including the full PR diff against its merge-base with current main.

The previous retained `failed/stale_generation` re-arm blocker is fixed. I reran the earlier schema-valid retained-record reproduction with the actual current TurnReportOwner, ServerRuntimeDurabilityStore and PostgreSQL durable-work store over PGlite; exact-claim re-arm now writes `running` and sends the report.

There is one new P2 regression in the expanded Client settlement changes: a second `submit` of the same report while its successful receipt write is awaiting acknowledgement can replace the committed success with `running`. The inline finding gives the sequence and fix boundary. An isolated A/B component test used the real FileRuntimeDurabilityStore, with a gate after the succeeded write committed but before its promise returned. Baseline main's TurnReportOwner kept disk/mirror `succeeded`, pending=0, sends=1. Current head produced `succeeded -> running`, pending=1, sends=2; after releasing the first acknowledgement, disk was still `running` while the mirror said `succeeded`. This is a persistence/owner component reproduction, not a full runtime E2E claim.

Fresh validation: all 129 affected committed tests passed (58 Server unit, 59 Client, 12 PostgreSQL integration including migrations), plus the independent retained-failure re-arm regression passed. The independent concurrent-submit test passes on baseline and fails on this head. Shared build, repository check, Server/Client typechecks and `git diff --check` passed. Review-only test files were removed from the clean detached worktree after saving their evidence.

Requesting changes for the confirmed success-settlement regression. No author branch changes or merge were performed.

yuezengwu (approved):
Technical approval for exact head `7a894a8a7ed723fc61b0e3430e9459402a293841`, reviewed against current main `1ac249bb75ae8efef6f87ca921576bae2c0f5c8d`.

The concurrent-submit blocker from my previous review is fixed. I reran the independent reproduction with the **actual FileRuntimeDurabilityStore**, delaying only acknowledgement after its succeeded write committed. During that window the existing pending identity remains fenced, a duplicate submission reuses its promise, the report is sent once, and there is no `succeeded -> running` write. After releasing acknowledgement, both disk and mirror are `succeeded`, the duplicate promise resolves without invoking its replacement confirmation callback, and pendingCount returns to zero. The same test also passes against main's owner. The committed acknowledgement/retry overlap regression passes too, so this fix has not reopened the previously tested lost-update path.

The earlier retained `failed/stale_generation` exact-claim re-arm reproduction also passes again with the actual current Client owner, ServerRuntimeDurabilityStore and PostgreSQL store over PGlite.

Fresh validation: **118 unique affected committed tests passed across 9 suites** — 58 Server tests including PostgreSQL integration, and 60 Client tests — plus 3 independent regression cases (current and baseline concurrent submit, and retained-failure re-arm). Shared build, repository check, Server/Client typechecks and `git diff --check` passed. The 39-entry migration/snapshot chain is consistent; this update does not change migrations. Review-only files were removed from the clean detached worktree after saving evidence.

I found no additional introduced blocker in the full PR diff. This approval covers the reviewed implementation and the reproduced fixes; it does not establish full production use of the Server durability adapter or resolve the separately documented SessionMessageInbox concurrency, future-clock repair rollback and per-write payload accounting limitations. Quota/retention policy and migration deployment remain the project owner's decisions. If #544 lands first, regenerate the colliding migration against the new main and revalidate.

Current-head ordinary CI is green; ownership-gate is still failing at review time. No author branch changes or merge were performed.

PR #544, opened 2026-09-05
fix(im): harden delivery durability with indexed retention sweeps

bestony:
> **MIGRATION COLLISION RESOLVED — this branch has been regenerated against the new `main`.** #543 merged first, so `main` now owns `0038_melted_katie_power.sql`. This branch's own `0038`/`0039`/`0040`/`0041` were **deleted and regenerated**, not renamed — a renamed migration keeps its snapshot's `prevId`, and two snapshots sharing a `prevId` make `drizzle-kit generate` abort permanently *with exit code 0*, which no gate catches. `main`'s `0038` is now the only `0038` here and is byte-identical to `main`'s copy; regeneration collapsed the four into a single **`0039_clean_squirrel_girl.sql`**, journal is **40** entries and `CURRENT_MIGRATION_COUNT` is **40**. History stayed append-only — `main` was merged in (`ffc587ba`), not rebased, so nothing was force-pushed and `bb84d9f0` is still an ancestor.
>
> **Still outstanding for whoever merges next:** sibling #540 also carries a `0038` and is currently CONFLICTING. It has to go through the same regenerate-don't-rename procedure, and it should be done *after* this PR lands — otherwise it would regenerate to `0039` as well and collide with this branch.

Makes IM delivery durable: bounded worker deadlines, a retention janitor, and expiry/retention indexes that stay batch-bounded regardless of how expiry timestamps are distributed.

Covers the P1 item from `todo.zh-CN.md`: *IM 投递持久化与重试*.

## Checklist

- [x] Permit/cancellation handling and bounded worker deadlines
- [x] Retention janitor module on a 60 s cadence
- [x] Expiry and retention indexes (`0039`, `0040`)
- [x] Scheduler-routing coverage
- [x] Documented relational retention bounds
- [x] Post-timeout path releases its scheduler slot within a bounded time
- [x] Expiry index ordered by `(expires_at, id)` so the pass stays batch-bounded at equal expiry

## Round 4 — the two findings @yuezengwu reproduced at `ded2baae`

**P1 — the post-timeout await pinned worker capacity.** `withOperationDeadline` recorded the timeout and then awaited the *original* operation with no second bound. An `AbortSignal` does not settle an arbitrary promise, so a stalled database or assembler call meant `KeyedTaskScheduler` never released its slot — head-of-line blocking across unrelated agents, not just a slow delivery. The previous implementation returned after recording the timeout; the current shape had traded a bounded return for an unbounded one.

Fixed with a bounded grace window (`DEFAULT_OPERATION_TIMEOUT_GRACE_MS = 1000`): after the deadline fires, the operation races a grace timer, and if the grace expires first the slot is released and `IM_DELIVERY_OPERATION_ABANDONED` is emitted. A late settlement still reports `onLateSettle` without double-resolving. Observability was explicitly *not* accepted as the bound — the committed test drives a promise that never settles at all.

**P2 — the expiry index was missing the ordering tie-breaker.** The expiry CTE orders by `(expires_at, id)` but `im_message_deliveries_expiry_idx` supplied only `expires_at`, so a tie group had to be read whole and sorted before `LIMIT 100` applied. `0041_friendly_starfox.sql` recreates the index on `(expires_at, id)`, keeping its partial predicate (`state = 'pending' and reason is null`) — matching `im_message_deliveries_retention_idx`, which round 2 already got right.

## A criterion of mine that was satisfiable without the property holding

Round 3's criterion demanded `EXPLAIN` at two different **table sizes**. That was the wrong axis: a tie-group pathology depends on the *distribution* of `expires_at`, not the row count, so two sizes with distinct timestamps could never expose it. The criterion was restated over the distribution — *bounded regardless of how `expires_at` values are distributed, including the degenerate case where every eligible row shares one timestamp* — and that is what the retained test now asserts.

| # | Criterion | Result |
| --- | --- | --- |
| 1 | Never-settling dependency: slot released within a bounded time and an unrelated agent proceeds | **PASS** — `maxConcurrent=1`, `operationTimeoutMs=100`, first delivery returns a never-settling promise; second task still blocked at 100 ms, proceeds after the 1 s grace, two deliveries dispatched, `IM_DELIVERY_OPERATION_ABANDONED` emitted |
| 2 | Expiry pass bounded by batch size regardless of `expires_at` distribution | **PASS** — 1,000 rows sharing one `expires_at`; plan shows `Index Scan using im_message_deliveries_expiry_idx`, no `Sort`, `rows=100` |
| 3 | Equal-expiry EXPLAIN regression retained in the repo | **PASS** — committed in `im-delivery-worker.test.ts` |
| 4 | New index in a new `0041_*`; `0038`/`0039`/`0040` untouched; `CURRENT_MIGRATION_COUNT` = 42 | **PASS** — this round's diff touches only `0041`, `_journal.json`, `migration-hashes.json` |
| 5 | `pnpm check` exits 0 | **PASS** — includes the migration drift check |
| 6 | `pnpm --filter @opentag/server test` exits 0 | **PASS** — 68 files, 796 tests |
| 7 | `pnpm --filter @opentag/server test:integration` exits 0 | **PASS** — 18 files, 314 tests |
| 8 | `pnpm typecheck` exits 0 | **PASS** — 8/8 |
| 9 | Docs keep their `docs/zh-CN/` mirror in sync | **PASS, vacuously** — this round changed no docs |

**Both protections were mutation-tested**, because a criterion passing does not prove the detector ships:

- Reducing the index to `(expires_at)` makes the equal-expiry test fail — the plan gains a `Sort` node.
- Restoring the unbounded `await operation.catch(...)` makes the never-settles test fail — it times out at 5 s.

Both files were restored immediately; the tree is clean at `bb84d9f0` and neither mutation was committed.

## Stated precisely, so it can be weighed

- The EXPLAIN assertions run on **PGlite** — the genuine PostgreSQL engine compiled to WebAssembly — not on a server PostgreSQL 17 instance. The reviewer's own millisecond measurements came from PostgreSQL 17; the committed test asserts plan *shape* (Index Scan, no Sort, rows examined), which is the property that matters, rather than timings.
- The **claim** query's plan is recorded as a deliberate no-change result: in that workload nearly every row was eligible, so adding a claim index would not have been evidence-led.
- `.dispatch/progress.md` was not rewritten for round 4 — its tail still describes round 1. Every criterion above was therefore re-run by the orchestrator directly rather than read from the lane's notes.

## Commit granularity

Round 4 landed as two coherent commits, one per finding — `d295ddf2 fix(db): order expiry index ties` and `bb84d9f0 fix(im): bound timed-out delivery cleanup`. Earlier rounds were coarser; round 1 in particular landed as a single large commit. History was deliberately not rewritten.

---

Produced by dispatch-codex run `d76ae8`, lane `im-delivery-durability-3`. Written by codex running under bypassed approvals in an isolated git worktree; every acceptance criterion above was re-run and checked by the orchestrator before this pull request was updated.


packages/server/src/runtime/im-delivery-deadline.ts
```diff
@@ -0,0 +1,47 @@
...
+  onTimeout: () => Promise<void>,
+  onLateSettle: () => void,
+): Promise<void> {
+  let timeoutFailure: unknown;
+  try {
+    await onTimeout();
+  } catch (error) {
+    timeoutFailure = error;
+  }
+  await settleAfterTimeout(operation, onLateSettle);
```
yuezengwu: [P1] Bound the post-timeout wait before it pins worker capacity

After the timeout is recorded, this awaits the original operation without a second bound. AbortSignal does not settle an arbitrary promise: the worker's database/assembler calls and recovery path are not cancelled through this signal. If one of them stalls, `withOperationDeadline` never returns and `KeyedTaskScheduler` never releases that active slot; enough stalled calls stop deliveries for unrelated agents too. With the real helper/scheduler, `maxConcurrent=1` and a 50 ms deadline, timeout and abort both fired, but at 300 ms the first task was still active and agent B remained queued; both completed only after explicitly releasing the dependency. The previous deadline returned after recording the timeout. Please provide bounded cancellation/grace with fencing or a bounded fail-stop path, and test a dependency that never settles; merely emitting timeout/late-settle metrics does not provide that bound.

packages/server/src/db/schema/im-messages.ts
```diff
@@ -114,6 +115,16 @@ export const imMessageDeliveries = pgTable(
     index("im_message_deliveries_session_id_idx").on(table.sessionId),
     index("im_message_deliveries_steer_target_idx").on(table.steerTargetDeliveryId),
     index("im_message_deliveries_pending_idx").on(table.state, table.nextAttemptAt),
+    index("im_message_deliveries_expiry_idx")
+      .on(table.expiresAt)
```
yuezengwu: [P2] Include the tie-breaker in the expiry index

The expiry CTE orders by `(expires_at, id)`, but this index only supplies `expires_at`. A backlog with many equal expiry timestamps therefore requires reading/sorting the entire tie group before `LIMIT 100` can be applied. Executing the actual `runImDeliveryExpiry` statement on PostgreSQL 17 with this migration's index and 50k/200k equal-expiry rows produced sequential scans of all 50k/200k rows plus Sort (the 200k sort spilled), taking 11.2/47.1 ms. A comparison `(expires_at, id)` partial index served both runs with an Index Scan of 100 rows, about 0.23 ms. This contradicts the claimed batch-bounded expiry pass and repeats every expiry tick for a large backlog. Please align the index with both ordering columns, regenerate the matching migration metadata, and retain an equal-expiry EXPLAIN regression case. The fixture uses a reduced table, so the timings are illustrative; the missing ordering column is the defect.

yuezengwu (changes requested):
Reviewed the full base-to-head diff at `ded2baaef42317b529dbea0760c5ac7238e793ad`. Two reproducible issues remain: the operation deadline can retain worker capacity indefinitely, and the expiry index does not supply the full sweep ordering.

Validation: 127 tests passed across the delivery-worker, keyed scheduler, PostgreSQL IM-binding, and migration-backfill suites; shared build, repository checks, server typecheck and diff checks passed. The 41-entry migration journal and snapshot predecessor chain are consistent. The deadline reproduction uses the actual helper and scheduler with a synthetic unsettled dependency; the expiry reproduction executes the actual janitor query on a reduced PostgreSQL 17 table at 50k and 200k rows. These are component/query validations, not production-load measurements.

The separately documented relational retention-scan bounds are acknowledged; the expiry finding concerns another scan boundary. No merge or deployment was performed.

yuezengwu (approved):
Technical approval for exact head `bb84d9f083ca9483e0d07c5b50c6a8c3f5ec1300` against current main `1ac249bb75ae8efef6f87ca921576bae2c0f5c8d`. I reviewed the complete PR diff again and independently rechecked both previous findings.

- **Post-timeout permit retention is fixed for the reported case.** The actual deadline helper plus KeyedTaskScheduler, with maxConcurrent=1 and an operation that remains unresolved, now records the timeout, aborts the signal, emits abandonment and lets unrelated agent B run after the 1 s grace. With a 50 ms deadline, by 1500 ms both scheduler slots were released while the original dependency was still unresolved. Releasing it afterward emitted late-settlement observation without reclaiming capacity. The committed worker-level never-settles regression also passed.
- **Equal-expiry index scanning is fixed.** I reran the actual expiry SQL and migrations 0038/0041 on PostgreSQL 17, using a reduced table with 50,000 and 200,000 eligible rows all sharing one expiry timestamp. Both plans used `im_message_deliveries_expiry_idx`, read exactly 100 candidate rows and had no Sort; measured execution was 0.404 ms and 0.522 ms respectively. These timings describe these fixtures, not a production latency guarantee.

Fresh validation: **129 tests passed across 4 suites**, covering the delivery worker, keyed scheduler, PostgreSQL IM binding integration and migration/backfill integration. Shared build, repository check, Server typecheck and `git diff --check` passed. The 42-entry migration chain is consistent, and prior migration SQL 0038–0040 is unchanged. The detached worktree is clean; review containers were stopped.

The deadline result is scoped to timeout bookkeeping completing: the grace begins after `onTimeout` returns, so it is not a wall-clock bound for a stalled failure-recording database call; main already has that awaited recording step. I found no additional introduced blocker in this review. Retention policy and migration deployment remain the project owner's decisions, including regenerating colliding migrations if #543 lands first.

Current-head CI Checks is still in progress and ownership-gate is failing at review time. This technical approval does not waive CI or authorize a merge; no author branch changes or merge were performed.

PR #545, opened 2026-09-06
fix: integrate Context Tree without project instruction writes

ggordonhall:
## Summary

Context Tree create/connect no longer modifies project AGENTS.md. Upgrade both exact Context Tree pins to 0.1.10 and add explicit `--json` to parsed connect, verify, list, resolve, and read calls and portable probes. OpenTag continues delivering memory instructions through its managed prompt. Update integration documentation and regression tests to verify connection resolution without project instruction files.

Uses the published Context Tree 0.1.10 release containing https://github.com/first-tree-ai/context-tree/pull/15. Both exact pins remain aligned. The lockfile now uses npm's published artifact integrity, and a frozen install resolves both consumers to that release.

## Testing

- Passed after release: frozen lockfile installation against npm 0.1.10 and `pnpm check`.
- Passed: `pnpm check`, `pnpm build`, `pnpm typecheck`, and all configured pre-push gates.
- Passed: 39 affected client tests, 65 affected CLI tests, and 39 portable build/runtime-dependency tests, including assembly and execution of the published npm 0.1.10 package. All 143 affected tests were rerun against the released artifact.
- Passed: `pnpm --filter @opentag/server test:integration` (320 tests).
- `pnpm test` reached unrelated failures in `daemon-ensure-service.test.ts:98` and `daemon-service-commands.test.ts:57`: the Commander wrappers produced exit code 0 where the tests expect 3. Both failures reproduced in an isolated rerun. The full-suite run is not green.
- Passed on rerun: `pnpm --filter @opentag/client test:agent-runtime:coverage` (359 tests; 100% statements, branches, functions, and lines). An initial Claude readiness assertion failed under concurrent load and passed in isolation and on the complete rerun.

## Breaking changes

Context Tree create/connect results no longer contain `pointer`. Existing instruction content, including old pointer blocks, is preserved. No cleanup migration is performed.

## Checklist

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -51,7 +51,7 @@
     "test": "vitest run --maxWorkers=1"
   },
   "dependencies": {
-    "@first-tree-ai/context-tree": "0.1.8"
+    "@first-tree-ai/context-tree": "0.1.10"
   },
   "devDependencies": {
     "@opentag/client": "workspace:*",
--- a/apps/cli/src/__tests__/context-tree.test.ts
+++ b/apps/cli/src/__tests__/context-tree.test.ts
@@ -33,6 +33,7 @@ async function fakeCli(responses: Readonly<Record<string, unknown>>): Promise<Co
   await writeFile(
     cliPath,
     `const table = ${JSON.stringify(responses)};
+if (!process.argv.includes("--json")) { process.stdout.write("Human-readable text"); process.exit(0); }
 const entry = table[process.argv[2]];
 if (entry === undefined) { process.stdout.write(JSON.stringify({ error: { code: "UNKNOWN", message: "" }, ok: false })); process.exit(1); }
 process.stdout.write(JSON.stringify(entry.payload));
--- a/apps/cli/src/core/context-tree/connect.ts
+++ b/apps/cli/src/core/context-tree/connect.ts
@@ -48,12 +48,12 @@ async function validateTarget(
 ): Promise<{ message: string } | undefined> {
   if (target.kind === "github") return undefined;
   if (target.kind === "path") {
-    const { failureCode } = await runContextTreeCli(assets, ["verify", "--tree-path", target.path]);
+    const { failureCode } = await runContextTreeCli(assets, ["verify", "--tree-path", target.path, "--json"]);
     return failureCode === undefined
       ? undefined
       : { message: `${target.path} is not a usable Context Tree (${failureCode}).` };
   }
-  const { payload, failureCode } = await runContextTreeCli(assets, ["list"]);
+  const { payload, failureCode } = await runContextTreeCli(assets, ["list", "--json"]);
   if (failureCode !== undefined) return { message: `Could not list managed Context Trees (${failureCode}).` };
   const trees = (payload as { trees?: readonly { name?: unknown }[] }).trees ?? [];
   return trees.some((entry) => entry.name === target.name)
--- a/apps/cli/src/core/context-tree/state.ts
+++ b/apps/cli/src/core/context-tree/state.ts
@@ -34,11 +34,11 @@ export async function readContextTreeState(deps: ContextTreeCommandDeps = {}): P
   if (!assets) return { configPath, target, tree: "unknown", detail: "the Context Tree package is missing" };
 
   if (configured.kind === "path") {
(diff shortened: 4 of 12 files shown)


yuezengwu (approved):
Reviewed the full change at `42b687fc60cb4a62bc35a8ccbfc4c3feb6f86bf3`. No blocking findings.

Both consumers resolve the published Context Tree 0.1.10 artifact, whose npm integrity matches the lockfile. Independent isolated-account probes confirmed that create/connect preserve existing AGENTS.md bytes (including legacy content), leave empty projects without AGENTS.md/CLAUDE.md, preserve existing CLAUDE.md, and handle an AGENTS.md symlink without modifying its target. The documented missing-CLAUDE.md symlink behavior remains intact. Parsed calls use JSON output, while the install command retains its JSON contract.

Validation: 111 local tests passed (39 client Context Tree, 18 CLI Context Tree, 39 portable, and 15 daemon command tests), plus seven real-package fixture probes. Full build, check, typecheck, and diff check passed. The two previously reported daemon command failures pass in this isolated run and the current CI; their earlier cause has not been established.

CI run 34041401612 attempt 1 is green. Its checkout log confirms merge commit `0cf7d86ff29024a0bf82c8b6083e9eec03c155f4`, combining this head with current main `b7ed979ec3a52d8fa78ff20bcfa4faaf91afa010`; its tree matches the independently computed conflict-free merge. Approval is for this exact head.

liuchao-001 (approved):
Reviewed the full change at `42b687fc60cb4a62bc35a8ccbfc4c3feb6f86bf3`. No blocking findings.

I verified the load-bearing assumptions against the published `@first-tree-ai/context-tree@0.1.10` artifact itself, installed standalone under an isolated `HOME`, rather than only against the tests:

- `TEXT_DEFAULT_COMMANDS` in the shipped bundle is exactly `create, connect, list, resolve, publish, read, verify`. Every one of those that OpenTag parses now passes `--json`, and I confirmed the un-flagged form really does emit prose (`verify` without `--json` prints `Context Tree INVALID.` on stdout), so `runContextTreeCli` would have degraded a real failure code into `CLI_FAILED` without this change.
- `install` is deliberately not in that set: both `install --host claude --project <cwd>` and `install --host codex` still emit their JSON payload with no flag, including the `skipped[].reason` shape `codexInstallSkipReason` depends on. Leaving those two calls unflagged is correct, not an omission.
- Error envelopes survive with `--json`: `connect no-such-tree --json` returns `{"error":{"code":...},"ok":false}` and `verify --json` returns the `ok:false` + `findings` shape, which are the two forms `contextTreeFailureCode` honours.
- `connect --json` returns `tree.path` and `list --json` returns `trees[].name` / `trees[].tree.path`, matching what `ContextTreeManager`, `apps/cli` state/connect, and the portable probe read.
- Project instruction behaviour matches the rewritten documentation exactly: `create` in an empty project leaves no `AGENTS.md` or `CLAUDE.md`; `connect` into a project whose `AGENTS.md` contains a legacy pointer block leaves the file byte-identical and only adds the best-effort `CLAUDE.md → AGENTS.md` symlink.

I also swept for parsed call sites the diff might have missed. `runContextTreeCli` has three product callers plus the manager, all covered; `sync`, `prepare-write`, and `finish-write` are JSON-only commands used from the packaged skills, not from OpenTag; both exact pins are the only two in the tree and move together. The compensating channel for the removed pointer is real — `renderContextTree` in `managed-instructions.ts` still states the tree path and the read/write skill guidance — so nothing silently loses the tree location. The new fake-CLI guard in `apps/cli` (text unless `--json`) is a good regression fence for exactly this class of bug.

Two non-blocking nits, worth folding in whenever this file is next touched:

- `apps/cli/src/core/context-tree/connect.ts:41` and `docs/design/context-tree-integration.md:85` still justify never connecting a throwaway directory partly because it would "write instruction files into it". Under 0.1.10 that half is no longer true — it would leave a stale connection-store record and, only when an `AGENTS.md` already exists, a `CLAUDE.md` symlink. The rationale still holds; the wording is now the pre-0.1.10 one this PR retires elsewhere.
- `docs/design/context-tree-integration.md:49` is 154 characters where the rest of the file wraps at 101.

CI is green on run 34041401612 for this head, including the real-CLI `context-tree.integration.test.ts` and the portable runtime probe, so the behaviour above is covered by the suite and not only by my local probes. Approving this exact head.

PR #551, opened 2026-09-07
chore(deps): bump context-tree to 0.1.11

ggordonhall:

PR #552, opened 2026-09-07
chore(deps): bump context-tree to 0.1.11

ggordonhall:
Supersedes #551, which failed CI at `Validate pull request branch name`: the `bump-context-tree` branch lacks an approved prefix from CONTRIBUTING.md. Same commit cd8a2f93 on `chore/bump-context-tree`.

Bumps `@first-tree-ai/context-tree` 0.1.10 → 0.1.11 across apps/cli, packages/client and the portable dependency contract, with the lockfile and fixtures updated.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -51,7 +51,7 @@
     "test": "vitest run --maxWorkers=1"
   },
   "dependencies": {
-    "@first-tree-ai/context-tree": "0.1.10"
+    "@first-tree-ai/context-tree": "0.1.11"
   },
   "devDependencies": {
     "@opentag/client": "workspace:*",
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -64,7 +64,7 @@ export default defineConfig({
   ],
   server: {
     proxy: {
-      "/api": "http://localhost:8000",
+      "/api": { target: "http://localhost:8000", ws: true },
       "/healthz": "http://localhost:8000",
       "/readyz": "http://localhost:8000",
     },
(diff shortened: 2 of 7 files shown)


yuezengwu (approved):
Reviewed exact head `cd8a2f93d00453179e0050e81908d1c924e7e2bd`. No blocking finding in the full seven-file change. The coordinated CLI/client pins and lockfile match the published 0.1.11 integrity. Its CLI bundle and postinstall script are byte-identical to 0.1.10; the update changes packaged skills/templates, and an isolated project install successfully includes all seven skills. The additional Vite proxy change preserves HTTP forwarding and successfully forwards WebSocket upgrades in an isolated loopback probe.

Validation: `pnpm build`, `pnpm check`, `pnpm typecheck`, and `git diff --check` passed; 75 tests passed (34 Client unit, 5 real-CLI integration, 18 CLI, 18 portable dependency tests). Current PR CI run 34085842881 passed on `chore/bump-context-tree`; its tested merge `eb7682faec4626619d1b2a592cc4c70b5bebc828` combines this head with current main `20bb14a8043f812d6ad41efcc16ebc008607865c`, and its tree matches the locally computed merge. The displayed older failed run 34084436480 belongs to the superseded `bump-context-tree` branch.

Non-blocking: please retain the removed installed-version/source-pin mismatch regression test, updating its expected source pin to 0.1.11. The production fail-closed check still exists, and this is useful coverage to preserve.

liuchao-001:
Reviewed at merged head `cd8a2f93` — no blocking finding, and I agree with @yuezengwu's read of the pin/lockfile/fixture coordination.

Picking up the one non-blocking item, since it merged unaddressed: the removed `the installed direct dependency must match the exact source pin` test covered a guard that is still live in `scripts/portable/runtime-dependencies.mjs`, so nothing in CI proves that branch still throws. Restored in #558, with the expected pin read from the fixture manifest so the next context-tree bump does not force an edit there.

One scope note for future bumps rather than a request here: the `vite.config.ts` proxy change (`ws: true`) is unrelated to the dependency bump and would be easier to bisect on its own branch.

PR #554, opened 2026-09-07
feat(release): serve portable installs from dl.opentag.build

bestony:
## What

Points the public portable-release download base URL at the newly bound `dl.opentag.build` custom domain.

| | |
|---|---|
| Old | `https://storage.googleapis.com/opentag-release/releases` |
| New | `https://dl.opentag.build/releases` |

The domain was verified to serve the same objects 1:1 before this change — `staging/install.sh` returns an identical `etag` and `x-goog-meta-sha256` on both hosts, and `staging/latest.json` plus a versioned `manifest.json` resolve on the new host. `prod/install.sh` 404s on both, so there is no regression there.

## What does *not* change

Only the public HTTP base URL moves. The Cloud Storage coordinates stay put: `upload-gcs.sh` still writes to the `opentag-release` bucket under the `releases` prefix, and those literals plus their ratchet assertions are deliberately untouched.

This is safe because nothing derives the public host from the bucket name. `$BUCKET` only ever appears inside `gs://` URIs, and both validators (`normalizeDownloadBaseUrl`, `validate_download_base_url`) check protocol and a trailing channel segment, never host identity. `upload-gcs.sh` needs no edit at all — `derive_download_base_url` recovers the base URL from the built `latest.json`, so the upload side follows the build automatically.

## Changes

- The five in-repo defaults: `build-portable.mjs`, `build-release.sh`, `release-gcs.sh`, `install.sh`, and the CLI's `portable-installer.ts`.
- The server-side Zod default in `packages/server/src/config.ts`, which drives both the channel-target poller and the `install.sh` URL baked into every Computer connect command.
- Test assertions that pin those defaults, plus pass-through fixtures, kept consistent.
- Docs (EN + zh-CN mirrors, sync dates refreshed).

### New test coverage

`scripts/portable/install.sh`'s `DOWNLOAD_BASE_URL` fallback was pinned by **no test at all** — it was the one default that could have silently kept the old host with `pnpm check` and `pnpm test` both green. It is now covered by the same ratchet that pins the bucket and prefix. Verified by regressing the literal and confirming the test fails.

## Operational notes

- **The CI repository variable has been repointed.** `vars.OPENTAG_PORTABLE_DOWNLOAD_BASE_URL` was explicitly set to the old URL and overrides every in-repo default in all four publish steps. Without that update this PR would have been completely inert for real releases, and silently so, since build and upload read the same variable and would have agreed with each other.
- **The cutover completes at the next release, not at merge.** The published per-channel `install.sh` is a *rendered* copy with the base URL baked in, and `latest.json` / `manifest.json` embed absolute asset URLs. Until a release is published, the new domain still serves artifacts pointing at the old host.
- **Both hosts must stay live.** Version prefixes are immutable (create-only), so every already-published `manifest.json` points at the old host permanently. Pinned installs (`--version <old>`) and rollbacks will keep using it. This is an additive cutover; the old host cannot be decommissioned.
- No ordering of (merge) × (variable update) × (next release) breaks an install or a self-update. Installers resolve the tarball from the manifest's absolute `asset.url` and validate only the scheme, so mixed-host states install correctly.
- Worth re-validating on the new domain before the first publish: it must honor the object `Cache-Control` headers, since channel pointers are written `no-cache` and re-read immediately by the post-publish verification.

## Testing

All green locally:

- `pnpm check`
- `pnpm typecheck`
- `pnpm test` (263 test files)
- `pnpm build`
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 100% statements/branches/functions/lines
- `pnpm --filter @opentag/server test:integration` — 18 files, 324 tests (this suite owns the `bootstrapCommand` assertion in `computers.test.ts`, which `pnpm check`/`pnpm test` do not cover)

No migration is added, so the `CURRENT_MIGRATION_COUNT` ratchet does not apply.

### DIFF
--- a/apps/cli/src/__tests__/manual-upgrade.test.ts
+++ b/apps/cli/src/__tests__/manual-upgrade.test.ts
@@ -269,7 +269,7 @@ describe("manual upgrade", () => {
     const home = await tempHome();
     const installed: string[] = [];
     const fetchFn = (async (url: string | URL | Request) => {
-      expect(String(url)).toBe("https://storage.googleapis.com/opentag-release/releases/staging/latest.json");
+      expect(String(url)).toBe("https://dl.opentag.build/releases/staging/latest.json");
       return jsonResponse({ channel: "staging", version: "0.0.3-staging.1.1" });
     }) as typeof fetch;
     const result = await runUpgrade({
--- a/apps/cli/src/core/update/portable-installer.ts
+++ b/apps/cli/src/core/update/portable-installer.ts
@@ -14,7 +14,7 @@ import type { ChannelName } from "@opentag/shared";
  * commit point. A version directory is never rewritten in place.
  */
 
-export const DEFAULT_DOWNLOAD_BASE_URL = "https://storage.googleapis.com/opentag-release/releases";
+export const DEFAULT_DOWNLOAD_BASE_URL = "https://dl.opentag.build/releases";
 const PORTABLE_APP_ENTRY = "app/cli/index.mjs";
 const MANIFEST_SCHEMA_VERSION = 1;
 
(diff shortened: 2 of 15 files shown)


bestony:
Reopening from a correctly-prefixed branch (`feat/release-domain-cutover`) — the branch-name gate rejects the `worktree-` prefix the worktree tool generated. Superseded by the PR that follows; no content change.

PR #555, opened 2026-09-07
feat(release): serve portable installs from dl.opentag.build

bestony:
## What

Points the public portable-release download base URL at the newly bound `dl.opentag.build` custom domain.

| | |
|---|---|
| Old | `https://storage.googleapis.com/opentag-release/releases` |
| New | `https://dl.opentag.build/releases` |

The domain was verified to serve the same objects 1:1 before this change — `staging/install.sh` returns an identical `etag` and `x-goog-meta-sha256` on both hosts, and `staging/latest.json` plus a versioned `manifest.json` resolve on the new host. `prod/install.sh` 404s on both, so there is no regression there.

## What does *not* change

Only the public HTTP base URL moves. The Cloud Storage coordinates stay put: `upload-gcs.sh` still writes to the `opentag-release` bucket under the `releases` prefix, and those literals plus their ratchet assertions are deliberately untouched.

This is safe because nothing derives the public host from the bucket name. `$BUCKET` only ever appears inside `gs://` URIs, and both validators (`normalizeDownloadBaseUrl`, `validate_download_base_url`) check protocol and a trailing channel segment, never host identity. `upload-gcs.sh` needs no edit at all — `derive_download_base_url` recovers the base URL from the built `latest.json`, so the upload side follows the build automatically.

## Changes

- The five in-repo defaults: `build-portable.mjs`, `build-release.sh`, `release-gcs.sh`, `install.sh`, and the CLI's `portable-installer.ts`.
- The server-side Zod default in `packages/server/src/config.ts`, which drives both the channel-target poller and the `install.sh` URL baked into every Computer connect command.
- Test assertions that pin those defaults, plus pass-through fixtures, kept consistent.
- Docs (EN + zh-CN mirrors, sync dates refreshed).

### New test coverage

`scripts/portable/install.sh`'s `DOWNLOAD_BASE_URL` fallback was pinned by **no test at all** — it was the one default that could have silently kept the old host with `pnpm check` and `pnpm test` both green. It is now covered by the same ratchet that pins the bucket and prefix. Verified by regressing the literal and confirming the test fails.

## Operational notes

- **The CI repository variable has been repointed.** `vars.OPENTAG_PORTABLE_DOWNLOAD_BASE_URL` was explicitly set to the old URL and overrides every in-repo default in all four publish steps. Without that update this PR would have been completely inert for real releases, and silently so, since build and upload read the same variable and would have agreed with each other.
- **The cutover completes at the next release, not at merge.** The published per-channel `install.sh` is a *rendered* copy with the base URL baked in, and `latest.json` / `manifest.json` embed absolute asset URLs. Until a release is published, the new domain still serves artifacts pointing at the old host.
- **Both hosts must stay live.** Version prefixes are immutable (create-only), so every already-published `manifest.json` points at the old host permanently. Pinned installs (`--version <old>`) and rollbacks will keep using it. This is an additive cutover; the old host cannot be decommissioned.
- No ordering of (merge) × (variable update) × (next release) breaks an install or a self-update. Installers resolve the tarball from the manifest's absolute `asset.url` and validate only the scheme, so mixed-host states install correctly.
- Worth re-validating on the new domain before the first publish: it must honor the object `Cache-Control` headers, since channel pointers are written `no-cache` and re-read immediately by the post-publish verification.

## Testing

All green locally:

- `pnpm check`
- `pnpm typecheck`
- `pnpm test` (263 test files)
- `pnpm build`
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 100% statements/branches/functions/lines
- `pnpm --filter @opentag/server test:integration` — 18 files, 324 tests (this suite owns the `bootstrapCommand` assertion in `computers.test.ts`, which `pnpm check`/`pnpm test` do not cover)

No migration is added, so the `CURRENT_MIGRATION_COUNT` ratchet does not apply.

### DIFF
--- a/apps/cli/src/__tests__/manual-upgrade.test.ts
+++ b/apps/cli/src/__tests__/manual-upgrade.test.ts
@@ -269,7 +269,7 @@ describe("manual upgrade", () => {
     const home = await tempHome();
     const installed: string[] = [];
     const fetchFn = (async (url: string | URL | Request) => {
-      expect(String(url)).toBe("https://storage.googleapis.com/opentag-release/releases/staging/latest.json");
+      expect(String(url)).toBe("https://dl.opentag.build/releases/staging/latest.json");
       return jsonResponse({ channel: "staging", version: "0.0.3-staging.1.1" });
     }) as typeof fetch;
     const result = await runUpgrade({
--- a/apps/cli/src/core/update/portable-installer.ts
+++ b/apps/cli/src/core/update/portable-installer.ts
@@ -14,7 +14,7 @@ import type { ChannelName } from "@opentag/shared";
  * commit point. A version directory is never rewritten in place.
  */
 
-export const DEFAULT_DOWNLOAD_BASE_URL = "https://storage.googleapis.com/opentag-release/releases";
+export const DEFAULT_DOWNLOAD_BASE_URL = "https://dl.opentag.build/releases";
 const PORTABLE_APP_ENTRY = "app/cli/index.mjs";
 const MANIFEST_SCHEMA_VERSION = 1;
 
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -96,7 +96,7 @@ server container.
 | `OPENTAG_JWT_SECRET` | At least 32 random characters, unique to Staging and distinct from `BETTER_AUTH_SECRET`; signs Slack OAuth state only |
 | `OPENTAG_ENCRYPTION_KEY` | Base64 32-byte key, unique to Staging |
 | `OPENTAG_AUTO_MIGRATE` | `true` so each rollout applies pending migrations |
-| `OPENTAG_PORTABLE_DOWNLOAD_BASE_URL` | Optional; defaults to `https://storage.googleapis.com/opentag-release/releases` |
+| `OPENTAG_PORTABLE_DOWNLOAD_BASE_URL` | Optional; defaults to `https://dl.opentag.build/releases` |
 | `OPENTAG_CHANNEL_TARGET_POLL_INTERVAL_MS` | Optional; defaults to `300000` |
 
 The two optional variables control how the Server learns the exact channel latest target it advertises to connected
--- a/docs/portable-release.md
+++ b/docs/portable-release.md
@@ -18,7 +18,7 @@ identity rewrite, and the same version coordinate as the npm package it accompan
 ## Installing
 
 ~~~bash
-curl -fsSL https://storage.googleapis.com/opentag-release/releases/prod/install.sh | sh
+curl -fsSL https://dl.opentag.build/releases/prod/install.sh | sh
 ~~~
 
 The installer resolves the channel's `latest.json`, downloads the tarball for the detected platform, verifies its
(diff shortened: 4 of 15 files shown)


yuezengwu (approved):
Reviewed exact head `d97cfb8fb6bf26721d10fb449a6f04f5e357f66e`, with clean merge applicability checked at current main `db870b63b5c87840c825ab63c5154c345a143479`. No blocking findings in the download-domain change.

Local validation passed: 140 tests (45 portable, 25 CLI upgrade/installer, 52 Server config/account API, 18 PostgreSQL Computer integration), plus build, check, typecheck, and diff check. Storage bucket/prefix and verification ordering remain unchanged; explicit download-base overrides are preserved.

Independent read-only probes found identical bytes/ETags at both hosts for the staging installer, latest pointer, and versioned manifest, with the expected no-cache/immutable headers. The repository variable and live staging `0.0.3-staging.100.1` metadata already use the new domain. Both prod installer endpoints currently return 404; this PR does not establish production-channel availability. Keep the old endpoint available for historical absolute URLs.

Current PR CI run `34093112562` passed. The failed run `34092755074` belongs to the former `worktree-release-domain-cutover` branch, not this PR's current CI run. No deployment or public-package execution was performed during this review.

liuchao-001 (approved):
Approving at head `d97cfb8f`. The change is a clean literal cutover — no leftover `storage.googleapis.com` reference survives anywhere in the tree, and the bucket/prefix literals plus their ratchets are untouched, so the `gs://` write path is unaffected.

Independently verified rather than taken from the description:

- **Repo variable really is repointed.** `vars.OPENTAG_PORTABLE_DOWNLOAD_BASE_URL` is `https://dl.opentag.build/releases`. This matters because it overrides the in-repo defaults in all four publish steps, so without it the PR would have been inert for real releases.
- **The cutover has in fact already happened for staging.** `releases/staging/latest.json` is version `0.0.3-staging.100.1` at `gitSha 8d47986e` (current `main`) with `manifestUrl` and every `asset.url` already absolute on the new domain. So the publish pipeline — including its post-publish public verification gate — has already run green against the new host; this PR is the in-repo defaults catching up.
- **Byte identity, not just reachability.** `staging/install.sh` and `staging/latest.json` return the same `etag` and `x-goog-meta-sha256` on both hosts, and `no-cache, max-age=0, must-revalidate` passes through the new domain intact — which is the property the channel-pointer re-read depends on. `prod/install.sh` 404s on both, so no regression there.
- **Large binaries survive the new edge.** Downloaded the 42 MB `darwin-arm64` staging tarball through `dl.opentag.build`: 42139030 bytes, sha256 `8ee6320a…a228`, matching the manifest's declared size and digest exactly. The new host is Cloudflare-fronted rather than a GCS-native domain mapping, so this was the check worth doing.
- **Mixed-host states are genuinely safe.** `install.sh` derives only `latest.json` / `manifest.json` from `DOWNLOAD_BASE_URL`; the tarball comes from the manifest's absolute `asset.url` and is sha256-verified before extraction, and `portable-installer.ts` validates only `^https?://` on the asset URL. An old install self-updating through the old host reads the same object and follows the new absolute URLs.

The new ratchet on `install.sh`'s `DOWNLOAD_BASE_URL` fallback is the right addition — that literal was the one default with no test behind it.

One operational follow-up, non-blocking: the deployed Server's env. If the Staging deployment sets `OPENTAG_PORTABLE_DOWNLOAD_BASE_URL` explicitly rather than relying on the schema default, the Computer connect command will keep advertising the old host after this merges. Harmless while both hosts serve the same objects, but worth clearing to finish the cutover.

PR #556, opened 2026-09-07
feat(docs): update readme

ggordonhall:


Gandy2025 (changes requested):
Reviewed `8d47986e7a20f74281d89ade7a3625b57ab5a9f9...85ec234e1a182920d1e05c867e531d342ed4e5ab` as a low-risk documentation change. Because the PR has no body or linked requirement, the Spec axis is the stated README-update goal checked against the current product behavior in this repository.

## Standards

No blocking findings. The canonical English documents and Simplified Chinese mirrors were updated together, the repository checks pass, and the changed documents' local links and anchors resolve.

## Spec

**Blocking — `README.md:26` and `README.zh-CN.md:26` advertise an unavailable product option.**

The overview says users can bring a Pi agent, but the production Server schema currently accepts only `codex` and `claude-code` (`packages/server/src/db/schema/agents.ts:6`), and the Quick Start itself only offers Codex or Claude Code. The Pi adapter and its smoke tests do not make Pi selectable or connectable through the documented product flow; `docs/design/agent-runtime-contract.md:72-85` also distinguishes that adapter from production composition. Please remove Pi from the current-availability claim, or explicitly label it as adapter-only/not yet available, and keep the Chinese mirror synchronized.

Gandy2025 (approved):
Re-reviewed the focused delta `85ec234e1a182920d1e05c867e531d342ed4e5ab..24aee4a205104e89fb2672a5348653cb2073ec7f`.

## Standards

The prior Standards evidence carries forward. This delta changes only the matching English and Simplified Chinese availability claims, keeps the mirrors synchronized, and passes `git diff --check`.

## Spec

The previous blocker is resolved at `README.md:26` and `README.zh-CN.md:26`: Pi is no longer advertised as a currently connectable product option. No remaining findings in the affected scope.

liuchao-001 (approved):
Reviewed `24aee4a205104e89fb2672a5348653cb2073ec7f` (docs only, 6 files). I re-ran the factual claims of the new Quick Start against the code at this head rather than against the old README.

**What I verified**

- Dropping `pnpm install` / `pnpm build` from Step 1 is correct: `scripts/dev-install.sh` runs `pnpm install --frozen-lockfile` and `pnpm build` itself, so `bootstrap:admin` and `start` (both `node dist/...`) have their artifacts.
- The Step 2 env set satisfies every gate in `packages/server/src/config.ts`: dev bypass needs an explicit `OPENTAG_ENV=dev` plus loopback `OPENTAG_HOST`/`OPENTAG_PUBLIC_URL` and both `OPENTAG_DEV_AUTH_*` vars (lines 192-206), `BETTER_AUTH_SECRET` must differ from `OPENTAG_JWT_SECRET`, and `openssl rand -base64 32` yields the canonical 32-byte base64 the encryption-key schema requires.
- `docker compose up -d --wait postgres` is a real improvement — `docker-compose.yml` has a TCP `pg_isready` healthcheck, so `--wait` actually gates on readiness before `bootstrap:admin`.
- "Developer sign-in" matches shipped copy (`apps/web/messages/auth/en.json:15`), and Step 3's order (create agent → connect computer → connect chat) matches the flow's own step order in `apps/web/src/onboarding-v2/flow.ts:15`.
- Every local link and anchor in all six changed files resolves, including the new `docs/README.md` / `docs/zh-CN/README.md` indexes; `docs/README.md#quick-start` and `README.zh-CN.md#快速开始` line up with the renamed headings, and no doc under `docs/` is missing from the new index. No file elsewhere in the repo links to the removed README anchors (`#quickstart`, `#runtimes`, `#chat-platforms`).

No blocking findings. Approving. Five non-blocking notes, in the order I'd care about them:

**1. Ephemeral secrets will break the setup on the second day.** All three secrets are generated inline into a shell that goes away. `OPENTAG_ENCRYPTION_KEY` is the key behind `ApplicationCipher`, which protects the IM binding credentials (`packages/server/src/services/im-bindings/*`) the reader creates in Step 3. Restart the server from a fresh terminal, regenerate, and the Slack/Lark binding they just finished is undecryptable — with `BETTER_AUTH_SECRET` rotated too, so their session is gone and the failure looks unrelated to the restart. One line would prevent it: write the exports into a file (`.env.local` or similar) and re-source it on restart, rather than re-running the generators.

**2. Node support range narrowed.** `README.md:39` says "Node.js 24 (24.15.0 or later)", but the root `package.json` engines field is `^22.22.2 || ^24.15.0 || ^26.0.0` and CI runs all three lines. Nothing here is false, but a reader on 22 or 26 will conclude they have to switch. Suggest "Node.js 22.22.2+, 24.15+, or 26".

**3. The zh mirror header drifted into the middle of the About section.** On `main` the `> 权威来源… · 同步日期…` blockquote sat directly under `</div>`; the reflow left it anchored above the GIF, so it now renders between the About bullets and the image (`README.zh-CN.md:29`) while the English file has nothing there. Moving it back under `</div>` restores it.

**4. Step 3's `export PATH` block is redundant.** The connect command the Web hands out already prefixes `PATH="$HOME/.local/bin${PATH:+:$PATH}"` and invokes the binary by absolute path (`packages/server/src/services/computers/machine-auth-service.ts:600`), so pasting it works in a bare terminal. Dropping the extra block (and the "after setting the PATH below" clause) makes the step one paste instead of two.

**5. Copy nits.** `README.md:21-22` "From Slack and Lark talk to AI agents that run on your own machine…" front-loads the prepositional phrase and reads as a fragment — "Talk to AI agents from Slack and Lark: they run on your own machine and use the model provider you choose" is the same claim. `README.md:7` uses `__…__` where the rest of the file uses `**…**`, and the first bullet uses a hyphen as a dash separator while others don't.

One editorial observation, entirely the author's call: the differentiators this trims — bring your own plan, no provider-key custody, work-area files stay on your machine, Apache 2.0 with no hosted-service carve-out — were the part of the README doing the "why this instead of the hosted thing" work, and they don't survive anywhere else now that Architecture moved to `DEVELOPMENT.md` and the docs table moved to `docs/README.md`. The new structure is much easier to follow and I'd keep it; I'd just consider keeping three or four of those bullets under About. Not a request for change.

PR #546, opened 2026-09-07
feat: define agent-managed Home layout in runtime prompts

yuezengwu:
## Summary

Refs #221 and #213.

- Following [First Tree's Agent Home/worktree instructions](https://github.com/agent-team-foundation/first-tree/blob/2e0c4c26d8f09ef7115e45a96695c612124356aa/packages/client/src/runtime/templates/agent-briefing.ejs#L108), describe persistent shared Home, Agent-managed `source-repos/`, independent task `worktrees/`, and on-demand `files/` in the managed prompt.
- Pass the resolved Home into the prompt and anchor Context Tree project commands to that Home when working in a task subdirectory. Keep Provider cwd, writable roots, and missing-memory behavior unchanged.
- Add prompt/runtime regression tests and synchronize English/Chinese development docs. No repository model, automatic directory creation, GC, migration, or new instruction files.

## Testing

- Passed: frozen install; `pnpm check`; `pnpm build`; `pnpm typecheck`; pre-commit and all pre-push hooks.
- Passed: 145 focused client tests; Agent Runtime coverage suite **360 tests, 100% statements/branches/functions/lines**; Server integration **324 tests**.
- Passed: full `pnpm test` (Shared 220, Client 987, Server 853, CLI 431, Web 665 tests, plus root script checks).
- Focused-local acceptance rerun at `1de1b883`: two Sessions share a Home across manager reconstruction; two task worktrees keep concurrent edits separate; each task writes to and reads from the same Context Tree 0.1.10 using `--project-path` for the Home. Real filesystem/Git/packaged CLI, isolated test account, preserved user instructions, final Tree valid and clean. No live model-provider E2E; prompt conventions are not an enforced sandbox.

## Breaking changes

None. `ManagedSessionContext.agentHome` is optional; existing cwd and filesystem policy remain unchanged.

## Checklist

- [x] The change is focused and contains no unrelated work.
- [x] Tests and documentation cover the changed behavior.
- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass.
- [x] No credentials, private data, or generated build output are included.
- [x] Canonical English docs and Chinese mirrors are synchronized when applicable.

Issue #221, opened 2026-08-28
[Feature]: define an explicit per-Agent Home and directory layout


bestony:
## TL；DR
1. 在 data 下放 context tree
2. 如果用户已经有 context-tree（如 ~/.context-tree），直接 soft link 过来；
3. agent Workspace 下 提供 AGENTS.md、.skill 目录，用于实现相关功能。

## Summary

Every Agent should have an explicit, stable, and documented filesystem Home. The Home should define which directories belong to the Agent, what each directory is for, how long its contents live, and which boundaries apply when multiple Tasks or Agents run on the same Computer.

The refined layout separates the shared Context Tree from each Agent's filesystem Home. Multiple Agents must be able to use the same Context Tree, but only through the Context Tree authorization model. The following staging layout is the proposed starting point. The workspace entries below are representative runtime examples; Agent-owned files and directories are created as needed:

```text
.opentag-staging/                         # configured OpenTag Home (staging example)
├── config/                               # OpenTag configuration
├── data/
│   ├── contex-tree/                       # shared Context Tree data for authorized Agents
│   ├── runtime/                           # platform-owned runtime and recovery data
│   │   ├── durability/
│   │   ├── effective-snapshots/
│   │   ├── provider-cli-validation/
│   │   ├── provider-credentials/
│   │   ├── session-bindings/
│   │   ├── session-cli-proofs/
│   │   └── workspace-states/
│   └── workspaces/                        # Agent filesystem Homes
│       ├── a-0d8524659519fcc2b1f39d2e9358327b9e44340b/
│       │   ├── .skills/
│       │   ├── AGENTS.md
│       │   ├── issue-419-image.png
│       │   └── opentag/
│       ├── a-2f86894609140977322348f18bddd8b2a1076310/
│       │   ├── .skills/
│       │   ├── AGENTS.md
│       │   └── files/
│       └── a-6d0e3782e96f72bbe59dc19c3d6198e7a08f2146/
│           └── opentag/
├── logs/                                  # sanitized operational logs
└── state/                                 # daemon and service state
    ├── daemon/
    └── service/
```

Entries under `data/workspaces/<agent-key>/` may include `AGENTS.md`, `.skills/`, repository checkouts such as `opentag/`, Task artifacts such as `issue-419-image.png`, and a `files/` directory. These entries are Agent-owned and may not all exist for every Agent at every point in time.

[...]

Issue #213, opened 2026-08-28
[Bug]: clone Git repositories into a workspace child directory


bestony:
## Summary

When cloning a Git repository for a task, the repository is currently cloned directly into the workspace directory. It should be cloned into a dedicated child directory named after the Git repository.

## Current behavior

The workspace itself becomes the Git repository directory:

```text
/Users/bestony/.opentag-staging/data/workspaces/a-1894556afeff994f8f84a69148a136b0431ec989/opentag
```

This mixes repository files with other workspace-level files and can make the workspace contents and task boundaries unclear.

## Expected behavior

Keep the workspace as the container and clone the repository into a child directory:

```text
/Users/bestony/.opentag-staging/data/workspaces/a-1894556afeff994f8f84a69148a136b0431ec989/opentag/[git-repo-name]/
```

The active working directory for repository operations should be that child directory.

## Impact

- Workspace metadata and repository files are mixed together.
- Multiple repositories or task artifacts can conflict in the same directory.
- Repository discovery, cleanup, and isolation become ambiguous.
- Instructions and diagnostics can report the wrong project root.

## Acceptance criteria

- Derive a safe child directory from the repository name, with a deterministic collision policy.
- Clone into that child directory for HTTPS and SSH repository URLs.
- Preserve the existing workspace root for workspace-level metadata and task artifacts.
- Set the task's repository working directory to the cloned child directory.
- Handle an existing child directory explicitly without overwriting unrelated files.
- Add tests covering repository names with `.git` suffixes, nested URLs, collisions, and multiple repositories in one workspace.

bestony:
I filed #221 to define the complete per-Agent Home contract around this concrete bug.

#213 should remain the focused implementation slice for repository placement: every checkout gets a dedicated child under `workspaces/`, with deterministic collision and non-overwrite behavior. The broader issue captures the related decisions that otherwise tend to become implicit in separate fixes: the canonical Agent Home root, `AGENTS.md` and `.skills` provenance, Context Tree boundaries, Task/Session isolation, cleanup, and migration from the current `files/AGENTS.md` layout.

@Gandy2025 @baixiaohang I would appreciate your view on whether #213 should target the existing `data/workspaces/<agent-key>` root or wait for the canonical root decision in #221. The child-directory behavior is useful either way, but the cwd and retention contract should be consistent.

yuezengwu:
[...]

PR #557, opened 2026-09-07
test: retain shared Agent Home and Context Tree acceptance

yuezengwu:
## Summary

Follow-up to merged #546. Refs #221 and #213.

- Retain the temporary shared Agent Home acceptance as an offline Client regression using real local Git and the packaged Context Tree CLI (currently 0.1.11).
- Cover two Sessions sharing a durable Home, Agent-created independent code worktrees, explicit Home project anchoring from task directories, sequential bidirectional memory writes, preserved user instructions, and manager reconstruction.
- Document the focused rerun command in English and Chinese. No production runtime, dependency, CI, repository-management, GC, migration, or sandbox changes.

## Testing

- Passed: frozen install, `pnpm check`, `pnpm build`, and `pnpm typecheck`.
- Passed: 107 directly related tests; independent focused rerun passed in 2.86 seconds.
- Passed: Agent Runtime coverage (360 tests; 100% statements, branches, functions, and lines).
- Passed: Server PostgreSQL integration (324 tests).
- Passed: full `pnpm test` (329 root script tests; Shared 220, Client 988, Server 853, CLI 431, Web 674).
- Passed: pre-commit and pre-push hooks; no bypasses.

The test uses a disposable account and child-process environment, no Provider credentials or public network. It tests filesystem/CLI behavior, not model compliance or enforced isolation. Tree writes are sequential, not a concurrent-write test.

## Breaking changes

None.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -77,6 +77,16 @@ pnpm --filter @opentag/server test:integration
 
 Use `pnpm lint` for lint-only feedback. Use `pnpm format` to apply Biome formatting.
 
+The shared Agent Home regression is included in the Client's offline test suite. To run it alone after building:
+
+```bash
+pnpm --filter @opentag/client exec vitest run src/__tests__/agent-home-context.integration.test.ts --maxWorkers=1
+```
+
+It uses real local Git worktrees and the packaged Context Tree CLI in a temporary account to check shared Home
+persistence, independent code edits, and sequential bidirectional memory writes from task directories anchored to Home.
+It needs no Provider credentials or network and does not establish model compliance or sandbox isolation.
+
 The separate `Unit Coverage` workflow runs `pnpm test:coverage` against `main` every Monday at 03:17 UTC and can also be
 started manually. It builds the workspaces and measures the offline unit tests for CLI, Web, Shared, Client, and Server,
 then retains the unified report for 14 days. Run the command locally when changing the root coverage configuration or
--- a/DEVELOPMENT.zh-CN.md
+++ b/DEVELOPMENT.zh-CN.md
@@ -74,6 +74,15 @@ pnpm --filter @opentag/server test:integration
 
 仅检查 lint 可运行 `pnpm lint`；应用 Biome 格式化可运行 `pnpm format`。
 
+共享 Agent Home 回归测试包含在 Client 离线测试中。构建后可单独运行：
+
+```bash
+pnpm --filter @opentag/client exec vitest run src/__tests__/agent-home-context.integration.test.ts --maxWorkers=1
+```
+
+测试在临时账户中使用真实本地 Git worktree 和配套 Context Tree CLI，验证共享 Home 持久性、代码修改互不混入，
+以及从任务目录显式指定 Home 后串行进行双向记忆读写。无需 Provider 凭据或网络；不证明模型遵守提示词或 sandbox 隔离。
+
 独立的 `Unit Coverage` workflow 会在每周一 03:17 UTC 针对 `main` 运行 `pnpm test:coverage`，也支持手动触发。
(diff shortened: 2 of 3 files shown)


Gandy2025 (approved):
Reviewed `8d47986e7a20f74281d89ade7a3625b57ab5a9f9...d1f98095939301b3c892b81679b04d9e92e95288` as a low-risk tests-only change.

**Standards:** Pass. The regression is isolated, uses the repository's existing production managers and packaged CLI seams, keeps host credentials out of child processes, cleans up its real-Git fixture, and keeps the English/Chinese development documentation synchronized.

**Spec:** Pass. The test covers two Sessions resolving the same durable Agent Home, independent Agent-created code worktrees, explicit Home anchoring from task directories, sequential write/read handoff in both directions through the same Context Tree, preservation of user-owned instructions, and workspace/session manager reconstruction. It does not claim or introduce repository management, concurrent Tree writes, model compliance, or sandbox enforcement.

No blocking or non-blocking findings. Local verification at the exact head passed: `pnpm build`, the documented focused regression command (1 test), `pnpm check`, `pnpm typecheck`, and `git diff --check`.

liuchao-001 (approved):
Reviewed at head `d1f98095`. Approving — no blocking findings.

**Verification (this head, local macOS):** `pnpm install --frozen-lockfile`, `pnpm exec turbo run build --filter=@opentag/client...`, then the documented focused command — 1 test passed in ~5s. Biome clean on the new file.

**Mutation check (the part I cared about most):** I relaxed `AgentWorkspaceManager.prepareAgent`'s non-empty-workspace guard to `if (workspaceExists && (await readdir(paths.workspaceRoot)).length > 0)` — i.e. a regression where a populated Home can no longer be re-prepared — and the test failed with `code: 'conflict'` on the manager-reconstruction step. So the headline claim (a durable Home survives manager restart with real Agent-created content in it) is genuinely load-bearing, not a filesystem tautology. Reverted afterwards.

Non-blocking notes, all optional:

1. The Codex-skill assertion cannot fail for the reason it looks like it covers. `isolatedExecFile` deliberately drops `options.env`, so the `HOME: dirname(codexHome)` override `ContextTreeManager` builds for `install --host codex` never reaches the child; the skills land in `<accountHome>/.codex` purely because the fixture's own env sets `HOME=accountHome`, which happens to equal `dirname(codexHome)`. If the manager stopped passing that override, this assertion would still pass. The env-drop is the right call for credential isolation and `context-tree.integration.test.ts` covers the redirect properly, so this is fine as-is — worth a one-line comment so a future reader does not read more into it.

2. The two `SKILL.md` assertions match `"context-tree sync"` / `"context-tree prepare-write"`, i.e. prose inside the pinned `@first-tree-ai/context-tree@0.1.11`. The neighbouring test matches only `"context-tree"`. A Dependabot bump that rewords the packaged skills turns this red on an unrelated PR; loosening to the command name alone would keep the same signal.

3. `seedAgentOwnedCode` asserts a few things that are properties of Git rather than of OpenTag (`rev-parse --is-bare-repository`, `remote get-url origin`, `branch --show-current`, reading back the file it just wrote). They cost little, but they cannot fail independently of the fixture, and the equivalent post-restart assertions later in the test already carry the real signal.

4. One `it` covers roughly three claims (shared Home, worktree independence, sequential Tree writes), so a failure needs the assertion line to localize. Given the ~5s real-Git/CLI fixture I would not split it just to split it — flagging it only so the choice is deliberate.

Everything else reads well: the acceptance path from #546 is now a real regression instead of a one-off manual rerun, the docs are synchronized in both languages and the "after building" caveat is accurate (I hit the `@opentag/shared` entry-resolution failure when I skipped the build), and nothing here touches production runtime, CI, or dependencies.

PR #558, opened 2026-09-07
test(portable): retain the installed-vs-source pin regression

liuchao-001:
Follow-up to #552, which removed the regression test for the installed-vs-source pin check while bumping `@first-tree-ai/context-tree` to 0.1.11. @yuezengwu asked in review on that PR to retain the coverage.

The production guard it covered is still live — `collectRuntimeDependencyClosure` fails closed when an installed runtime dependency's version differs from the exact source pin (`scripts/portable/runtime-dependencies.mjs`). Without the test, nothing in CI proves that branch still throws.

This restores the test with one change to why it was dropped: the expected pin is now read from the fixture manifest instead of hard-coded, so the next context-tree bump does not force an edit here. The deliberately-wrong installed version is `0.0.1`, which no future pin will collide with.

## Validation

- `node --test scripts/__tests__/portable-runtime-dependencies.test.mjs` — 19/19 pass (was 18).
- `node --test scripts/__tests__/*.test.mjs` — 330/330 pass.
- `pnpm check` — exit 0.
- Mutation check: disabling the pin comparison in `runtime-dependencies.mjs` makes exactly this test fail (18 pass / 1 fail), confirming it observes the guard rather than passing vacuously.

### DIFF
--- a/scripts/__tests__/portable-runtime-dependencies.test.mjs
+++ b/scripts/__tests__/portable-runtime-dependencies.test.mjs
@@ -220,6 +220,23 @@ test("materializing the closure copies published content and never runs lifecycl
   assert.deepEqual(readdirSync(join(appDir, "node_modules")).sort(), ["@first-tree-ai", "commander"]);
 });
 
+test("the installed direct dependency must match the exact source pin", async (t) => {
+  const root = await mkdtemp(join(tmpdir(), "opentag-portable-rdep-"));
+  t.after(() => rm(root, { force: true, recursive: true }));
+  const cliRoot = fixtureCliRoot(root);
+  const sourcePin = JSON.parse(readFileSync(join(cliRoot, "package.json"), "utf8")).dependencies[
+    "@first-tree-ai/context-tree"
+  ];
+  writeContextTreePackage(join(cliRoot, "node_modules"), { version: "0.0.1" });
+  assert.throws(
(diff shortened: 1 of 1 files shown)


bestony (approved):
Reviewed exact head 808bc992896aa822c39595db62938dfb0a4cf7e6. No blocking findings.

Standards: the focused test follows the repository Node test and fixture patterns; the branch and Conventional Commit comply with repository policy.

Spec: the restored test reads the expected pin from the fixture manifest, installs a deliberately mismatched version, and exercises the live fail-closed comparison requested in the #552 review. No missing requirement or scope creep found.

Validation: node --test scripts/__tests__/portable-runtime-dependencies.test.mjs (19/19); node --test scripts/__tests__/*.test.mjs (330/330); pnpm check; git diff --check.

ggordonhall:
this is a crazy nit, but you go for it agents 😅

yuezengwu (approved):
Reviewed exact head `808bc992896aa822c39595db62938dfb0a4cf7e6` against main `db870b63b5c87840c825ab63c5154c345a143479`. No blocking findings.

The restored test reads the expected pin from its fixture manifest and checks the specific installed-version mismatch error. It retains the coverage requested on #552 without adding another copy of the current expected version to the assertion.

Independent validation passed: 19 focused tests, all 330 script tests, `pnpm check`, and `git diff --check`. A paired mutation check in a separate temporary copy passed with the real guard and failed when only the pin comparison was disabled. Current PR CI run `34104739436` also passed.

PR #562, opened 2026-09-08
fix(client): grant runtimes access to Context Tree configuration

ggordonhall:
## Summary

Agents can run `opentag context-tree connect` before Context Tree is configured. Every provider runtime receives the dedicated `<OPENTAG_HOME>/config/context-tree/` writable directory, created before provider startup. The configuration now lives at `config/context-tree/config.json`. The directory grant supports file creation on Linux while keeping Computer identity and credential files outside the grant.

Both visible and internal Sessions can change this Computer-wide target, including through direct configuration edits. Later provider runtime starts consume the selected target. Workspace, Slack, and connected-tree grants are preserved. Claude Code already runs with unrestricted filesystem access; this PR does not add isolation for it.

Session PATH composition preserves the Context Tree shim and provider executable directories. Shim preparation is shared, bounded by the Session start budget, and retried after failures; preparation failures remain diagnosable. The stalled-preparation test waits for the shared write instead of assuming filesystem preparation completes within 10 ms.

## Breaking changes

Existing `<OPENTAG_HOME>/config/context-tree.json` files are ignored. There is no automatic migration or read fallback. Previously configured Computers must reconnect using their previous target:

- `opentag context-tree connect <managed-name>`
- `opentag context-tree connect OWNER/REPO`
- `opentag context-tree connect --tree-path <path>`

Restart the daemon and affected Sessions, then verify with `opentag doctor`. The old configuration and existing tree data remain on disk. When both configuration files exist, the new path wins. The schema and connection command syntax are unchanged.

## Testing

- CI passed on reviewed head `0dacb63515f5c1a78ec853d5d2b22df3b45613a3`, including Checks, patch coverage, and Node 22/24/26 compatibility: https://github.com/first-tree-ai/opentag/actions/runs/34302866500.
- Independent review validation on that head passed check, build, typecheck, 1,004 Client tests, 434 CLI tests, and the 368-test Agent Runtime coverage gate at 100% across all four metrics.
- The earlier real Linux Codex sandbox probe established that the exact-file grant could not create the configuration; this motivated granting an existing directory. The final implementation grants only the dedicated Context Tree configuration directory on both platforms. That earlier probe is not a fresh sandbox run of the final layout.
- Runtime and CLI tests cover an old-only configuration and new-path precedence when both files exist. Independent upgrade validation confirmed manual reconnect writes the new configuration with private permissions and leaves legacy bytes untouched.
- Review follow-up: all 41 tests in `packages/client/src/__tests__/context-tree.test.ts` passed locally after fixing the timing-sensitive assertion; `pnpm check` passed. CI must run again for the follow-up commit.

## Checklist

- [x] Check, build, typecheck, and tests passed for the reviewed head as recorded above.
- [x] Canonical documentation describes the directory grant and breaking upgrade; no Chinese mirror exists for this design document.

### DIFF
--- a/packages/client/src/storage/home-layout.ts
+++ b/packages/client/src/storage/home-layout.ts
@@ -4,6 +4,7 @@ import { join, resolve } from "node:path";
 export interface OpenTagHomeLayout {
   config: string;
   contextTreeBin: string;
+  contextTreeConfigDir: string;
   contextTreeConfigFile: string;
   contextTreePreparationFile: string;
   daemonState: string;
@@ -33,7 +34,8 @@ export function resolveOpenTagHomeLayout(home = resolveOpenTagHome()): OpenTagHo
     config,
     // Holds the shim the Context Tree skills invoke by name; OpenTag's own calls bypass it.
     contextTreeBin: join(resolvedHome, "context-tree", "bin"),
-    contextTreeConfigFile: join(config, "context-tree.json"),
+    contextTreeConfigDir: join(config, "context-tree"),
+    contextTreeConfigFile: join(config, "context-tree", "config.json"),
     contextTreePreparationFile: join(state, "context-tree-preparation.json"),
     daemonState: join(state, "daemon"),
     data,
--- a/apps/cli/src/__tests__/context-tree.test.ts
+++ b/apps/cli/src/__tests__/context-tree.test.ts
@@ -178,29 +178,32 @@ describe("readContextTreeState", () => {
     });
   });
 
-  it("reports a recorded unavailable GitHub preparation as invalid", async () => {
-    const home = await temporaryDirectory("opentag-ct-state-preparation-");
-    await mkdir(join(home, "config"), { mode: 0o700, recursive: true });
-    await mkdir(join(home, "state"), { mode: 0o700, recursive: true });
-    await writeFile(
-      configFile(home),
-      JSON.stringify({ schemaVersion: 1, target: { kind: "github", repository: "acme/missing" } }),
-      "utf8",
-    );
-    await writeFile(
-      join(home, "state", "context-tree-preparation.json"),
-      JSON.stringify({
-        schemaVersion: 1,
-        target: "acme/missing",
-        status: "unavailable",
-        reason: "GITHUB_AUTH",
-        at: new Date().toISOString(),
-      }),
-      "utf8",
-    );
+  it.each(["GITHUB_AUTH", "SHIM_UNAVAILABLE", "PACKAGE_MISSING"])(
+    "reports recorded %s preparation instead of not cloned",
+    async (reason) => {
+      const home = await temporaryDirectory("opentag-ct-state-preparation-");
+      await mkdir(join(home, "config"), { mode: 0o700, recursive: true });
+      await mkdir(join(home, "state"), { mode: 0o700, recursive: true });
+      await writeFile(
+        configFile(home),
+        JSON.stringify({ schemaVersion: 1, target: { kind: "github", repository: "acme/missing" } }),
+        "utf8",
+      );
+      await writeFile(
+        join(home, "state", "context-tree-preparation.json"),
+        JSON.stringify({
+          schemaVersion: 1,
+          target: "acme/missing",
+          status: "unavailable",
+          reason,
+          at: new Date().toISOString(),
+        }),
+        "utf8",
+      );
 
-    await expect(
-      readContextTreeState({ home, contextTreePackage: await fakeCli({ list: listing([]) }) }),
-    ).resolves.toMatchObject({ target: "acme/missing", tree: "invalid", detail: "GITHUB_AUTH" });
-  });
+      await expect(
+        readContextTreeState({ home, contextTreePackage: await fakeCli({ list: listing([]) }) }),
+      ).resolves.toMatchObject({ target: "acme/missing", tree: "invalid", detail: reason });
+    },
+  );
 });
--- a/apps/cli/src/__tests__/context-tree.test.ts
+++ b/apps/cli/src/__tests__/context-tree.test.ts
@@ -77,6 +77,34 @@ describe("opentag context-tree connect", () => {
     expect(text()).toContain("team-context-tree");
   });
 
+  it("requires reconnecting an old-only configuration and reads the new target afterward", async () => {
+    const home = await temporaryDirectory("opentag-ct-upgrade-");
+    const oldFile = join(home, "config", "context-tree.json");
+    const oldConfig = JSON.stringify({ schemaVersion: 1, target: { kind: "managed", name: "old-tree" } });
+    await mkdir(join(home, "config"), { recursive: true });
+    await writeFile(oldFile, oldConfig);
+    const contextTreePackage = await fakeCli({ list: listing(["new-tree"]) });
+
+    await expect(readContextTreeState({ home, contextTreePackage })).resolves.toEqual({
+      configPath: configFile(home),
+      tree: "unknown",
+    });
+    await expect(readFile(configFile(home))).rejects.toMatchObject({ code: "ENOENT" });
+    await expect(
+      runContextTreeConnect({ ...capture().deps, home, contextTreePackage, target: "new-tree" }),
+    ).resolves.toEqual({ exitCode: 0 });
+    expect(JSON.parse(await readFile(configFile(home), "utf8"))).toEqual({
+      schemaVersion: 1,
+      target: { kind: "managed", name: "new-tree" },
+    });
+    await expect(readContextTreeState({ home, contextTreePackage })).resolves.toEqual({
+      configPath: configFile(home),
+      target: "new-tree",
+      tree: "valid",
+    });
+    await expect(readFile(oldFile, "utf8")).resolves.toBe(oldConfig);
+  });
+
   it("accepts a GitHub target without network work, and says who clones it", async () => {
     const home = await temporaryDirectory("opentag-ct-github-");
     const { deps, text } = capture();
--- a/apps/cli/src/__tests__/context-tree.test.ts
+++ b/apps/cli/src/__tests__/context-tree.test.ts
@@ -53,7 +53,7 @@ function capture(): { deps: { stdout: (chunk: string) => void; stderr: (chunk: s
 const listing = (names: readonly string[]) => ({
   payload: { schemaVersion: 1, trees: names.map((name) => ({ name, tree: { kind: "local", path: `/t/${name}` } })) },
 });
-const configFile = (home: string) => join(home, "config", "context-tree.json");
+const configFile = (home: string) => join(home, "config", "context-tree", "config.json");
 const invalid = { findings: [{ code: "MISSING_ROOT" }], ok: false };
 
 describe("opentag context-tree connect", () => {
@@ -159,7 +159,7 @@ describe("readContextTreeState", () => {
   ])("reports %s", async (_label, target, responses, expected) => {
     const home = await temporaryDirectory("opentag-ct-state-");
     if (target !== undefined) {
-      await mkdir(join(home, "config"), { mode: 0o700, recursive: true });
+      await mkdir(join(home, "config", "context-tree"), { mode: 0o700, recursive: true });
       await writeFile(configFile(home), JSON.stringify({ schemaVersion: 1, target }), "utf8");
     }
 
@@ -170,7 +170,7 @@ describe("readContextTreeState", () => {
 
   it("treats unreadable configuration as unknown rather than failing", async () => {
     const home = await temporaryDirectory("opentag-ct-state-bad-");
-    await mkdir(join(home, "config"), { mode: 0o700, recursive: true });
+    await mkdir(join(home, "config", "context-tree"), { mode: 0o700, recursive: true });
     await writeFile(configFile(home), "{ not json", "utf8");
 
     await expect(readContextTreeState({ home, contextTreePackage: await fakeCli({}) })).resolves.toMatchObject({
@@ -182,7 +182,7 @@ describe("readContextTreeState", () => {
     "reports recorded %s preparation instead of not cloned",
     async (reason) => {
       const home = await temporaryDirectory("opentag-ct-state-preparation-");
-      await mkdir(join(home, "config"), { mode: 0o700, recursive: true });
+      await mkdir(join(home, "config", "context-tree"), { mode: 0o700, recursive: true });
       await mkdir(join(home, "state"), { mode: 0o700, recursive: true });
       await writeFile(
         configFile(home),
--- a/apps/cli/src/__tests__/doctor.test.ts
+++ b/apps/cli/src/__tests__/doctor.test.ts
@@ -695,9 +695,9 @@ describe("doctor report and exit contract", () => {
   it("never lets a Context Tree fault change the doctor exit code", async () => {
     const home = await createHome();
     const cases = [
-      { configPath: resolve(home, "config", "context-tree.json"), tree: "unknown" as const },
+      { configPath: resolve(home, "config", "context-tree", "config.json"), tree: "unknown" as const },
       {
-        configPath: resolve(home, "config", "context-tree.json"),
+        configPath: resolve(home, "config", "context-tree", "config.json"),
         target: "team-context-tree",
         tree: "invalid" as const,
         detail: "DIRTY_TREE",
@@ -720,7 +720,7 @@ describe("doctor report and exit contract", () => {
     const home = await createHome();
     const result = await runHealthyDoctor(home, {
       inspectContextTreeState: vi.fn().mockResolvedValue({
-        configPath: resolve(home, "config", "context-tree.json"),
+        configPath: resolve(home, "config", "context-tree", "config.json"),
         tree: "unknown" as const,
       }),
     });
@@ -735,7 +735,7 @@ describe("doctor report and exit contract", () => {
     const home = await createHome();
     const result = await runHealthyDoctor(home, {
       inspectContextTreeState: vi.fn().mockResolvedValue({
-        configPath: resolve(home, "config", "context-tree.json"),
+        configPath: resolve(home, "config", "context-tree", "config.json"),
         target: "acme/shared-context",
         tree: "not-cloned" as const,
       }),
@@ -802,7 +802,7 @@ async function runHealthyDoctor(home: string, overrides: Partial<DoctorOptions> 
 
 function configuredContextTree(home: string) {
   return {
-    configPath: resolve(home, "config", "context-tree.json"),
+    configPath: resolve(home, "config", "context-tree", "config.json"),
     target: "team-context-tree",
     tree: "valid" as const,
   };
--- a/apps/cli/src/core/context-tree/shared.ts
+++ b/apps/cli/src/core/context-tree/shared.ts
@@ -38,7 +38,7 @@ export function resolveContextTreeAssets(deps: ContextTreeCommandDeps = {}): Con
 }
 
 export function contextTreeConfigPath(home: string): string {
-  return join(home, "config", "context-tree.json");
+  return join(home, "config", "context-tree", "config.json");
 }
 
 export async function readContextTreeConfig(home: string): Promise<ContextTreeConfig | undefined> {
--- a/docs/design/context-tree-integration.md
+++ b/docs/design/context-tree-integration.md
@@ -157,11 +157,16 @@ What this costs, stated rather than left implicit:
 
 ### The CLI shim
 
-`<OPENTAG_HOME>/context-tree/bin/context-tree` is a generated `0700` shim that execs the installed
+`<OPENTAG_HOME>/context-tree/bin/context-tree` is a generated `0700` shim that execs the bundled
 CLI with the same Node.js runtime OpenTag itself uses, so a Session cannot resolve a different one
 from the user's shell configuration. That directory is prepended to the Provider `PATH` during
 Client composition, unconditionally — it is a stable OpenTag-owned path, and a directory that does
 not exist yet is inert on `PATH`.
+
+The shim is prepared before checking Computer configuration, so an unconfigured Computer can run
+the bundled command without creating or connecting a tree. Preparation failures retain the existing
+unavailable statuses. Managed instructions identify command availability failures as runtime setup
+problems; a global install is unnecessary.
 
 The package's own `node_modules/.bin/context-tree` is not used for this: npm populates
 `<consumer>/node_modules/.bin` but pnpm's virtual store does not, so the location is not portable
@@ -171,6 +176,12 @@ resolve whatever `node` the Session's `PATH` happens to find.
 It is prepended at composition rather than through per-Session workspace environment because a
 Session-level `PATH` would replace the value the factory composes, including the discovered
 executable directory that lets `codex` and `claude` resolve at all.
+
+Visible Sessions supply their tool directory through workspace `pathPrepend`. Every Provider factory
+prepends it after composing its environment, preserving the Context Tree and executable directories.
+Internal Sessions retain Context Tree without adding visible-Session tools.
+
+Rollout requires restarting the daemon and affected Sessions after release; no migration is needed.
 
 OpenTag's own invocations never rely on the shim: they exec the resolved CLI path directly, so a
 broken or shadowed shim cannot change what OpenTag executes.
--- a/docs/design/context-tree-integration.md
+++ b/docs/design/context-tree-integration.md
@@ -74,7 +74,7 @@ There is deliberately no `inspect` subcommand. `opentag doctor` already owns dia
 the injectable-inspector seam, and two surfaces over one piece of state means every future reason
 code has to be rendered twice.
 
-The target is recorded machine-locally in `<OPENTAG_HOME>/config/context-tree.json`, mode
+The target is recorded machine-locally in `<OPENTAG_HOME>/config/context-tree/config.json`, mode
 `0600`, credential-free. The Server is not involved. The three target kinds mirror
 `context-tree connect`'s own argument shape, so OpenTag passes the target through rather than
 reinterpreting it.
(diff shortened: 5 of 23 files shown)


yuezengwu (commented):
Reviewed exact head `af9deae2521cf839214bd0ffd3f180f32e435828`; changes since its base on current main `42362fd034e7e451f15c97e71f49d3f1e8f4e538` were separately checked for overlap and merge applicability.

Independent validation passed: 988 Client tests, the isolated PostgreSQL custody test that failed in CI, and the 360-test runtime coverage suite (100% in all four metrics; these tests overlap the Client suite). Frozen install, build, check, typecheck and diff checks passed.

Approval remains pending. Current CI run 34196500563 attempt 1 failed `Checks / Test PostgreSQL integration`: `im-binding.test.ts:5645`, "The Client custody fixture was not dispatched" (323 passed, 1 failed). A single isolated local pass does not establish flakiness or replace a passing current CI run.

The permission scope also needs an explicit maintainer decision: both visible and internal Sessions can now write the Computer-wide Context Tree configuration. The grant names one file, but changing it changes the Tree selected for other Agents when their provider runtimes next start. Please confirm that this shared configuration authority is intended. Workspace, Slack and connected-tree grants are preserved; Claude Code remains unrestricted as documented in the PR.

ggordonhall:
Yes, I confirm that both sessions can write to the context tree configuration file.

liuchao-001 (commented):
Reviewed at head `d3e124e84642969cc87e91e0e1d1ea78fa991127` (merge base `dcae5e42`), plus an independent local run of the four touched suites (`context-tree`, `context-tree.integration`, `session-runtime-manager`, `agent-runtime-exhaustive`): 79 tests, all passing.

The second commit is the more valuable half of this PR. A visible Session used to set `workspace.environment.PATH`, which replaced the whole factory-composed `PATH` and silently dropped both the Context Tree shim directory and the discovered executable directory. Moving that to `pathPrepend` and composing it inside each factory fixes a real regression, and the per-provider tests that assert the actual spawned process `PATH` are the right level to pin it at. No comments on that mechanism.

Five things on the rest.

**1. The one thing I would like answered before merge: this is the first writable root that is a file, and one that usually does not exist.**

Until now every root handed to a Provider Runtime was an existing directory — the workspace `cwd`, the Slack config leaf, the connected tree root. `homeLayout.contextTreeConfigFile` is neither: on the exact Computer this PR targets (unconfigured), `config/context-tree.json` is absent at Runtime start, and the parent directory is deliberately *not* granted.

The macOS evidence in the PR description is convincing for seatbelt, where `(subpath …)` on a non-existent path still authorises creating it. Codex on Linux sandboxes with Landlock instead, where a path rule is opened per path and creating a new file is authorised by `MakeReg` on the **parent directory**, not by a rule naming the file. So the plausible Linux outcomes are that the rule is dropped and `context-tree connect` is still denied — the goal of the PR unmet on Linux, with no test able to see it — or that sandbox setup errors on the unopenable path. Could you confirm the behaviour on Linux (the repo already runs Linux containers in CI) before this merges? If Landlock does drop it, the options are granting the `config` directory on Linux or pre-creating the file, and both deserve to be a deliberate choice rather than a discovery after release.

**2. `mkdir` at `session-runtime-manager.ts:264` is unguarded and sits outside the wrapped `try`.**

Everything else Context Tree related is explicitly optional: `context-tree.ts` states that nothing in the class throws into Session start, and `prepareContextTree` never throws. This `mkdir` does. `EACCES`, `EROFS`, `ENOSPC`, or a `config` path that is not a directory now fails `#startRuntime` with a raw `Error` — and because it is before the `try` that wraps failures in `ClientRuntimeProviderStartError`, `agent-turn-runner.ts:461` cannot classify it: the Turn is reported as `turn_state_unknown` / `executionEffects: "may_have_occurred"` even though no provider process was ever spawned. Every other pre-spawn failure in that function reports `provider_start_failed` / `not_started`.

Moving the call inside the existing `try`, or treating a failure as "do not grant the file" and continuing, would keep the classification honest.

**3. The early returns in `ensureAgent` bypass `#recordPreparation`, so `opentag doctor` loses a diagnostic it used to have.**

`PACKAGE_MISSING` and `SHIM_UNAVAILABLE` now return before `#joinPreparation`, which is the only caller of `#recordPreparation`. For a configured GitHub target that has not cloned, `readContextTreeState` reads that record to explain *why*; with no record written it falls through to `tree: "not-cloned"` — "the first Agent Session has yet to clone it" — permanently, for a Computer whose shim cannot be written. A stale record from an earlier reason is served in the same situation. Recording those two statuses before returning would restore it.

Related and smaller: an unconfigured Computer whose shim write fails now reports `unavailable(SHIM_UNAVAILABLE)` where it previously reported `unconfigured`. That is arguably more accurate — telling an Agent to run a command that is not on `PATH` helps nobody — but `docs/design/context-tree-integration.md` still says an unconfigured Computer is a normal state and not an error, and win32 (where the shim is unconditionally unsupported) now takes the fault branch for every Session regardless of configuration, logging a warning per Runtime start.

**4. The shim write moved onto the per-Runtime-start path and out from under the start budget.**

`#writeShim` used to run inside `#ensureAgentOnce`: serialized, cached per workspace, and raced against `#withinSessionStartBudget`. It now runs on every `ensureAgent` call ahead of the cache — an `ensurePrivateDirectory` walk plus a `writeDurableFile` with two `fsync`s per Provider Runtime start — and, being outside the budget, a stalled home directory stalls Session start with no 5s ceiling. Correct as written; just no longer bounded the way the surrounding design says it is.

**5. Documentation landed in the wrong section.**

The shim paragraph was updated, but the decision this PR actually changes lives two sections up ("Setup is explicit, once": *OpenTag never creates a Context Tree. A user creates or connects one…*) and in the writable-roots paragraph, which still reads as if the resolved tree path and the Slack config root are the only entries. Neither mentions that every Session now holds write authority over the Computer-wide configuration.

On the shared-authority question already raised: worth noting for the decision that the grant is on the raw file, not on the command, so an Agent can write `context-tree.json` directly and never go through `opentag context-tree connect` or its validation. Whatever lands there is what every other Agent's next Runtime start consumes, and what `doctor` reports. That does not change whether the capability is wanted — it changes what "an Agent may connect a tree" costs, and it belongs in the same maintainer answer.

liuchao-001 (changes requested):
Re-reviewed at head `bc3b8a4b1800c29d6f06ac8f18e5f747c0371ab4`. Ran the touched suites locally again: 68 Client tests (`context-tree`, `context-tree.integration`, `session-runtime-manager`) and 20 CLI tests (`context-tree`), all passing.

Four of the five earlier points are properly closed, and one of them is closed by real evidence rather than argument. Taking the Linux question to an actual sandboxed Codex run was the right move, and the run answered it: the file-named grant does not work there. The `ContextTreeManager` rework is also better than what I asked for — shared shim preparation, the whole pipeline back under the 5-second budget, the cooldown retry, and the record-replacement tests are all solid, and `bounds stalled shim preparation and shares its background work` is a genuinely good test.

Requesting changes on what the Linux fix grants, which I think is larger than the decision that was approved.

**Blocking: on Linux the grant is `<OPENTAG_HOME>/config`, which holds the Computer's identity and its machine credential.**

The authority that was raised, debated and confirmed in this PR was one file: `context-tree.json`. `prepareConfigurationRoots` now returns `layout.config` on Linux, and that directory also contains:

- `computer.json` — `computerId` and `serverUrl` (`computer-identity.ts:15`)
- `computer-credentials.json` — `machineToken`, `installationId`, `computerId` (`machine-credentials.ts:20`)

So on Linux every Agent Session, visible and internal, can now write the Computer's bearer credential and its server binding. The daemon and the provider process run as the same OS user, so the sandbox was the only thing standing between a Session and those files. At best a Session corrupts them and the Computer drops off until a human runs `opentag computer connect` again; at worst it substitutes an identity or a credential of its choosing. That reaches a Session under prompt injection from a repository or an IM message, and it applies to Computers that never use Context Tree at all, since the root is granted unconditionally.

The design doc names this as "changes to sibling configuration files", which is accurate but reads as a minor consequence. The siblings are the Computer's identity and credential, and I do not think the maintainer answer given earlier covers them.

The shape that removes the problem rather than documenting it is to stop sharing a directory with credentials: give the Context Tree config its own leaf — `<OPENTAG_HOME>/config/context-tree/config.json` — and grant that directory on Linux. The grant then contains exactly one file's worth of authority on both platforms, the platform split disappears, and Landlock has an existing directory to attach a rule to. It costs a read-fallback to the old path for Computers already configured, which is a smaller and more reviewable thing than a credential-adjacent grant. Pre-creating the file instead also works mechanically, but a zero-byte `context-tree.json` makes `readConfig` log an unreadable-configuration warning and `doctor` report "the configuration file is unreadable or invalid" on every fresh install, so I would not take that route.

**The platform split has no test authority.** Eight assertions in `session-runtime-manager.test.ts` now compute their expected value with `process.platform === "linux" ? "config" : "config/context-tree.json"` — the same expression the implementation branches on. Neither runner can fail on the platform decision, and no test states "on Linux the granted root is the directory" as an independent fact. `ContextTreeManager` already takes an injectable `platform`; `prepareConfigurationRoots` reading `process.platform` directly is what forces the tautology. Injecting it lets both branches be asserted on every runner.

**The mkdir failure log drops the only useful detail.** `err: error instanceof Error ? error.name : "unknown"` records `"Error"` for every `ErrnoException`. Whether the directory could not be created because of `EACCES`, `EROFS`, `ENOSPC` or `ENOTDIR` is the whole content of that event, and `error.code` carries it.

**The dated smoke appendix belongs in the PR, not the design document.** Container flags, a Codex build number, an aarch64 host and "This was not an unsandboxed container run" are evidence for this review; they are already stale the next time Codex changes its sandbox. The durable statement — Landlock authorises creating a file through its parent directory, so Linux grants the directory while macOS grants the file — is in the writable-roots section, which is where a future reader will look.

liuchao-001 (changes requested):
Re-reviewed at head `775dafc5f052d994855faac806d528001bac1e4c`. Ran the touched suites again: 71 Client tests (`context-tree`, `context-tree.integration`, `session-runtime-manager`, `home-layout`, `agent-home-context.integration`) and 67 CLI tests (`context-tree`, `doctor`), all passing.

The credential-adjacency problem is gone, and gone properly. `config/context-tree/` is granted on both platforms, so the platform split and its tautological assertions disappeared with it — the eight `writableRoots` assertions now state a single unconditional fact. The `code` on the mkdir warning and the removal of the dated smoke appendix are both right.

One thing came with the move that I do not think was intended, and I own half of it: I described the read-fallback as a cost of this shape without saying plainly that it is required. It is.

**Blocking: `context-tree.json` shipped in a published release, and this relocates it with no fallback.**

`open-tag@0.0.2` is on npm, and tag `v0.0.2` contains #455 — `contextTreeConfigPath` in that release is `<home>/config/context-tree.json`. Nothing in this branch reads that path any more; a repository-wide search for the old filename returns nothing.

Against this branch's built client, with a home holding exactly what 0.0.2 writes:

```
readConfig()   -> undefined
ensureAgent()  -> {"status":"unconfigured"}
```

So every Computer that has run `opentag context-tree connect` on the released build reports itself as never configured after this upgrade. There is no error, no warning, and no prompt: Sessions simply start without durable memory, `doctor` says the Computer has no tree, and the tree the user connected is still sitting on disk one directory up. Recovery is re-running `connect`, but nothing tells anyone to do that. It also reaches our own machines — the ones running source builds with trees connected today. The PR summary still says "The config location, schema, and connection command are unchanged" and "Breaking changes: None"; both were true two commits ago and are not true now.

The small fix is a read fallback in `readContextTreeConfig` and `ContextTreeManager.readConfig`: when the new path is absent, read the legacy one, and let the next `connect` write the new location. That keeps this PR's grant story intact and costs a few lines plus a test per reader.

If the maintainers would rather declare the released path unsupported given how new 0.0.2 is, that is a legitimate call — but it needs to be the stated decision, with the PR body and the design doc saying the location changed and what a configured Computer must do, rather than arriving as a silent regression. Either way I would like the claim in the summary corrected.

Two small things, neither blocking:

- The commit message is `fix`. `AGENTS.md` asks for Conventional Commits; the two commits before it in this branch follow that form.
- `config/context-tree/` sits beside `<OPENTAG_HOME>/context-tree/bin/`, which is the shim rather than configuration. Not worth churn now, but if a third Context Tree location ever appears, the naming will need a moment's thought.

packages/client/src/storage/home-layout.ts
```diff
@@ -33,7 +34,8 @@ export function resolveOpenTagHomeLayout(home = resolveOpenTagHome()): OpenTagHo
     config,
     // Holds the shim the Context Tree skills invoke by name; OpenTag's own calls bypass it.
     contextTreeBin: join(resolvedHome, "context-tree", "bin"),
-    contextTreeConfigFile: join(config, "context-tree.json"),
+    contextTreeConfigDir: join(config, "context-tree"),
+    contextTreeConfigFile: join(config, "context-tree", "config.json"),
```
yuezengwu, line 38: [P2] Preserve configured targets when relocating the configuration

This switches the runtime to `config/context-tree/config.json`, while the CLI reader makes the same switch and neither reads or migrates `config/context-tree.json`. For an existing Computer configured at the former path, an isolated probe on this exact head returns `undefined` from both readers, `{ status: "unconfigured" }` from `ensureAgent`, and no target from `readContextTreeState`. The old file remains intact and the CLI is never invoked; putting the identical JSON at the new path makes both readers recognize it. Thus restarting after upgrade silently removes durable memory from new Sessions on already-configured Computers. Preserve the existing target through a migration or compatible read path in both readers, test an old-only configuration and the precedence when both paths exist, and update the rollout documentation that currently says no migration is needed.

yuezengwu (changes requested):
Independently reviewed exact head `775dafc5f052d994855faac806d528001bac1e4c`. The isolated upgrade probe confirms the persisted-target regression also identified in the latest review; see the inline finding.

Validation: frozen install, build, check, typecheck and diff check passed. 1,003 Client tests and 433 CLI tests passed. One CLI suite initially could not resolve the Client build while a concurrent typecheck rebuilt dependencies; its 49 tests passed when rerun after the build settled, with no source changes.

The required Agent Runtime coverage command fails locally despite all 368 tests passing: branch coverage is 99.96% (2840/2841), below the required 100%. The uncovered branch is the non-Error arm at `session-runtime-manager.ts:516`. CI run [34212761092, attempt 1](https://github.com/first-tree-ai/opentag/actions/runs/34212761092) fails the same coverage step and percentage; its unit and PostgreSQL integration steps passed. Please resolve this check as well.

The current main (`42362fd0`) has seven non-overlapping changes; a merge-tree source audit preserves both sides. Local tests above exercised the exact PR head, not that merged tree.

liuchao-001:
Agreeing with both parts of @yuezengwu's review, and adding the shape I would use for the coverage failure, since the uncovered branch is one I asked for.

The uncovered arm at `session-runtime-manager.ts:516` is `error instanceof Error ? … : undefined`. `mkdir` rejects with an `ErrnoException`, so the false arm is unreachable, and the guard is not narrowing anything — the cast to `NodeJS.ErrnoException` is unchecked either way. Dropping the ternary keeps the diagnostic and removes the branch:

```ts
logger.warn(
  { code: (error as NodeJS.ErrnoException).code },
  "Context Tree configuration directory could not be created",
);
```

That is preferable to writing a test that throws a non-`Error` at `mkdir` to reach an arm production cannot produce.

On the relocation: the precedence case @yuezengwu asks for is the one worth pinning hardest. If both paths exist, the new one has to win, or a Computer that re-runs `connect` after upgrading keeps being served its pre-upgrade target.

liuchao-001 (changes requested):
Re-reviewed at head `0dacb63515f5c1a78ec853d5d2b22df3b45613a3`. CI is green at this head, including the coverage gate.

The relocation decision is now made properly rather than arriving by accident: `fix(client)!`, an upgrade section with the three exact `connect` forms, the "no migration is needed" line corrected, and both readers tested for an old-only home and for the case where both files exist — which is the precedence @yuezengwu asked for, since the new path wins while the old file stays untouched. No production code references the old filename any more. The coverage ternary is exactly the shape I suggested. Declaring the released path unsupported was one of the two legitimate answers and it is now a stated decision, so I have no remaining objection to the substance.

Two things left, both small.

**Blocking, and it is the one I asked for last round: the PR description still contradicts the change.** It currently says the runtime is granted "the exact `<OPENTAG_HOME>/config/context-tree.json` writable root", that "the config location, schema, and connection command are unchanged", and under Breaking changes, "None". The location moved to `config/context-tree/`, the commit is marked `!`, and every already-configured Computer must reconnect. This is the artifact a maintainer reads at merge and the one users will be pointed at afterwards, so it should say the location changed, what a configured Computer has to do, and that the grant is now the directory. The Testing section is also still describing only the macOS exact-file evidence; the Linux run and the relocation are the interesting parts now.

**`bounds stalled shim preparation and shares its background work` is time-dependent and will flake.** `expect(writes).toHaveBeenCalledTimes(1)` at `context-tree.test.ts:352` asserts, at one instant, that a side effect has already happened inside a 10 ms budget. But `#writeShim` runs `ensurePrivateDirectory` — a mkdir chain plus two `realpath` calls — before it reaches the gated `writeDurableFile`, and the budget clock starts before all of it. When that filesystem work does not fit in 10 ms, both `ensureAgent` calls still resolve `PREPARING`, so the first assertion passes and the next one sees zero calls.

I reproduced it once in eight runs of that single test locally, though under a load average around 42, so treat the rate as an upper bound rather than a prediction for CI. The mechanism does not depend on the load, only on the filesystem work fitting inside the budget. Awaiting the effect instead of sampling it keeps the intent — one shared write across two concurrent starts — without that dependency:

```ts
await vi.waitFor(() => expect(writes).toHaveBeenCalledTimes(1));
```

The rest of the suite is unaffected: 71 of 72 Client tests and 68 CLI tests pass at this head, and `shares successful shim preparation across concurrent and repeated starts` is safe because it runs on the default budget.

Non-blocking, and worth considering now that reconnecting is the sanctioned path: `readContextTreeState` returns a bare `tree: "unknown"` for a Computer upgrading from the released layout, so `doctor` reports no tree and says nothing about why. Noticing a legacy `config/context-tree.json` and adding a `detail` that names the reconnect command would turn a confusing upgrade into a self-explaining one, for about four lines in the one place users look first.

I will approve as soon as the description matches the change.

yuezengwu (commented):
Re-reviewed exact head `0dacb63515f5c1a78ec853d5d2b22df3b45613a3`.

The previous Runtime coverage failure is resolved: the local coverage gate now passes at 100% statements, branches, functions, and lines (368 tests). Local `pnpm check`, `pnpm build`, `pnpm typecheck`, Client tests (1004), and CLI tests (434) also pass.

I independently exercised the upgrade using an isolated temporary home and an injected bundled CLI:
- An existing `config/context-tree.json` alone is ignored; runtime preparation returns `unconfigured`, and doctor state is `unknown` without a reconnect explanation.
- The bundled CLI shim is available before configuration.
- Manual reconnect writes `config/context-tree/config.json` with private directory/file permissions. The next preparation uses the new target when both files exist, and the legacy bytes remain untouched.
- Session PATH composition preserves the factory's tool directories.

This confirms the behavior now explicitly documented by the breaking-change commit: upgrading an already configured Computer requires reconnecting, then restarting the daemon and affected Sessions. It does not migrate the existing connection automatically.

Before merge, the PR description must match that contract and the actual permission grant. It still says the exact old file is granted, the location is unchanged, and “Breaking changes: None.” I agree with the description blocker in review #5149274261; please describe the new directory grant, manual reconnect/restart steps, and current validation. That review's timing-sensitive test observation also remains applicable in the source; my successful suite run does not refute it.

The permission decision remains explicit: both visible and internal runtimes receive write access to the Computer-wide Context Tree configuration directory, so either can change the shared target consumed by later runtime starts. Acceptance of this authority and the breaking upgrade belongs to the maintainer; the technical checks above do not supply that decision. I am therefore leaving this head without approval.

Validation limits: this probe used no live provider credentials or real Context Tree. Current main advanced to `696b2de9d55842a3a7110fe6444cd07aeb3c65ed`; its overlapping composition-file change adds the outgoing-reply collector separately, and Git's merge simulation is conflict-free. The test counts above apply to the PR head, not a combined-current-main checkout.

liuchao-001:
The description rewrite is accurate and complete on everything I blocked on: the directory grant, the relocated path, why identity and credential files stay outside it, the breaking upgrade with the three reconnect forms and the restart/verify steps, and new-path precedence when both files exist. That blocker is cleared.

It has moved ahead of the branch, though. The Summary says the stalled-preparation test "waits for the shared write instead of assuming filesystem preparation completes within 10 ms", and Testing records 41 tests passing "after fixing the timing-sensitive assertion" — but the head is still `0dacb635`, and at that commit the assertion is unchanged:

```
$ git show 0dacb635:packages/client/src/__tests__/context-tree.test.ts | sed -n '353p'
      expect(writes).toHaveBeenCalledTimes(1);
```

So the fix is described in the present tense and the local run is recorded, while the commit carrying it has not been pushed. Testing does note that CI still has to run for the follow-up commit, so I read this as the push simply not having landed rather than anything more.

Push it and I will approve. Note separately that @yuezengwu is holding approval on the maintainer accepting the write authority and the breaking upgrade, which is a decision rather than a technical item, so that will still need an answer after my approval lands.

liuchao-001 (approved):
Approving at head `8556fad8fa9d29543f6976381d1113012e052d6c`.

The follow-up commit is the one-line change and nothing else, and it removes the dependency rather than widening a window: the assertion no longer samples the spy at an instant inside the 10 ms budget, it waits on it with vitest's default one-second ceiling, so filesystem preparation taking longer than the budget is no longer a failure. That reasoning is the substance of the fix; the runs support it without carrying it — 12 sequential and 6 concurrent runs of that test passed, but at a load average around 6, where the original failure came at around 42, so I would not read the count as a rate. All 72 tests across the five touched Client suites pass at this head.

Over the four rounds this PR went from a real regression fix wrapped in three quieter ones to something I am glad to have on main. Worth recording what got it there: the Linux question was answered by running an actual sandboxed Codex probe rather than by argument, and the answer overturned the original approach; the credential-adjacency finding was fixed by moving the configuration rather than by documenting the exposure; and the relocation that followed was turned into a stated breaking change with real upgrade instructions instead of being left to surprise people. The `ContextTreeManager` rework — shared shim preparation, the whole pipeline back under the start budget, cooldown retry, record replacement — is better than what I asked for.

Two things remain outside my approval:

- CI is still running for this head. The previous head was fully green including the coverage gate, and this commit touches one line of one test, but the merge gate is CI's to give.
- @yuezengwu is holding approval on the maintainer accepting the write authority and the breaking upgrade. That is a decision, not a technical item, and my approval does not answer it. Both visible and internal Sessions can change the Computer-wide target, and every already-configured Computer has to reconnect — someone with the authority to accept both should say so before this merges.

The non-blocking `doctor` suggestion from my last review still stands on its own: an upgrading Computer currently gets a bare `tree: "unknown"` with no mention of the reconnect it now needs. A follow-up is fine.

liuchao-001:
I reviewed this change from the other end — it arrived in #577's range before that PR was retargeted — and read this thread afterwards. I'm not reopening anything: @yuezengwu is already withholding approval on maintainer acceptance of the write authority, and my own approval above explicitly carves that out. This is one piece of input for whoever makes that call, because I think the consequence is sharper than "either can change the shared target," and the maintainer accepting it should be accepting a known shape.

The target is not only data. Reading it back does three things at once.

`readConfig()` validates against `ContextTreeConfigSchema` — shape, not provenance. `connectArguments()` turns a `github` target into `connect <owner/repo>`, which clones over the network on the 120s budget. The resolved tree path is then added to `writableRoots`, and its content is rendered into the system prompt as authoritative durable memory: "Read the decisions that bear on a task before planning or changing code, and record durable decisions there."

So one Session writing `{"schemaVersion":1,"target":{"kind":"github","repository":"attacker/tree"}}` into the granted leaf causes, at the next Session start for **any** Agent on that Computer including internal ones, an unattended clone of an attacker-chosen repository, a write grant to it, and its content presented to every Agent as trusted memory. The trigger is ordinary untrusted input — a repo README, an issue body, a dependency's output. No human is in the loop, and `doctor` reports the substituted target as healthy, so the substitution is not visible as a fault.

The part I'd weigh most: the blast radius is not the Session that wrote the file. A single injected Session relocates the shared memory of every Agent on the machine, and the natural place to notice would be `doctor`, which by design reports the new target as fine.

Two narrower shapes preserve the capability without the channel, if the maintainer wants the former but not the latter: grant the leaf read-only, or route the write through a daemon operation so command-level target validation still applies to an Agent-initiated change. There's also a plausible reading in which no grant is needed at all — OpenTag's own operations run as the daemon and aren't sandboxed, the CLI's connection store lives elsewhere, and `managed-instructions.ts` already tells the Agent not to reconnect a tree on its own.

None of that is a request to change this PR. The technical work here is good, and the thread shows the Linux question and the credential-adjacency finding were both settled by evidence rather than argument. I'd just rather the acceptance decision be made against the full chain than against the one-line description of it.

yuezengwu (approved):

liuchao-001:
With @yuezengwu's approval this is now mergeable: two approvals at `8556fad8`, CI green.

Flagging one thing before it merges, not to block it. The approval that opened the gate is empty-bodied, and it reverses an explicit position taken 75 minutes earlier — "Acceptance of this authority and the breaking upgrade belongs to the maintainer; the technical checks above do not supply that decision. I am therefore leaving this head without approval." My own approval carved out the same question. @bestony and @Gandy2025 were asked to review 30 minutes ago and have not commented.

So the gate is satisfied while the question both reviewers were holding for — a maintainer accepting that both Session kinds can rewrite the Computer-wide target, and that every configured Computer must reconnect — has no visible answer. If that decision was made out of band, saying so on the thread is enough and I have nothing further. If it was not, the approval count now conceals it rather than resolving it, and the acceptance is worth a sentence from someone who owns it.

@yuezengwu, was the empty approval a considered reversal on that point, or the head-level technical sign-off with the maintainer question still open?

liuchao-001:
Tracked as #578 for the hardening question, so it does not live only in a merged thread. It is scoped as hardening, not a revert: it takes the accepted authority as given and asks only that an injected Session not be able to point every Agent on the machine at attacker-controlled content that is then fetched, granted write access, and injected as trusted memory. The open question above about whether the acceptance was considered is unaffected and still stands on its own.

PR #571, opened 2026-09-08
fix(tasks): show actual sent Lark replies in Task history

### LINKED ISSUE #564: [Bug]: Task reply history must show actual Lark messages instead of Agent finalText
Task currently renders the Agent runtime's `finalText` as its reply. During production Lark regression, two successful Lark replies were missing from the DM Task because the Agent ended with an empty final answer. The page simultaneously showed `Completed` and “工作正在进行中。” (“Work is in progress.”).

**Confirmed product expectation: Task must display the actual replies sent to Lark, rather than substitute the Agent's terminal text output.** Similar wording in these two outputs does not make them the same source of truth.

### Reproduction steps

1. Connect a Lark bot to a Codex-backed Agent and send a DM.
2. Have the Agent reply using `lark-cli` and finish the turn without a separate nonempty final answer. This occurred naturally in both observed DM turns.
3. Confirm the Lark send returned success and a provider message ID.
4. Open the corresponding Task detail page: the Agent reply body is missing, although the turn is completed.
5. Compare another turn where the Agent both sends a Lark reply and produces a final answer: Task displays the final answer, which may only resemble the actual Lark message.

Affected production Task: [hi introduce urself](https://app.opentag.build/agents/092a06ad-2f5a-473b-b46e-e14d1d7f3dbe/tasks/3c4c4137-8438-4b05-9186-9002c84cd7fe), Agent `my-prod-tag-lark`.

### Expected behavior

- Task's reply history uses the actual Lark outbound messages associated with the task, preserving their content, order, and conversation/thread association. A sent reply remains visible whether the Agent's final answer is empty, identical, different, or absent.
- Agent terminal text must not be presented as a Lark reply unless it was actually sent. If retained as an execution summary, it must be clearly separate from the reply history.
- A completed turn with no reply or unavailable reply data has an accurate empty state; it must not show “Work is in progress” or invent a successful send.
- Correcting the placeholder alone is insufficient: the actual Lark reply content must be represented.

### Actual behavior

Evidence from 2026-09-08, UTC+8:

| Turn | Lark send | Codex final answer | Persisted OpenTag report | Task display |
| --- | --- | --- | --- | --- |
| DM `continue` | Success at 16:33:38, message ID returned | Empty | `outcome=completed`, `finalText` absent; report upload succeeded | Completed + “Work is in progress.” |
| DM work-summary request | Success at 17:09:02, message ID returned | Empty | `outcome=completed`, `finalText` absent; report upload succeeded | Completed + “Work is in progress.” |

A 16:41 topic-creation turn shows why other Tasks can appear correct:

- Actual Lark message: `可以，我们就在这个话题里聊。你想先讨论什么？`
- Codex final answer and persisted `finalText`: `我们就在这个话题里聊，你想先讨论什么？`

These are two separate outputs with different wording. The DM failure is not caused by losing a nonempty Codex final answer: the native final answer was genuinely empty after a successful tool send.

### Environment

- Production: `https://app.opentag.build`, macOS, OpenTag CLI `0.0.2`.
- Production CLI source: `db870b63b5c87840c825ab63c5154c345a143479`.
- Codex CLI upgraded to `0.153.4` before these successful turns; model `gpt-6-astra`.
- Message platform: **Lark, not Feishu**. The UI's separate branding defect is outside this issue.
- Live web entry: `index-iUEPvyTI.js`; Task bundle: `tasks-page-BEXTRuf8.js`.

### Logs and diagnosis

Sanitized local completion records:

```text
turn a188b5d6-8d00-43d4-97e5-b64af98657dc
  outcome=completed; hasFinalText=false; reportUpload=succeeded
turn 48f3b7a1-feea-4a8e-ad25-df77f04b9c23
  outcome=completed; hasFinalText=false; reportUpload=succeeded
```

The corresponding Codex records each contain a successful `lark-cli` send (`exit_code=0`, `ok=true`, provider message ID returned), followed by an empty `final_answer`. Send receipts establish platform acceptance, not that the recipient read the message.

Implementation evidence:

- [Codex final-text selection](https://github.com/first-tree-ai/opentag/blob/db870b63b5c87840c825ab63c5154c345a143479/packages/client/src/providers/codex/agent-runtime.ts#L1356) selects Agent messages, not the content sent by tools.
- [Turn report construction](https://github.com/first-tree-ai/opentag/blob/db870b63b5c87840c825ab63c5154c345a143479/packages/client/src/runtime/agent-turn-runner.ts#L404) includes `finalText` only when nonempty; [Task API projection](https://github.com/first-tree-ai/opentag/blob/db870b63b5c87840c825ab63c5154c345a143479/packages/server/src/services/tasks/task-service.ts#L243) exposes this report field.
- [Task reply rendering](https://github.com/first-tree-ai/opentag/blob/db870b63b5c87840c825ab63c5154c345a143479/apps/web/src/features/tasks-page.tsx#L540) falls back from absent final text/error to `delivery.state === "accepted"` and incorrectly displays ongoing work even with a completed report. The same branch was verified in the actual live Task bundle.
- The current [direct-provider-CLI contract](https://github.com/first-tree-ai/opentag/blob/db870b63b5c87840c825ab63c5154c345a143479/docs/direct-provider-cli.md#L38) says OpenTag does not receive outbound content, message IDs, or results. That contract and the necessary data path must be reconciled with the confirmed Task expectation above; changing wording or asking the Agent to repeat its reply in `finalText` does not satisfy it.

### Acceptance coverage

- Successful Lark reply with empty, identical, and deliberately different final output: Task shows the actual sent reply in every case.
- Nonempty final output without a successful Lark send: no fabricated Lark reply appears.
- Multiple replies and supported non-text replies retain their actual content or usable representation and correct task association across DM, group, and topic conversations.
- Completed, failed, and genuinely running turns have accurate state/empty-state rendering.

Investigation used the user-provided screenshot, scoped local execution/report records, and live static web code. Authenticated Task API JSON was not captured. No application code or production configuration was changed during this diagnosis.

yuezengwu:
## Summary

Task history previously used the Agent's final text, so a successful Lark send with empty final output appeared unfinished and a different final output replaced the actual reply. Report-bearing Feishu Turns now capture provider-confirmed sent messages through the official CLI launcher and display them as sent replies. Model output is separately labelled **Execution summary**.

- Keep receipts private by Home/Session/Run; match conversation and bound sender, deduplicate message IDs, preserve send order, and explicitly represent incomplete or unavailable content. Successful sends retain their exit behavior even when a read-only body lookup fails.
- Freeze the negotiated Turn report contract, preserve legacy hashes, canonicalize native payload keys for JSONB, and budget the full 64 KiB frame. Durable v2 reports wait for a current v2 connection across rollback/reconnect; a capability mismatch is nonfatal.
- Show work in progress only when the server's effective running predicate is true, including expiry, ended Sessions, and superseding deliveries. Slack Tasks do not show a permanent unsupported reply-data notice.

Closes #564.

## Review fixes

The launcher now exits naturally after stdout drains. Real launcher tests compare every byte of 3 MiB output through a slow consumer with capture enabled and disabled; read commands and non-capture plans inherit stdio. Capture requires an explicit immutable Feishu plan opt-in, so visible inbox callbacks leave no receipt Run.

Receipt collection re-reads capture status after inflight writers drain, and equal provider timestamps use capture sequence before message ID. Unknown success envelopes become incomplete evidence. Recovery preserves uninspectable nested evidence; opportunistic seven-day cleanup protects active Runs, inflight writers, unsafe files, and unknown entries. Regression tests also cover false/literal dry-run arguments, reconnect races, missing capability after restart, and active-plan conflicts during cleanup.

## Compatibility

The supported external official `lark-cli` range is tightened to **`>=1.0.92 <2.0.0`**, matching the required receipt envelope and raw API JSON surface. A 1.0.23 executable is rejected during selection. Optional report fields remain version-negotiated, existing report hashes remain valid, and no database migration is required.

## Testing

At `4c2a229ecdaaff06c41985d53a959a5e0549cc71`:

- `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass locally; workspace unit suites contain 271 files / 3,306 tests, plus 330 root script tests.
- `pnpm --filter @opentag/client test:agent-runtime:coverage` passes: 24 files / 370 tests, all four dimensions 100%.
- Current-head CI PostgreSQL integration passes: **18 files / 329 tests**. Local Docker could not start containers (`procReady not received` / resource unavailable), so local database assertions did not execute; CI supplies this validation.
- [Current-head CI](https://github.com/first-tree-ai/opentag/actions/runs/34304203147), [patch coverage](https://github.com/first-tree-ai/opentag/actions/runs/34304203218), browser/container smoke, CLI packaging, Node 22/24/26 compatibility, and PR image build all pass. Ownership approval remains pending.
- Web component tests cover effective running states, Slack gating, reply content, summaries, literal rendering, and accessibility. Earlier desktop/mobile screenshots use synthetic data and belong to the prior head; no new production visual acceptance is claimed.

Capture covers recognized launcher mutations. Historical sends, other-process/SDK sends, edits, and reactions are not backfilled. A crash before durable report handoff cannot reconstruct the original Task scope. Uninspectable evidence is intentionally retained. Real-account Lark send/reply and permission-failure acceptance remains outstanding; no production sends, merge, or deployment were performed.

## Checklist


liuchao-001 (changes requested):
Reviewed at head `7d5502c`. The contract work is careful — version-pinned negotiation, legacy hash preservation, key-canonical hashing, explicit unavailable/truncated states, and the frame budget that prefers real replies over the model summary. One blocking issue, then non-blocking notes.

## Blocking: the launcher can now silently truncate provider CLI output

`executeProviderCliTurnPlan` replaced `stdio: "inherit"` with a piped stdout that this process re-writes (`turn-runner.ts:143`, `outgoing-reply-process.ts:103`), and the launcher still ends with `process.exit(code)` (`turn-runner.ts:350-354`). Writes to a pipe are asynchronous, so whatever is still queued in the launcher's own `process.stdout` when it exits is discarded. Under `stdio: "inherit"` that loss was impossible: the child wrote to the inherited fd directly.

Reproduced locally with the same forward/pause/drain shape as `createStdoutCapture`, child emitting 5 MiB, consumer reading slowly:

```
# new shape (pipe + forward + process.exit)
pending=196608          # process.stdout.writableLength at close
received=5046272        # of 5242880 -> 192 KiB lost

# old shape (stdio: "inherit")
received=5242880
```

Backpressure pausing bounds the queue but does not empty it, so the tail is lost whenever the consumer is slower than the CLI — and the launcher's stdout is a pipe in production, since the Agent runtime captures tool output. The Agent then reads a truncated `lark-cli`/`slack api` result with no error. Any read command with a large JSON body (`--page-all` listings, docs fetches) is exposed, on both providers.

Suggested fix: set `process.exitCode = code` and let the process end after the streams flush, or await a stdout flush before exiting.

Related, and cheap: `spawnCapturedProcess` is used for every provider, while `withOutgoingReplyInflight` is `enabled` only for feishu. Slack invocations pay the pipe change plus a 256 KiB buffer for capture that never runs. Keeping the inherit path for non-feishu plans would shrink the blast radius of this change to the provider that benefits.

## Non-blocking

1. **Every Slack Task now shows "Reply data is unavailable."** `TaskCapturedReplies` (`tasks-page.tsx:577`) has no provider gate, and `#collectOutgoingReplies` returns `undefined` for any non-feishu provider, so `outgoingReplies` is permanently `null` there. For legacy Feishu reports that notice is accurate and self-resolving; for Slack it is a permanent notice about a capture that will never exist. `TaskTurnView` already receives `task`, so `task.source.provider` can gate the block (or select provider-specific copy). Same applies to Turns with no Turn plan (internal Sessions).

2. **Orphaned receipt directories are never reclaimed.** `recover()` keeps a Session directory whenever any run still holds evidence (`turn-plan-manager.ts:191`), and `cleanupOutgoingReplyRun` only removes the Run that reported. A crash between capture and durable report handoff leaves `runs/r-*` with message text on disk, and every later `recover()` re-preserves it — the retention is unbounded in both size and time. An age or count based sweep of Run directories inside a live Session would close it.

3. **`expectedSenderIds` never arrives in production.** Only `outgoing-reply-capture.test.ts:263` supplies it; no caller of `executeProviderCliTurnPlan` sets it, so `applySenderPolicy` (`outgoing-reply-capture.ts:154`) degrades to the `senderType` check and the bound-sender guarantee rests entirely on `isOutgoingReplyInScope` at report time. Either wire it from the providerRef or drop the parameter so the enforcement point is single and obvious.

4. **Dead branch** in `inspectLarkCliEnvelope`: line 235 is fully subsumed by line 236, and `parseLarkCliSuccessEnvelope` then hard-codes `identity: "bot"`. Worth collapsing to one check, and worth confirming the intent that an envelope without `identity` is rejected.

## Verified locally

Against the real `@larksuite/cli`, `lark-cli im +messages-send --dry-run --json` returns exactly the envelope this code expects (`ok: true`, `identity: "bot"`, `dry_run: true`, `data`), and every boolean flag on `+messages-send` / `+messages-reply` (`--dry-run`, `--json`, `--reply-in-thread`, `-h`) is in `BOOLEAN_FLAGS`, while the only pre-subcommand global flags are `--profile <v>`, `-h`, `-v`. So argv classification and envelope gating hold for the current CLI.

Not verified, and still the main live risk: the shape of a *successful* send/reply response. `extractSendReceipt` accepts `data.message_id` or `data.items[0].message_id`; if lark-cli nests the created message differently, every Turn degrades to "Reply history is incomplete" with no content. That is the one thing exact-head live QA has to exercise.

Gandy2025 (changes requested):
Reviewed range: `42362fd034e7e451f15c97e71f49d3f1e8f4e538...7d5502c4f1ef6d75a2f8c845f633193b890ad81b`  
Exact head: `7d5502c4f1ef6d75a2f8c845f633193b890ad81b`  
Risk tier: **HIGH** — this crosses provider output, local private storage, runtime protocol negotiation, persistent report data, server projection, and user-visible history.

Changes requested. The existing exact-head review's stdout-drain blocker remains valid; I am not duplicating it below. I found these additional blocking issues.

1. **[P1][Spec] Preserve capture order when provider timestamps tie** (`packages/client/src/runtime/provider-cli/outgoing-reply-store.ts:308-321`). The comparator uses provider `createTime` and then jumps directly to `messageId`, ignoring `sequenceHint`. The official Lark shortcut response formats `create_time` only to seconds, so two replies sent in one second can tie; if the body lookup fails, a first reply with ID `om_z` and a later reply with ID `om_a` is rendered in reverse order. Issue #564 explicitly requires multiple actual replies to preserve send order. Use `sequenceHint` as the tie-break before `messageId`, and cover same-second, reverse-lexical IDs with a failed body lookup.

2. **[P1][Spec] Render the real running predicate, not persisted `accepted`** (`packages/server/src/services/tasks/task-service.ts:439-445,697-718`; `apps/web/src/features/tasks-page.tsx:551-573`). The summary query correctly excludes expired, ended-session, and superseded accepted deliveries from `is_running`, but the Turn-detail projection does not expose that result and `TaskUnreportedBody` treats every unreported `accepted` Turn as “Work is in progress.” A Task can therefore be `Expired` at the top while its history says it is still running, contrary to the acceptance criterion's “genuinely running” distinction. Project the effective running state (or an equivalent terminal reason) into `TaskTurn` and cover expiry, ended Session, and a later accepted Turn.

3. **[P1][Compatibility/Spec] Use a raw-API output flag supported by the declared external Lark range** (`packages/client/src/runtime/provider-cli/catalog.ts:81-89`; `packages/client/src/runtime/provider-cli/outgoing-reply-capture.ts:321-346`). External `lark-cli` versions `>=1.0.0 <2.0.0` are accepted, but the follow-up message GET unconditionally appends `--json`. A compatible 1.0.23 binary rejects that command at argument parsing (`unknown flag: --json`); it exposes JSON through the default format or `--format json`. Because shortcut send receipts do not contain the message body, every otherwise successful send on that supported CLI degrades to unavailable content. Use the compatible raw-API form or tighten/probe the supported surface, with a lowest-supported-version regression.

4. **[P1][Data lifecycle] Do not capture receipts for visible collaboration Runs that have no report owner** (`packages/client/src/runtime/provider-cli/turn-runner.ts:137-176`; `packages/client/src/runtime/session-message-inbox.ts:307-322,350-358,386-404`). Capture is enabled for every Feishu Turn plan. Visible Session collaboration callbacks create those plans, but their normal terminal path only cleans up the plan and credentials; unlike `AgentTurnRunner`, it never collects or deletes the receipt Run. A normal successful send therefore leaves private message/body data on disk indefinitely, and startup recovery deliberately preserves it. Scope capture to report-bearing Task Turns, or give this path a durable consume/finalize cleanup, and assert that a successful visible callback leaves no receipt Run.

5. **[P1][Concurrency] Re-read capture status after inflight writers drain** (`packages/client/src/runtime/provider-cli/outgoing-reply-store.ts:136-160`). Collection reads `capture-status.json` before waiting for the inflight marker. A writer can then mark the capture incomplete and release its marker; collection sees no receipt but keeps the stale pre-wait status and returns `complete` with zero sends. That converts a known capture failure after a successful provider mutation into a false complete-zero-send claim. Read/merge the status after the wait and add a deterministic delayed-writer test.

6. **[P1][Protocol compatibility] Do not replay a durable v2 report on a cap1 connection** (`packages/client/src/runtime/agent-turn-runner.ts:122-130`; `packages/client/src/runtime/turn-report-owner.ts:330-350`). A Turn correctly freezes v2 when it starts, but the report owner sends every pending report after any registration without checking the current `runtime.turnReport` capability. On reconnect to an older/cap1 server, the strict legacy schema rejects `outgoingReplies`; the server closes with 4400 and the client treats that close as fatal. Thus a mixed-version reconnect or rollback can kill the whole runtime, contradicting the negotiated no-breaking-change claim. Keep v2 reports durable but send them only over a current cap2 connection; cover cap2 Turn -> cap1 reconnect -> cap2 reconnect. The server should also reject a capability mismatch without turning one report into a transport-fatal failure.

7. **[P1][Recovery safety] Preserve evidence when the nested receipt directory cannot be inspected** (`packages/client/src/runtime/provider-cli/outgoing-reply-store.ts:197-222`; `packages/client/src/runtime/provider-cli/turn-plan-manager.ts:187-196`). Recovery preserves an unavailable Run directory at the first level, but treats an unavailable `outgoing-replies` listing as “no evidence.” It then returns false and recursively removes the Session. Transient I/O/permission trouble or an unsafe nested directory can therefore make recovery delete provider-confirmed evidence it could not inspect. Only `missing` should mean absent; `unavailable` must preserve the Run. Add an unreadable/unsafe nested-directory recovery test.

Validation performed on this head:

- focused Client suites: 4 files / 46 tests passed;
- focused Web suites: 2 files / 61 tests passed;
- repository CI and patch coverage passed; the ownership gate is blocked by requested changes;
- external `lark-cli` 1.0.23 reproduced the raw-API `--json` rejection locally.

The JSONB/account-isolation path and complete-frame budgeting did not reveal another blocker. No production Lark send was performed, so the successful live send/reply response shape remains a QA requirement after the deterministic issues above are fixed.

liuchao-001:
Following up on @Gandy2025's point 3, with the older CLI actually installed. I confirmed it and found the failure is worse than degraded content.

Managed `lark-cli` is 1.0.92 (`catalog.ts`), where `api GET ... --json` works and the envelope is exactly what the code expects. But `compatibility: ">=1.0.0 <2.0.0"` also admits a user-selected 1.0.23, and on that binary:

```
$ lark-cli api GET /open-apis/im/v1/messages/om_probe --as bot --json --dry-run
Usage:
  lark-cli api <method> <path> [flags]
Flags:
      --format string    output format: json|ndjson|table|csv (default "json")   # no --json

$ lark-cli api GET /open-apis/im/v1/messages/om_probe --as bot --dry-run
=== Dry Run ===
{ "api": [ ... ] }        # no ok/identity envelope at all

$ lark-cli im +messages-send --chat-id oc_probe --text hi --dry-run
=== Dry Run ===
{ "api": [ { "method": "POST", "url": "/open-apis/im/v1/messages", ... } ] }
```

So on 1.0.23 the shortcuts exist and `classifyLarkOutgoingMutation` still recognizes the send, but no output on that CLI carries `ok: true` / `identity: "bot"`. `inspectLarkCliEnvelope` returns `rejected`, and `captureFeishuOutgoingReply` maps `rejected` straight to `return "ignored"` — **without** `markIncomplete`. A successful send on a supported CLI therefore produces no receipt and no incomplete marker, and the snapshot reports `status: "complete"` with zero replies. Task history then states "No sent replies recorded." for a Turn that did send. That is the exact false claim this PR exists to remove, so it should not be reachable through a version the catalog declares compatible.

The narrow fix is to separate the two meanings currently folded into `rejected`: a dry run or a non-bot identity legitimately means "not our send", but an unparseable or unrecognized envelope after a recognized, successful mutation is evidence loss and must mark the Run incomplete. Tightening the compatibility range or probing the envelope surface would close the rest.

`--json` on the shortcuts is unaffected (1.0.23 accepts it there); only the raw-API path and the envelope contract break.

Gandy2025:
Confirmed against the explicit external 1.0.23 binary (the managed 1.0.92 launcher is now earlier on this machine's PATH). My finding 3 understated the impact.

On 1.0.23, the shortcut exists but emits the legacy document without the `ok` / `identity` envelope. For a recognized mutation with exit code 0, `inspectLarkCliEnvelope` therefore returns `rejected`, and `captureFeishuOutgoingReply` returns `ignored` without marking the Run incomplete. The raw follow-up cannot recover it either: `api ... --json` is unsupported, while the default JSON output is also the legacy non-envelope shape.

So the exact failure is not merely unavailable content: the capture can report `complete` with zero replies after a real send, and Task history can state “No sent replies recorded.” Point 3 should be read with that stronger consequence.

The fix needs both parts described above: a recognized successful mutation whose receipt envelope is missing/unrecognized must become incomplete evidence rather than ignored, and the advertised external compatibility range must be tightened or probed for the envelope/raw-API surface the capture requires.

Gandy2025 (approved):
Re-reviewed current base / merge-base `8cc14e38786b53fdd9c876167eecf2fc3f40e890` through exact head `4c2a229ecdaaff06c41985d53a959a5e0549cc71`.

Risk tier remains **HIGH**. I reviewed the fix delta from `7d5502c` and the affected adjacent capture, storage, runtime-protocol, server-projection, and Task-rendering boundaries. I also reused the unaffected evidence from the original full review.

## Standards: pass

The prior stdout-drain blocker is fixed: non-capture/read paths inherit stdio again, the capture launcher exits naturally after flushing, and the real-launcher regression verifies every byte of 3 MiB output through a slow consumer with capture both enabled and disabled.

The data-lifecycle and protocol corrections also hold. Capture is now an explicit immutable Feishu/report-v2 plan opt-in, so visible collaboration callbacks do not create orphan receipt Runs. Durable v2 reports wait for a current cap2 connection, capability races are fenced at both the report owner and transport, and server-side mismatches return a nonfatal business result.

## Spec: pass

All seven blockers from my prior review are fixed on this head:

- equal provider timestamps use capture sequence before message ID;
- Task history uses the server's effective running predicate;
- external Lark support is narrowed to `>=1.0.92 <2.0.0`, and unrecognized success output marks capture incomplete;
- visible collaboration Runs no longer capture Task reply receipts;
- capture status is read after inflight writers drain;
- v2 reports are not replayed onto cap1;
- uninspectable nested receipt evidence is preserved during recovery.

The Slack-only reply-history notice and the earlier dead/unused sender-policy branch are also resolved. No new blocking correctness, security, or compatibility finding was found.

## Validation

- focused Client: 9 files / 185 tests passed;
- focused Web: 3 files / 89 tests passed;
- focused Server: 2 files / 55 tests passed;
- `pnpm check` and `git diff --check` passed locally;
- current-head CI, PostgreSQL integration, patch coverage, browser/container smoke, CLI packaging, Node 22/24/26 compatibility, and PR image build pass.

Non-blocking lifecycle note: the seven-day sweeper treats every nonempty inflight directory as active. A launcher crash after writing its PID marker but before releasing it can therefore preserve that Run indefinitely; the PID/marker age is not currently checked. This is the remaining narrow form of the earlier orphan-retention note, not an Issue #564 acceptance blocker.

Real-account Lark send/reply and permission-failure acceptance remains outstanding as the PR states. This approval covers the reviewed code and deterministic evidence; it does not claim that live provider QA was performed.

liuchao-001 (changes requested):
Re-reviewed at exact head `4c2a229e`. Every finding from my previous review is fixed in shipped code, and I verified each one rather than reading the summary:

- launcher exits with `process.exitCode` + `flushStdout()` instead of `process.exit(code)`, and non-capture plans / non-mutation argv go back through `spawnInheritedProcess` with `stdio: "inherit"`, so read commands never touch the pipe at all — a tighter scope than I suggested;
- `TaskCapturedReplies` is gated on `task.source.provider === "feishu"`, so Slack Tasks no longer carry a permanent unavailable notice;
- `expectedSenderIds` is gone, leaving `isOutgoingReplyInScope` as the single enforcement point;
- the dead `identity` branch is collapsed, and unknown envelopes now return `malformed` → `markIncomplete`, which is exactly the split I asked for in the follow-up comment. With `>=1.0.92 <2.0.0` also rejecting 1.0.23 at selection, that false "complete, zero sends" path is closed twice over.

`is_running` being defined once in the `executions` CTE and consumed by both the topic aggregate and the Turn projection is the right shape — Task status and Turn history now cannot disagree.

One blocking item remains, and it is narrow.

## Blocking: the 3 MiB drain regression cannot fail when production regresses

The new test spawns `prepared.launcherPath`, and in tests `runnerInvocation` resolves to `__tests__/fixtures/provider-cli-turn-runner-entry.ts`. That fixture got its own copy of the fix in this commit:

```ts
// fixtures/provider-cli-turn-runner-entry.ts
void runProviderCliTurnRunner(argv, { plansRoot }).then(
  async (code) => { process.exitCode = code; await flushStdout(); },
  async () => { process.exitCode = 1; await flushStdout(); },
);
```

Production ships a second, hand-duplicated copy: `resolveProviderCliTurnRunnerInvocation()` returns `[execPath, <turn-runner.js>]`, so the real launcher enters `turn-runner.ts:357` under `isProviderCliTurnRunnerMain` — and nothing spawns that entry in any test.

I reverted **only** the production block back to `process.exit(code)`, changing nothing else, and ran the suite:

```
$ vitest run src/__tests__/provider-cli-turn-runner.test.ts
Test Files  1 passed (1)
     Tests  16 passed (16)
```

Green, including both `drains multi-MiB stdout to a slow consumer with capture=true|false`. So the guard on the one blocker from my last review asserts a twin of the shipped code, and the exact 192 KiB silent-truncation regression can return with a fully green suite. The PR body's "Real launcher tests compare every byte of 3 MiB output" reads as coverage of the shipped exit path; it is coverage of the fixture's.

To be precise about what *is* covered: the `capture=true` variant does exercise production forwarding and backpressure in `spawnCapturedProcess`, and the `capture=false` variant does exercise the real inherit-vs-pipe selection in `executeProviderCliTurnPlan`. Only the exit/flush half is duplicated, and that half is what caused the bug.

Suggested fix, roughly five lines: export the terminal handler from `turn-runner.ts` —

```ts
export async function finishProviderCliTurnRunner(code: number): Promise<void> {
  process.exitCode = code;
  await flushStdout();
}
```

— call it from both the production `isProviderCliTurnRunnerMain` block and the fixture, so the drain test covers the shipped path. Alternatively, spawn the production entry directly in one case. Either way the duplicate disappears.

## Verified locally at this head

- `provider-cli-turn-runner` suite: 16 tests pass as shipped; 16 still pass with the production exit path reverted, which is the finding above.
- Read the full delta for the sweep, plan opt-in, capture-status ordering, receipt tie-break, recovery `unavailable` handling, and the cap2 send gate in `#canSendReport`; all consistent with their stated intent, and pending v2 reports correctly flush when a cap2 connection returns.

I agree with @Gandy2025's non-blocking note that a nonempty `inflight` directory is treated as active with no marker-age or liveness check, so a launcher crash inside that window still pins a Run forever. That is a much smaller residue of my earlier retention finding and I do not consider it a merge blocker.

Live Lark send/reply acceptance is still outstanding, as the PR states; nothing here claims otherwise.

Gandy2025 (changes requested):
Correction to my approval at the same exact head `4c2a229ecdaaff06c41985d53a959a5e0549cc71`.

I independently reproduced liuchao-001's remaining blocker. The launcher tests use `__tests__/fixtures/provider-cli-turn-runner-entry.ts`, which duplicates the terminal `process.exitCode` + `flushStdout()` logic instead of entering the production `isProviderCliTurnRunnerMain` block in `turn-runner.ts:357-367`.

I changed only the production block back to `process.exit(code)`, left the fixture and all other shipped code unchanged, rebuilt the shared package, and ran:

```text
vitest run src/__tests__/provider-cli-turn-runner.test.ts
Test Files  1 passed (1)
Tests       16 passed (16)
```

Both 3 MiB slow-consumer cases remained green. The test therefore covers the production forwarding/inherit selection but cannot fail when the shipped exit/flush path regresses—the exact half that caused the original silent truncation.

My previous statement that the real-launcher regression covered the shipped terminal path was incorrect. **Spec remains pass and the eight product blockers remain fixed, but Standards is blocked until this regression executes the production terminal handler.** Exporting one terminal-finisher function and calling it from both production and the fixture, or spawning the production entry directly, is sufficient.

This review changes my verdict; it is not an additional distinct product finding. The earlier non-blocking stale-inflight-marker note and outstanding live Lark QA remain unchanged.

PR #573, opened 2026-09-08
fix(runtime): preserve Provider CLI retry budgets

### LINKED ISSUE #490: fix(runtime): settle permanently unrepairable Provider CLI artifacts
## Implementation status

Implemented by #573, merged into `main` on 2026-09-09 as [`acf7609b9bfdd5fc999ebbc23ef68fc2eb1a7108`](https://github.com/first-tree-ai/opentag/commit/acf7609b9bfdd5fc999ebbc23ef68fc2eb1a7108). This issue was automatically closed as completed by the merge. The acceptance checklist below records implementation and automated-test completion; deployment and live-provider validation are separate.

## Summary

After #477, active-Integration recovery correctly re-drives transient Provider CLI failures instead of retaining a non-expiring unavailable observation. However, a diagnostic that OpenTag cannot repair automatically can now enter a slow dispatch loop rather than settling in a stable terminal state.

## Current mechanism

The Client reconciler treats diagnostics such as unsupported_platform and global_bin_unavailable as unrepairable and publishes unavailable without running another managed ensure.

The Server recovery path treats a retained unavailable artifact as recoverable. Once the bounded retry is no longer armed, the next demand read retires the request and dispatches a new one. That new request receives a fresh retry budget. Repeated onboarding or readiness reads can therefore produce this cycle:

1. dispatch;
2. inspect an unrepairable diagnostic;
3. publish unavailable;
4. exhaust the bounded retry;
5. retire and dispatch again with a fresh budget.

This was identified as a non-blocking follow-up in the exact-head review of #477.

## Expected behavior

Distinguish transient Provider failures from terminal or manual-action diagnostics at the Server protocol boundary.

A terminal observation remains stable under the existing Computer, Agent, Integration, credential-generation, request, and connection fences. An explicit retry or a real connection, credential, or placement change may start a new bounded recovery. Periodic read-only Client inspection can discover a local repair on the same connection, but an inspection or ordinary readiness read must not replenish exhausted repair/validation budgets. No new local-environment generation is required.

## Acceptance criteria

- [x] Terminal or manual-action diagnostics do not automatically receive a fresh retry budget on every later readiness read.
- [x] Transient failures, including operation_in_progress and connection loss, continue to converge automatically.
- [x] An explicit repair/retry action can re-arm a terminal observation.
- [x] Relevant placement, credential, connection-instance, or local Provider-state changes can revalidate the observation with the existing fencing rules.
- [x] Web and Agent-facing status expose a stable allowlisted diagnostic with actionable guidance derived from its reason, rather than an endless checking/unavailable cycle. No additional Provider CLI `nextAction` DTO is required.
- [x] Tests cover both the terminal path and recovery after an explicit or relevant state change.

## References

- #477
- Exact-head reviewer follow-up: https://github.com/first-tree-ai/opentag/pull/477#pullrequestreview-5089355028

yuezengwu:
## Summary

Repeated readiness reads could replace a failed Provider CLI requirement and restart its retry budget. Retain terminal observations and cumulative artifact/credential budgets during one unsuccessful recovery. Client periodic inspection can still discover a local repair on the same connection; explicit retry opens a new bounded recovery.

- Serialize requirement transitions per integration, invalidate retired work immediately, and fence delayed binding reads, queued dispatches, and send failures against their actual owner. Passive GETs do not invalidate explicit retries, and unrelated Agents remain independent.
- Set artifact readiness to `unavailable` when delivery failures exhaust the budget. Duplicate failure frames do not consume additional scheduled retries.
- Negotiate artifact reasons through runtime reconcile v2 and the separate HTTP opt-in `x-opentag-provider-cli-reason: 2`. Computer list, Agent setup, diagnostics, and direct handoff preserve the old strict response shapes for unmarked/v1/unknown callers, including existing credential reasons. The readiness header remains `x-opentag-provider-readiness: 1`.
- Keep allowlisted reasons and their user-facing copy; remove the redundant new Provider CLI `nextAction` fields and exports. Existing local preparation actions remain available.

Fixes #490.

## Validation

- `pnpm check`, `pnpm build`, and `pnpm typecheck` passed.
- `pnpm test` passed: 3,258 workspace tests and 330 root script tests.
- Client agent-runtime coverage passed: 360 tests, 100% statements/branches/functions/lines in the required scope.
- PostgreSQL integration passed: 324 tests against disposable containers.
- Independent controlled runtime checks passed: 15 cases covering delayed cancel/query/send completion, generation/placement/unbind changes, stale queued work, duplicate notifications, explicit retry vs GET, terminal delivery failure, and runtime v1 reasons.
- HTTP injection passed 22 compatibility checks using the full frozen pre-PR Shared parsers from `8cc14e38` across all four affected surfaces.
- Earlier cold-start/read-only inspection, late grant, TTL, credential backoff, and concurrent Agent retry regressions passed again.

Provider observations are synthetic; no live provider credentials were used. No database migration.

## Checklist


Gandy2025 (changes requested):
Requesting changes on exact range `8cc14e38786b53fdd9c876167eecf2fc3f40e890..d28f4daf865458a36025019dd0de74b8ef86916f`.

## Standards

- **[P1] Preserve the strict HTTP v1 response shapes during the server-first rollout.** The old `ComputerImCliReadinessSchema` and `ImBindingDiagnosticsSchema` are strict, but the new server unconditionally projects `reason` / `providerCliReason` / `providerCliNextAction` when the artifact is unavailable (`packages/server/src/services/computers/provider-readiness.ts:58`, `packages/server/src/services/im-bindings/im-binding-service.ts:300`, and `:1062`). The Computer list still uses the existing v1 request marker, while Agent setup and binding diagnostics have no response-version opt-in at all. Therefore an already-deployed CLI or stale Web bundle talking to the new server rejects exactly the failure responses this change is meant to explain as invalid responses. This conflicts with the repository's frozen-v1/server-first rolling-upgrade contract and with the PR's “no breaking changes” claim. Please add an explicit v2 opt-in to every HTTP surface that can return the new fields (Computer list, Agent setup, and binding diagnostics), keep v1/unmarked responses on the old shape, and make the new Client/Web request v2. Relaxing only the new parsers cannot protect deployed strict parsers.

## Spec

- **[P1] Do not let an older retry overwrite a newer requirement generation.** `#dispatchRequirement` deletes the keyed request and then awaits the cancel send before installing its replacement (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:448`; `#retireRequest` at `:1147-1151`). During that await, a credential-generation or placement change can install a newer request; when the older continuation resumes it overwrites the newer entry. A controlled delayed-cancel reproduction produced `requirement(gen1) -> cancel(gen1) -> requirement(gen2) -> requirement(gen1)`, after which a valid gen2 ready result was rejected and readiness remained `checking`. An ordinary readiness read in the same window also emitted an extra requirement. Please serialize the transition per Computer/integration or add a monotonic transition epoch with post-await stale-commit checks, and cover delayed cancel against generation, placement, and readiness-read races.

- **[P1] Persist terminal `unavailable` when requirement delivery exhausts its budget.** Dispatch writes artifact `checking` and credential `unconfirmed` before sending (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:468-507`), but the exhausted branch in `#scheduleArtifactRetry` only marks the in-memory episode stopped (`:933-936`). With `maxRetries = 0` and a failed send callback, later readiness reads send nothing and the observations remain `checking` / `unconfirmed` even though no work is in flight; TTL expiry still cannot recover because the episode is stopped. Please settle the exact current artifact observation to generic terminal `unavailable` before stopping, and cover both zero-budget and final-retry send failures.

- **[P2] Charge the artifact budget once per actual retry, not once per replayed status.** Every accepted transient `unavailable` reaches `#scheduleArtifactRetry`, which increments the attempt and replaces the timer without checking `artifactRetryPending` (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:930-954`). Three identical statuses for one request with `maxRetries = 2` exhausted the episode before any replacement requirement was sent. The Client can produce this legitimately when concurrent periodic-refresh continuations reach status publication before the first send records its dedupe state. Please make an already-armed retry idempotent for repeated transient status on the same current request, preserve manual/state-changing transitions, and add a replay test proving two real redispatches remain available.

Non-blocking compatibility note: the server currently consumes `artifact.reason` semantics even when `runtime.providerCliReconcile` negotiated v1, and the new test fixture explicitly exercises that path. The direct-provider CLI contract says reasons are a v2-negotiated addition and v1 should retain a generic bounded episode. Please gate or explicitly document that receiver behavior while addressing the compatibility work above.

All functional checks reported on this exact head are green; the ownership gate remains pending/failed for the required code-owner approval. The three runtime findings above were reproduced against the exact head with controlled timing and the worktree remained clean.

liuchao-001 (commented):
Reviewed exact head `d28f4daf865458a36025019dd0de74b8ef86916f` against base `8cc14e38786b53fdd9c876167eecf2fc3f40e890` in a fresh worktree.

I read this after @Gandy2025's review. I independently reproduced the reasoning behind all three of their blocking findings, so I am not re-filing them — this comment records what I confirmed and adds one finding they did not cover.

## Confirmed independently

- **Strict HTTP v1 response shapes (their Standards P1).** Confirmed. `ImBindingDiagnosticsSchema` is `.strict()` and is parsed by both deployed surfaces — `packages/client/src/api.ts:512` and `apps/web/src/api.ts:370` — while `packages/server/src/services/im-bindings/im-binding-service.ts:1062` now projects `providerCliReason` / `providerCliNextAction` with no version gate. `ComputerImCliReadinessSchema` is likewise strict and gains `reason`, and the `x-opentag-provider-readiness` header on the Computer list gates *which* readiness is returned, not the new field. The existence of that header is itself the precedent: it was introduced to keep old strict CLI parsers working when Provider readiness was first added, and the same protection is missing here. `ProviderCliHandoffProgressSchema.reason` is additionally *widened* from `IntegrationCredentialExecutionReasonSchema` to `ProviderCliPublicFailureReasonSchema`, so an old peer fails on the enum value even before it fails on the unknown key.

- **Overwrite of a newer requirement across the awaited cancel (their Spec P1).** Confirmed by reading. `#dispatchRequirement` awaits `#retireRequest(key)` — which awaits `#sendCancel` on the wire — before `this.#requests.set(key, …)` (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:448-467`). Nothing holds the key across that await. The same window also affects the send-failure path: the `catch` at `:504-506` calls `#scheduleArtifactRetry(this.#requests.get(key))`, so a failed send for the older dispatch can charge the *newer* request's budget.

- **Terminal state not persisted when delivery exhausts its budget (their Spec P1).** Confirmed. Dispatch writes artifact `checking` + credential `unconfirmed` at `:468-492` before the send, and the exhausted branch of `#scheduleArtifactRetry` (`:933-936`) only flips `artifactRepairStopped`. Because `providerCliArtifactReadiness` exempts `unavailable` from TTL but not `checking` (`packages/server/src/runtime/connection-registry.ts:302`), the stranded `checking` never ages out into anything terminal either, and `#refreshRequirement` returns early on `#episodeStopped`. The result is a permanent "checking" — which is precisely the endless-checking state #490's acceptance criteria rules out.

## Additional finding

**[P2] `nextAction` is write-only across the whole change.** `ProviderCliPublicNextAction` is added to three response schemas (`ProviderCliHandoffProgressSchema.nextAction`, `ImBindingDiagnosticsSchema.providerCliNextAction`) and is computed on every unavailable artifact via `classifyProviderCliArtifactFailure`, which always returns a value (`"repair_cli"` whenever the failure is not manual). No consumer renders it: `grep -rn "nextAction|providerCliNextAction" apps/web/src packages/client/src` (excluding tests) returns nothing. The Web's "concrete action" text comes entirely from `providerCliWaitingCopy` mapping `progress.reason` to a copy string; the only reader of `providerCliNextAction` anywhere is the raw tab-separated dump in `apps/cli/src/core/agent/im.ts:89`, which prints the bare enum token rather than an action.

So the PR's "manual failures expose allowlisted reasons and concrete actions in Web and Agent diagnostics" is delivered by `reason` alone. `nextAction` currently adds new wire surface — and is one of the two fields that breaks the deployed strict parsers in the finding above — for no behavior. Either render it (the summary/detail rows in `preparation-readiness.ts` are the natural place) or drop it and keep the reason→copy mapping as the single source of the next step. Note that `ImBindingDiagnostics.providerCliReason` is in nearly the same position: only the CLI diagnostics dump reads it.

## Non-blocking notes

- `#armTimer` gives artifact retries and grant retries a single shared `current.retryTimer` slot, but `artifactRetryPending` is cleared only by the artifact timer's own callback (`:941-955`). Arming a grant retry while an artifact retry is pending would therefore strand `artifactRetryPending = true` forever, and both `#refreshRequirement` (`:1071`) and `#shouldHoldForceRedispatch` (`:530`) return early on that flag with no terminal settle. I could not construct a reachable path on this head — `#invalidateCredentialOnArtifactFailure` forces `grantConsumed`, which blocks the validation results that would arm a grant retry — so this is latent rather than live. Separate timer slots, or clearing `artifactRetryPending` inside `#armTimer`, would remove the class.
- `#handleValidation` re-checks `frame.status === "ready" && !this.#artifactIsReady(current)` (`:675`) immediately after `#acceptsGrantResult` already returned `false` for exactly that condition (`:1014`). The second check is unreachable.
- Client-side, `refreshPublishedImCliReadiness` now inspects every *active-requirement* provider on each capability-monitor tick, not just published ones (`packages/client/src/runtime/provider-cli/reconciler.ts:139-145`). That is what makes local-repair discovery work, but it also means a steady-state healthy Computer now spawns a CLI inspection per provider per tick indefinitely. Worth confirming that cadence is intended.

## What I checked and found sound

The client-side overlap guards read carefully to me: `#passiveOverlappedRepair` bails on a repair job present at either end, on a `#readySelection` identity change, and on a `lastStatus` change under the same `requestId`, and `#publishPassiveInspection` re-checks liveness after the async `readySelectionFromInspect`. Preserving `requirement.lastReason` over a fresh non-manual `inspectReason` is the right call — publishing a reasonless `unavailable` over a terminal one would restart the server budget. The inspect/ensure asymmetry for `version_incompatible` is applied consistently on both sides, and the `superRefine` on `ProviderCliArtifactStatusFrameSchema` composes correctly inside `ClientRuntimeBusinessFrameSchema` under Zod 4.

liuchao-001 (approved):
Re-reviewed exact head `04dd3f17741e1cf41c6746ec10970cf0c5fa2a5b`, reading `d28f4daf..04dd3f17` in a fresh worktree.

All four findings from the previous round are resolved. Details of what I checked, since several of these are the kind of fix that can look right in the summary and still leave a hole:

**My [P2] `nextAction` write-only — resolved by removal.** `PROVIDER_CLI_PUBLIC_NEXT_ACTIONS`, the schema, the type, both `index.ts` exports, the `ProviderCliHandoffProgressSchema.nextAction` and `ImBindingDiagnosticsSchema.providerCliNextAction` fields, the `providerCliHumanNextAction` mapping, the `classifyProviderCliArtifactFailure` return member, and the CLI dump line are all gone. Removal was the right call over rendering it — the reason→copy mapping in `messaging-readiness-copy.ts` already carries the user-facing next step, so the field was pure duplication.

**Strict HTTP v1 shapes — resolved, and it fails safe.** The new `x-opentag-provider-cli-reason: 2` opt-in is separate from the readiness header, and `requestsProviderCliReasonV2` matches only the exact string `"2"`, so unmarked, `1`, and unknown values all fall back to the old shape rather than the new one. All four affected surfaces are gated at their single `.parse().send()` site: `account.ts:171`, `agents.ts:138`, `im-bindings.ts:63`, `im-bindings.ts:122`. I checked that those are in fact the only send sites — `ComputerImCliReadinessCollectionSchema` reaches HTTP only via `account.ts` and `agent-setup.ts`, both covered. `projectProviderCliHandoffProgressForHttp` correctly keeps credential reasons (valid in v1) while stripping artifact reasons, which is the distinction the widened `ProviderCliHandoffProgressSchema.reason` enum would otherwise break. Web requests v2 on all four; the CLI requests it on the two it consumes, and the two it does not are safe by construction.

I also spot-checked the frozen-parser test rather than trusting the label: `LegacyImBindingDiagnosticsSchema` in `provider-cli-http-v1.test.ts` is a faithful transcription of the real `8cc14e38` schema — same key set, same `.strict()`, only `slackIdentityClosure` relaxed to `z.unknown()`, which is not the field under test.

**Requirement transition race — resolved.** `#withRequestGate` serializes per `computerId:integrationId`, and the gate itself is correct: each waiter chains on the previous promise and releases in a `finally`, with the map entry deleted only by the last holder. `#retireRequest` now bumps the transition epoch and drops the map entry *synchronously*, then takes the gate only for the cancel send, so the window Gandy reproduced is closed from both sides — a dispatch queued before the bump fails `#dispatchMayProceed`, and one queued after captures the new epoch. `#dispatchRequirementLocked` no longer calls `#retireRequest` at all (it installs, then cancels the displaced request directly), which is what keeps it from invalidating itself. `#dispatchStillOwns` re-checks identity, epoch, and connection after each await. The send-failure path now charges `current` rather than re-reading the map, so the "failed send for the older dispatch charges the newer request" facet is gone too.

**Terminal settle on exhausted delivery — resolved.** `#settleTerminalArtifact` writes a generic `unavailable` observation before stopping the episode, and it is reached from both the retry-exhaustion branch and the send-failure path, so the `maxRetries = 0` case Gandy described now settles on the first failure instead of stranding `checking`. The credential side ages out of `unconfirmed` on its TTL and the projection then reports `needs_attention`, so the surfaced state is stable rather than perpetually checking.

**Duplicate status charging the budget — resolved.** `#scheduleArtifactRetry` now returns early on `artifactRetryPending`, and `artifactRetryPending` is set after `#armTimer` rather than before. Two genuine sequential retries still work, because the timer callback clears the flag before redispatching and `#installRequest` carries `artifactRetryAttempt` forward.

That last change also removes the latent class I raised as a non-blocking note: `#armTimer` now clears `artifactRetryPending` whenever it displaces an existing timer, so a grant retry can no longer strand the flag. The unreachable duplicate check in `#handleValidation` is gone as well, and gating the accepted `frame.reason` on negotiated reconcile v2 in `#handleArtifact` — re-validated through `publicProviderCliArtifactReason` rather than trusted from the frame — closes the v1-receiver gap Gandy noted.

## One new non-blocking observation

`#transitionEpoch`, `#agentReconcileSeq`, and `#observedGeneration` are only ever written, never deleted or cleared — not on `#retireRequest`, not on `#resetComputer`, not in `close()`. `#requestGates` self-prunes in its `finally`, but those three accumulate one entry per distinct `computerId:integrationId` / `computerId:agentId` pair for the lifetime of the process, so a Computer removed long ago still holds entries.

Worth saying that a naive fix would be worse than the leak: `#transitionOf` and `#currentAgentReconcile` default to `0` on a missing key, so deleting those two is safe only because a stale captured epoch `N > 0` still compares unequal — but `#observedGeneration` defaulting to `0` would re-admit a superseded credential generation, which is exactly the guard it exists to provide. If this is worth pruning at all, it should be tied to the integration actually being retired (unbind), not to reconnect or request retirement. Fine to leave as is.

## Status

Approving on code-review merit. This does not clear the merge gate: @Gandy2025's `CHANGES_REQUESTED` from the previous head still stands and only they can dismiss it, the ownership gate needs @bestony, and the functional jobs (Checks, Node 22/24/26, Enforce Patch Coverage) were still pending when I finished — I did not run the suite myself, so I am not asserting anything about test results on this head.

Gandy2025 (changes requested):
Re-reviewed exact range `8cc14e38786b53fdd9c876167eecf2fc3f40e890..04dd3f17741e1cf41c6746ec10970cf0c5fa2a5b`, with focused inspection of the corrective delta `d28f4daf865458a36025019dd0de74b8ef86916f..04dd3f17741e1cf41c6746ec10970cf0c5fa2a5b`.

## Standards

The previous HTTP rolling-compatibility blocker is resolved. The exact `x-opentag-provider-cli-reason: 2` opt-in gates all four affected response surfaces, unmarked/v1/unknown callers keep the frozen pre-#490 shapes (including existing credential reasons), and new Client/Web callers request the extension. The redundant `nextAction` schema/type/exports/projections are removed. I found no new blocking Standards issue.

## Spec

The prior delayed-cancel generation overwrite, terminal delivery-exhaustion state, duplicate-artifact budget charging, and reconcile-v1 reason issues are resolved. Two new concurrency blockers remain:

- **[P1] Keep the whole binding snapshot bound to the sequence that read it.** `onActiveBindingChanged` checks `seq` only once after loading requirements, then awaits retirement and later calls `#dispatchRequirement`, which captures the *current* sequence instead of the sequence that owns the snapshot (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:315-353`). In a controlled exact-head reproduction, the initial active requirement was A; an older reconcile read stale snapshot C and blocked while cancelling A; a newer reconcile read current B and dispatched it; after the cancel was released, the older call borrowed the newer sequence and dispatched stale C. The observed requirement order was `A -> B -> C`. C then remains the keyed current request, and an ordinary readiness GET does not remove it because GET only iterates current database requirements. Please carry the originating sequence through the entire reconcile, re-check it after each awaited retirement and before each dispatch/mutation, and pass that original sequence into the gated dispatch rather than rereading the latest value. Audit the analogous multi-requirement retry/read loops for the same snapshot-lifetime rule.

- **[P1] Ignore a validation-grant send failure after that grant loses ownership.** The grant send catch mutates `current` and calls `#scheduleGrantRetry` without rechecking request/grant ownership (`packages/server/src/runtime/provider-cli-reconcile-owner.ts:933-938`). A controlled reproduction held the old grant send, started an explicit `prepareComputer` retry that installed a new same-generation requirement and wrote credential `unconfirmed`, then failed the old send. With `maxRetries = 0`, the stale catch overwrote the new recovery with `needs_attention/provider_unreachable`; after the new artifact reported ready, no new grant was issued. Please capture the sent grant identity and, before charging budget or writing terminal state in the catch, require that the keyed request is still `current`, `grantRequestId` still matches the sent frame, and the connection fence still owns it. A lost-owner send failure should be ignored.

Targeted local validation passed after building Shared: Shared compatibility/schema tests (38), Server reconcile/HTTP tests (59), Client API tests (9), and Web API tests (25). Independent exact-head checks also passed the old blocker regressions and reproduced both findings above; the worktree is clean. All functional CI and patch coverage are green; the ownership gate still requires the designated code owner.

liuchao-001 (changes requested):
Withdrawing my approval of `04dd3f17741e1cf41c6746ec10970cf0c5fa2a5b`. I verified both of @Gandy2025's new findings against the head and both are real. I read these code paths last round and called them sound; that was wrong.

## Both findings confirmed

**[P1] Borrowed sequence on a stale snapshot.** Confirmed. `onActiveBindingChanged` checks `seq` exactly once, at `:315`, and then awaits at `:328` (`#retireRequest`, which can block on the per-key gate for the duration of a cancel send) and again per iteration at `:352`. Neither continuation re-checks. The mechanism that should have caught it does not, because `#dispatchRequirement:464` does `const seq = this.#currentAgentReconcile(...)` — it re-reads the *latest* sequence rather than receiving the caller's. `#dispatchMayProceed` therefore compares latest against latest, which is trivially true at capture time; the check only ever catches a bump that lands while the dispatch is queued on the gate, never a caller whose own snapshot went stale across an earlier await. Gandy's `A -> B -> C` ordering follows directly.

**[P1] Grant send failure after losing ownership.** Confirmed. The catch at `:934-937` mutates `current` and calls `#scheduleGrantRetry` with no ownership recheck of any kind — not `this.#requests.get(key) === current`, not a `grantRequestId` match against the frame it just tried to send, not the connection fence. The asymmetry is visible within the same method: `#performIssueGrant` opens with `if (this.#requests.get(key) !== current) return;` and re-checks `stillCurrent !== current` after the grant-issue await, but the send catch inherits none of that. At exhaustion `#scheduleGrantRetry` writes a terminal `needs_attention` credential observation keyed by agent/integration/generation, which `matchObservation` will match against a new same-generation request, and `#shouldSkipGrant` then latches `grantRepairStopped = true` on that new request. The stall is permanent, not transient.

## Root cause is one line, not an audit

Gandy asks to "audit the analogous multi-requirement retry/read loops." Having walked them, I think the audit collapses to a single defect, which should make the fix smaller and more verifiable than it sounds.

Four snapshot-driven loops share the identical shape — read requirements, check `seq` once, then `await` per iteration:

- `onComputerRegistered:231`
- `onActiveBindingChanged:352`
- `#retryActiveRequirements:689`
- `#refreshRequirement:1238` and `:1254`, reached from the `#refreshActiveReadiness:1213` loop

Every one of them is defeated by the same line, `#dispatchRequirement:464`. Thread the originating sequence in as a parameter instead of re-reading it, and all four are fixed at once; leave that line as it is, and adding re-checks to the loops still lets a stale caller borrow the current sequence at the moment it finally dispatches.

The two remaining call sites, `#scheduleInternalRetry:1020` and `#scheduleArtifactRetry:1070`, should keep re-reading: they are timer continuations with no originating reconcile, and they are already fenced by `if (options.retry && !sameEpisode) return;`.

`onComputerRegistered` is worth noting as the near-miss — it is the one loop that does re-check per iteration, via `#agentReconcileUnchanged` against a snapshot map, and it is still defeated downstream by the same re-read. That is good evidence the guard belongs inside `#dispatchRequirement` rather than at each call site.

## On the second finding

The narrow fix Gandy describes is right. I would add that the general rule the two findings share is worth applying once rather than twice: every continuation in this class that resumes after an await and then mutates `current` or writes an observation needs to re-establish that it still owns the request. `#performIssueGrant` already demonstrates the pattern; the send catch is the one place inside it that skips it. It is worth grepping the remaining `catch` blocks in this file for the same omission while making this change, since a catch is exactly where the guard is easiest to forget.

## Status

Changing my review to request changes. My previous approval should not be read as independent sign-off on the concurrency rework — I checked the guards that were present and did not enumerate the await points that lacked one, which is the specific thing that would have caught both of these.

PR #574, opened 2026-09-08
feat(web): report the activation funnel to Google Analytics

bestony:
## What

Adds Google Analytics 4 (`G-RMDF361W1B`) to the Web App and instruments the activation funnel.

The three transitions asked for are measured as the drops between adjacent steps:

| | Step | Event | Raised where |
| --- | --- | --- | --- |
| 1 | signed in | `login` / `sign_up` | `routes/_authenticated.tsx`, on the first resolved Account |
| 2 | created an Agent | `agent_created` | `onboarding-v2/page.tsx`, after the Server returns an id |
| 3 | connected a Computer | `computer_connected` | `features/computer-connect/computer-connect.tsx`, when the redeemed Computer comes online |
| 4 | held a first conversation | `first_conversation_observed` | `features/agents/agents-page.tsx`, when the Agent list first shows a Task |

Supporting events: `agent_create_failed` (`name_conflict` vs `error`), `computer_connect_started`,
`agent_setup_stage_reached`, `agent_setup_completed`, and a `page_view` per resolved route template.
Every milestone carries `funnel: "activation"` and a contiguous `funnel_step`, so a GA4 exploration is
built from the step number rather than a hand-ordered list of events. A `computer_connected` with
`mode: "repair"` carries **no** step — reconnecting a Computer the Account already had is not somebody
reaching step 3 again.

Full detail, including the GA4 admin settings this needs, is in
[`docs/web-analytics.md`](./docs/web-analytics.md) (with a synchronized `docs/zh-CN/` mirror).

## Configuration

**Production is `app.opentag.build`** (confirmed by the deployer).

Measurement requires a production build **and** a hostname under `opentag.build` — an allowlist, not a
loopback exclusion. OpenTag is open source and meant to be self-hosted, so excluding only loopback
would mean every self-hosted deployment quietly reported its operators and their readers into this
property. An unlisted host is a silent gap rather than a silent leak, and it excludes the e2e stack
and local previews for free. **A self-hosted OpenTag loads no tag and reports nothing.**

> ⚠️ If the site ever moves domain, measurement stops silently — it fails closed. `MEASURED_HOST_SUFFIX`
> in `apps/web/src/analytics/config.ts` is the one place to change.

The measurement ID is a constant, not configuration: one image is built per commit and promoted
unchanged, so a build-time value cannot differ between environments. Injecting a `<meta>` at boot does
not work either — `@fastify/static` is registered with `wildcard: false`, so `/` and `/index.html` are
served from disk and never pass through the cached string. Staging therefore shares the property and
sends `traffic_type: "internal"`, so one GA4 Internal Traffic filter separates it.

## Privacy

- **The path comes from the route, never the address.** `page_path`/`page_location` are projected from
  the matched TanStack route's declared path (`/agents/$agentId/settings/$section` →
  `/agents/:agentId/settings/:section`). Classifying the address by shape was tried first and does not
  work here: routes declare free-form parameters, so `/agents/<uuid>/settings/<anything>` matches and
  renders, and no rule over segment length or character set tells a section name from a secret. Anything
  unmatched — `/invites/<token>` included, which renders the not-found page rather than failing to
  match — becomes the single constant `/(not-found)`.
- **Only the Account uuid**, as `user_id`. Never an address, display name, or Agent name.
- **Campaign parameters are allowlisted by key and still not trusted by value** — a value that is
  address-shaped or implausibly long is dropped. The "no address" promise is about what *this
  application* sends; campaign values are caller-supplied, and the docs say so.
- **Referrers are origins only**; an opaque origin reports nothing rather than the string `"null"`.
- **Identity ends with the session, by either exit** — Sign out, and expiry via the `/me` 401 branch.
  Not on a pending read or a dropped connection, which are also "no Account" but are not sign-outs.
- **Nothing from `/internal`**, enforced at the tag with Google's `ga-disable-<id>` flag rather than only
  by refusing this app's own calls — Enhanced Measurement raises scroll/click/form events of its own
  accord and `send_page_view: false` does not touch them.
- `allow_google_signals` and `allow_ad_personalization_signals` are off.

## Known limits, stated rather than hidden

**Step 4 is an observation and undercounts by construction.** Conversations happen in Slack/Feishu and
the Task views do not poll, so it fires when the Agent list is next open. Three biases, all lossy:
lag/survivorship (a delighted user who never returns reads as a step-3 drop-off), a **rolling 30-day
window** on `usage.tasks` (so this is *not* lifetime activation), and acceptance-rather-than-completion.
Treat it as a floor. Making it exact means server-side Measurement Protocol — deliberately out of scope.

Step counts are **per object, not per reader**: an Account with five conversing Agents contributes five
step-4 events. Funnel *rates* are unaffected; raw event-count reports read high.

`page_view` is keyed on the route template, so `/agents/A` → `/agents/B` is one page view.

## Breaking / operational

- **CSP change** (`packages/server/src/web-app.ts`): `script-src` stated explicitly to allow
  `googletagmanager.com`; `connect-src`/`img-src` widened for the regional collectors. Host allowance
  only — no `'unsafe-inline'`, so the GA console's snippet cannot run and the queue is built in-bundle.
- **GA4 admin, required before the numbers mean anything:** turn off *"Page changes based on browser
  history events"*; define the Internal Traffic filter and set it Active; register event-scoped custom
  dimensions (`funnel`, `funnel_step`, `method`, `runtime_provider`, `mode`, `stage`, `reason`) — these
  take 24–48h to become queryable.
- **Diagnostics behaviour change:** a third-party resource that fails to load is no longer reported as
  an application diagnostic. A blocked analytics tag is a reader's content blocker, not a defect, and
  reporting it would fail every browser test through the e2e console gate.
- One assertion in `agent-tasks.test.tsx` used "no storage read at all" as a proxy for "no workspace
  scope is remembered"; narrowed to the keys it actually means.

## Non-goals

No cookie banner or consent manager — the repo has none today, and the enablement switch is the control
a deployer has. GA Consent Mode v2 would belong in `installAnalytics` if that changes.

## Validation

- `pnpm check`, `pnpm build`, `pnpm typecheck` pass (complexity ratchet included); `pnpm test` 8/8 tasks.
- Coverage: 7 colocated suites under `apps/web/src/analytics/` plus
  `apps/web/src/__tests__/analytics-funnel.test.tsx`, which drives the real components and the
  **generated route tree** — that is how two defects were caught that a fake router agreed with: the
  router resolves asynchronously and holds no matches until it has, and the identity survived session
  expiry. The expiry regression restores production `refetchOnReconnect` and was confirmed to fail
  without its fix.
- Verified in the production bundle that the queue is filled with an `arguments` object (not an array)
  after minification, and that the host allowlist is present.

`AnalyticsReporter` is inert until armed, so no existing suite reports by importing a page; tests arm it
with a recording sink. No module mocking.


liuchao-001 (changes requested):
Reviewed at `acf6fc25`, in a fresh worktree off the PR head. The follow-up commit already fixes the one thing I had written up as blocking on `15a8914a` — the loopback-only guard meant every self-hosted deployment would have reported into `G-RMDF361W1B`. The allowlist is the right shape (`opentag.build.evil.test` correctly rejected), and the `/invites/<token>` reduction closes a leak I had missed: that path still renders the SPA not-found page, so `installRouteAnalytics` was reporting a 43-character token as `page_location`.

Independently verified on this head: `pnpm --filter @opentag/web test` → **753/753 passing, 72 files** (matches the PR body). The `@opentag/shared` build is a prerequisite; without it the suite fails to resolve `@opentag/shared/browser`.

---

## Should fix before merge

### `user_id` survives sign-out — post-sign-out page views are attributed to the signed-out Account

`AnalyticsReporter.identify(null)` is implemented and unit-tested (`analytics.test.ts`, *"clears the identity when the session ends"*), but it has **no call site**. `useSignedInReport` returns early on `!userId`, and `logout()` in `features/shell/app-shell.tsx` ends the session with a *client-side* `navigate({ replace: true, to: "/login" })` — no document reload — so `gtag('set', { user_id })` stays in force.

I confirmed this by driving the real components rather than reading the code. Rendering `<App />` signed in at `/agents`, then Sign out, the sink receives exactly:

```
["set",   {page_location: ".../agents", ...}]
["event", "page_view", {...}]
["set",   {user_id: "53e2babe-e4ac-4e2c-b7d1-d092d5a4568e"}]
["set",   {page_location: ".../login", page_referrer: ".../agents", ...}]
["event", "page_view", {...}]          ← still carries the user_id above
```

No `user_id: null` is ever sent. Blast radius is bounded — both sign-in paths do a full document load (`window.location.assign` for password, an `<a href>` for the providers), so a *different* Account cannot inherit the identity in the same document — but it is still wrong, and it contradicts an invariant the codebase states explicitly. `session-context.tsx` says of `endSession`: *"drops everything read under it … so nothing answered for this Account can reach the next one."* The analytics identity is the one piece of Account-derived state that isn't ended.

One line, in `logout()` beside `endSession()` (or inside the `endSession` implementation, which is the better home given the comment above):

```ts
analytics.identify(null);
```

---

## Non-blocking

**1. The charset guard weakens the 16-character rule.** `TOKEN_SEGMENT = /^[A-Za-z0-9_-]+$/` means a long segment containing anything else — a `.`, a `~`, a percent-escape — is *not* reduced. The justification for the rule is that this application has no legitimate segment at or past 16 characters (`integrations`, the longest, is 12), and that argument does not depend on the charset. Dropping the test and reducing every segment ≥ 16 is strictly safer and loses nothing.

**2. The sign-in intent is recorded on the press, not on the outcome.** `LoginProviderLink`'s `onClick` writes the intent unconditionally, before anything succeeds. Abandon the consent screen, come back, then land on an authenticated surface with a session you already had, and `_authenticated` takes the stale intent and reports a `login`/`sign_up` that never happened. The password path doesn't have this — it writes only after the API call resolves. Step 1 is the base of every ratio in this funnel, so it is the step most worth not inflating. Cheapest fix is to stamp the intent with `Date.now()` and have `takeSignInIntent` ignore anything older than a few minutes.

**3. `page_view` collapses sibling detail navigations, and the doc doesn't say so.** `installRouteAnalytics` keys `lastReported` on the *sanitized* location, so `/agents/A` → `/agents/B` is one page view, not two — and clicking through ten tasks is one. `route-analytics.test.ts` asserts this deliberately (*"treats two Agents as the same page"*), so it's a decision rather than a bug; but `docs/web-analytics.md` sells the event as "per resolved route", which is what a reader building an engagement report would rely on. Worth a clause in the doc.

**4. `analyticsPageReferrer` can emit the literal string `"null"`.** A referrer with an opaque origin (`android-app://…`, `about:…` — both give `origin === "null"`, verified) fails the same-origin test and falls through to `return url.origin`, so GA receives `page_referrer: "null"` rather than nothing. Guard on `url.origin === "null"` → `undefined`.

**5. `first_conversation_observed` is deduped per Agent, so step 4 scales with Agent count.** `reportMilestoneOnce(\`first-conversation:${agent.id}\`)` means an Account with five conversing Agents emits five step-4 events, all on the first list read. GA4 funnel explorations are user-scoped so the conversion rate is unaffected, but any event-count report on step 4 reads high. Steps 2 and 3 are per-object too, so it's at least self-consistent — flagging it so the number isn't read as "readers who reached step 4".

---

## Needs a human answer before merge

`MEASURED_HOST_SUFFIX = "opentag.build"` cannot be verified from the repository. `docs/deploying.md` puts the deployed host behind the `CAPROVER_*` secrets, and `analyticsEnabled` is hard-false under test (`import.meta.env.PROD`), so no suite can catch a mismatch. The failure mode is silence: wrong suffix, no tag, no events, no error anywhere. You already flag this in the PR body — I'm confirming it as the one open item a reviewer can't close, not adding to it. Someone who knows the CapRover app domains should confirm both staging and production sit under `opentag.build` before this merges.

---

## Checked and satisfied

- **UUID reduction actually works here.** `routeTemplate`'s `UUID_SEGMENT` requires version `[1-5]`, which would miss a v7 id — but every id in the URL space comes from Drizzle `uuid().defaultRandom()` → `gen_random_uuid()` → v4, and routes take ids rather than names (`agents.$agentId`, `agentSetupLink` puts the id in a dropped query parameter). No leak.
- **`page_title` is safe.** Nothing writes `document.title`; it is the static `"OpenTag"` from `index.html`, so the one free-text field in the page payload cannot carry an Agent or Account name.
- **The host guard covers e2e.** `playwright.config.ts` pins `baseURL` to `http://127.0.0.1:${port}`, which the allowlist excludes.
- **CSP is a host allowance only** — no `'unsafe-inline'` added, `script-src` narrowed to `www.googletagmanager.com` while the wildcard is confined to `connect-src`/`img-src`.
- **Branch is current** with `origin/main` (`7b9efff3`), no drift.

Gandy2025 (changes requested):
Reviewed the exact range `7b9efff3...acf6fc25` as a high-risk analytics/privacy change. I reused liuchao-001's same-head review for the missing `user_id` clear on logout instead of duplicating that finding. The items below are additional.

## Standards / privacy

### 1. `/internal` is not actually excluded from Google Analytics collection

`installAnalytics()` is called before every production render (`apps/web/src/main.tsx:14-16`) and installs/configures gtag without considering the route (`apps/web/src/analytics/analytics.ts:92-110`). The `/internal` check at `analytics.ts:77-79` only suppresses this app's own `page()` and `track()` calls; `identify()` bypasses it as well.

That does not stop the loaded Google tag. Google documents that Enhanced Measurement emits scroll, outbound-click, site-search, file-download, and form events independently, and that these event types are enabled by default. `send_page_view: false` and the documented history-page-view setting do not disable them. A direct visit or SPA navigation to `/internal/agent-setup` can therefore send rehearsal interactions, associated with the Account UUID on the authenticated internal route. This contradicts the explicit “Nothing from `/internal`” privacy contract in `docs/web-analytics.md:89-90`.

Smallest credible correction: enforce the exclusion at the tag/transport lifecycle for both direct entry and SPA transitions (Google's supported `window['ga-disable-G-RMDF361W1B']` control is one option), and test that an internal route neither configures/sends through the tag nor sets identity. References: [Enhanced Measurement](https://support.google.com/analytics/answer/9216061), [turn off Google Analytics programmatically](https://developers.google.com/tag-platform/security/guides/privacy#turn_off_google_analytics).

### 2. The path sanitizer still fails open on current dynamic routes

`apps/web/src/analytics/page-location.ts:45-49` preserves every non-UUID/non-integer segment shorter than 16 characters. Current routes accept arbitrary strings in `$legacySection` and `$section` (`routes/.../agents.$agentId.$legacySection.tsx` and `agents.$agentId.settings.$section.tsx`) before their components render a not-found state. For example, the current code turns:

```text
https://opentag.build/agents/<uuid>/secret
```

into `https://opentag.build/agents/:id/secret`, not a route template. Long percent-encoded or punctuation-bearing segments also bypass `TOKEN_SEGMENT`. This is broader than the already-noted charset weakness: a short raw dynamic value leaks today. It violates the PR's promise that an unanticipated URL shape cannot leak and the docs' “No raw URL” guarantee.

Smallest credible correction: derive `page_path`/`page_location` from the matched TanStack route template, and map unmatched/not-found routes to one constant path. Do not preserve unmatched raw segments.

### 3. Allowlisted campaign keys still forward arbitrary PII values

`apps/web/src/analytics/page-location.ts:60-62` copies the value of every allowlisted campaign parameter verbatim. A public URL such as `?utm_source=alice@example.com` is currently reported unchanged, despite the PR promise that no address is sent. Query strings are caller-controlled; allowlisting the key does not make its value trusted.

Google's own guidance specifically says that URL paths and parameters must be free of PII and warns not to put PII in `utm_source`, `utm_medium`, `utm_term`, `utm_campaign`, or `utm_content`. Treat these values as untrusted: either drop free-form campaign values from `page_location`, restrict them to an actually controlled value set, or add a fail-closed redaction/validation boundary plus property-side redaction as defense in depth. Reference: [Best practices to avoid sending PII](https://support.google.com/analytics/answer/6366371).

## Spec

### 4. Step 4 only observes conversations from the rolling 30-day window

`apps/web/src/analytics/milestones.ts:117-128` treats `agent.usage.tasks > 0` as “has held a first conversation”, but that count is explicitly bounded to 30 days (`packages/shared/src/agent.ts:120,144-149`; `packages/server/src/services/agents/agent-service.ts:596,625-639`). A reader who first returns on day 31 emits no step-4 event. The PR documents undercount for readers who never return, but says the event fires when the list is next opened; the rolling-window loss is an additional, materially different limitation.

Use a durable lifetime-existence signal for this milestone, or rename/redefine the event and document the 30-day observation window so the reported funnel is not interpreted as “ever held a first conversation.”

## Validation and open evidence

- Targeted analytics/funnel/computer-connect tests pass on this head: 62/62 across 8 files.
- The repository cannot verify that both deployed staging and production hosts are under `opentag.build`; the allowlist otherwise fails silently. This remains an operational confirmation before merge.

liuchao-001:
Following up on @Gandy2025's review — I re-verified the additional findings against `acf6fc25` rather than taking them as read. Three of the four hold, one of them changes the severity of something I filed as non-blocking. Consolidating so the two reviews read as one set of work rather than two.

## Confirmed, and I was wrong to call this one non-blocking

**My non-blocking #1 (the `TOKEN_SEGMENT` charset guard) and Gandy's #2 are the same defect, and it is worse than either of us stated.** The charset test doesn't merely weaken the 16-character rule — it exempts exactly the values most likely to be credentials. Run against the current code:

```
/invites/<43-char base64url>   ->  /invites/:id                     ✅
/invites/<98-char JWT>         ->  /invites/eyJhbGciOiJIUzI1NiJ9.…  ❌ untouched
/agents/<uuid>/secret          ->  /agents/:id/secret               ❌ verbatim
/agents/<uuid>/settings/my-private-note -> …/settings/my-private-note ❌ verbatim
```

A JWT is longer, more structured, and more obviously a credential than the base64url token the rule was added for — and it passes through in full, because it contains `.`. The rule catches the one shape it was written against and misses the neighbouring one. Please treat this as blocking alongside the `identify(null)` item.

Both of us landed on the same root cause from different directions: `analyticsPagePath` sanitizes the **value** rather than projecting onto the **matched route** template. `$section` and `$legacySection` declare no `params.parse`, so any string matches and renders. That is what makes the PR's own claim — *"rebuilt … so a URL shape nobody anticipated cannot leak through a filter that failed to match"* — not yet true: the path is still value-driven. Deriving `page_path`/`page_location` from the TanStack matched route (Gandy's suggestion) fixes my item and theirs at once, and is the only version of this that stays correct as routes are added.

**Gandy's #4 (the 30-day window) is correct and I missed it.** `AgentUsageSummarySchema` pins `windowDays: z.literal(AGENT_USAGE_WINDOW_DAYS)` = 30, and the aggregate is bounded by `gte(imMessageDeliveries.acceptedAt, now - 30d)` in `agent-service.ts`. So `usage.tasks < 1` does not mean "has never held a conversation" — it means "has held none in the last 30 days". A reader whose Agent last talked 31 days ago emits no step 4, and the docs describe the undercount as being about readers who never return, which is a different and smaller bias than this. Worth fixing or renaming before the number gets read as lifetime conversion.

**Gandy's #1 (`/internal`) is correct.** `identify()` is not behind `#reporting()`, and the guard only covers this application's own calls — it cannot stop the tag's own Enhanced Measurement hits, which are on by default and are not disabled by `send_page_view: false`. Worth noting the second-order effect: because `page()` returns early on `/internal`, the `gtag('set', …)` location is never refreshed there, so an automatic hit raised on an internal route is attributed to *the last real page the reader was on*. So it is a data-integrity bug as well as a broken privacy claim.

## Where I'd weigh it differently

**Gandy's #3 (campaign values forwarded verbatim) is factually right — `?utm_source=alice@example.com` does survive, I checked — but I would not gate merge on it.** The values are written by whoever built the link, not by this application or its readers, and the realistic case is a marketer's own sloppy tagging rather than a leak of a reader's data. Google's guidance is aimed at the site owner not putting PII there in the first place. A sentence in `docs/web-analytics.md` softening "Never an address" to name the campaign allowlist as caller-controlled would close the gap between claim and behaviour honestly; a fail-closed validation boundary reads as more machinery than the risk earns.

## Net

Two things I'd hold merge for beyond my original review: the route-template sanitization (which subsumes my #1 and Gandy's #2), and the `/internal` exclusion being real rather than partial. The 30-day window needs a decision — fix the signal or rename the event — but either is fine. The hostname confirmation remains the one item neither reviewer can close from the repository.

yuezengwu (changes requested):
Request changes at exact head `acf6fc2596cd53500d178b875ae237ce608340d7`, against `7b9efff3fa78683aa8630dc22b92f410f8979cca`.

I independently reproduced the three existing blockers with synthetic local data and the real application/reporter code:

1. **Route values still reach the report.** `/agents/<uuid>/settings/private-note` becomes `/agents/:id/settings/private-note`; a punctuation-separated synthetic credential under `/invites/` survives in full, while a 43-character base64url control is reduced. The current `$section` and `$legacySection` routes accept these dynamic values. Project the matched route template and use a constant fallback for unmatched routes, rather than classifying arbitrary segment values by length and charset.
2. **The `/internal` exclusion does not cover installation or identity.** With the production flag and a synthetic allowed host, direct internal entry queues `config`, creates the tag script, and queues `user_id`. The application milestone is suppressed, but the tag is not disabled. This is a local queue/lifecycle reproduction, not a claim that I sent a real Google hit. Google's [privacy control documentation](https://developers.google.com/tag-platform/security/guides/privacy#turn_off_google_analytics) and [Enhanced Measurement documentation](https://support.google.com/analytics/answer/9216061) confirm that blocking application events or default page views does not disable the tag's independent collection. Cover direct entry and SPA transitions at the tag/transport boundary.
3. **A successful sign-out leaves analytics identified.** The real App test performs the logout POST and reaches the login page. The sink records the original Account UUID and a subsequent login-page view, with no `user_id: null`. Clear identity as part of ending the Account session, including session invalidation.

These corroborate [liuchao-001's review](https://github.com/first-tree-ai/opentag/pull/574#pullrequestreview-5143642807) and [Gandy2025's review](https://github.com/first-tree-ai/opentag/pull/574#pullrequestreview-5143721839); I am keeping one consolidated review rather than duplicating their inline threads.

Two remaining contract details are also confirmed: an allowlisted campaign key forwards an arbitrary synthetic address unchanged, so the unconditional no-address promise is not enforced; `usage.tasks` is explicitly a 30-day aggregate, so step 4 cannot represent lifetime conversation activation. Resolve the data boundary and metric definition in code/docs before presenting those guarantees.

Independent validation: `pnpm check`, `pnpm build`, and `pnpm typecheck` pass; all **753 Web tests** and the **server static/CSP test** pass. Five additional local observation probes reproduce the behaviors above. Existing tests passing therefore does not establish the claimed privacy guarantees. No author source edits, real Google script execution, or analytics transmission were performed.

Before rollout, the deployer still needs to confirm the measured production/staging hostnames, intended Account-UUID collection scope, and GA property configuration. This review does not approve enabling third-party collection or a deployment.

bestony:
Thanks — all three reviews were right, and the route-template point in particular was right in a way I'd underestimated. Addressed at `920f93d9`. Taking them in the order they'd break something.

## The hostname, which none of you could close from the repo

**Production is `app.opentag.build`** (confirmed by the deployer). It is a subdomain of `MEASURED_HOST_SUFFIX`, so the allowlist already covered it — no change was needed, and there is now an explicit assertion for it so it can't drift silently.

I also used the fact to make the shared-property tradeoff actually manageable: every host that is **not** `app.opentag.build` now sends `traffic_type: "internal"`, so one GA4 Internal Traffic filter excludes staging from every report, rather than each report remembering to segment by hostname. Until that filter exists the parameter is inert, so it is safe for it to land first.

## Fixed — blocking

**1. `user_id` survived sign-out** (@liuchao-001, @yuezengwu). `analytics.identify(null)` is now called from `endSession` in `_authenticated.tsx` — the home you suggested, and the right one given that comment about nothing read under a session reaching the next. There is a test that drives the real account menu → Sign out → login page and asserts the sink sees `[userId, null]`.

**2. The path sanitizer classified values instead of projecting routes** (@liuchao-001 #1, @Gandy2025 #2, @yuezengwu #1). You were all correct and the charset rule is gone entirely rather than widened — it was treating a symptom. `page_path`/`page_location` are now projected from the matched TanStack route's declared path (`/agents/$agentId/settings/$section` → `/agents/:agentId/settings/:section`), and anything unmatched becomes the single constant `/(not-found)`. A value reaches a report only because a route file names the parameter, so the JWT case, the `/agents/<uuid>/secret` case, and the percent-encoded case all close together.

Worth reporting back: **verifying this against the generated route tree rather than a fake router found a second defect.** The router resolves asynchronously and holds `matches: []` until it has, so the eager report at install was filing the first page of *every* visit as `/(not-found)`. Both real-route assertions are now in `analytics-funnel.test.tsx` (`/agents/<uuid>` → `/agents/:agentId`; `/invites/<token>` → `/(not-found)` with the token absent) precisely because a fake router would have kept agreeing with me.

**3. `/internal` was not actually excluded** (@Gandy2025 #1, @yuezengwu #2). Correct — refusing this app's own calls says nothing about Enhanced Measurement, and `send_page_view: false` doesn't touch it. Now enforced at the tag with Google's `ga-disable-<id>` flag, set before the tag can send anything and kept in step with the route on every resolve, so direct entry and SPA transition are both covered. @liuchao-001's second-order point (stale `set` location on internal routes) is moot once no hit is sent at all.

## Fixed — non-blocking

**4. Sign-in intent recorded on the press** (@liuchao-001 #2). Your fix, taken as written: the intent carries a timestamp and expires after ten minutes. Long enough for a consent screen and a second factor, short enough that an abandoned OAuth flow can't manufacture a `login` at the step every ratio divides by.

**5. `analyticsPageReferrer` could emit the literal `"null"`** (@liuchao-001 #4). Fixed. Referrers are now origin-only in all cases, and an opaque origin returns `undefined`.

**6. Campaign values forwarded verbatim** (@Gandy2025 #3, @liuchao-001's weighting). I took both: the doc now says plainly that the no-address promise is about what *this application* sends and that campaign values are caller-supplied — and a value that is address-shaped (`@`) or implausibly long (>100 chars) is dropped rather than forwarded. Cheap enough that it didn't seem worth arguing the risk down.

## Documented rather than changed

**7. The 30-day window on step 4** (@Gandy2025 #4, @yuezengwu). You're right and I had it wrong: `usage.tasks` is a rolling 30-day aggregate, so this is not lifetime activation. I kept the signal and fixed the claim — `docs/web-analytics.md` now lists three distinct biases (lag/survivorship, the 30-day window, acceptance-not-completion) and says explicitly that step 4 must not be read as "readers who have ever held a conversation". Renaming felt worse: the event *is* the milestone that was asked for, and a misleading name plus honest docs beats an accurate name nobody maps back to the funnel. Happy to rename if you'd rather.

**8. `page_view` collapses sibling detail navigations** (@liuchao-001 #3) and **9. step counts are per object, not per reader** (@liuchao-001 #5). Both now stated in the docs, the second under its own heading, since it's the one most likely to be misread off a raw event-count report.

## Still not done, deliberately

No consent mechanism or privacy policy — unchanged from the original non-goal, and the enablement switch remains the deployer's control. The exact-conversion version of step 4 (server-side Measurement Protocol) is still out of scope.

## Validation

`pnpm check`, `pnpm build`, `pnpm typecheck` pass; `pnpm test` 8/8 tasks. Bundle budget clear.

liuchao-001 (approved):
Approving the code at `920f93d9`. I re-verified each fix by driving the real application rather than reading the diff, since the whole point of the route-template change is that reasoning about it from a fake router is what produced the original defect.

## Verified, not taken on trust

**Sign-out clears identity.** Real account menu → Sign out → login page, capturing the ordered sink:

```
SET, EVENT:page_view,                       ← /agents
IDENTIFY:53e2babe-…, IDENTIFY:null,         ← endSession, before the navigation
SET, EVENT:page_view                        ← /login, no longer attributed
```

The ordering is the part that matters and it is right: `identify(null)` lands before the login page view, not after.

**Route projection holds against the generated route tree.** `/agents/<uuid>` reports `page_path: "/agents/:agentId"` with the uuid absent from the entire payload. `/invites/<98-char JWT>` reports `/(not-found)` with the token absent — the exact case the charset rule missed. Both checked by rendering `<App />` at those addresses.

**`/internal` is off at the tag.** `ga-disable-G-RMDF361W1B` reads `true` on a direct entry to `/internal`, with zero events queued. That closes it where Enhanced Measurement actually decides, which the previous guard could not.

**The smaller ones.** A fresh intent is returned and an 11-minute-old one is discarded; `utm_source=alice@example.com` is dropped while `utm_campaign=ok` survives; `:agentId` and `(not-found)` both build valid URLs rather than throwing in `new URL`.

**Suite:** 764/764 web tests, 72 files, on this head — up from 753, so the new behaviour came with coverage rather than borrowing existing assertions.

## On the two you asked about

**The async-matches defect you found is the more valuable half of this commit.** Filing the first page of every visit as `/(not-found)` would have been invisible in the data — a plausible-looking bucket, not an error — and it only surfaced because you tested against the real route tree. That is the right lesson to have drawn from the review, and a better outcome than the fix I asked for.

**Keep the name.** I agree with your reasoning on step 4: the event is the milestone that was requested, and `first_conversation_observed` already says "observed". An accurate name nobody maps back to the funnel is worse than an honest docs section, and you now list all three biases where the number is read. No change wanted.

## One thing left, and it is not code

The **PR description is now stale** — it still says measurement requires "a non-loopback hostname" and describes the 16-character segment rule as the mechanism, both of which this commit replaced. The description is what gets read at merge and afterwards; worth a pass before this lands so the durable record matches the code.

## Scope of this approval

This approves the implementation. It is not the decision to enable third-party collection, to identify readers by Account UUID, or to accept staging and production sharing one property — those are @liuchao-001's to make, and I have raised them with him separately. @Gandy2025 and @yuezengwu still have changes requested against `acf6fc25` and will want to confirm their own items against this head.

Gandy2025 (approved):
Approved at exact head `920f93d9cdc72ca61a4abaa0d4b48959c1bca7e3`, after a proportional re-review of `acf6fc25..920f93d9` and re-checking each item from my previous review.

The implementation now closes the reviewed boundaries:

- Page locations are projected from the matched TanStack route declaration, with a constant `/(not-found)` fallback. The real generated-route-tree tests cover both a dynamic Agent route and an invite token that must not survive in the payload.
- `/internal` is disabled at the tag boundary with `ga-disable-G-RMDF361W1B`, set before tag installation for direct entry and synchronized on resolved SPA routes.
- `endSession` clears `user_id` before the client-side login navigation; the real account-menu flow verifies the ordered identity clear.
- Caller-controlled campaign values now have a small fail-closed boundary and the documentation states its limit honestly.
- The 30-day observation window and the per-object semantics of step 4 are now explicit, so the event is no longer documented as a lifetime-conversion signal.
- The deployer confirmed production is `app.opentag.build`, which is covered by the allowlist and now asserted in tests.

Independent validation on this head:

- `pnpm --filter @opentag/web test`: 72 files, 764/764 tests passed.
- `pnpm check`: passed, including the complexity ratchet (existing lint diagnostics remain warnings).
- `git diff --check 7b9efff3..920f93d9`: passed.

No additional code findings from this pass. The PR description still describes the superseded non-loopback/16-character mechanisms; it should be refreshed before merge so the durable record matches the implementation. The documented GA property settings remain rollout prerequisites and are outside this implementation approval.

apps/web/src/analytics/milestones.ts
```diff
@@ -0,0 +1,150 @@
...
+ * Attach the Account, and report the sign-in that produced it.
+ *
+ * This runs wherever an authenticated surface first resolves the Account, which is every load of
+ * every signed-in page — so the sign-in itself is told apart from a return visit by the intent the
+ * sign-in path left behind. Without one, this is somebody arriving with a session they already had,
+ * and only the identification is worth doing.
+ */
+export function useSignedInReport(userId: string | undefined): void {
+  useEffect(() => {
+    if (!userId) return;
```
yuezengwu, line 115: [P2] Clear the previous analytics identity on authoritative session loss

This hook returns without clearing `user_id` when the authenticated Account becomes unavailable. The new clear in `endSession` covers the Sign out button, but the existing `/me` 401 redirect never calls it. A reachable production sequence is: open `/agents` while signed in, let that session expire or be revoked, then reconnect the network. `createQueryClient()` enables reconnect revalidation in production, `/me` returns 401, and the real App navigates to `/login` with the old identity still installed. My regression records `set({user_id: fixtureUserId})` followed by the `/login` `page_view`, with no intervening `set({user_id: null})`; the identity-clear assertion fails. That attributes unauthenticated activity to the expired Account until another identity change or document reload. Handle confirmed session loss before reporting its redirect, distinguishing a 401 from temporary loading/network errors, and add a regression alongside explicit logout. This follows the [GA User-ID lifecycle](https://developers.google.com/analytics/devguides/collection/ga4/user-id#clear_the_user-id_upon_logout), where a previously identified user must be cleared when no longer signed in.

yuezengwu (changes requested):
Reviewed exact head `920f93d9cdc72ca61a4abaa0d4b48959c1bca7e3` against `7b9efff3fa78683aa8630dc22b92f410f8979cca`.

The route-template projection, direct and resolved SPA `/internal` suppression, and explicit logout ordering pass independent probes on this head. One identity-lifecycle case remains: an already identified session that becomes unauthorized after a reconnect reaches `/login` without clearing `user_id`. The inline finding includes the trigger and the observed command order.

Validation on the detached review checkout:

- `pnpm check`, `pnpm build`, `pnpm typecheck`, and `git diff --check`: pass.
- Full Web suite: 764 tests across 72 files pass.
- Server static/CSP test: 1 test passes.
- Seven independent probes using the real App/router: six pass; the production reconnect → `/me` 401 regression fails because the identity-clear command is absent.

The current technical CI is green; the ownership gate remains red. The failing independent regression is not covered by the existing suite. These checks use synthetic API responses and a recording analytics sink; they do not verify Google collection requests or the live GA property's settings.

Non-blocking documentation follow-up: the PR description still describes the old non-loopback guard and segment-length sanitizer. Please align it with the hostname allowlist and route projection now implemented. The documented GA admin settings and deployment/collection decision remain rollout prerequisites.

liuchao-001 (changes requested):
Withdrawing my approval of `920f93d9`. @yuezengwu is right, and this one is mine to own: I reported the class of defect — analytics identity outliving the Account session — then verified the fix at the single call site I had named and approved it without checking the other way out of an authenticated session. My own review quoted `endSession`'s promise that "nothing answered for this Account can reach the next one" and I did not go on to ask which paths actually reach `endSession`.

## Reproduced

Their sequence, driven through the real `<App />` with production reconnect behaviour restored (`createQueryClient` sets `refetchOnReconnect` only when `MODE !== "test"`, which is why a naive probe shows nothing):

```
SET, EVENT:page_view:/agents, IDENTIFY:53e2babe-…, SET, EVENT:page_view:/login
```

No `IDENTIFY:null`. The login page view after an expired session is attributed to the Account that just lost it — the same defect as the sign-out case, on the path the fix did not cover.

## The full enumeration, since this is what the last round was missing

There are exactly two ways an identified session ends, and the clear sits on one of them:

| Exit | Path | Identity cleared |
| --- | --- | --- |
| Sign out | `app-shell.tsx:82` → `endSession()` | yes |
| Session expired or revoked | `_authenticated.tsx:69` → `<Redirect to="/login">` | **no** |

`grep` for `endSession` returns those two call sites and nothing else, so this is the whole surface rather than the next instance of an open-ended list. Worth saying plainly: **expiry is the more common of the two.** Most sessions end by lapsing, not by somebody pressing Sign out, so the uncovered path is the one that will carry most of the mis-attributed traffic.

## What I'd ask for

Clear the identity where session loss is *confirmed*, at the 401 branch, rather than wherever the Account happens to be absent — a loading state and a network error are also "no Account" and must not clear anything. `resource-state.tsx:9` already distinguishes an authoritative failure from a transient one, so the distinction exists in the codebase. And a regression that drives the real reconnect → 401 → `/login` path, alongside the sign-out one, since the sign-out test passing is exactly what let this through.

The rest of my approval stands — route projection, `/internal` tag suppression, intent expiry, campaign filtering, and 764/764 all re-verified on this head and unaffected. This is one gap in one fix, not a re-litigation of the commit. The stale PR description is still outstanding too.

Gandy2025 (changes requested):
Correcting my earlier approval at the same exact head, `920f93d9cdc72ca61a4abaa0d4b48959c1bca7e3`: [yuezengwu's inline finding](https://github.com/first-tree-ai/opentag/pull/574#discussion_r3963328350) is valid and is the remaining blocker.

I had checked for an explicit caller of `reloadMe` and incorrectly concluded that an identified `/me` query could not become a 401 in the mounted application. That missed the production query-client behavior: `createQueryClient()` sets `refetchOnReconnect: true` outside tests. Because `/me` is active and stale, reconnect revalidates it. A terminal 401 then produces the `_authenticated.tsx` redirect to `/login`; `useSignedInReport(undefined)` returns without clearing identity, and the only `analytics.identify(null)` call remains inside explicit `endSession()`.

The consequence is the reported one: an expired or revoked Account session can leave its `user_id` installed on the subsequent unauthenticated page views. Please clear the analytics identity on authoritative session loss before the redirect, without treating loading or transient network errors as sign-out, and add the reconnect → `/me` 401 regression beside the explicit-logout coverage.

The other previously reviewed fixes remain valid and do not need to be reopened. Re-review can be limited to this identity-lifecycle path and its adjacent page-view ordering.

bestony:
Fixed at `b466a5f2`. @yuezengwu found it, @liuchao-001's enumeration is what made it obviously complete rather than "one more instance", and the point that **expiry is the commoner exit** is the part that matters — the path I covered is the rarer one.

## The gap

I fixed the class of defect at the call site I'd been handed and stopped there. There are exactly two ways an identified session ends, and I covered one:

| Exit | Path | Was cleared |
| --- | --- | --- |
| Sign out | `app-shell.tsx` → `endSession()` | yes |
| Session lapsed or revoked | `_authenticated.tsx` 401 → `<Redirect to="/login">` | **no** |

Both now run through one hook (`useAccountIdentityReport`) that owns the identity for the life of the session. The refusal is computed once and drives the redirect *and* the release, so the two cannot drift apart about whether the Account is gone.

Taking @liuchao-001's point about *where* precisely: the release is keyed on an **authoritative refusal**, not on the Account being absent. A read that has not answered and a dropped connection are also "no Account" and neither is a sign-out — clearing on those would discard the identity of a session that is still perfectly alive. There is a second test that holds the identity through a transient `/me` failure, so the fix cannot be satisfied by clearing too eagerly.

## On the regression

Written the way you specified, driving the real reconnect → `/me` 401 → `/login` path, with `MODE` stubbed to production so `createQueryClient` actually sets `refetchOnReconnect` — without that the probe sees nothing and passes for the wrong reason, exactly as you both noted.

**I confirmed it fails without the fix** rather than assuming it: disabling the clear turns that one test red and leaves the other fifteen green. Worth saying plainly, since a sign-out test passing is what let this through the first time — the value was never in the assertion, it was in the path it drives.

Ordering is asserted too: the identity is released before the `/login` page view, so that page view does not carry the Account that just lost its session.

## Documentation

`docs/web-analytics.md` and the zh-CN mirror now state both exits, and explicitly that a pending read and a network error clear nothing.

**The stale PR description is rewritten** — thanks for flagging it, all three of you. It described the non-loopback guard and the segment-length sanitizer, both of which are gone. It now describes the hostname allowlist, the route-template projection, `app.opentag.build` as production, the `traffic_type` split for staging, the 30-day window on step 4, and the GA4 admin prerequisites.

## Validation

`pnpm check`, `pnpm build`, `pnpm typecheck` pass; `pnpm test` 8/8 tasks. Re-review can stay on this path, as @Gandy2025 suggested — nothing else was touched.

Gandy2025 (approved):
Approved at exact head `b466a5f21f1e07c9b82bce033444a5e26427df8a`, after a targeted re-review of `920f93d9..b466a5f2`.

**Standards:** pass. The identity lifecycle now has one explicit hook boundary, and the 401 decision is computed once and shared by both the identity release and redirect. It does not infer sign-out from loading or a transient network failure.

**Spec/privacy:** pass. The production reconnect → `/me` 401 path clears `user_id` before the `/login` page view, while the transient-failure control keeps the existing identity. This closes the single remaining blocker from my previous review. The PR description and both analytics documents now match the implemented hostname, route-projection, identity-lifecycle, and step-4 contracts.

Validation on this head:

- `pnpm --filter @opentag/web exec vitest run src/__tests__/analytics-funnel.test.tsx --maxWorkers=1`: 16/16 passed.
- `pnpm check`: passed, including the complexity ratchet (existing diagnostics remain warnings).
- `git diff --check 7b9efff3..b466a5f2`: passed.

The prior full-range evidence for the unaffected analytics boundaries carries forward. No remaining code findings from this review.

liuchao-001 (approved):
Approving at `b466a5f2`. Re-review scoped to the identity-lifecycle path as agreed; nothing else was touched and the prior evidence carries forward.

## Verified

**Both exits now clear, and the ordering is right.** The reconnect → `/me` 401 → `/login` path, driven through the real `<App />` with production `refetchOnReconnect` restored:

```
SET, EVENT:page_view:/agents, IDENTIFY:53e2babe-…, IDENTIFY:null, SET, EVENT:page_view:/login
```

`IDENTIFY:null` lands before the login page view. That ordering was the thing worth checking rather than assuming — the release happens in the gate's effect while the redirect renders in a child, and child effects run first in React, so it could plausibly have gone the other way. It doesn't.

**The clear is not over-eager.** A transient `/me` network failure on the same path leaves the identity installed and stays on `/agents` — so the fix cannot be satisfied by clearing on "no Account".

**766/766 web tests, 72 files** on this head.

## The regression earns its keep

I checked your "confirmed it fails without the fix" claim rather than taking it: disabling only the `identify(null)` in `useAccountIdentityReport` turns `analytics-funnel.test.tsx` **1 failed / 15 passed** — the lapsed-session test red, the transient-failure control still green. That is the right shape. It is also the check that was missing the first time round: the sign-out test passed both before and after the defect existed, so it never had the power to catch anything.

## What this round was actually about

Worth recording, since it is the transferable part. The defect survived two independent reviews because both of us verified a *path* rather than a *boundary* — I checked the one call site I had named, @Gandy2025 checked one trigger and concluded the 401 was unreachable. Neither of us asked "how many ways does this state end, and does the guard sit on all of them?" The two-row table was not extra rigour; it was the minimum that would have been correct, and `grep endSession` would have produced it in seconds.

The structural fix is better than the one I asked for. I suggested clearing at the 401 branch; routing both exits through a single hook that owns the identity for the session's life, with the refusal computed once and shared by the release and the redirect, means the two cannot drift apart later. That is a boundary rather than a second patch.

**The rewritten PR description** now matches the implementation — I checked it for the superseded non-loopback and segment-length claims and both are gone, with the allowlist, route projection, `app.opentag.build`, the `traffic_type` split, the 30-day window and the GA4 admin prerequisites all stated.

## Scope, unchanged

This approves the implementation. Enabling third-party collection, identifying readers by Account UUID, and staging sharing production's property remain @liuchao-001's decisions, and the GA4 admin settings in your description are rollout prerequisites that no review here can discharge.

yuezengwu (approved):
Reviewed current head `b466a5f21f1e07c9b82bce033444a5e26427df8a`. The authoritative-session-loss finding from my previous review is resolved: a production-mode reconnect that receives `/me` 401 clears `user_id` before the login page view. A separate controlled fetch probe confirms that identity remains attached while revalidation is pending and after a transient network failure. Explicit sign-out still clears before its destination is reported.

Fresh validation at this head: check, build, typecheck, all 766 Web tests, the server static-app test, and eight independent integration/boundary probes pass. The probes also cover route-template sanitization, invite/unknown URLs, direct and SPA internal-preview suppression, sign-in intent expiry, campaign filtering, and referrer reduction. No further code blocker found in the reviewed flows.

A merge simulation against current main `e56659a37934b94f3307e0da98f84846cbe42046` is conflict-free; local tests above ran on the PR head, not the simulated merge. Current technical CI passes. This is code approval; collection enablement and the documented GA4 property settings remain deployment decisions, and live collection was not exercised.

PR #577, opened 2026-09-09
chore: upgrade context tree to 0.1.12

ggordonhall:
Upgrade the bundled Context Tree dependency from 0.1.11 to 0.1.12 for the CLI and client runtime. The portable-release dependency contracts now assert the upgraded version.

Validation:

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm --filter @opentag/client exec vitest run src/__tests__/context-tree.integration.test.ts src/__tests__/context-tree.test.ts` (47 tests)
- `pnpm test` (fails in two unrelated CLI daemon-wrapper tests: expected exit code 3, received 0 after the local `opentag-dev` service restarts)

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -51,7 +51,7 @@
     "test": "vitest run --maxWorkers=1"
   },
   "dependencies": {
-    "@first-tree-ai/context-tree": "0.1.11"
+    "@first-tree-ai/context-tree": "0.1.12"
   },
   "devDependencies": {
     "@opentag/client": "workspace:*",
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -29,7 +29,7 @@
     "zod": "^4.5.1"
   },
   "devDependencies": {
-    "@first-tree-ai/context-tree": "0.1.11",
+    "@first-tree-ai/context-tree": "0.1.12",
     "@types/semver": "^7.8.0",
     "@types/ws": "^8.18.1",
     "vitest": "^4.1.11"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -48,8 +48,8 @@ importers:
   apps/cli:
     dependencies:
       '@first-tree-ai/context-tree':
-        specifier: 0.1.11
-        version: 0.1.11
+        specifier: 0.1.12
+        version: 0.1.12
     devDependencies:
       '@opentag/client':
         specifier: workspace:*
@@ -189,8 +189,8 @@ importers:
         version: 4.5.4
     devDependencies:
       '@first-tree-ai/context-tree':
-        specifier: 0.1.11
-        version: 0.1.11
+        specifier: 0.1.12
+        version: 0.1.12
       '@types/semver':
         specifier: ^7.8.0
         version: 7.8.0
@@ -983,8 +983,8 @@ packages:
   '@fastify/websocket@11.3.0':
     resolution: {integrity: sha512-g89ag4BCcD9YP5wBZXixzoLnuf5j89p/sXFcfpCiv2pdEkYYukBEoK3heVzqsp0EAtszVDc2BBZG0KZqeAShIA==}
 
-  '@first-tree-ai/context-tree@0.1.11':
-    resolution: {integrity: sha512-VLlnDqQS3ysYicQnuT8AVYH0toezbs8jw/K1t1uPAvC2QxbtaLriP0cNiHYA7C+pTtHvdDv9iTt+FssYKa7fHQ==}
+  '@first-tree-ai/context-tree@0.1.12':
+    resolution: {integrity: sha512-kJ3fNpkCqgDNzjcMPAmrCUJ6iZQBpvXnvU3WAZ5ahvz3Zk+h6ZeyKilE3yrmBfozRYz6yxKp//y59/65Qo7cDQ==}
     engines: {node: '>=22.13.0'}
     hasBin: true
 
@@ -5339,7 +5339,7 @@ snapshots:
       - bufferutil
       - utf-8-validate
 
-  '@first-tree-ai/context-tree@0.1.11':
+  '@first-tree-ai/context-tree@0.1.12':
     dependencies:
       commander: 15.0.0
       mdast-util-from-markdown: 2.0.3
--- a/scripts/__tests__/portable-runtime-dependencies.test.mjs
+++ b/scripts/__tests__/portable-runtime-dependencies.test.mjs
@@ -179,7 +179,7 @@ test("the portable runtime dependency contract accepts only the supported exact 
 test("the checked-in apps/cli manifest is accepted under the portable shipping contract", () => {
   const cliManifest = JSON.parse(readFileSync(join(repoRoot, "apps", "cli", "package.json"), "utf8"));
   assert.deepEqual(readPortableDirectDependencyPins(cliManifest), [
-    { name: "@first-tree-ai/context-tree", version: "0.1.11" },
+    { name: "@first-tree-ai/context-tree", version: "0.1.12" },
   ]);
 });
 
@@ -531,13 +531,13 @@ test("the real frozen install closure assembles, verifies, and runs the embedded
   });
   assert.equal(closure.direct.length, 1);
   assert.equal(closure.packages[0].name, "@first-tree-ai/context-tree");
-  assert.equal(closure.packages[0].version, "0.1.11");
+  assert.equal(closure.packages[0].version, "0.1.12");
   assert.ok(closure.packages.length >= 20, `real closure is unexpectedly small: ${closure.packages.length}`);
   writeJson(join(appDir, "package.json"), {
     name: "open-tag",
     version: "0.0.2",
     type: "module",
-    dependencies: { "@first-tree-ai/context-tree": "0.1.11" },
+    dependencies: { "@first-tree-ai/context-tree": "0.1.12" },
   });
   writeDependencyClosureFile(appDir, closure);
   writeText(join(appDir, "cli", "index.mjs"), "export {};\n");


liuchao-001 (changes requested):
Reviewed at head `af5db90`. Ran `pnpm --filter @opentag/client exec vitest run src/__tests__/context-tree.test.ts src/__tests__/session-runtime-manager.test.ts src/__tests__/home-layout.test.ts` — 65/65 pass.

First, the good part: the `pathPrepend` refactor is a real fix, not just plumbing. The old `visibleProviderCliPath` returned a `PATH` into `workspace.environment`, which **replaced** the composed factory value and therefore silently dropped both `contextTreeBin` and the discovered executable directory for every visible Session. Composing at the factory and prepending last is the right shape, and driving the new per-provider tests through an actual spawned process rather than a stubbed client is the right level to assert it at.

The rest of the review is about scope. I unpacked `@first-tree-ai/context-tree` 0.1.11 and 0.1.12 side by side: the upstream delta is a new `context-tree-schedule-cleanup` skill, an updated cleanup skill, and CLI dist changes. Nothing upstream forces a config relocation — `<OPENTAG_HOME>/config/context-tree.json` is OpenTag's own file, written by `apps/cli/src/core/context-tree/shared.ts`. So the relocation, the new sandbox grant, and the shim-ordering change are independent behavior changes riding in a dependency bump, and the PR title and body describe only the bump. The branch itself is more honest than the description — it carries `fix(client)!: require Context Tree reconnect after config relocation`. Please retitle and rewrite the body; a reviewer scanning "chore: upgrade to 0.1.12" will not go looking for a machine-wide sandbox grant.

---

## Blocking

### 1. The new config write grant is a cross-Agent tree-hijack channel

`session-runtime-manager.ts:264,295,510` adds `<OPENTAG_HOME>/config/context-tree` to `writableRoots` for **every** Session, visible and internal, unconditionally — including when Context Tree is not configured and when `contextTree` is not wired at all. `writableRoots` is the Codex sandbox's writable set (`providers/codex/agent-runtime.ts:1232`), so this is a genuine write grant to the Agent, not just a path OpenTag tracks.

That directory holds `config.json`, the Computer-wide Context Tree target. On read, `ContextTreeManager.readConfig()` validates against `ContextTreeConfigSchema` only — shape, not provenance — and `connectArguments()` turns a `github` target into `connect <owner/repo>`, which clones over the network on a 120s budget.

Failure scenario: one Session is prompt-injected through ordinary untrusted input — a repo README, an issue body, a dependency's postinstall output. It writes `{"schemaVersion":1,"target":{"kind":"github","repository":"attacker/tree"}}` into the granted leaf. At the next Session start on that Computer, for **any** Agent including internal ones, OpenTag clones the attacker's repository, adds the resolved tree path to `writableRoots`, and renders its content into the system prompt as authoritative durable memory — "Read the decisions that bear on a task before planning or changing code, and record durable decisions there." No human is in the loop, `doctor` reports the new target as healthy, and the previous target is gone. The blast radius is the whole Computer, not the Session that wrote the file.

The design doc discloses the mechanism ("direct edits bypass command-level target validation") but treats it as a convenience and doesn't reason about it being a cross-Agent channel.

What makes me think the grant should just be dropped rather than hardened: OpenTag's own operations run as the daemon and are not sandboxed, so they never needed it. The CLI's connection store lives in the CLI's own home, not this leaf. And `managed-instructions.ts` already tells the Agent the opposite of what the grant enables — "do not create or reconnect a tree just because the task cwd changed", plus this PR's own new line, "report a runtime setup problem; do not install it globally." The grant's only purpose is to let an Agent repoint a Computer-wide target that the system prompt explicitly tells it not to repoint.

If there is a driving use case I'm missing, could you state it in the PR body? If it's "the Agent should be able to run `opentag context-tree connect`", the narrower shapes are a read-only grant, or routing the write through the daemon so command-level target validation still applies.

### 2. Every existing installation silently loses its tree, and nothing says so

The relocation is deliberate and documented, and I'm not arguing the destination — moving to a leaf directory so the grant never touches `<OPENTAG_HOME>/config`, where the machine credential lives, is sound. What I'd push back on is the silence.

An existing user with a connected managed tree upgrades. Their next Session's prompt reads "Context Tree: not configured on this Computer" and "Do not assume earlier decisions were recorded." Agents keep working, without durable memory, until a human happens to notice. `doctor` reports plain unconfigured — indistinguishable from a Computer that never connected one.

The old file is deliberately left on disk, so the detection is one `access()` call. Either read it as a one-time migration, or keep the hard cut and have `readContextTreeState`/`doctor`/the prompt distinguish "a legacy configuration is present, run `opentag context-tree connect <previous target>`" from "never configured." Right now the upgrade note lives only in a design doc, which is the one place an affected user won't be looking. This also needs to be in the PR body.

---

## Non-blocking

**`context-tree.ts:240-251` — unconfigured Computers now report a fault.** `executableFailure` is computed before the config read, so a Computer that never configured Context Tree gets `unavailable/SHIM_UNAVAILABLE` on win32, or `PACKAGE_MISSING`, instead of `unconfigured`. The prompt then reads "Context Tree unavailable (SHIM_UNAVAILABLE) ... do not attempt to repair the tree yourself" rather than the actionable "not configured on this Computer (`opentag context-tree connect`)". For `PACKAGE_MISSING` that's an improvement — it's a real install fault. For win32 it turns "you have never used this feature" into a permanent fault message with no remedy. It's intentional and tested, so I may be missing the motivation; worth confirming you meant it for the win32 case specifically.

**`session-runtime-manager.ts:76,96` — `home` becoming optional is a footgun.** `resolveOpenTagHomeLayout(undefined)` falls back to `OPENTAG_HOME` or the real `~/.opentag`, so a manager constructed without `home` would `mkdir` and grant a directory in the developer's actual home while `OPENTAG_HOME` is not exported to the workspace. Production always passes it (`client-runtime-composition.ts:622`), and the new test needed `vi.stubEnv("OPENTAG_HOME", …)` in `beforeEach` to avoid exactly this. Making `home` required would remove the class.

**`context-tree.ts:331-333` — the two guards say one thing.** The `Date.now() < this.#shimRetryAt` check is subsumed by the `=== 0` check plus the cooldown; `if (this.#shimPreparation && (this.#shimRetryAt === 0 || Date.now() < this.#shimRetryAt))` reads as the single rule it is. The logic is correct as written — success caches forever, failure retries after the cooldown, concurrent callers share the in-flight promise.

Gandy2025 (changes requested):
Reviewed the immutable range `dcae5e42ccd6200f612dc5a3d2d24c14fc759832...af5db90b0e9c3b63a16a499930f41a1f87184e52` as high risk.

I independently confirmed the two current security/upgrade blockers in @liuchao-001's review: the new Computer-wide configuration writable root is a cross-Agent target-replacement channel, and the hard path relocation makes an existing connection indistinguishable from a never-configured Computer. I will not duplicate that review line by line. This review adds one 0.1.12-specific blocker and one separate merge-scope blocker.

## Standards

**Blocking — the current PR range also contains the five commits under review in #562.** The stated change is a four-file dependency pin/lock/portable-contract update, but the actual `main...head` range is six commits and 27 files. Merging this PR now would also merge #562's breaking relocation, sandbox grant, and runtime lifecycle work while #562 is still open with changes requested. That conflicts with the repository requirement to keep PRs focused and keep unrelated behavior out of the same review object.

Please either retarget #577 to `fix/context-tree-config-runtime-access` while #562 is open, or recreate the version bump from `main` so this PR contains only `af5db90b`'s four files. The latter is preferable if 0.1.12 does not depend on #562.

**Non-blocking contract gap — `scripts/portable/runtime-dependencies.mjs:37-44`.** `CONTEXT_TREE_SKILL_DIRECTORIES` says it enumerates the skills installed by the packaged CLI, but 0.1.12 installs eight skills while this contract names six; it omits both `context-tree-cleanup` and the new `context-tree-schedule-cleanup`. Consequently `assertContextTreeAssets` and its synthetic fixture can pass without either asset. If 0.1.12 remains the target, update the asset contract to cover the actual installed skill set.

## Spec

**Blocking — 0.1.12's new scheduled-cleanup skill is not usable from a managed Codex Session.** `packages/client/src/runtime/context-tree.ts:291-295` installs every packaged skill into Codex. The published 0.1.12 installer now includes `context-tree-schedule-cleanup`, whose CLI writes private state under `$HOME/.context-tree/cleanup` and native scheduler files under `~/Library/LaunchAgents` on macOS or `~/.config/systemd/user` on Linux. OpenTag's Codex policy remains `workspace-write` with `approvals: never` (`packages/client/src/providers/codex/runtime-policy.ts:4-9`), and the roots assembled at `packages/client/src/runtime/session-runtime-manager.ts:288-297` include none of those locations.

Therefore `cleanup schedule` and `cleanup remove` fail under managed Codex. More subtly, the CLI's successful `create`/`connect`/`sync`/`read`/`prepare-write`/`finish-write` post-action tries to refresh the same private activity state and catches every error; an externally created schedule can silently age past the 24-hour activity window even while the tree is actively used from OpenTag Codex.

The safe correction is not a broad account-home grant. Either hold this dependency for Codex / avoid installing or advertising the scheduling skill there, or route scheduling and activity updates through an explicitly authorized daemon operation and cover both Codex and Claude behavior end to end.

The isolated version-bump commit itself correctly updates both manifests, the lockfile, and the three exact-version assertions. The portable dependency suite passes 19/19, the installed package resolves as 0.1.12, and its lockfile integrity matches the published package. Those checks do not close the blockers above.

liuchao-001:
Independently verified @Gandy2025's two new blockers before endorsing them. Both hold, and one of them corrects the reach of something I wrote.

**The branch-scope blocker is exact, and slightly worse than stated.** `gh api .../pulls/562/commits` and `.../pulls/577/commits` return the same five SHAs — `af9deae2`, `d3e124e8`, `bc3b8a4b`, `775dafc5`, `0dacb635` — and #562 is OPEN with CHANGES_REQUESTED. Both PRs target `main`, so merging #577 merges #562's work while #562 is still under review. The extra detail: #577 forked from #562 at `0dacb635`, *before* #562's most recent commit `8556fad8 test(client): await shared Context Tree shim preparation`. So this PR would land #562's relocation and sandbox grant at a stale point, without a test fix its author has since made. Recreating the bump from `main` so #577 is only `af5db90b`'s four files is the right call.

**The scheduled-cleanup blocker is real and it's the 0.1.12-specific one.** From the published 0.1.12 tarball, `skills/context-tree-schedule-cleanup/SKILL.md` states plainly: "macOS uses user LaunchAgents; Linux uses systemd user timers/services", and `dist/cli/index.mjs` carries the matching `LaunchAgents` and `.context-tree` string references. `codex/runtime-policy.ts:5-8` is `fileSystem: "workspace-write"` with `approvals: "never"`, and the roots assembled in `session-runtime-manager.ts` include none of those locations. `install --host codex` installs the whole packaged skill set, so the skill ships into managed Codex Sessions already unable to execute it.

The activity-window consequence is the sharp end, and the skill documents it against itself: "Initial scheduling starts a 24-hour activity window. Successful ordinary create, connect, sync, read, prepare-write, and finish-write use refreshes activity for scheduled trees" and "Missing or old activity skips before network or model work." Those are exactly the operations OpenTag drives. If their activity refresh can't write under the sandbox and the failure is swallowed, a tree under constant use from managed Codex looks idle to the scheduler and cleanup silently stops. Agreed the correction is not a broad account-home grant.

One nuance on the non-blocking contract gap: `CONTEXT_TREE_SKILL_DIRECTORIES` (`scripts/portable/runtime-dependencies.mjs:35-43`) does name six against 0.1.12's eight, but `context-tree-cleanup` shipped in 0.1.11 too, so that half of the gap predates this PR — 0.1.12 widens an existing hole rather than opening it. Worth fixing here either way, since `assertContextTreeAssets` currently passes without the assets it claims to cover.

Finally, to correct the reach of my own review: I wrote that the upstream 0.1.11→0.1.12 delta is a new scheduling skill, an updated cleanup skill, and CLI dist changes, and that nothing upstream forces the config relocation. That's accurate, but it answered only the question I was asking — whether the relocation was externally driven. It should not be read as clearing 0.1.12 itself. I listed the new skill and drew no conclusion from it; @Gandy2025 asked what it *does*, which was the question the bump actually deserved. My two blockers stand as written, with that one scoped down.

liuchao-001 (commented):
The retarget resolves my scope objection completely. `main...head` is now the single commit `af5db90b` over four files — both manifests, the lockfile, and the three exact-version assertions — which is exactly what the title has said all along.

**Both of my blockers are withdrawn from this PR.** The config relocation and the configuration writable root are #562's content and are no longer in this range, so neither is a reason to hold #577. I should have checked whether those commits belonged to an open PR before framing them as undisclosed changes riding in a bump; @Gandy2025 asked the question I skipped, and retargeting was the better of the two remedies.

Having now read #562's thread, I also want to be accurate about novelty: the write authority is not an unnoticed defect there. @yuezengwu is explicitly withholding approval on maintainer acceptance of it, and the approving review carves it out by name. My silent-upgrade point was likewise already raised there and consciously deferred to a follow-up. I've moved the part of my analysis that is actually additive — the concrete consequence chain — to #562, where the decision lives.

My CHANGES_REQUESTED stays on this PR, but for @Gandy2025's reason rather than mine: 0.1.12 installs `context-tree-schedule-cleanup` into managed Codex Sessions that cannot execute it, and the activity-refresh path can silently age out a schedule on a tree that is actively in use. I verified that against the published tarball and `codex/runtime-policy.ts` in the comment above and I still think it's the right hold on the bump itself.

The bump mechanics are clean: lockfile integrity matches the published package, and the four files are internally consistent.

Gandy2025 (changes requested):
Re-reviewed after the base was retargeted. The new review object is `0dacb63515f5c1a78ec853d5d2b22df3b45613a3...af5db90b0e9c3b63a16a499930f41a1f87184e52`; merging it onto the current base head `8556fad8` is conflict-free and changes only the two manifests, lockfile, and portable dependency test.

The retarget fully closes my earlier scope blocker. The configuration writable-root and reconnect findings are #562's content and no longer count against #577. Changes remain requested for two issues directly introduced or exposed by the 0.1.12 bump.

## Standards

**Blocking — the portable required-assets contract does not cover 0.1.12's runtime assets.** `scripts/portable/runtime-dependencies.mjs:36-46,502-515` claims to enumerate the packaged skills the CLI installs, but the 0.1.12 installer returns eight while the contract names six. It omits:

- `skills/context-tree-cleanup/SKILL.md`
- `skills/context-tree-schedule-cleanup/SKILL.md`
- the cleanup skill's required `skills/context-tree-cleanup/references/editorial.md`

Materialization currently copies the whole locked tarball, so today's artifact contains them, but verification still accepts an artifact with any of these files removed. The runtime probe exercises only version/help/list, and the synthetic fixture is built from the same incomplete list, so all portable checks remain green while installed cleanup is broken. Because this PR explicitly advances the portable dependency contract to 0.1.12, please add both skills and the required nested resource to the asset contract, with deletion-failure coverage.

## Spec

**Blocking — the new scheduled-cleanup skill is installed into managed Codex Sessions that cannot execute it.** The base behavior at `packages/client/src/runtime/context-tree.ts:291-295` installs every packaged skill. Version 0.1.12 newly installs `context-tree-schedule-cleanup`, whose implementation writes state under `$HOME/.context-tree/cleanup` and scheduler files under `~/Library/LaunchAgents` on macOS or `~/.config/systemd/user` on Linux. OpenTag Codex remains `workspace-write` with `approvals: never` (`packages/client/src/providers/codex/runtime-policy.ts:4-9`), and its assembled writable roots contain none of those locations.

Consequently Codex `cleanup schedule` and `cleanup remove` fail. The ordinary `create`/`connect`/`sync`/`read`/`prepare-write`/`finish-write` post-action also catches and discards activity-write failures, so an externally created schedule can silently age past its 24-hour activity window while the tree is actively used from OpenTag Codex.

The safe correction is not a broad account-home grant. Either avoid installing/advertising the scheduling skill in managed Codex until the upstream installer supports that boundary, or route scheduling and activity updates through an explicitly authorized daemon operation and test both Codex and Claude behavior.

The four-file bump mechanics are otherwise clean: the manifests, lockfile, and exact-version assertions agree; the portable dependency suite passes 19/19; CI is green; and the lock integrity matches the published 0.1.12 package.

liuchao-001:
Verified the escalated asset-contract blocker, and it corrects something I wrote earlier in this thread.

**My "predates this PR" nuance was wrong where it counts, so please don't lean on it.** I said the `context-tree-cleanup` half of the contract gap pre-existed and that 0.1.12 only widened an existing hole. The skill *directory* did ship in 0.1.11, but I checked both tarballs again and 0.1.11's cleanup skill contains only `SKILL.md` and `agents/openai.yaml`. `references/editorial.md` is **new in 0.1.12**, and 0.1.12's `SKILL.md:19` makes it mandatory — "Read and follow [the shared editorial instructions](references/editorial.md)", with line 28 calling it "its required editorial resource." So the missing required resource is introduced by this bump, not inherited, and the fix belongs in this PR rather than a follow-up.

**A structural note on the fix, beyond adding the two names.** `assertContextTreeAssets` (`runtime-dependencies.mjs:503-515`) builds its asset list as CLI path + `skills/<name>/SKILL.md` + templates. It only ever checks one file per skill, so adding `context-tree-cleanup` and `context-tree-schedule-cleanup` to `CONTEXT_TREE_SKILL_DIRECTORIES` still would not cover `references/editorial.md` — the asset shape cannot express a nested required resource for any skill. A cleanup skill whose `SKILL.md` is present but whose required editorial resource is gone would pass verification and be broken at runtime. The contract needs both the new entries and a way to name per-skill required files.

**The fixture point is exact.** `portable-runtime-dependencies.test.mjs:78-79` builds the synthetic package by looping over `CONTEXT_TREE_SKILL_DIRECTORIES` and writing a `SKILL.md` for each. The fixture is generated from the very list under test, so it cannot fail for a skill the list omits — by construction, not by oversight. That's why all portable checks stay green while installed cleanup is broken, and it's the reason deletion-failure coverage is the right ask rather than just a longer list.

No disagreement on the scheduled-cleanup blocker; I verified it against the published tarball and `codex/runtime-policy.ts` above and endorse it as written. Both remaining blockers are @Gandy2025's and both are correctly scoped to this bump.

liuchao-001:
Mechanical note, since this PR's diff silently re-expanded and my earlier withdrawal now reads as false.

#562 was **squash-merged** (`e56659a3`, single parent `52438e06`), so its five commits are not ancestors of `main` — `git merge-base --is-ancestor af9deae2 origin/main` fails. GitHub auto-retargeted this PR from the merged base back to `main`, and because the branch still carries those five commits as its own, `main...head` recomputes to 27 files and the PR is now `CONFLICTING`.

To be clear about what that does and doesn't mean: the relocation and the configuration writable root are **on main** as of `e56659a3`. They are not being reintroduced here. The 27-file diff is a rebase artifact, and my withdrawal of those two blockers stands — please don't re-review them against this PR.

A rebase onto current `main` should drop the five now-redundant commits and return this to `af5db90b`'s four files. The two open blockers from @Gandy2025 — the portable asset contract missing 0.1.12's cleanup skills and the required `references/editorial.md`, and the scheduled-cleanup skill installed into managed Codex Sessions that cannot execute it — are unaffected by the rebase and still apply to the bump itself.

yuezengwu (changes requested):
**Review decision update (2026-09-09):** The two remaining P2 findings are accepted as non-blocking follow-up work for #577. They do not need to be fixed before this PR merges. My request for changes on these findings is withdrawn; I am approving current head `f0a31f44b621cd5e067abbd3b3a6e6f187c56724`. The findings remain recorded below for follow-up.

---

Historical review below; the merge conflict was resolved by the subsequent rebase.

Reviewed head `af5db90b0e9c3b63a16a499930f41a1f87184e52` against current main `e56659a37934b94f3307e0da98f84846cbe42046` after #562 was squash-merged and this PR was retargeted again.

**Integration is currently blocked.** The branch still carries the pre-squash #562 history. A fresh `git merge-tree --write-tree` against main reports a content conflict in `packages/client/src/__tests__/context-tree.test.ts`, where main has the newer `vi.waitFor` assertion. The earlier conflict-free four-file merge result against branch `8556fad8` is now historical. Reconcile this branch with current main, retaining the timing fix, before merging. The two findings below concern the 0.1.12 dependency change itself; #562's already-merged configuration policy is not a new finding on the bump.

**[P2] The portable verifier accepts missing resources required by the upgraded package.** I materialized the actual frozen dependency closure, then separately removed each of:
- `skills/context-tree-cleanup/SKILL.md`
- `skills/context-tree-schedule-cleanup/SKILL.md`
- `skills/context-tree-cleanup/references/editorial.md`

Both `verifyPortableDependencyGraph()` and `runContextTreeRuntimeProbe()` still passed after every deletion. The complete package passes too. An isolated `install --host codex` installs eight skills, while `CONTEXT_TREE_SKILL_DIRECTORIES` enumerates six; the cleanup skill now requires the nested editorial resource, which the asset list cannot express. Today's complete tarball includes these files, but the shipping gate cannot detect their loss. Extend the required-assets contract to the installed skill set and required nested resources, with independent deletion-failure coverage.

**[P2] The new scheduling skill is installed into managed Codex without the permissions its implementation requires.** The isolated install confirms that `context-tree-schedule-cleanup` is advertised. The published CLI writes cleanup state under `$HOME/.context-tree/cleanup` and registers files under `~/Library/LaunchAgents` or `~/.config/systemd/user`. OpenTag's default Codex policy is `workspace-write` with `approvals: never`, and the assembled grants cover none of those locations. The ordinary-command activity hook catches write failures; the scheduled runner skips when recorded activity is missing or older than 24 hours. Thus scheduling is unsupported under that policy, and a separately created schedule can lose activity tracking. This conclusion is from the published implementation and runtime policy; I did not invoke a real host scheduler or authenticated Codex session. Gate the scheduling capability for managed runtimes or provide an explicitly authorized integration, and verify activity updates as well as schedule/remove.

These independently support the two remaining findings in review #5149861637.

Validation: registry tarball SHA-512 matches the lockfile; check, build, typecheck, 19 portable tests, and 47 Context Tree tests pass. The negative asset probe above reproduces the validation defect. Tests apply to the PR head; integration into current main is blocked by the conflict above.

liuchao-001:
Verified at the new head `f0a31f44`. **The rebase is correct and complete — no need to redo it.**

- `main...head` is one commit over four files, `+12/-12`, all of it version strings plus the integrity hash. `MERGEABLE` against current main.
- @yuezengwu's conflict is cleanly resolved rather than papered over: the branch no longer touches `packages/client/src/__tests__/context-tree.test.ts` at all, so main's `vi.waitFor(() => expect(writes).toHaveBeenCalledTimes(1))` timing fix stands untouched at line 353. Dropping the five squashed commits was the right resolution.

**Both blockers are unchanged at this head, so my requested changes stand.** `scripts/portable/runtime-dependencies.mjs` is not in the four files, and `CONTEXT_TREE_SKILL_DIRECTORIES` still enumerates exactly the same six names — no `context-tree-cleanup`, no `context-tree-schedule-cleanup`, and no way to express the nested `references/editorial.md` that 0.1.12's cleanup skill now requires. Nothing gates the scheduling skill for managed Codex either.

To be explicit about what's left, since three reviewers are holding on the same two items:

1. Extend the required-assets contract to the installed skill set and to per-skill required nested resources, with deletion-failure coverage. @yuezengwu's probe is the acceptance test — after the fix, removing any of the three files should fail verification, where today all three pass.
2. Gate or authorize the scheduling capability for managed runtimes, covering activity updates as well as `schedule`/`remove`.

The bump mechanics remain clean at this head.

yuezengwu (changes requested):
**Review decision update (2026-09-09):** The two remaining P2 findings are accepted as non-blocking follow-up work for #577. They do not need to be fixed before this PR merges. My request for changes on these findings is withdrawn; I am approving current head `f0a31f44b621cd5e067abbd3b3a6e6f187c56724`. The findings remain recorded below for follow-up.

---

Re-reviewed current head `f0a31f44b621cd5e067abbd3b3a6e6f187c56724` against main `e56659a37934b94f3307e0da98f84846cbe42046`. The rebase resolves the merge conflict and leaves only the four dependency/version files in scope. The two dependency-compatibility findings remain:

**[P2] Required-resource validation still accepts an incomplete portable package.** I rebuilt the real frozen dependency closure at this head and separately deleted `skills/context-tree-cleanup/SKILL.md`, `skills/context-tree-schedule-cleanup/SKILL.md`, and `skills/context-tree-cleanup/references/editorial.md`. Both `verifyPortableDependencyGraph()` and the runtime probe passed after each deletion. The complete tarball contains the files; this is a missing validation gate. The actual isolated Codex install exposes eight skills, but the contract still checks only six and cannot express the required nested editorial resource. Please extend the required-assets contract and add independent deletion-failure coverage.

**[P2] Managed Codex still receives a scheduling capability its default policy cannot execute.** Fresh installation again includes `context-tree-schedule-cleanup`. The published CLI writes activity/configuration under `$HOME/.context-tree/cleanup` and scheduler files under `~/Library/LaunchAgents` or `~/.config/systemd/user`. The unchanged managed Codex policy uses `workspace-write` with `approvals: never`; its assembled writable roots cover none of those locations. Activity-update failures are swallowed, and the scheduled runner skips missing or older-than-24-hour activity. Gate this capability for managed runtimes or provide an explicit supported integration, including activity tracking and schedule/removal validation. This finding is based on the installed package and runtime policy; no real scheduler or authenticated Codex session was invoked.

Fresh validation: check, build, typecheck, 19 portable tests, and 47 Context Tree tests pass. The independent negative-asset probe reproduces the first defect. The current head merges cleanly with main; current technical CI passes. The previous merge-conflict finding is withdrawn. The two P2 findings above are retained as non-blocking follow-up work, as recorded in the decision update above.

yuezengwu (approved):
Approved at `f0a31f44b621cd5e067abbd3b3a6e6f187c56724`. I have updated my earlier review comments: the remaining P2 findings (portable required-resource validation and managed Codex scheduling compatibility) are accepted as non-blocking follow-up work and do not block merging this PR. My previous request for changes on these findings is withdrawn. The earlier merge conflict was already resolved by the rebase.

liuchao-001 (approved):
Approving at head `f0a31f44`, and withdrawing my request for changes.

My hold was explicitly derivative — I wrote that it stayed "for @Gandy2025's reason rather than mine." @yuezengwu verified both findings more rigorously than either of us argued them, and has now accepted them as non-blocking follow-up. That is a risk-tolerance call within their remit, and I don't have grounds to hold the PR against it.

To be clear about what I am and am not saying: my technical assessment of the two findings hasn't changed, and I'm not re-characterising them as harmless. I'm deferring on whether they must be fixed *before this bump lands*, and I think deferring is defensible on both — the shipped tarball is complete, so the asset-contract gap is a missing gate rather than a broken artifact; and the scheduling capability is new upstream surface nobody depends on yet, not a regression in something that worked.

**I've opened #586 to carry both findings**, because "accepted as non-blocking follow-up" with nothing tracking it is how findings evaporate. It records both with their evidence and, for the asset contract, @yuezengwu's probe as the acceptance test: after the fix, deleting any one of `skills/context-tree-cleanup/SKILL.md`, `skills/context-tree-schedule-cleanup/SKILL.md`, or `skills/context-tree-cleanup/references/editorial.md` must fail verification, where all three pass today.

Verified at this head myself: four files, `+12/-12`, version strings and the integrity hash only; `MERGEABLE`; and the rebase correctly left main's `vi.waitFor` timing fix untouched by dropping the five squashed commits rather than resolving into them.

The merge gate is not clear yet — @Gandy2025's request for changes still stands at this head, and it is theirs to release.

Gandy2025 (approved):
Approved at exact head `f0a31f44b621cd5e067abbd3b3a6e6f187c56724` against base `e56659a37934b94f3307e0da98f84846cbe42046`. My previous request for changes is withdrawn.

**Standards:** pass. The corrected range is one commit over four dependency/version files, is mergeable, and leaves the timing fix already on `main` untouched. The manifests, lockfile integrity, and exact-version assertions agree; the technical and ownership checks are green.

**Spec:** pass for this PR. The two P2 findings remain technically valid, but the maintainers have explicitly accepted them as non-blocking follow-up work. That risk disposition is reasonable here: the published artifact is complete, so the portable-assets issue is a validation-gate gap rather than a broken shipped package; and the managed-Codex scheduling issue affects newly exposed upstream surface rather than regressing an existing working capability.

#586 now carries both findings, including the independent deletion-failure acceptance test for the required assets and the requirement to gate or explicitly authorize scheduling plus activity tracking. With that follow-up recorded, I have no remaining blocker on #577.

PR #580, opened 2026-09-09
docs(readme): update

ggordonhall:
Improve user-facing docs


liuchao-001 (commented):
Reviewed at head `243e444f`. I checked the rewritten docs by running every command, script name, env var and link against the tree rather than reading the prose — the development guide holds up well. One thing outside the stated scope I'd like your call on before approving.

## Please confirm: removing `TRADEMARKS.md` is deliberate

`756e60b8` deletes `TRADEMARKS.md` and `TRADEMARKS.zh-CN.md`, and strips the pointer to them from `brand-mark.tsx`. The PR title is `docs(readme): update` and the body is "Improve user-facing docs", so this reads as incidental cleanup — but all ten vendor marks it covered are still shipped in `apps/web/src/assets/` (`slack.svg`, `feishu.svg`, `claude.svg`, `openai-blossom-{black,white}.svg`, `google-g.png`, `integration-{github,google-drive,linear,notion}.svg`, plus sentry/figma).

After this PR the only "trademark" text left in the repo is Apache-2.0 §6 boilerplate in `LICENSE`, which covers the *Licensor's* marks and says nothing about third-party marks we redistribute. Specifically, these no longer exist anywhere:

- The non-endorsement disclaimer for the marks we display.
- The statement that third-party marks and their asset files are **not** offered under Apache-2.0 — the deleted doc was explicit that recording provenance establishes provenance, not permission.
- The reservation of OpenTag's *own* name and mark as outside the repository licence.
- The per-vendor record of which published guidelines we're keeping to (Slack media kit, OpenAI Marks usage terms, Google identity branding guidelines, GitHub logo guidelines, Linear, Notion, Sentry, Figma).
- The "Adding another" procedure telling contributors to take files from the publisher's brand kit unmodified.
- Provenance for `google-g.png`. The SVGs each keep an in-file source comment (`slack.svg` → `<!-- Official Slack mark sourced from ... on 2026-08-29. -->`), so those survive, but bitmap provenance lived only in the deleted table.

I'm not asserting a legal position, and this may well be an intentional owner decision — but it's a different kind of change from a README rewrite and isn't disclosed in the PR. Could you either (a) confirm it's deliberate and split it into its own PR so it gets reviewed on its merits, or (b) keep the file? If the goal was just to declutter the repo root, `docs/trademarks.md` with the nav link would get you that without dropping the content.

## Non-blocking

The removed `## Project status` section is mostly correctly retired — I verified `app.opentag.build` serves 200 and `open-tag@0.0.3` is published to npm, so "There is no hosted OpenTag service" and "install channels are not yet publicly available" were both stale. It also dropped "pre-alpha: public APIs and package boundaries may change before the first stable release" and "the Skills and Integrations pages use demo data". Pointing new users at hosted signup while retiring the maturity warning is a positioning call — just flagging it's now gone, entirely your call.

## Verified working

For the record, everything the new `DEVELOPMENT.md` instructs actually resolves:

- `scripts/dev-install.sh` does run `pnpm install --frozen-lockfile` and `pnpm build`, matching the step 1 description.
- `bootstrap:admin` and `start` both exist; `admin/bootstrap-cli.ts` calls `migrateDatabase` itself, so running it before the first server start (step 3) works against an empty database.
- The dev-auth trio is right, and it's a subtle one: `config.ts` requires `OPENTAG_DEV_AUTH_BYPASS_ENABLED` and `OPENTAG_DEV_AUTH_EMAIL` *together* plus an explicitly-set `OPENTAG_ENV=dev`. The `.env.local` template sets all three.
- `docker compose up -d --wait postgres` is safe — the service has a healthcheck, so `--wait` has something to wait on.
- Nothing in the repo auto-loads `.env` or `.env.local`, so `set -a; source .env.local` is genuinely required and correctly repeated in the restart section. `.gitignore` `.env.*` covers the file, and `umask 077` is a nice touch.
- `daemon start|stop|restart|status`, `computer connect` and `doctor` all exist as documented.
- Troubleshooting paths are right for the dev channel: `channel-config.json` gives `serviceId: opentag-dev`, so the systemd unit really is `opentag-dev.service` and the home really is `~/.opentag-dev`.
- All 47 local links across the four root docs resolve, with no dead anchors. `DEVELOPMENT.md#git-hooks-and-worktrees` is preserved, which matters because `CONTRIBUTING.md:13` links to it.
- The `DEVELOPMENT.zh-CN.md` code blocks are byte-identical to the English ones, and the README mirrors are structurally in sync.

Also: `243e444f` already fixed the dead `[Docs](#documentation)` anchor that `696cbd4a` left behind when it removed the `## Documentation` section — good catch, and the English side now matches what the Chinese mirror was already doing.

liuchao-001:
Re-checked at `60ec3950`. All 47 local links across the four root docs still resolve, and the two layout commits were applied symmetrically to both README mirrors — no drift.

One small thing in the block you're currently editing: swapping `# OpenTag` for `<h3>OpenTag</h3>` leaves the README with no `h1`, and the outline now runs `h3` → `h2` (`## About`), which is out of order for screen readers and for GitHub's generated document outline. If the goal is a visually smaller title under the logo, `<h1>` in that same raw-HTML spot renders at whatever size you style it and keeps the hierarchy intact.

The `TRADEMARKS.md` question from my earlier review is still open — that's the one I'd like a call on before approving.

yuezengwu (commented):
Reviewed head `ace657acde36e275a2d0f655d7a5a4998f95f900` against main `acf7609b9bfdd5fc999ebbc23ef68fc2eb1a7108`.

The README/development-guide rewrite and AGENTS corrections check out in the reviewed paths. Both README titles are now H1, so the earlier heading-level concern is resolved. All 51 local references and anchors resolve, all 14 Bash blocks parse, and the English/Chinese instructions agree. I checked the documented commands against package scripts, the installer, generated connect commands, bootstrap/migrations, loopback auth, Compose defaults and channel configuration. In particular, the generated dev connect command supplies an absolute CLI path and PATH prefix, so removing the manual PATH step is supported. Fresh `pnpm check`, `pnpm build`, `pnpm typecheck`, and both dev-installer tests pass. This is source and isolated-check validation, not a fresh real-account setup run.

**The remaining approval question is the removal of the asset provenance and license-scope documentation.** I independently confirmed that all 12 assets listed in the deleted `TRADEMARKS.md` remain byte-for-byte unchanged. The PR removes both language versions, the README notice, and the pointer in `brand-mark.tsx`; the previously recorded Google bitmap source URL also disappears from the tracked tree. This removes repository information beyond the stated “Improve user-facing docs” scope. The existing SVG source comments do not replace that central record or the distinction it made between the code licence and the marks/assets.

Please record the intended treatment of those notices and asset provenance as an explicit maintainer decision in the PR description. Retaining or relocating the two documents would preserve that information while allowing the README cleanup. I am leaving approval pending that scope decision; this is a documentation/ownership observation, not a legal conclusion. This reconfirms the earlier review question on the current head, rather than treating its previous-head evidence as current.

yuezengwu (approved):
Approved at `ace657acde36e275a2d0f655d7a5a4998f95f900`. I defer the documentation scope and remaining review decisions on this PR to the author and liuchao-001. My earlier approval hold is withdrawn.

PR #582, opened 2026-09-09
fix: polish floating navigation and enable local Internal tools preview

Gandy2025:
## Summary

The floating navigation felt heavy, account-menu hover feedback was faint, and long Agent names could expand the switcher beyond the viewport. This keeps the floating frame and stable content position while making the navigation lighter and the menus easier to use. Local acceptance also needs the existing Internal tools entry, which now has an explicit development configuration gate.

- Remove the extra Home/Account dividers; refine frame borders, shadows, avatar states, popup spacing, and focus/tooltip behavior. Bound the Agent switcher and give mobile menu items 44px touch targets. Both menus share a 24px icon column with 16px glyphs.
- Keep popup menus within their existing sidebar/header landmark, with accessible names inherited from their triggers. Do not introduce a transient navigation landmark. Sign out intentionally uses neutral styling and its own separated group: it ends the session without deleting Account data.
- Preserve Computers, Account, Internal tools, and Sign out. Add an All Agents return link to Internal tools.
- Offer the existing Internal tools during local acceptance through an explicit, loopback-only development opt-in. Staging behavior stays the same; production remains closed. Document the setting in English and Chinese and exercise the real tools in E2E tests.

## Testing

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test` with a fresh temporary `OPENTAG_HOME`: all workspace suites passed before the review follow-up; the isolated directory prevents CLI tests from reading workstation credentials. Directly affected tests were rerun after the follow-up (results below).
- `pnpm --filter @opentag/client test:agent-runtime:coverage`: **378 passed**; statements, branches, functions, and lines all **100%** on the completed rerun.
- `pnpm --filter @opentag/server test:integration`: **330 tests verified across the full run and a serial retry**. The full run passed 209 tests; four suites hit a local Docker port-binding timeout during setup. All four suites passed on serial retry (**121 passed**). CI will also validate the complete suite.
- Complete Playwright journey and smoke suites: **38 passed** (25 journey, 13 smoke), using an isolated local server and database. The five navigation polish tests were also rerun after the final menu open-state adjustment.
- Directly affected frontend tests: **64 passed**; server configuration tests: **27 passed**.
- Internal navigation preview tests restore the original Server-wide settings in fixture teardown, including assertion-failure paths. A local fault-injection probe verifies that a subsequent test sees the original settings.
- Final code review and `git diff --check` completed.

## UI validation

- Desktop evidence: locally inspected workspace account popup and bounded Agent switcher captures generated by `e2e/tests/navigation-polish.spec.ts` (`workspace-account.png`, `agent-switcher.png`, under each test's output directory).
- Mobile evidence: locally inspected `mobile-account.png` at 320px and 390px, with all four destinations visible and no horizontal overflow.
- Widths checked: **320px / 390px / 768px / 1440px**; additional visual review at 1920px and short desktop viewport heights.
- States verified: workspace/Agent transitions, Overview, Tasks, Usage, settings, account destinations, long names, both Internal tools switches, both reset-dialog cancellations, and Agent Setup lab navigation. The workspace dock remains 64 × 128px; the expanded floating frame remains 240px wide with a 12px outer inset.
- Keyboard/accessibility checks: menu Home/End navigation, Escape and focus restoration, outside-click dismissal, tooltip/menu mutual exclusion, route history, dirty-settings blockers, reduced motion, and automated axe checks.
- New reusable block, theme value, or CSS exception: shared `MenuItemIcon` keeps both icon columns aligned; scoped account/Agent popup styles use existing theme tokens. Fixed menu positioners use z-index 50 so later content cannot cover them within their owning landmark. No new dependencies or theme tokens.

## Breaking changes

None. `OPENTAG_DEV_INTERNAL_TOOLS_ENABLED` defaults to false and requires explicit `OPENTAG_ENV=dev` plus loopback bind/public URL. It does not enable production access. No database migrations.

## Checklist


liuchao-001 (commented):
Reviewed at head `5a30a69c` in a fresh worktree off `origin/main` (`acf7609b`).

The dev opt-in gate is the part I looked at hardest, and it holds up: `OPENTAG_DEV_INTERNAL_TOOLS_ENABLED` mirrors the existing `OPENTAG_DEV_AUTH_BYPASS_ENABLED` gate exactly — explicit `OPENTAG_ENV=dev`, loopback `OPENTAG_HOST`, loopback `OPENTAG_PUBLIC_URL` — and `offersInternalTools()` re-derives the same answer independently of the refinement, so there is no path where `prod` or `staging`-with-flag reaches `true`. `OnboardingResetService.enabled` still re-checks per request, `internalNavigationService` is still only wired when `setupResetService` exists, and the `stagingSetupReset` → `internalTools` rename leaves no stragglers.

Verified locally at this head: `@opentag/web` vitest 802 passed (73 files), `packages/server` `config.test.ts` 27 passed, `biome check apps/web packages/server e2e` clean (warnings only).

## Must fix

**1. `apps/web/messages/common/zh.json:27` — stray space left by the `staging` removal**

```json
"common_internal_navigation_save_failed": "无法保存 导航设置。"
```

Deleting `staging` from `"无法保存 staging 导航设置。"` left the space behind. Should be `"无法保存导航设置。"`. This is user-visible Chinese copy on the Internal tools page save-failure path.

## Non-blocking

**2. `apps/web/src/features/shell/account-menu.tsx:86-87` — `render={<nav />}` / `role="navigation"` land on the positioner, not the menu**

`DropdownMenu.Content` spreads `...props` onto Base UI's `Menu.Positioner` and passes only `className` down to `Menu.Popup`, so these two props decorate the wrapper. I rendered it to confirm:

```html
<nav role="navigation" aria-label="Account" style="position: absolute; ...">
  <div role="menu" aria-labelledby="base-ui-_r_4_" aria-orientation="vertical">…</div>
</nav>
```

Two consequences worth a second look:
- `aria-label="Account"` does **not** name the menu — the menu keeps `aria-labelledby` pointing at the trigger. The label names the wrapper instead.
- A transient `navigation` landmark now wraps a menu whose last item is **Sign out**, which is an action rather than a destination. Landmarks are page structure; adding one inside a popup shows up in landmark navigation for the lifetime of the open menu.

axe passes either way and the menu behaves correctly, so nothing is broken — but the source reads as if the popup becomes a `nav`, and it doesn't. I'd drop both props unless the landmark is deliberate.

**3. `navigation-polish.spec.ts:62-77` — process-wide toggles are only reset on the happy path**

The Internal tools test turns Skills and Integrations on, then off again several assertions later. `internalNavigationService` is process-wide and the Playwright run shares one server across all tests (`workers: 1`), so any failure between the two points leaves both unreleased nav links visible for every subsequent test in the run — one real failure becomes a cascade of unrelated ones. Worth restoring them in a `finally` or fixture teardown.

**4. Icon size divergence between the two menus**

`MenuItemIcon` in `account-menu.tsx:139` went `size-6` → `size-4`, while the identical helper in `agent-shell.tsx:190` stays `size-6`. Both menus now share the same `[role="menuitem"]` padding/height rules in `app.css`, so the icon columns will not line up between them. Probably worth picking one.

**5. `packages/server/src/index.ts:127` — `StagingInternalNavigationVisibilityService` still says "Staging"**

Everything else got renamed off "staging"; this class name and its comment did not. One line. (The `docs/staging-onboarding-reset.md` filename is the same story, but renaming a doc mid-PR is probably not worth the churn.)

**6. Sign out lost `variant="danger"`**

Intentional, I assume, given the rest of the visual pass — but it is the one destructive item in the menu and the description doesn't mention it, so flagging in case it was collateral.

## Scope note

The title is `fix(web)` and the checklist says "contains no unrelated work", but this bundles a navigation visual pass with a new server configuration surface. They are separable and the config half is the half that deserves careful review attention. Not asking for a split now — just noting for the next one.

yuezengwu (approved):
Independently reviewed exact head `5a30a69c0b960d305188e5bd3366e3e0938b4f1b` against `acf7609b9bfdd5fc999ebbc23ef68fc2eb1a7108`. No blocking regression found.

The local Internal tools opt-in defaults off, requires explicit dev plus loopback bind/public URL, and rejects hosted opt-in. Reset eligibility and authenticated Account scope remain enforced; the PostgreSQL tests cover both local reset modes without touching the other Account.

Verified in a clean detached worktree:
- `pnpm check`, `pnpm build`, `pnpm typecheck`.
- Web: 802 tests; server config: 27; Account API/services: 66; PostgreSQL onboarding-reset integration: 21.
- All 5 new navigation browser tests and all 13 `app.spec.ts` journey tests passed, including keyboard focus, dirty-settings/history behavior, 320/390px touch menus, long names, reset cancellation and axe checks. Inspected desktop/mobile/switcher screenshots.

My first filtered browser run omitted the existing serial app fixture creators and left its IDs undefined; running the complete app file in its intended order passed 13/13. Both disposable local stacks were cleaned up. Full unrelated workspace suites were not rerun locally; technical CI on this head is green.

Two non-blocking cleanups already noted in the earlier review are valid: remove the extra space in the Chinese navigation-save error, and restore the process-wide preview toggles in `finally`/fixture teardown so a failed test does not contaminate later tests. The popup wrapper retains an inner menu with working keyboard behavior; I found no functional accessibility blocker.

This approval covers the reviewed code and local validation. It does not establish production or real-provider acceptance.

liuchao-001 (approved):
Approving at `5a30a69c`. No blocking finding stands.

Re-grading my own item: I labelled the `zh.json:27` stray space a "must fix", and on reflection that was one notch too high — it is test-environment-only copy on a page that never reaches production, and @yuezengwu independently reached the same non-blocking read. It is still worth correcting, so along with the e2e teardown item it is the pair I would fold into a follow-up commit if you are still touching this branch, or a small follow-up PR if you are not:

- `apps/web/messages/common/zh.json:27` → `"无法保存导航设置。"`
- `e2e/tests/navigation-polish.spec.ts` → restore the process-wide preview toggles in a `finally`/fixture teardown, so one failed test does not contaminate the rest of the run.

The remaining points in my earlier review (positioner-vs-popup ARIA, icon size divergence, the leftover `Staging` class name, the dropped `variant="danger"`) are observations, not requests — take or leave them.

liuchao-001 (approved):
Re-reviewed at `87fef366`. Approving again — this approval is bound to the current head.

Worth stating plainly: the two approvals already on this PR (mine and @yuezengwu's) are both recorded against `5a30a69c`, and `ownership-gate` is currently satisfied by those. This branch has since taken a non-trivial commit — the popup portal target and positioning strategy changed — so nothing was re-bound automatically. Hence this pass.

Every item from the earlier round is addressed, and two of the fixes are better than what I asked for: the `internal-tools-page` comment no longer claims *every* mutation on that page is reflexive, which was untrue of the Server-wide navigation toggles, and the e2e fixture restores the original preview state rather than a hardcoded off. The new `toHaveAccessibleName("Account menu")` and `getByRole("navigation", { name: "Account" })` assertions lock the semantics fix in place.

I re-rendered the account menu to confirm the new behavior rather than take it from the diff: the popup now lands inside the owning `<header>` (ancestor chain `DIV → DIV#_r_7_ → HEADER`), there is no `navigation` landmark left, and the menu takes its name from the trigger via `aria-labelledby`. Positioner style is `position: fixed; z-index: 50`.

Verified locally at this head: `@opentag/web` 802 tests passed (73 files), `tsc --noEmit` clean, `biome check apps/web packages/server e2e` clean (warnings only).

Two observations on the new approach. Neither blocks.

**1. `container={triggerRef.current?.closest(...)}` reads a ref during render**

`account-menu.tsx:96` and `agent-shell.tsx:132`. On the mount render `triggerRef.current` is still `null`, so this only resolves because opening the menu is state-driven and re-runs the render with the ref populated. That makes the controlled `open`/`setOpen` state newly added to `AgentSwitcher` load-bearing for the portal target, which the comment above it doesn't say — it explains the landmark intent only. Drop back to an uncontrolled `<DropdownMenu>` in a later refactor and `container` silently becomes `undefined`, the popup reverts to `document.body`, and no test fails, because a popup in `body` still works. The property this commit exists to establish would be gone with nothing to catch it.

Kumo's `PortalContainer` already accepts `RefObject<HTMLElement | ShadowRoot | null>`, so this can be a ref rather than a render-time read:

```tsx
const landmarkRef = useRef<HTMLElement | null>(null);
useLayoutEffect(() => {
  landmarkRef.current = triggerRef.current?.closest<HTMLElement>("aside, nav, header, main") ?? null;
}, []);
// …
container={landmarkRef}
```

Base UI reads `.current` when it portals, so the container resolves without depending on a re-render, and the controlled state stops being secretly required.

**2. `position: fixed` now renders inside a transformed ancestor on mobile**

Kumo's mobile sidebar applies `translate-x-0` to the sheet while it is open (`openMobile ? "translate-x-0" : "-translate-x-full"`), and a non-`none` `translate`/`transform` establishes a containing block for fixed-position descendants. So on mobile the popup's `position: fixed` is resolved against the sheet rather than the viewport, and its clipping boundary becomes the sheet.

Floating UI does the coordinate math for `strategy: "fixed"` against a transformed offset parent, and the new 320/390px sheet test walks exactly this path with real taps plus `expectWithinViewport` — a tap carries a hit-target check, so occlusion and mis-positioning would both surface. I raise it only so the reason that test earns its keep is written down somewhere: it is the thing standing between this and a silently clipped menu, so please don't let it get trimmed as redundant with the desktop case.

Also noting the retitle to `fix: polish floating navigation and enable local Internal tools preview` — that reads accurately for what the branch actually does.

PR #600, opened 2026-09-11
feat(server): expose website browser session status

Gandy2025:
## Summary

The public website cannot distinguish a returning signed-in visitor from an anonymous one, so its navigation always says “Log in.” Add a cookie-only session-status read that lets the official website display “Open app” after an authoritative check.

- `GET /api/v1/auth/browser/session-status` returns only an `authenticated` boolean, without Account identity or tokens.
- Register the route only for `https://app.opentag.build`, with credentialed CORS restricted to the exact `https://opentag.build` and `https://www.opentag.build` origins. Other APIs retain their current origin rules.
- Check the live Better Auth session and active Account without refreshing session lifetime or trusting the cookie cache. Bearer credentials are not forwarded.
- Keep unavailable services distinct from anonymous/expired/suspended sessions, and mark all route responses `no-store`.
- Document rollout order in both deployment guides.

This narrowly scoped endpoint serves the official marketing website. Hardcoded production origins deliberately keep staging and self-hosted sessions outside that website's reach; same-origin app clients continue using the existing Account APIs. A configurable deployment-wide session-sharing feature is outside this change.

Companion website PR: https://github.com/first-tree-ai/opentag-site/pull/5

Merge this Server support before the website change. Production deployment is separate from this repository's automatic Staging deployment. The website retains its ordinary login entry until this endpoint is deployed; merging both PRs alone does not establish that production session detection is live.

## Testing

Current head: `2a01c70bbc547270e8e886e85ef7092150a2ffd6`.

- [CI passed](https://github.com/first-tree-ai/opentag/actions/runs/34577456461): formatting/lint, build, type-check, the complete unit suite, PostgreSQL integration, Agent Runtime coverage, all three supported Node versions, browser smoke, CLI pack smoke, and container smoke.
- [Patch coverage passed](https://github.com/first-tree-ai/opentag/actions/runs/34577456476). The new route has 100% statement, branch, function, and line coverage.
- Local focused authentication tests: 64 passed. Agent Runtime coverage: 386 tests passed with 100% coverage.
- Real browser validation used the compiled route and current website on separate loopback origins with synthetic session cookies. Origin strings were mapped in the temporary QA harness. Verified signed-in, signed-out, failure, link targets, and browser Back after logout. This was not a production account test.

Earlier local full runs encountered a Client Context Tree timeout and PostgreSQL integration failures; isolated rechecks passed, and the complete current-head CI unit and integration suites have now passed. The local failures are not represented as successful local runs.

## Breaking changes

None. Self-hosted and staging deployments do not register the website session endpoint.

## Checklist

- [x] Formatting/lint, build, type-check, unit tests, integration tests, and coverage gates pass in current-head CI.


liuchao-001 (approved):
Reviewed at head `2a01c70b`. Approving — the security-sensitive parts hold up under independent checking, and the failure mode if anything is off is a graceful "keeps showing Log in".

## What I verified rather than took on trust

- **The cookie actually arrives in production.** `https://opentag.build` → `https://app.opentag.build` is cross-origin but *same-site* (registrable domain `opentag.build`), and better-auth 1.7.2 defaults the session cookie to `SameSite=Lax` with no partitioning and no `crossSubDomainCookies` configured here, so a credentialed `fetch` does carry it. Worth stating explicitly because the loopback QA harness (`127.0.0.1` vs `localhost`) is *cross-site* — strictly more restrictive than production — so that evidence could not have proven this either way.
- **`disableRefresh` does what the comment claims in this exact version.** In 1.7.2's `/get-session`, the `disableRefresh` branch returns immediately after `findSession` and never reaches `setSessionCookie` or `setCookieCache`, so no lifetime extension and no cookie writes; the route also never asks for `returnHeaders`, so nothing could reach the browser regardless. One behavior worth knowing (not changing): an *already expired* token still has its row deleted inside `getSession`, so the endpoint is not strictly read-only for expired sessions. That is harmless cleanup, but it is a write triggered by a website visit.
- **The error classification is exhaustive, not just plausible.** `#resolveActiveUser` can raise only `AUTH_INVALID_TOKEN` (missing row) and `AUTH_USER_SUSPENDED` — exactly the two codes folded into `authenticated: false`. Every other failure surfaces as 500. So an outage cannot be reported to the website as "signed out", which is the property that matters most here.
- **The reflected `Access-Control-Allow-Origin` cannot be duplicated or widened.** There is no global CORS plugin anywhere in the server, so this route is the only emitter, and `publicOrigin` reaches it already normalized through `PublicUrlSchema` (`url.origin`), so the exact-string registration gate cannot be defeated by a trailing slash or path.
- **Bearer isolation is load-bearing, not decorative.** The better-auth `bearer` plugin promotes `Authorization` into a session cookie, so constructing a fresh `Headers` with only `cookie` is precisely what stops this from becoming a second bearer-authentication surface. Good that a test pins it.
- **No route-surface conflict.** `/api/v1/auth` is published as explicit routes, not a catch-all, so `browser/session-status` cannot shadow or be shadowed by the library surface.
- **CI is green at this head**, including the `Checks` job (unit + PostgreSQL integration) and patch coverage. The local `pnpm test` / integration failures recorded in the description read as environment artifacts; the gate that counts has passed.

## Non-blocking, in priority order

1. **The hardcoded `https://app.opentag.build` gate means no environment this repository deploys can ever execute this code.** `docs/deploying.md` states production is not deployed from here, and the test file itself treats `https://staging.opentag.build` as a deployment that must *not* register the route. So the rollout ordering the new documentation describes can never be rehearsed, and the only browser-level evidence came from a harness where these constants were replaced — the code that was exercised in a browser is not the code that merges. Consider a config-driven allowlist instead (default empty = feature off, validated as HTTPS and same registrable domain as `OPENTAG_PUBLIC_URL`): identical posture for anyone who does not set it, plus a staging rehearsal and self-host parity for an open-source server. If you would rather keep the decision closed in code, that is defensible — but then say *why* in the route comment, so the next person does not "fix" it into config without understanding the tradeoff.
2. **Know the polling cost before the marketing site scales.** The companion website rechecks once per minute per visible tab plus on focus, `pageshow`, and reconnect; each check carrying a valid cookie is two database round trips, and this route has no limiter while every other unauthenticated browser-auth entry point in `browser-auth.ts` does. It is the same cost profile as `/api/v1/me`, and a forged cookie fails the signature check before any query, so this is not a new class of exposure — just a load characteristic that is easy to forget once tabs stay open.
3. **Same-origin callers get 403.** A request without an `Origin` header — which is what the app's own web client would send — is rejected, so this path is permanently unusable from `app.opentag.build` itself. Intended, and I would keep it, but one line in the route comment saves someone a confusing debugging session later.

Nothing above needs to land before merge.

yuezengwu (approved):
Reviewed head `2a01c70bbc547270e8e886e85ef7092150a2ffd6` independently in a clean detached worktree. No blocking finding.

The route is registered only for the official app origin, allows only the two explicit website origins, forwards cookies without bearer credentials, and returns only session presence. Active-Account checks preserve the distinction between anonymous/suspended sessions and service failures. Responses retain `no-store`, and this route does not widen the other API routes' CORS permissions.

I also checked the installed Better Auth 1.7.2 implementation: `disableCookieCache` bypasses cached session acceptance, and `disableRefresh` returns before valid-session renewal. Expired-session cleanup can still delete its database row; the route does not forward Better Auth response cookies.

Fresh local validation: all 64 website-session/auth API/browser-auth/Better Auth surface tests passed, as did `pnpm check`, Shared build, and Shared/Server typechecks. The head merges cleanly with current main `947c79fc6eded43b6db8d328d06f163b2632ae63`.

Limits: the route tests use the repository's mocked Better Auth fixture; I did not run a real-account browser acceptance, the full monorepo suite, or PostgreSQL integration locally. CI and the deployment sequence remain separate from those local results; production rollout should install Server support before the website indicator.

PR #617, opened 2026-09-14
fix(context-tree): share the account Context Tree home

ggordonhall:
Moves OpenTag’s selected Context Tree target to the standalone CLI’s account-scoped `~/.context-tree/opentag.json`, while retaining shims and preparation diagnostics under `OPENTAG_HOME`. Sessions now receive the shared directory as a writable root and use the same account environment for Context Tree subprocesses.\n\nValidation: `pnpm install --frozen-lockfile`; `pnpm check`; `pnpm build`; `pnpm typecheck`; targeted Context Tree and CLI suites; `pnpm --filter @opentag/client test:agent-runtime:coverage --testTimeout=30000`; `pnpm --filter @opentag/server test:integration`. The full `pnpm test -- --testTimeout=30000` run was interrupted after shared and server suites passed, before client and web completion.


liuchao-001 (changes requested):
Reviewed at head `ffed6061`, in a fresh worktree off `origin/main` (`c14da22c`). Ran locally: `pnpm install --frozen-lockfile`, `pnpm --filter @opentag/shared build`, `pnpm --filter @opentag/client build`, `pnpm --filter @opentag/client test` (87 files / 1268 tests pass), `pnpm --filter open-tag test --testTimeout=30000` (43 files / 520 tests pass), `pnpm check` (exit 0). The new `apps/cli` integration case really does drive the standalone CLI across two `OPENTAG_HOME`s, and it left my own `~/.context-tree` untouched — test isolation via stubbed `HOME` holds.

The writable-root half of this is clearly right: the Context Tree skills run the CLI *inside* the sandbox, and the CLI writes `~/.context-tree`, so a grant scoped to `<OPENTAG_HOME>/config/context-tree` could never work. Threading `env` end-to-end (connect → state → `ContextTreeManager` → subprocess) is consistent, the symlink/non-directory rejection matches `ensureManagedDirectory` in the standalone CLI, and the `chmod` after `writeFile` is a genuine fix — `writeFile`'s `mode` only applies on creation, so a pre-existing loose-mode config was never tightened.

Two things I'd like resolved before merge, plus nits.

---

## 1. The selection file move makes the target per OS account, which collides with channels (blocking)

`docs/design/context-tree-integration.md:245` states the new rule plainly: "Changing `OPENTAG_HOME` does not change the selected default." But `packages/shared/src/channel-config.json` ships three channels that are *designed* to coexist on one machine — `dev` → `~/.opentag-dev` (server `127.0.0.1:8000`), `staging` → `~/.opentag-staging`, `prod` → `~/.opentag` — each with its own home, its own daemon `serviceId`, its own server, and its own Computer identity and machine credential under `<OPENTAG_HOME>/config`. After this change all three read and write one `~/.context-tree/opentag.json`, so the last `opentag context-tree connect` wins across channels: pointing a dev Computer at a scratch tree silently repoints the prod Computer at it too, and `opentag doctor` reports the same `configPath` on all three with no way to tell them apart.

That also runs against the durable decision in `opentag/system/context-tree-durable-memory.md`: "A person names an existing tree once per Computer; from then on every Agent Workspace **on that Computer** is connected to it." Per-Computer choice is no longer expressible on a single account.

Worth separating the two changes in this PR: the permission bug is fixed entirely by granting `~/.context-tree` as a writable root. Nothing inside a Session reads `opentag.json` — only `ContextTreeManager.readConfig` and the CLI do, both daemon/CLI-side — so relocating the selection file isn't required for the fix. My preference is to keep the selection under `OPENTAG_HOME` and land the grant change alone. If account scope is genuinely wanted, then it needs to be a recorded decision rather than a side effect, and the filename should be channel-namespaced (`opentag-<channel>.json`) so the three homes don't overwrite each other.

## 2. The grant's blast radius is wider than the doc admits (blocking on the doc, at least)

The removed paragraph argued minimality explicitly — one leaf directory, exactly one file, "carries no more authority on Linux than the macOS file grant." The replacement at `docs/design/context-tree-integration.md:242-247` describes the new grant as "the shared trees, connections, and OpenTag default selection." That list is incomplete. `~/.context-tree` in `@first-tree-ai/context-tree@0.1.12` also holds `cleanup/launchers/<schedule-id>/context-tree-cleanup` — per its README, "macOS uses user LaunchAgents, each invoking an executable named `context-tree-cleanup` at `~/.context-tree/cleanup/launchers/<schedule-id>/`", with systemd user timers on Linux. So any visible or internal Session can now overwrite an executable that the user's own scheduler later runs **outside** the provider sandbox. Registering a *new* schedule stays out of reach (the LaunchAgent plist lives in `~/Library/LaunchAgents`), but tampering with an existing one does not.

Alongside that, every Session gets write access to every tree under `trees/`, including trees belonging to other Agents on other Computers, and to `connections.json`, which maps every project on the account to a tree. The current durable constraint is narrower: "The resolved tree root is granted explicitly, including its Git directory."

I don't think there's a narrower grant that still lets the in-session CLI work — `writableRoots` can't express exclusions — so this is likely the right trade. But it should be argued, not dropped: name the launcher executable and the cross-Agent tree/connection reach in the doc, and say why the grant is still acceptable. If the decision is that Sessions get account-wide Context Tree write authority, that belongs in the Context Tree node too, which today says the opposite.

## 3. Upgrade guidance was deleted rather than updated

`<OPENTAG_HOME>/config/context-tree/config.json` has shipped since #455, which is in 0.0.3, 0.0.4, and 0.0.5. Every installation that ran `opentag context-tree connect` will silently read as unconfigured after this upgrade, with a stale config file and directory left behind. The diff removes the "**Upgrade requires reconnecting**" paragraph (which covered the *previous* relocation) instead of rewriting it for this one, so there is now no instruction anywhere. Please restore it, pointed at the new path.

## Nits (non-blocking)

- `packages/client/src/runtime/context-tree.ts:221` — `join(dirname(resolveContextTreeHome(this.#environment).directory), ".codex")` is an obscure spelling of "the account home", and it quietly changes the default from `os.homedir()` to a realpath'd, env-derived home. A small `resolveAccountHome(env)` that both `resolveContextTreeHome` and this call use would read straight and keep one definition of the account home.
- `apps/cli/src/core/context-tree/shared.ts:46,59` — plain `readFile`/`writeFile` follow symlinks, while the repo already has `readSecureFile` / `writeDurableFile` (`O_NOFOLLOW` temp + `O_EXCL` + atomic rename + `assertReplaceableFile` + `chmod`), and `ContextTreeManager.readConfig` reads this exact file through `readDurableJson`. Pre-existing, but it matters more now: after this PR a Session can plant `~/.context-tree/opentag.json` as a symlink, and the user's own unsandboxed `opentag context-tree connect` would write through it. Reusing the existing helpers closes it for ~two lines, and it also matches the standalone CLI, which `lstat`-checks `connections.json` before reading.
- `packages/client/src/runtime/session-runtime-manager.ts:311-312` — `writableRoots` now carries both `~/.context-tree` and the resolved tree path, and for a managed target the latter is nested inside the former. Harmless, just redundant.
- `apps/cli/src/__tests__/context-tree.test.ts:63` — `configFile = (_home: string) => ...` ignores its parameter while every call site still passes `home`. Dropping the parameter would stop the call sites implying a per-home path that no longer exists.

packages/client/src/storage/context-tree-home.ts
```diff
@@ -0,0 +1,33 @@
...
+    // Like the standalone CLI, retain the account path when it cannot be canonicalized.
+  }
+  const directory = join(home, ".context-tree");
+  return { directory, configFile: join(directory, "opentag.json") };
+}
+
+export async function prepareContextTreeHome(environment: NodeJS.ProcessEnv = process.env): Promise<string> {
+  const { directory } = resolveContextTreeHome(environment);
+  try {
+    await mkdir(directory, { mode: 0o700 });
```
bestony: **[suggestion]** `prepareContextTreeHome` creates `~/.context-tree` with `mkdir(..., { mode: 0o700 })` but never `chmod`s the directory after it exists. `ensureOneDirectory` in `durable-file.ts` always follows mkdir with `chmod(path, 0o700)` so a pre-existing 0755/0777 directory is tightened. Here an already-created standalone CLI home (or a user-made `~/.context-tree`) keeps its looser mode, then SessionRuntimeManager grants that directory as a writable root. The design claims mode `0700`.

**Suggestion:** After the real-directory check, `chmod(directory, 0o700)` the same way `ensureOneDirectory` does, and add a test that a pre-existing 0755 managed directory is tightened before being returned.

docs/design/context-tree-integration.md
```diff
@@ -74,19 +74,11 @@ There is deliberately no `inspect` subcommand. `opentag doctor` already owns dia
 the injectable-inspector seam, and two surfaces over one piece of state means every future reason
 code has to be rendered twice.
 
-The target is recorded machine-locally in `<OPENTAG_HOME>/config/context-tree/config.json`, mode
+The target is recorded per OS account in `~/.context-tree/opentag.json`, mode
 `0600`, credential-free. The Server is not involved. The three target kinds mirror
```
bestony, line 78: **[suggestion]** The selected target now lives only at `~/.context-tree/opentag.json`. Existing `<OPENTAG_HOME>/config/context-tree/config.json` files are ignored, so already-connected Computers become `unconfigured` until someone reconnects. The previous design called that out as a required reconnect; this revision deletes that note, and the tests that described the old-path upgrade were removed rather than replaced with a migration or a documented reconnect case.

**Suggestion:** Either migrate when the new file is absent and the old OpenTag-home config is valid, or restore an explicit upgrade note (`opentag context-tree connect …`, restart daemon/Sessions, `opentag doctor`) so operators are not surprised by silent durable-memory loss. The leftover "as described in the upgrade instructions above" sentence later in this doc now points at nothing.

bestony (commented):
## Summary

The PR correctly relocates the selected Context Tree target from `OPENTAG_HOME` to the account-scoped `~/.context-tree/opentag.json`, threads that account environment into Context Tree subprocesses, and grants Sessions the shared managed directory after rejecting a symlink or non-directory. Shim and preparation diagnostics stay under `OPENTAG_HOME`, and the split is covered by unit and cross-home integration tests. The remaining gaps are permission enforcement on an already-existing managed directory, and a silent break for installs that still have the previous OpenTag-home config.

CI (Checks, Node 22/24/26, patch coverage) is green. This is a COMMENT review, not an approval: existing Computers that already connected under `<OPENTAG_HOME>/config/context-tree/config.json` will look unconfigured until someone reconnects, and that upgrade path is no longer documented.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

## Also noted

- `docs/design/context-tree-integration.md` still says rollout requires reconnecting "as described in the upgrade instructions above", but those instructions were deleted in this PR. Restore them next to the new path, or drop the leftover sentence.
- `opentag context-tree connect` calls `writeContextTreeConfig` → `prepareContextTreeHome` with no catch. Session start swallows a symlink/non-directory and continues without the grant; the CLI rethrows a generic `Error` instead of exiting 1 with a message (`apps/cli/src/commands/context-tree.ts` only maps `ContextTreeUsageError` to exit 2).

liuchao-001:
Following up on @bestony's review — I verified both suggestions against head `ffed6061` and reproduced them, and one claim in the summary needs correcting so the fix lands in the right place.

**The 0755 case is real, with a reproduction.** Pre-existing directory keeps its mode, and that directory is what `SessionRuntimeManager` then hands Sessions as a writable root:

```
$ mkdir -m 0755 "$T/.context-tree"
before: 755
$ HOME=$T OPENTAG_HOME=$T/.opentag opentag context-tree connect --tree-path <tree>
Recorded in /private/tmp/.../.context-tree/opentag.json
EXIT=0
after dir: 755   config: 600
```

The selection file itself is correctly `0600`, so the target is not world-readable; what leaks is directory traversal and listing of `~/.context-tree` — tree checkouts under `trees/` are created with ordinary git permissions. Worth stating precisely, but the doc's "OpenTag creates `~/.context-tree` with mode `0700`" is still not true for this path.

**The cleanest fix is reuse, not an added `chmod`.** `prepareContextTreeHome` is `ensureOneDirectory` from `packages/client/src/storage/durable-file.ts:139-147` with the `chmod` dropped — same `mkdir(mode 0o700)`, same EEXIST catch, same real-directory assertion. And `assertRealDirectory` (`durable-file.ts:124`) is already exported from `@opentag/client` (`index.ts:511`) and does the identical `lstat` check that `context-tree-home.ts` reimplements. Exporting `ensureOneDirectory`, or building `prepareContextTreeHome` on top of `assertRealDirectory`, closes the permission gap and removes the duplicate symlink check in one move — and keeps the two from drifting the next time one is hardened.

**Correction on the CLI error path.** The summary says the CLI "rethrows a generic `Error` instead of exiting 1 with a message (`apps/cli/src/commands/context-tree.ts` only maps `ContextTreeUsageError` to exit 2)." That subcommand handler does only map `ContextTreeUsageError`, but the entrypoint at `apps/cli/src/cli/index.ts:36-44` catches everything else through `toCommandError` / `presentCommand`. Observed behaviour with `~/.context-tree` as a symlink:

```
$ ln -s "$T/real" "$T/.context-tree"
$ HOME=$T OPENTAG_HOME=$T/.opentag opentag context-tree connect --tree-path <tree>
INTERNAL_ERROR: Context Tree managed directory must be a real directory: /private/tmp/.../.context-tree
EXIT=1
```

So it does exit 1 with a message. The genuine defect is narrower: this is classified `INTERNAL_ERROR`, which tells an operator OpenTag is broken when the actual fault is one they own and can fix in one command — `~/.context-tree` is a symlink, replace it with a real directory. A typed error carrying that remedy (the way `ContextTreeUsageError` does for usage faults) is the fix; wrapping the call site in a `catch` would not be.

**Agreed on the dangling reference** — `docs/design/context-tree-integration.md:194-195` still reads "Rollout requires reconnecting existing installations and restarting the daemon and affected Sessions as described in the upgrade instructions above," and those instructions are deleted in this diff. That is the same gap as point 3 in my review, and I missed the leftover sentence; it makes restoring the note rather than dropping it the clearly better of @bestony's two options.

My two blocking points from the earlier review still stand and are untouched by this round: the per-account selection colliding with the dev/staging/prod channels, and the widened grant reaching `cleanup/launchers/<id>/context-tree-cleanup`, which the user's own scheduler runs outside the sandbox.

bestony:
Agree with both blocking items on `ffed6061`. I missed the channel collision and the grant blast-radius write-up; they should land before merge. I am not approving.

1. Selection file vs channels — confirmed. `packages/shared/src/channel-config.json` gives three coexisting homes (`~/.opentag-dev`, `~/.opentag-staging`, `~/.opentag`). `resolveContextTreeHome` ignores `OPENTAG_HOME` and always returns `~/.context-tree/opentag.json`, so the last `opentag context-tree connect` wins across channels. The same design doc still says “The target is machine-local, so it is set once per Computer” (`docs/design/context-tree-integration.md:158`) while also saying changing `OPENTAG_HOME` does not change the selected default. Those cannot both be true on one account.

   Relocating the selection file is not required for the permission fix. Session start never reads `opentag.json`; `ContextTreeManager.readConfig` and the CLI do, both outside the sandbox. The in-session CLI needs a writable `~/.context-tree` because it writes trees/connections there. Keep the selection under `OPENTAG_HOME` and land the grant alone. If account scope is actually wanted, namespace the file per channel.

2. Grant blast radius — confirmed as a doc gap at least. `prepareConfigurationRoots` returns the whole managed directory, and the replacement paragraph lists trees, connections, and the OpenTag default. It does not name launcher executables under `cleanup/launchers/` or write reach into every tree and `connections.json`. If `writableRoots` cannot exclude those, the trade should be stated, including why a Session may rewrite an executable the user’s own scheduler later runs outside the sandbox.

3. Upgrade guidance — same finding as my COMMENT: the reconnect paragraph was deleted rather than retargeted, while `<OPENTAG_HOME>/config/context-tree/config.json` has already shipped.

Nits on `resolveAccountHome`, `readSecureFile`/`writeDurableFile` for the selection file, nested writable roots, and the unused `configFile(_home)` parameter all look right. The symlink-following CLI write is sharper after this grant: a Session can plant `opentag.json` as a symlink and the unsandboxed CLI will write through it.

bestony:
@liuchao-001 You are right on the CLI path, and I am retracting that part of the summary.

`apps/cli/src/commands/context-tree.ts` only maps `ContextTreeUsageError`, but `apps/cli/src/cli/index.ts:36-44` runs everything else through `toCommandError` / `presentCommand`. A generic `Error` from `prepareContextTreeHome` becomes `INTERNAL_ERROR` with the original message and exit 1. Wrapping the connect call site in `catch` would not be the fix. The remaining defect is the classification: this is an operator-owned, one-command repair (replace the symlink with a real directory), not an OpenTag internal failure. A typed error that carries that remedy is the right shape.

Agreed that `prepareContextTreeHome` should reuse `ensureOneDirectory` (or at least `assertRealDirectory` plus the existing `chmod`) rather than grow a second copy of the mkdir/EEXIST/lstat sequence. `assertRealDirectory` is already a public export; `ensureOneDirectory` is the missing chmod. Your 0755 reproduction matches the code: the selection file is `0600`, the directory stays `0755`, and that is the writable root Sessions receive.

Also agreed that restoring the reconnect paragraph is better than dropping the leftover “as described in the upgrade instructions above” sentence.

Still not approving. The two blocking items from your earlier review are unchanged: per-account `opentag.json` colliding with the three channel homes, and the grant reaching `cleanup/launchers/<id>/context-tree-cleanup`.

ggordonhall:
Addressed in `8cf43410`. Thanks for the detailed review — the verification work and the 0755 reproduction made this straightforward to land.

**1. Account-scoped selection — keeping it, now as a recorded decision.**
All OpenTag channels on one OS account share the standalone CLI's account home, and that home already holds `trees/` and `connections.json` account-wide. A shared selection is consistent with the CLI OpenTag delegates to, so this is deliberate rather than a side effect. I've stated it explicitly in `docs/design/context-tree-integration.md` and replaced the now-false "machine-local … once per Computer" line with "account-local … shared by every OpenTag channel (`dev`, `staging`, `prod`)". If we ever want per-channel targets, the file should be channel-namespaced (`opentag-<channel>.json`); that is not the intent now.

**2. Grant blast radius — documented, not narrowed.**
`writableRoots` cannot express exclusions, so the trade is now argued in the doc instead of dropped. It names every checkout under `trees/`, `connections.json` (every project on the account), and the launcher executable at `cleanup/launchers/<schedule-id>/context-tree-cleanup` that the user's own LaunchAgent/systemd timer runs **outside** the sandbox — including that a Session can rewrite an existing launcher but cannot register a new schedule, because the plist/unit lives outside the directory.

**3. Upgrade guidance — restored and retargeted.**
The paragraph now points at `<OPENTAG_HOME>/config/context-tree/config.json`, which has shipped since #455. The leftover "as described in the upgrade instructions above" reference now resolves.

**Nits + permission enforcement**
- `prepareContextTreeHome` reuses `ensureOneDirectory` (`durable-file.ts`), so a pre-existing `0755` directory is tightened to `0700`; a new test covers exactly that case. The duplicated `mkdir`/`EEXIST`/`lstat` sequence is gone. The specific message is preserved as a `RuntimeStorageError` (`unsafe`).
- The selection file now goes through `readDurableJson`/`writeDurableJson`, so a Session-planted symlink can no longer be followed by the unsandboxed `opentag context-tree connect`.
- `resolveAccountHome` defines the account home once, shared by `resolveContextTreeHome` and the default Codex home.
- Nested `writableRoots` are deduplicated (new test with a managed tree under `~/.context-tree/trees`), and the unused `configFile(_home)` parameter is gone.

**Validation:** `pnpm check` exits 0. `@opentag/client` (`context-tree-home`, `session-runtime-manager`) and `open-tag` (`context-tree`, `doctor`, 69 tests) pass. On this machine two `daemon ensure-service` CLI tests and one `pi-agent-runtime-exhaustive` test also fail/flake at `ffed6061`, so they are pre-existing and unrelated.

Still open from @bestony (not expanded here): a symlinked `~/.context-tree` now surfaces as a `RuntimeStorageError` rather than a raw `Error`, but it is not yet a dedicated operator-repair error. Happy to add that typed classification if you want it in this PR.

docs/design/context-tree-integration.md
```diff
@@ -74,14 +74,15 @@ There is deliberately no `inspect` subcommand. `opentag doctor` already owns dia
...
 `context-tree connect`'s own argument shape, so OpenTag passes the target through rather than
-reinterpreting it.
+reinterpreting it. The file does not depend on `OPENTAG_HOME`, so every OpenTag channel on the
+account (`dev`, `staging`, `prod`) reads and writes the same selection.
```
bestony, line 81: **[suggestion]** Recording the channel collision is better than leaving it implicit, but it does not resolve the isolation problem. `dev` / `staging` / `prod` still share this one file, so the last `connect` retargets every Computer on the account. The grant/permission fix does not need that. Keep the selection under `OPENTAG_HOME`, or namespace the file per channel.

bestony (commented):
## Re-review at `8cf43410`

Most of the last round is fixed: `prepareContextTreeHome` now goes through `ensureOneDirectory` (0700 on an existing dir, with a test), the selection file uses `readDurableJson` / `writeDurableJson`, Codex home shares `resolveAccountHome`, nested managed-tree writable roots are deduped, upgrade reconnect guidance is restored, and the grant paragraph now names `trees/`, `connections.json`, and `cleanup/launchers/<id>/context-tree-cleanup`. The leftover “as described in the upgrade instructions above” sentence now has a target.

Still not an approval. The remaining blocker is the one that was not a docs gap: selection is still one `~/.context-tree/opentag.json` for every OpenTag channel on the account. This commit records that as intended rather than isolating it. Last `opentag context-tree connect` still wins across `dev` / `staging` / `prod`. Relocating the file was not required for the grant/permission fix. Keep the selection under `OPENTAG_HOME`, or namespace it (`opentag-<channel>.json`). Documenting the collision does not make a staging connect safe for prod.

Partial, not blocking: a symlink/non-directory still surfaces as `RuntimeStorageError` code `unsafe`, which `toCommandError` classifies as category `internal`. The message is now usable; the code still tells an operator OpenTag is broken.

liuchao-001:
Re-verified at `8cf43410`. Everything I raised as a concrete defect is fixed, and I reproduced each one rather than reading the diff for it.

**Verification run:** `pnpm check` exit 0; `@opentag/client` 87 files / 1270 tests pass; `open-tag` 43 files / 520 tests pass. On the flakes you mentioned — I see 520/520 on the CLI suite at both `ffed6061` and `8cf43410`, including `daemon ensure-service` and `pi-agent-runtime-exhaustive`, so those look local to your machine rather than pre-existing in the branch.

**Reproductions, before → after:**

```
A. pre-existing 0755 managed directory
   ffed6061: before 755 → after dir 755, config 600
   8cf43410: before 755 → after dir 700, config 600          ✓ tightened

B. symlinked ~/.context-tree
   ffed6061: INTERNAL_ERROR: ...must be a real directory   EXIT=1
   8cf43410: unsafe: ...must be a real directory           EXIT=1   ✓ code no longer says "internal"

C. Session-planted symlink at ~/.context-tree/opentag.json → ~/victim.txt
   8cf43410: unsafe: Runtime storage destination must be a regular file   EXIT=1
             victim contents: ORIGINAL-SECRET (untouched)    ✓ write-through closed
```

C was the one that sharpened under the widened grant, and `writeDurableJson` closes it exactly.

## One new finding: two exported `resolveAccountHome` functions

Extracting `resolveAccountHome` fixed the duplication I pointed at, but it collides with a function of the same name that already exists and is already public:

- `packages/client/src/runtime/provider-cli/account-layout.ts:42` — `resolveAccountHome(account: { homedir: string } = userInfo())`, exported from `packages/client/src/index.ts:247` and used by `apps/cli/src/core/provider-cli/shared.ts`.
- `packages/client/src/storage/context-tree-home.ts:7` — `resolveAccountHome(environment: NodeJS.ProcessEnv = process.env)`, new.

They answer the same question and disagree. `userInfo().homedir` reads the passwd entry and ignores `$HOME`; the new one prefers `$HOME`/`USERPROFILE` and canonicalizes with `realpath`. Any environment where those differ — `e2e/tests/fixtures.ts:147` sets `HOME` to a temp dir, plus sudo, containers, and the daemon under launchd/systemd — gets two different account homes from one name, and a call site gives the reader no way to tell which. Nothing is broken today because the import paths differ, which is also why this will stay invisible until it bites.

Both semantics are individually right: Context Tree has to match the standalone CLI's `realpathSync(homedir())`, and provider-cli's injectable account record is deliberately different. So keep two functions, just not two of this name — the new one is internal, so renaming it (`resolveContextTreeAccountHome`) is free and non-breaking, while the existing one is the public export.

## On the account-scoped selection

I have made this argument once and you have answered it with an explicit, documented decision, so I am not going to re-litigate the call itself. Two things are worth putting on the record.

Your rationale does not quite carry, on one specific point: `connections.json` being account-wide did not already make the selection shared. It is keyed by project path, and Agent Workspaces live under `<OPENTAG_HOME>/data/`, so `dev` and `prod` workspaces are distinct paths with independent connection records. Per-channel targets were genuinely expressible before this PR; the selection file is what removes that, not something the CLI had already decided.

Second — and this is why it is not purely a PR-local call — `opentag/system/context-tree-durable-memory.md` still reads "A person names an existing tree once per Computer; from then on every Agent Workspace **on that Computer** is connected to it." `docs/design/context-tree-integration.md:167` now says the opposite. One of the two has to move. If the account scope stands, that Context Tree node needs updating as part of landing this, and I am happy to take that write once this merges. If it does not stand, `opentag-<channel>.json` is the cheap version you already named.

My preference is still the namespaced file, because the hazard lands on whoever runs two channels — which is mostly us — and a stale target is only obvious if someone thinks to run `opentag doctor`. But it is your call to make, and it is now a recorded one rather than a silent one, which was the part that mattered most.

@bestony's remaining point about the `unsafe` code landing in category `internal` is accurate at the envelope level, and lower impact than it looks: `context-tree connect` does not accept `--json` (it exits 2 with `USAGE_ERROR`), so the category is not reachable from this command today. The string an operator actually sees is now `unsafe: Context Tree managed directory must be a real directory: <path>`.

bestony (approved):
Approved at exact head `8cf43410`.

The follow-up records the remaining product call instead of leaving it as a side effect: one OS account, one standalone Context Tree home, one OpenTag selection, shared across `dev` / `staging` / `prod`. That matches the CLI OpenTag already delegates to (`trees/`, `connections.json`). Per-channel files are deferred, not implied. Grant reach and the reconnect upgrade path are now stated.

Nits from the earlier round are in: `ensureOneDirectory` (0700 on an existing dir), durable JSON for the selection file, shared `resolveAccountHome`, nested writable-root dedup.

The symlink/`RuntimeStorageError` classification can stay out of this PR. `unsafe` plus the path in the message is enough to repair; a dedicated operator-repair code is a follow-up, not a merge gate.

bestony:
@ggordonhall Accepting the recorded decision on account-scoped selection. The standalone CLI's home is already account-wide, so one OpenTag default there is consistent; per-channel `opentag-<channel>.json` is a later change if we ever want it.

Do not add the typed operator-repair error in this PR. `RuntimeStorageError` `unsafe` with the path is enough to find and replace a symlink; a dedicated code can wait.

Approving `8cf43410`.

liuchao-001 (approved):
Approving at exact head `8cf43410`, clearing my earlier `CHANGES_REQUESTED`.

Both of my blocking items are resolved on the terms they were raised. The grant blast radius is now argued in the doc rather than dropped — it names `trees/`, `connections.json`, and the `cleanup/launchers/<id>/context-tree-cleanup` executable that runs outside the sandbox, says a Session can rewrite an existing launcher but cannot register a new schedule, and states the deployment consequence. That was the whole ask; I never thought a narrower grant was available.

On account-scoped selection: my objection was that it was an undocumented side effect contradicting the doc itself and a durable decision. It is now an explicit recorded decision with the self-contradiction fixed, and @bestony has accepted it. I still mildly prefer the namespaced file, and the record now shows why, which is the part that survives. Your call, made in the open — that is the outcome I wanted.

Everything I verified at this head stands: 0755 tightened to 0700, the symlinked managed directory rejected with a usable message, and the planted-symlink write-through closed with the victim file untouched. `pnpm check` exit 0, client 1270 tests, CLI 520 tests.

Not merge gates, listed so they are not lost:

1. **`resolveAccountHome` name collision** — two exported functions of that name with different semantics (`userInfo().homedir` vs `$HOME`+`realpath`). It is invisible today because the import paths differ, which is exactly why it will surface at an inconvenient time. Cheapest to rename the new internal one while this code is fresh; a follow-up is fine if you would rather land this as-is.
2. **Context Tree node** — `opentag/system/context-tree-durable-memory.md` still says "once per Computer" and now disagrees with the shipped behaviour. I will take that write once this merges, so it does not fall to you.

Agreed with @bestony that the typed operator-repair error stays out of this PR.

bestony:
@liuchao-001 Agreed on both leftover items, and neither changes the Approve at `8cf43410`.

The two `resolveAccountHome` functions do answer the same question with different sources (`userInfo().homedir` vs `$HOME`/`USERPROFILE` + `realpath`). The new one is internal, so `resolveContextTreeAccountHome` is the free rename. Fine as a follow-up; not a merge gate.

Also agreed that `connections.json` being keyed by project path meant per-channel OpenTag targets were expressible before this PR — that was the part of the rationale that did not carry. The recorded decision still stands.

If you take the `opentag/system/context-tree-durable-memory.md` write after merge, that is the right place. I will not hold this PR for it.

The `unsafe` / `internal` envelope point is accurately lower impact given `context-tree connect` has no `--json`.

liuchao-001 (approved):
Re-approving at exact head `ae4b7b3a`. My previous approval was bound to `8cf43410`, so it no longer carried.

The new commit replaces the general N×N `deduplicateWritableRoots` with `dropCoveredRoots(contextTree.writableRoots, configurationRoots)`. I checked it rather than waving it through, since it touches what Sessions are granted:

- The trailing `sep` in the prefix test avoids the classic sibling bug — `/a/bc` is not treated as covered by `/a/b`.
- Narrowing it to Context Tree roots against the configuration grant is the right scope. That nested managed checkout was the only real case; my original note was just that the two overlapped.
- Both directions stay covered by tests: an external `--tree-path` tree is still granted explicitly, and a managed checkout under `~/.context-tree/trees/` is dropped in favour of the parent grant while the prompt still reports the tree path.
- The `configFailure` path composes correctly — when directory preparation fails, `configurationRoots` is empty, so a managed tree is still granted directly rather than silently losing its grant along with the parent.

One theoretical difference: an exact match between a tree root and the account directory is no longer dropped, because `startsWith(parent + sep)` is false for equality. It is unreachable — `~/.context-tree` is not itself a valid tree root and would fail `verify` — and the worst case is a duplicate entry, which was cosmetic to begin with. Not worth code.

Verification at this head: `pnpm check` exit 0; `@opentag/client` 87 files / 1270 tests; `open-tag` 43 files / 520 tests. All three reproductions re-run and still hold — 0755 tightened to 0700, symlinked managed directory rejected, planted `opentag.json` symlink refused with the victim file untouched.

The two follow-ups are unchanged and neither gates this: the `resolveAccountHome` rename, and the `opentag/system/context-tree-durable-memory.md` "once per Computer" update, which I will take once this merges.

yuezengwu (commented):
Independently validated exact head `ae4b7b3a65e635383a44f85def7e1c300514d935` in a clean detached worktree. `pnpm check`, shared/client builds, client tests (1,270), CLI tests (520), Agent Runtime coverage (388 tests and its coverage gate), and client/CLI typechecks all pass. Current CI is green; the previous-head coverage failure no longer applies.

One permission decision remains open on my side before approval. In `SessionRuntimeManager`, `prepareConfigurationRoots(environment)` runs unconditionally, independently of whether a Context Tree manager/default is available. The current tests explicitly retain the whole `~/.context-tree` grant for an internal Session without a manager and for an unavailable tree. Therefore leaving Context Tree unconfigured does not implement the documentation's advice to avoid enabling it when account-wide write authority is unacceptable.

The documented account-wide scope also shares the selected target across dev/staging/prod and permits rewriting an existing cleanup launcher that a host scheduler later executes outside the provider sandbox. The other reviews have discussed and accepted that tradeoff; I still need an explicit decision on accepting this scope, including the lack of an effective opt-out for unconfigured Sessions. If the documented opt-out is intended to be a supported alternative, please identify its concrete configuration and test that it suppresses the account-directory grant.

Validation here establishes the configured grant and its tested behavior. I did not start a host scheduler, modify a real cleanup launcher, or exercise a sandbox escape.

liuchao-001 (changes requested):
@yuezengwu is right, and this withdraws my approval at `ae4b7b3a`. I verified the claim independently rather than deferring to it.

When I released my block on the grant blast radius, the thing I accepted was the doc stating the trade *and its deployment consequence*: "Deployments that cannot grant account-wide Context Tree write authority should not enable Context Tree." That sentence is the entire mitigation. I did not check whether the opt-out it names exists. It does not.

`prepareConfigurationRoots(this.#environment)` at `session-runtime-manager.ts:281` runs unconditionally. The line above it, `prepareContextTree`, is properly guarded — no manager or no status yields no roots — and then the grant is taken anyway, consulting neither `this.#contextTree` nor whether a target is configured. The branch's own tests assert the resulting behaviour:

- `starts internal Sessions with configuration directory failure=false` constructs `SessionRuntimeManager` with no `contextTree` at all, then asserts `writableRoots` is `[cwd, ~/.context-tree]` and that the directory `isDirectory()` is `true`. A Computer that never touches Context Tree still has `~/.context-tree` created and handed to every Session as a writable root.
- The unavailable-tree case asserts `[cwd, ~/.context-tree]` with `Context Tree unavailable (DIRTY_TREE)` in the prompt. Tree unusable, grant still issued.

In production `client-runtime-composition.ts` always wires a `ContextTreeManager`, so the no-manager shape is unit-test-only and the real fail-open is the unconfigured-target case. The only thing that currently removes the grant is `prepareConfigurationRoots` *throwing* — the `configFailure` path, where `~/.context-tree` is deliberately a non-directory. That is a fault path, not an opt-out. A user who deliberately leaves Context Tree unconfigured in OpenTag, but who uses the standalone CLI on the same account, still exposes their `trees/`, `connections.json`, and `cleanup/launchers/` to every Session.

This matters more than a docs correction because it is the sentence the accepted trade rests on. Either the opt-out becomes real, or the doc should stop offering one and say the grant is unconditional for every Session on the account — and that is a materially bigger ask than what @bestony and I accepted.

**Suggested shape.** Gate the grant on a *configured target*, not on tree readiness:

```ts
const contextTree = await prepareContextTree(this.#contextTree, managed.cwd);
const configurationRoots = this.#contextTree && (await this.#contextTree.readConfig())
  ? await prepareConfigurationRoots(this.#environment)
  : [];
```

Readiness is the wrong condition and would break first use: a GitHub target that has not cloned yet is not `ready`, and the in-session CLI still needs to write under `~/.context-tree` to get there. "A target is recorded" is the condition that makes the documented opt-out true — no target, no grant, no directory created — while every configured path keeps exactly the reach it has today. The two tests above would then assert the opposite of what they assert now, which is the point: they currently lock in the fail-open.

Everything else I verified at this head still stands and is unaffected: `pnpm check` exit 0, client 1270 tests, CLI 520 tests, and the three reproductions (0755 tightened, symlinked directory rejected, planted-symlink write-through refused).

ggordonhall:
@liuchao-001 @yuezengwu — this is correct, and we are resolving it by dropping the promise rather than building the opt-out. Decision recorded in `dfc408b0`.

The finding holds: `prepareConfigurationRoots(this.#environment)` at `session-runtime-manager.ts:281` runs unconditionally, so leaving a target unconfigured does not remove the grant — and the tests at `session-runtime-manager.test.ts:101` and `:402` lock that in. The deleted sentence was the entire mitigation, and no configuration implemented it.

**Decision:** Context Tree is a built-in part of OpenTag and is always enabled. Every Session receives the `~/.context-tree` grant whether or not a target is configured. The grant is unconditional by design, not by oversight.

The doc now says that plainly instead of offering an opt-out:

> The grant is unconditional — OpenTag issues it for every Session whether or not a Context Tree target is configured, because Context Tree is a built-in part of OpenTag rather than an optional add-on. No configuration withholds it today; an explicit opt-out may be added later.

No code change in this commit, deliberately: the current behavior is what we want, and the fix was the documentation claiming otherwise. A real opt-out (gating the grant on a recorded target, as you sketched, with `readConfig` added to the `Pick`) is a reasonable follow-up but is not part of this PR.

The other parts of your verification stand unchanged — this only touches the design doc.

liuchao-001 (approved):
Approving at exact head `dfc408b0`, clearing my `CHANGES_REQUESTED`.

I offered two ways out — make the opt-out real, or stop offering one and state the grant is unconditional — and flagged the second as a materially bigger ask. You took the second and said so plainly. I accept it. The defect was never the grant's width; it was the doc naming a control that no configuration implemented. A reader can now see exactly what they get, which is the property that matters.

Verification carries forward rather than being re-run, and here is why that is sound rather than convenient: `git diff --name-only ae4b7b3a..dfc408b0` is one `.md` file, and `git diff --quiet ae4b7b3a dfc408b0 -- packages apps e2e` confirms the code tree is byte-identical to the head where I ran the full suites. `pnpm check` exits 0 here. The 1,270 client tests, 520 CLI tests, and the three reproductions (0755 tightened to 0700, symlinked managed directory rejected, planted-symlink write-through refused with the victim file untouched) all still describe this head's behaviour.

Two things I am carrying forward, neither gating:

1. **The real opt-out.** Gating the grant on a recorded target — with `readConfig` added to the `Pick` — remains worth doing, and the two tests that currently assert the fail-open would need to flip with it. Agreed it is a follow-up, not this PR.
2. **The Context Tree write is now larger than I first offered.** `opentag/system/context-tree-durable-memory.md` has to absorb three things, not one: account-scoped selection, the unconditional account-directory grant, and "Context Tree is built-in and always enabled." That last one sits awkwardly next to the node's existing "Durable memory is optional" and "a Computer with no named tree is a normal state." Both stay literally true — Sessions still start, and failures still only change what the prompt reports — but the node currently reads as though an unconfigured Computer is untouched by Context Tree, and after this it is not. I will make that distinction explicit when I take the write after merge, so the next reader does not have to rediscover it from this thread.

Thanks for taking the finding straight rather than papering it. @yuezengwu's catch was the sharpest of the review — it was aimed at the sentence the whole accepted trade rested on, and I had read past it.

PR #618, opened 2026-09-14
chore(dev): add test runner skills to lower testing token usage

ggordonhall:
## Summary

Adds a `test-runner` subagent plus a companion skill for both Claude Code and Codex so that test
execution can be delegated out of the main agent context.

Running `pnpm test` / `pnpm check` / integration suites in the parent context floods it with
verbose Vitest, Turborepo, and typecheck output that is rarely needed after the fact. The new
subagent runs only the requested command, keeps the verbose output in its own context, and returns
a compact report: command, exit status, elapsed time, and per-failure details (test or stage, source
file, assertion/error, smallest useful excerpt). It explicitly separates environmental/setup
failures from real test failures and never edits source or attempts fixes.

Files added:

- `.claude/agents/test-runner.md` — Claude Code agent definition (Haiku, `maxTurns: 12`) with reporting contract.
- `.claude/skills/test-runner/SKILL.md` — fork-context skill that delegates validation and mirrors the `AGENTS.md` test guidance (narrow scope while developing, all required checks before a PR).
- `.codex/agents/test-runner.toml` — Codex agent definition (`test_runner`, low reasoning effort, `workspace-write` sandbox).
- `.codex/skills/test-runner/SKILL.md` — Codex skill that routes test execution to the `test_runner` subagent and tells the parent how to consume pass/failure results.

No source, runtime, or product behavior changes.

## Testing

- `pnpm check` — passed (format, typecheck, lint, check) via the `pre-push` lefthook gate.
- No tests added: the change is agent/skill configuration only and does not touch product code paths.

## Breaking changes

None. Developer tooling only; no impact on published packages, server behavior, or UI.

## Checklist

- [ ] Tests and documentation cover the changed behavior. — N/A: no behavior change; guidelines live in `AGENTS.md` and are referenced by both skills.
- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. — `pnpm check` verified by the pre-push gate; build/test unaffected by non-code files.
- [x] Canonical English docs and Chinese mirrors are synchronized when applicable. — N/A, no documentation changes.


bestony (approved):
## Summary

Approving. This is developer-tooling only: Claude Code and Codex `test-runner` agents plus companion skills so `pnpm test` / `pnpm check` / integration output stays in a child context and comes back as a compact report. Frontmatter and TOML match the current Claude (`context: fork` + `agent`) and Codex custom-agent schemas. The reporting contract and no-edit rule are present on both sides. No product, runtime, or test-code changes.

Ownership: three markdown files are `*.md` territory; `.codex/agents/test-runner.toml` is `*` gate. One owner approval from this account covers both rules.

Optional follow-ups, not merge blockers:

1. Put WHEN in the always-loaded `description` fields (`pnpm test`, `pnpm check`, typecheck, Vitest, pre-PR checks). Right now they state what the runner does, not when the parent should delegate, which is the variance risk for this whole change.
2. On the Claude agent, deny Write/Edit (`disallowedTools`) so “do not edit source” is a capability, not only a prompt. Codex should stay `workspace-write` so Vitest can write cache.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

liuchao-001 (changes requested):
## Summary

I verified both halves of this change against the actual runtimes installed here (Claude Code 2.1.252, Codex 0.149.0), from a detached worktree at `dcf99cc`.

**The Claude half works end-to-end.** **The Codex half is inert** — `.codex/agents/test-runner.toml` is never loaded, so the `test_runner` subagent the Codex skill delegates to does not exist.

## Blocker: Codex does not read repo-local `.codex/agents/`

Codex 0.149.0 resolves agent roles from `$CODEX_HOME/agents/` only. A/B run from this PR's worktree (project explicitly trusted via `-c projects."<worktree>".trust_level="trusted"`, so trust is not the variable):

| `~/.codex/agents/` | cwd | roles in `spawn_agent` tool definition |
|---|---|---|
| empty | worktree containing `.codex/agents/test-runner.toml` | `NONE` |
| one probe role copied from this PR's TOML | same worktree | `<probe>, default, explorer, worker` — `test_runner` still absent |

The TOML itself is schema-valid (Codex requires non-empty `name`, `description`, `developer_instructions`; all three are present, and `gpt-5.6-luna` is a real slug). It is simply never read from the repo.

The skill, by contrast, *is* picked up — `codex debug prompt-input` in the worktree lists `test-runner: Delegate OpenTag test execution to the low-cost test_runner subagent ... (file: r0/test-runner/SKILL.md)`. So the shipped state is a loaded skill instructing Codex to delegate to a role that does not exist. Two likely outcomes, both of which defeat the PR's purpose:

1. Codex spawns a default-role agent instead. Its own `spawn_agent` description says full-history forks (`fork_turns` omitted or `"all"`) *inherit the parent model and reasoning effort and do not accept overrides* — so both the `gpt-5.6-luna` and `low` effort economy and the context isolation are lost.
2. Codex takes the skill's own escape hatch — "unless subagent delegation is unavailable" — and runs the suite in the parent context, which is exactly the behavior this PR exists to remove.

Either fix works:

- Keep the role, but ship it to `$CODEX_HOME/agents/` — an install step in `DEVELOPMENT.md` plus a line in the skill, since a repo file alone does nothing.
- Or drop `.codex/agents/test-runner.toml` and rewrite the Codex skill to not name a role: have the parent call `spawn_agent` with `fork_turns: "none"` and an explicit `model` / `reasoning_effort`, which is the documented way to get a cheap, context-isolated child without a role definition.

Worth a sentence in the PR description either way — a reader currently has no way to tell the Codex path was never exercised.

## Verified working (Claude side)

- Both files parse: `context`, `agent`, `background`, `user-invocable`, `permissionMode`, `maxTurns` are all in Claude Code 2.1.252's canonical frontmatter key list, and the agent + skill both load in this worktree.
- End-to-end fork: invoking the skill with a command in `args` spawned the fork, ran only that command, and returned the compact contract (`Command` / `Exit status` / `Output` / `Result`) inline, as `background: false` intends.
- `pnpm check` passes at this head (exit 0).

## Suggestions

1. **Claude skill: tell the caller to name the command.** `context: fork` means the fork starts with no parent context, but the body opens with "Run the validation requested in the current task" — text the fork is the only reader of, describing something it cannot see. It works today because a sensible caller passes `args`, but nothing instructs it to. The Codex twin gets this right ("Tell the subagent which command to run and why that scope is appropriate"); mirroring that one line into the Claude skill closes the gap.
2. **Seconding @bestony on `description` carrying WHEN** — and it binds harder on the Claude side than the review notes: `user-invocable: false` means model routing off `description` is the skill's *only* entry point. Naming `pnpm test`, `pnpm check`, typecheck and the pre-PR checks there is what makes it fire.
3. **Seconding the tool restriction, with evidence.** The agent resolves to `Tools: All tools` — `Write`/`Edit` included — so "do not edit source files" is prompt-only on a Haiku agent with 12 turns. `tools: Bash, Read, Glob, Grep` (an allowlist) is stronger than `disallowedTools` and still covers everything a test run needs. Agreed that Codex should stay `workspace-write` for Vitest cache.
4. **Nit:** the reporting contract is restated four times in four slightly different wordings. One of them is going to drift. Not worth blocking on for 47 lines.

bestony (changes requested):
@liuchao-001 Confirmed on Codex 0.154.0 (newer than your 0.149.0) from a detached worktree at `dcf99cca`. Retracting my Approve.

**Codex half is inert.** I reproduced the skill side with `codex debug prompt-input` in the worktree: `test-runner` is listed as `r0/test-runner/SKILL.md` (`r0` = the repo `.codex/skills`). The `test_runner` role is not. That matches your A/B.

Root cause is loader shape, not TOML schema. Current Codex `load_agent_roles` only discovers `*.toml` under each **config layer's** `agents/` directory. The user layer (`$CODEX_HOME`) is always present, even with no `config.toml`, so `~/.codex/agents/` works. The project layer is the in-repo `.codex/config.toml`. This PR never ships that file, so `.codex/agents/test-runner.toml` is never a layer folder and is never scanned. Skills walk the repo independently; roles do not. Official docs that say “drop a TOML in `.codex/agents/`” are describing a layout the loader does not implement without a project config layer. #14579 is the same class of bug even when `config.toml` *is* present.

So the shipped Codex skill tells the parent to delegate to a role that does not exist. Your two outcomes (default-role spawn inheriting parent model/effort, or the skill's “unless unavailable” escape hatch running in-parent) both defeat the PR.

I am not picking a fix. Your two options both close it:

- Keep the named role, but do not pretend the repo file is enough: install into `$CODEX_HOME/agents/` (and document it).
- Drop the TOML and have the skill spawn with `fork_turns: "none"` plus explicit `model` / `reasoning_effort`.

A committed `.codex/config.toml` might create the missing project layer so `agents/` is scanned; that is unverified here, and #14579 says even that path has failed to reach `spawn_agent` before. Do not take it as a third “just add the file” fix without a rerun of your A/B table.

Also agreed on your Claude suggestions (command in `args`; WHEN in `description` because `user-invocable: false`; Bash/Read/Glob/Grep allowlist). Those were nits on my first pass; they stay non-blocking relative to this Codex loader miss.

Claude half still looks good. Codex half is the blocker. Approve withdrawn until that path is actually loadable.

liuchao-001:
@bestony Ran the A/B you asked for before anyone treats `.codex/config.toml` as a third fix. **It is not one** — and the reason is stronger than "the PR forgot the file."

I also installed 0.154.0 into a scratch prefix so we are not comparing across versions. Everything below is from a detached worktree at `dcf99cca`, project trust forced via `-c projects."<worktree>".trust_level="trusted"`.

| Codex | setup | roles in `spawn_agent` |
|---|---|---|
| 0.149.0 | PR as-is (`.codex/agents/test-runner.toml` only) | `NONE` |
| 0.149.0 | + bare `.codex/config.toml` (`[features] multi_agent = true`) | `NONE` |
| 0.149.0 | + `[agents.test_runner]` with `config_file = "agents/test-runner.toml"` | `NONE` |
| 0.154.0 | all of the above together | `NONE` |
| 0.154.0 | probe role copied into `~/.codex/agents/` | `lc618b_probe, default, explorer, worker` |

**The project config layer does not exist at all.** I put `model = "gpt-5.6-luna"` and `model_reasoning_effort = "high"` in the worktree's `.codex/config.toml`; on both 0.149.0 and 0.154.0 the session still started `model: gpt-5.6-sol / reasoning effort: low`. The same two keys passed as `-c` overrides *do* move the header, so the instrument is sensitive and the keys are valid — the file is simply not read.

So the causal story inverts: it is not that `.codex/agents/` fails to be scanned *because* the PR omitted `.codex/config.toml`. There is no project config layer to create in the first place, which is why adding the file changes nothing. Shipping one would be cargo cult.

One correction on the citation: **#14579 is CLOSED as COMPLETED** (2026-03-14, filed against 0.114.0 — older than either of our builds), so it cannot carry weight as evidence of current behavior. Worth noting its repro also used a different shape than this PR — `[agents.<name>]` + `config_file` + `model_instructions_file`, not a standalone role file with `name` / `description` / `developer_instructions`. I tested that shape too (row 3); it is equally inert here.

Net effect on the PR: unchanged, and now with one option fewer. The two live fixes are the ones already on the table — install the role into `$CODEX_HOME/agents/` and document it, or drop the TOML and have the skill spawn with `fork_turns: "none"` plus explicit `model` / `reasoning_effort`. My `CHANGES_REQUESTED` stands; the Claude half still verifies clean.

bestony:
@liuchao-001 Thanks for running that table on both 0.149.0 and 0.154.0 with trust forced. I accept it.

The third path is dead. It is not “the PR forgot `.codex/config.toml`.” There is no project config layer to create: a worktree `config.toml` does not move `model` / `reasoning_effort` while the same keys as `-c` do. Shipping that file would be cargo cult. I should not have left it on the table even as unverified.

Retracting the #14579 citation. Closed as completed against 0.114.0, different role shape, and your row 3 shows that shape is inert here too. It does not describe current behavior.

Live fixes remain the two you listed: install the role into `$CODEX_HOME/agents/` and document it, or drop the TOML and spawn with `fork_turns: "none"` plus explicit `model` / `reasoning_effort`. `CHANGES_REQUESTED` stands. Claude half still fine.

yuezengwu (approved):
Approved at exact head `dcf99ccaca36f9dc13d51212c7381e0dfaa4d508` against base `c14da22c4ff289dc0eb31696b627ae41503d0a92`. I found no confirmed merge-blocking defect in these four tooling files.

I independently checked the Codex loader objection and obtained a contrary result. The repo-local `test_runner` role **is advertised to the model**, with the configured `gpt-5.6-luna` / `low` settings. This is the role catalog accompanying `spawn_agent`, not merely the skill description.

| Runtime and checkout | Parent model used in the request probe | `test_runner` in the role catalog |
| --- | --- | --- |
| Installed Codex 0.153.4, PR head | gpt-6-astra | Present, Luna / low |
| Fresh npm Codex 0.154.0, PR head | gpt-6-astra | Present, Luna / low |
| Fresh npm Codex 0.154.0, PR head | gpt-5.6-sol | Present, Luna / low |
| Same npm Codex 0.154.0, base control checkout | gpt-6-astra | Absent |

All checkouts were detached and clean. The PR checkout has no `.codex/config.toml`; my user `agents/` directory contains no TOML files, and I did not install this role globally. The final 0.154.0 probes used the normal user configuration with no temporary trust profile selected, explicitly enabled `agents.enabled`, and routed the first Responses request to a loopback capture server. The capture server returned an intentional error before inference, so no model or test command ran. A separate temporary trust-profile experiment produced the same positive discovery result and was removed afterwards.

The captured role declaration includes:

```text
test_runner: {
Runs the relevant OpenTag test command and returns a concise, actionable failure report to the parent agent.
- This role's model is set to `gpt-5.6-luna` and its reasoning effort is set to `low`. These settings cannot be changed.
}
```

This supports the project-local layout described in the [official custom-agent documentation](https://learn.chatgpt.com/docs/agent-configuration/subagents#custom-agents). It does not establish why the other environments omitted the role, but the claim that the shipped repository role is universally inert, or must be installed globally, does not hold in this reproduction. Please compare the effective invocation and model-visible role catalog before asking the author to replace the layout. My approval does not withdraw anyone else's change request.

Validation: all three Markdown frontmatters and the TOML parse; the four-file diff passes `git diff --check`; current technical CI is green. The prompts define bounded command execution, no source edits, and compact failure reporting. I did not independently run a real Claude/Codex child or the repository test suite in that child; role discovery and configured settings are the runtime behavior verified here.

Non-blocking improvements: name the triggering commands in descriptions; make the Claude caller's command/arguments explicit; optionally use a fresh Codex child context for these narrowly scoped commands to avoid copying unnecessary parent history. None requires introducing a global agent installation step.

liuchao-001:
@yuezengwu Your method is better than mine was, and one of my words was wrong. I redid everything at the wire level and I can now reproduce your capture exactly — but only under a condition that changes what it means.

First the correction I owe you: **"universally inert" was too strong.** There is a configuration where the shipped repo role *is* advertised, and your capture is real.

## Wire-level rerun

I dropped the model self-report method — you were right that it is the weaker instrument — and captured the first Responses request on a loopback server that errors before inference, same as you. Detached worktrees at head `dcf99cca` and base `c14da22c`.

| Codex | `CODEX_HOME` | extra config | `Available roles:` |
|---|---|---|---|
| 0.149.0 | normal | — | absent |
| 0.149.0 | normal | `agents.enabled=true` | absent, payload byte-identical (82749 B) |
| 0.154.0 | normal | — | absent |
| 0.154.0 | normal | `agents.enabled=true` | absent, byte-identical (81667 B) |
| 0.154.0 | normal | `+ features.multi_agent=true` | absent, byte-identical |
| 0.154.0 | **pristine empty dir** | `agents.enabled=true` | absent |
| 0.154.0 | normal, role copied into `~/.codex/agents/` | — | **present** |
| 0.154.0 | **pointed at `<checkout>/.codex`** | — | **present** |

The instrument is sound: the two "present" rows produce exactly the block you quoted, Luna / low and all. `agents.enabled=true` produced a byte-identical payload on both versions, so that flag is not the discriminator here. The pristine-`CODEX_HOME` row rules out my personal config.

## What reconciles our results

The last row reproduces your capture verbatim. If `CODEX_HOME` resolves to the checkout's `.codex`, then `.codex/agents/` **is** the user agents directory — so the role is found by the same `$CODEX_HOME/agents/` mechanism, not by project-local discovery.

That configuration also reproduces your base-control row: base has no `.codex` at all, so a `CODEX_HOME` pointed there finds nothing and the role reads "Absent". **Your base control cannot separate the two explanations** — it comes out identical either way.

I am not asserting that is your setup; I am saying it is a configuration that yields your exact bytes. The discriminating datum is the effective `codex_home` in your probe runs, not the checkout. Could you check it? If it is the default `~/.codex` and the role still appears on 0.154.0, then we have a real environment difference worth chasing, and I will drop this line.

## Why the change request stands either way

Even on the most favorable reading, the role resolves in your environment and not in a default install at the same two versions. A tool committed to the repo so contributors get cheaper test runs cannot depend on which of those a contributor happens to have — the failure is silent, and the skill's own escape hatch then runs the suite in the parent context, which is the cost this PR exists to remove.

So I am not asking for a global install step on aesthetic grounds. I am asking that the Codex path be reachable from a clean clone by construction: either the documented `$CODEX_HOME/agents/` install, or your third suggestion — spawn a fresh child with `fork_turns: "none"` plus explicit `model` / `reasoning_effort`, which needs no role definition and no global state. That last one is the better fix and it is yours; I would support it.

Agreed on your non-blocking three. Claude half unchanged and still verifies clean.

yuezengwu (commented):
@liuchao-001 @bestony I checked the home and isolated the difference in my reproduction. **`CODEX_HOME` was unset; the effective default was `/Users/yuezengwu/.codex`, not either checkout.** That directory has no agent TOML files. The role was never installed globally.

I then repeated the loopback capture on npm Codex **0.154.0**, head `dcf99ccaca36f9dc13d51212c7381e0dfaa4d508`, with inherited `CODEX_*` variables removed, `--ignore-user-config`, `sandbox_mode="workspace-write"`, the same parent model, and no temporary profile selected:

| Configuration | Model-visible `test_runner` role |
| --- | --- |
| User config ignored, no trust override | Absent |
| User config ignored, trust supplied through dotted keys containing literal quotes around the path | Absent |
| User config ignored, trust supplied as a TOML inline table | **Present, Luna / low** |
| Same inline-table invocation, base `c14da22c` control checkout | Absent |

The discriminating argument was this shape (substitute your worktree and primary checkout paths; retain your loopback fixture):

```sh
-c 'projects={"/absolute/path/to/worktree"={trust_level="trusted"},"/absolute/path/to/primary-checkout"={trust_level="trusted"}}'
```

There is an explanation in the exact 0.154.0 source:

- [`apply_toml_override`](https://github.com/openai/codex/blob/rust-v0.154.0/codex-rs/config/src/overrides.rs#L17-L26) splits the key on dots; it does **not** parse TOML quoting in the key. If the actual argv contains `projects."/path".trust_level`, those quote characters become part of the map key. Shell quoting that gets removed before argv is different. My earlier dotted-key trust probe had this mistake too.
- [Project discovery](https://github.com/openai/codex/blob/rust-v0.154.0/codex-rs/config/src/loader/mod.rs#L1800-L1812) explicitly creates an empty project layer when `.codex/config.toml` is absent. That file is not needed merely to discover the directory.
- The [layer iterator](https://github.com/openai/codex/blob/rust-v0.154.0/codex-rs/config/src/state.rs#L478-L484) excludes disabled layers; the [role loader](https://github.com/openai/codex/blob/rust-v0.154.0/codex-rs/agent-roles/src/loader.rs#L78-L87) scans each enabled layer's `agents/` directory.

This establishes a working path independent of my personal `config.toml`, using ordinary project trust and the shipped repository files. It does not prove the exact argv in your environment; please rerun with the inline-table form before treating project-role discovery as broken. Ignoring user config also ignores the temporary profile's trust settings, so selecting a profile together with that flag is not a valid positive trust control.

I am retaining my current-head approval. A global role install or replacement of the role is not required by this evidence. A short note about trusting the checkout could help contributors. A fresh child context remains a useful cost optimization.

Scope remains precise: these requests were captured and rejected locally **before inference**. They verify role discovery and the declared Luna / low settings, not an actual child test run or measured token savings. I did not change either checkout or persistent user configuration.

liuchao-001 (approved):
@yuezengwu You are right and I was wrong. Withdrawing my change request and approving at `dcf99ccaca36f9dc13d51212c7381e0dfaa4d508`.

You found the actual bug, and it was in my instrument, not in the PR. I reproduced your result on npm Codex 0.154.0 with `--ignore-user-config`, `CODEX_HOME` unset, loopback capture before inference:

| Configuration (head `dcf99cca`) | `Available roles:` |
|---|---|
| no trust argument | absent (80598 B) |
| trust via `-c 'projects."<path>".trust_level="trusted"'` — my form | absent (80598 B) |
| trust via `-c 'projects={"<path>"={trust_level="trusted"}}'` — your form | **present, `test_runner`, Luna / low** |
| your form, base `c14da22c` control | absent |

Rows 1 and 2 are the same length and differ only in per-request UUIDs — **my trust override was a no-op in every probe I ran.** `apply_toml_override` splits the key on dots without parsing TOML quoting, and my worktree path contains `.first-tree-staging`, so the key was shredded. Every "absent" I reported came from an untrusted project layer, which is exactly the condition that disables role discovery. My `CODEX_HOME`-points-at-the-checkout hypothesis was a wrong explanation for a wrong measurement; sorry for sending you to check it.

Confirmed on the shipped files as they stand: the PR head has **no** `.codex/config.toml`, and the role is still discovered. So project discovery does create the layer without that file, gated on trust, exactly as your source links say.

## A correction I owe @bestony

My earlier claim that "the project config layer does not exist at all" and that shipping `.codex/config.toml` "would be cargo cult" was the same artifact, and you retracted your position on the strength of it. Re-running with trust actually applied:

- project `.codex/config.toml` with `model = "gpt-5.6-luna"` / `model_reasoning_effort = "high"` → session starts **`gpt-5.6-luna` / `high`**
- identical run with my broken dotted-key form → `gpt-5.6-sol` / `low`

The project config layer is read. Please un-retract on that point — my refutation was invalid. For the record the precise mechanism is neither of the ones we argued: `.codex/config.toml` is not required to create the layer, and the gate is project trust. Your instinct that a layer was missing was closer than my "there is no layer."

## Position

No confirmed merge-blocking defect in these four files. The Codex path is reachable from a clean clone with ordinary project trust, no global install and no role replacement — so my ask for an install step or a rewrite is withdrawn, and so is my objection to the layout.

Endorsing your non-blocking notes, in priority order: a line telling contributors to trust the checkout (it is the difference between this working and silently not working, and it is invisible when it fails); WHEN in the `description` fields; the Claude caller naming its command explicitly.

Same scope caveat as yours: my probes reject before inference, so they verify role discovery and the declared Luna / low settings, not a real child test run or measured token savings. The Claude half I did exercise end to end earlier and it behaves as intended. Neither checkout nor my persistent user config was modified; all scratch state is removed.

yuezengwu:
My preference for this PR is to first add a logging convention to the existing repository work guidelines:

- Run test commands directly and save the full output to a log file.
- Preserve the actual exit status and return a concise result with relevant failure excerpts; read more of the saved log when needed.
- Consider subagent delegation only for long-running suites, substantial log analysis, or validation that can run independently alongside the main task.

liuchao-001:
@yuezengwu Support the convention — it is useful independently of what happens to this PR. One concrete thing to nail down before it lands in `AGENTS.md`, because the natural reading of your first bullet defeats your second.

"Save the full output to a log file" is usually written `pnpm test | tee test.log`, and a pipeline reports the **last** command's status, so a failing suite comes back green:

```
$ false | tee out.log >/dev/null ; echo $?
0        # bash and zsh both
```

The safe forms:

```
pnpm test > test.log 2>&1 ; echo $?          # 1 — no pipe, status preserved
set -o pipefail; pnpm test | tee test.log    # 1 — bash/zsh opt-in
pnpm test | tee test.log; echo ${PIPESTATUS[0]}   # zsh: ${pipestatus[1]}
```

Plain redirect is the one I would write into the guideline: it needs no shell option, and nothing in this workflow needs live output on the terminal — the whole point is that the log is read selectively afterwards. If the guideline says `tee`, it should say `pipefail` in the same breath.

Worth being explicit about this in the text rather than leaving it to the reader. A convention whose failure mode is "the agent reports PASS on a red suite" is worse than no convention, and it is invisible exactly when it matters.

On sequencing: your three bullets and this PR are not really in tension — your own third bullet (delegate long suites and heavy log analysis) is the case these four files implement. Whether the convention lands first is a scope call for @ggordonhall and the owners, not something I have a technical objection to either way. My approval stands as is; I am not re-blocking on direction.

Also note `AGENTS.md` already carries the adjacent rule ("Run directly affected tests during development and all commands above before opening a pull request"), so the convention probably belongs in that same Commands section rather than a new one.

PR #622, opened 2026-09-15
fix(server): hold Computer presence through a runtime reconnect

bestony:
## Problem

A dropped runtime WebSocket cleared the Computer's `current_instance_id` the instant it closed:

- `runtime-session.ts` `#onClose` awaited `ComputerService.disconnect()` directly after removing the socket from the `ConnectionRegistry`.
- `disconnect()` nulls `current_instance_id` and `connected_at`.
- `listAccountComputers` reads `connectionStatus` as online only while `current_instance_id` is set **and** `last_seen_at` is inside `presenceTimeoutMs` (90s), so the 90s timeout never got a chance to act.
- `projectComputerProviderReadiness` / `projectComputerImCliReadiness` map an offline Computer to `unavailable` for every provider.

The client reconnects on a capped exponential backoff starting around a second. An ordinary blip therefore published a definite offline and a full `unavailable` readiness set, then reverted seconds later — a flicker in the Computer card, not a real outage.

## Change

Presence is released on a grace window (default 15s) instead of immediately.

- New `ComputerPresenceGrace`: one pending release per Computer, keyed and replaceable, cancelled by a later registration, drained on shutdown.
- `ComputerService.releaseConnection()` schedules it; `register()` cancels; `close()` drains. `disconnect()` keeps its existing immediate contract and is what the window ends in.
- `RuntimeSession` calls `releaseConnection()` at both close paths.

What is deliberately unchanged: the `ConnectionRegistry` still drops the socket the moment it closes, so nothing is dispatched to a dead connection during the window. The dispatch fences in `im-delivery-worker` compare the registry's instance against the row's, and the registry returns nothing while disconnected, so a lingering `current_instance_id` cannot route work anywhere.

A release that fires after a replacement registered is fenced by `disconnect()`'s existing `current_instance_id = $instanceId` predicate, so an old connection can never take a newer instance offline.

## Tests

- `services/computers/__tests__/presence-grace.test.ts` (new): window elapses, reconnect inside the window, replacement of a pending release, per-Computer cancellation scope, failing release swallowed, shutdown drain.
- `server-account-services.test.ts`: a Computer stays online across release + re-register, and goes offline once a window with no reconnect ends.
- `runtime.test.ts`: the close paths now assert `releaseConnection`, including the pending-replacement fencing case.

## Scope

This is one of two fixes from an incident where messages sat in "queued" while the Computer status flickered. It addresses the flicker only. The `queued` projection is a separate change; delivery retry windows and the accepted/report reconciliation window are not touched here.


packages/server/src/services/computers/presence-grace.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * This instance's connection ended. The newest close is the one that matters, so a release
+   * already pending for the Computer is replaced rather than kept alongside.
+   */
+  schedule(computerId: string, instanceId: string): void {
+    this.cancel(computerId);
+    const timer = setTimeout(() => {
+      this.#pending.delete(computerId);
+      void this.#clearPresence(computerId, instanceId).catch(() => undefined);
```
yuezengwu, line 47: [P1] Fence a fired release against a same-instance reconnect. `RuntimeConnection.#connectOnce()` reuses `this.#options.instanceId` on every reconnect (`packages/client/src/runtime/runtime-connection.ts:461`), so a new socket does not necessarily have a new instance ID. Once this timer fires, the entry is removed and `register()`'s later `cancel()` cannot stop an in-flight database update. If registration commits while that update is waiting, `disconnect()` still matches the reused `current_instance_id` and clears the newly connected Computer. I reproduced: register instance A → fire/reach the release callback → delay its write → register A again → allow the original disconnect; `listAccountComputers()` becomes offline. Replacing A with B is the passing control. This can also make the new connection's next heartbeat fail its instance check. Please fence the actual release write by a connection generation, or otherwise serialize/fence it with re-registration; cancelling only the timer is insufficient. Add the same-instance, already-fired-callback regression rather than generating a fresh instance for every reconnect test.

packages/server/src/runtime/runtime-session.ts
```diff
@@ -635,7 +635,9 @@ export class RuntimeSession {
       this.#instanceId &&
       this.#registry.remove(this.#computerId, this.#instanceId, this.#socket)
     ) {
-      await this.#computers.disconnect(this.#computerId, this.#instanceId).catch(() => undefined);
+      // The registry drop above is what fences the dead socket. Presence is released on a grace
+      // window so a reconnect within it never reads as an offline Computer.
+      this.#computers.releaseConnection(this.#computerId, this.#instanceId);
```
yuezengwu, line 640: [P2] Update the remaining ComputerService test double for this new close contract. `runtimeComputerService()` in `packages/server/src/observability/__tests__/observability.test.ts:1149` still supplies `disconnect` but no `releaseConnection`, so closing the real RuntimeSession in that suite throws an uncaught TypeError here. The full local server run has 956 passing assertions but still exits 1, and CI run 34916767255 reports the same exception from `traces real RuntimeSession failure, overload, and dropped-queue terminal paths`. Please update this double and any other RuntimeSession fixtures to supply the new method, then obtain a clean test/CI exit; passing assertion counts alone do not establish a passing suite.

yuezengwu (changes requested):
Requesting changes at `c4195e1ecb86d44a06956469dbd9e7c49c7320d6`: one reconnect race and one test-double regression, detailed inline.

Validation in an isolated detached worktree: frozen offline install, `pnpm check`, shared build, and server typecheck passed. The complete server unit run reported 956 passing tests in 74 files but exited 1 because of the unhandled `releaseConnection` exception, matching CI run 34916767255. An additional deterministic database-backed probe failed for a same-instance reconnect and passed for the fresh-instance control. The probe delays only the old callback's database write; both registration and the eventual disconnect use the real ComputerService/database implementation.

No author-branch changes or remote CI reruns.

PR #623, opened 2026-09-15
fix(server): stop an overtaken pending delivery holding a Task at queued

bestony:
## Problem

`topicCtes` aggregated the queue as `bool_or(state = 'pending')` over every delivery in a topic, and `taskStatus` ranks `hasPending` above the latest execution's outcome. Any pending row therefore pinned the whole topic at `queued`.

`ImDeliveryWorker.#recordFailure` treats runtime-unavailable, stale-placement and dispatch failures as retryable: it writes `last_error_code` and a `next_attempt_at`, leaving `state = 'pending'`. There is no attempt ceiling and no retry-window ceiling, so the only thing that ends such a row is the TTL — **seven days** for a direct delivery (`DIRECT_TTL_MS`). For that whole week the topic read as `queued` even after later messages in it ran to a report, and the outcome the Account was waiting for never surfaced.

(Correcting one detail from the incident report that prompted this: `#recordFailure` does not set `reason`, so these rows *are* inside `runImDeliveryExpiry`'s `reason is null` predicate and the janitor does sweep them. The defect is that the sweep is seven days away, not that it never comes.)

## Change

The queue is read from the arrival order of the topic's messages rather than from the presence of a pending row. `topics` now exposes two maxima and `page` derives `has_pending` from them:

- `live_pending_message_at` — newest `occurred_at` among pending deliveries still inside their TTL.
- `settled_message_at` — newest `occurred_at` among everything else.

A topic is queued when a live pending delivery exists and its message is no older than the newest settled one. Pending wins an exact tie, so a message that arrives while an earlier Turn is still finishing keeps reading as `queued` rather than being swallowed by that Turn's report — the case `"reads cancelled after withdrawing a follow-up queued behind a Turn that finished later"` already covers, and which still passes unchanged.

A pending row past its TTL now reads as `expired`, which is exactly what the janitor is about to record.

No schema change; `expires_at` and `im_messages.occurred_at` are both already in the CTE's reach.

## What this does not do

The retry loop itself is untouched — no attempt ceiling, no retry-window ceiling, no terminal state for unrecoverable errors. A stuck delivery still retries for seven days; it just no longer masks the rest of its topic. That is a separate change.

## Tests

- `lets a topic's newest message decide once an overtaken delivery is still pending` (new): a stuck pending root under a follow-up that completed reads `completed`; under one that failed, `failed`.
- `stops counting a pending delivery as queued once its TTL has lapsed` (new).
- Both were confirmed to fail against the previous `bool_or` aggregate.
- Full server unit suite (952) and the PostgreSQL integration suite (334, incl. `integration/tasks.test.ts`) pass.

## Scope

Second of two fixes from an incident where messages sat in "queued" while the Computer status flickered; the flicker is #622.


packages/server/src/services/tasks/task-service.ts
```diff
@@ -981,7 +996,10 @@ export class TaskService {
           t.channel_id,
           t.topic_key,
           t.has_running,
-          t.has_pending,
+          (
+            t.live_pending_message_at is not null
+            and (t.settled_message_at is null or t.live_pending_message_at >= t.settled_message_at)
```
yuezengwu, line 1001: [P1] Preserve cancellation of deliveries that remain eligible for execution. With an unexpired pending root at t0 and a completed reply at t2, this comparison makes `has_pending` false and the Task becomes completed. But `TaskService.cancel()` only proceeds when the projected status is queued, so it now returns `TASK_NOT_QUEUED` / 409 without withdrawing the root. The worker still selects that root from its pending queue; it has no corresponding 'overtaken' exclusion. I reproduced the full sequence with the real worker: completed summary → refused cancel → `runOnce()` delivers and accepts the old root. Thus this changes more than a display label: a still-retryable command becomes impossible to cancel through the Task API and can execute after the UI says the Task finished. Keep remaining live work visible/cancellable, or durably retire explicitly superseded work before treating the queue as finished; a newer message's report alone does not retire the older delivery. Cover list/get, cancellation, and subsequent worker eligibility together.

yuezengwu (changes requested):
Requesting changes at `9478599b1566fd1ca0d722c37cd51663fa0956b8`: hiding the older pending delivery also removes the user's ability to withdraw it, while the delivery worker still executes it.

Validation in an isolated detached worktree: frozen offline install, `pnpm check`, shared build, server typecheck, and all 33 existing TaskService tests passed. A new cancellation regression passed against the exact base TaskService (`71e1d172`) but failed at this head with `TASK_NOT_QUEUED` / HTTP 409; the equal-message-timestamp control passed on both. An additional witness using the real ImDeliveryWorker, database, custody store, and existing fake runtime domain confirmed that after the Task reports completed and rejects cancellation, `runOnce()` still delivers the old pending root and records its acceptance (`turn-1`). This was a local synthetic fixture, not real user traffic.

No author-branch changes or remote CI reruns.

PR #625, opened 2026-09-15
feat: enable Pi across Local Agent workflows

yuezengwu:
## Summary

OpenTag had a Pi adapter, but normal Agent creation and Client runtime composition did not admit it. This change makes Pi selectable and runnable through existing Local Web, CLI and Server paths, with the existing Session lifecycle.

- Register Pi in Shared schemas, the existing PostgreSQL provider enum, Server admission, Client readiness, configuration, usage reporting and CLI/Web setup. Preserve Codex and Claude Code.
- Keep Provider-readiness v1 frozen to Codex/Claude Code. Pi requires explicit v2 opt-in, independently of transport v1/v2. New Client and Web callers offer both versions; Server WebSocket and HTTP responses respect the negotiated vocabulary. A v1-only daemon projects Pi as unavailable rather than checking indefinitely.
- Preserve a Pi Session UUID after an interrupted first Turn. Add the session-file fingerprint only after history exists; resume any history saved before the binding update. Once materialized, missing or empty history fails before a new prompt. Only a settled `stop` terminal plus successful child-process cleanup completes a Run.
- Extend the authoritative CLI/Server/Web onboarding fixture with Pi success. Update production-support, Context Tree, recovery and setup contracts, including existing Chinese mirrors.

## Validation

Validated source: `25df76eaa41c16058e272aca2d0452b58031b4f5`.

- `pnpm check`, `pnpm build` (including bundle report), `pnpm typecheck`, and pre-push checks: passed without disabling gates.
- `pnpm test`: 333 script tests and 4,073 CLI/Web/Shared/Client/Server tests passed.
- `pnpm --filter @opentag/client test:agent-runtime:coverage`: 416 tests; statements, branches, functions and lines all 100%.
- `pnpm --filter @opentag/server test:integration`: 335 tests passed against real PostgreSQL.
- Real Server WebSocket probes compiled the actual base-revision schemas and verified old Client acceptance of welcome, registration and heartbeat under transport v1 and v2. Tests also cover new-Client/old-Server readiness fallback, HTTP opt-in and v1-only daemon Pi availability.
- Four real Pi 0.84.2 fault probes used a loopback model with the real SessionRuntimeManager and disk-backed binding store: interruption before an assistant response, pause before first JSONL creation, interruption after tool history exists, and a failed OpenTag binding write after Pi has saved history. All recovered the same UUID with one session file, retained history and successful owned-process cleanup. No external model calls were made by these fault probes.
- `node scripts/e2e/cloud-computer.mjs local-pi`: 14/14 passed on the clean committed source with real Pi 0.84.2 and configured DeepSeek Flash. Verified file tools, same-Session recall, Client restart, foreground child shutdown, durable cancellation receipt and cleanup. Owned PIDs, container and loopback port were independently confirmed absent.
- Committed regression tests exercise same-runtime retry and a new Client composition reading the first interrupted Turn's binding from disk. An independent actual-Pi probe verifies missing/empty materialized history is rejected before forwarding a model prompt.

GitHub [CI](https://github.com/first-tree-ai/opentag/actions/runs/34937453519) and [Patch Coverage](https://github.com/first-tree-ai/opentag/actions/runs/34937453588) passed on this exact commit. CI includes Node 22.22.2, 24 and 26 compatibility, Browser/Container/CLI smokes, PostgreSQL integration and the mandatory Agent Runtime coverage gate. Changed-line coverage is 360/362 lines (99.45%; unchanged threshold 80%).

The E1 acceptance harness uses real Server/CLI/Pi and PostgreSQL, with synthetic IM ingress and isolated dummy Slack binding/probe responses. It does not establish external IM delivery. The active-stop case closes the Client daemon; reliable Server per-Session stop/status and Cloud resource management remain later increments.

## UI validation

Rendered Web sources are unchanged by the review fixes. The initial Pi admission was inspected in the built `/internal/agent-setup` presentation at 320px, 390px, 768px and 1440px: Pi renders with the other providers, exposes its selected state and enables Create Agent. Tab reaches Back and Enter activates Create Agent. This was a focused accessibility check.

The browser lab uses an in-memory adapter; actual API creation/execution is covered by real acceptance and Web creation tests. At 320px, the existing global minimum width with a 15px browser scrollbar causes slight horizontal overflow; the same rule exists on the base branch. No new CSS or reusable UI block is introduced.

## Compatibility and boundaries

Migration `0041_colossal_magdalene.sql` only adds `pi` to the existing provider enum; no new table is added. Pi requires a compatible installed CLI and usable configured model credentials. Local Pi uses unrestricted filesystem/network access with no approval gate; unsupported stricter execution restrictions are rejected. Packaged Context Tree skills are passed explicitly and ambient discovery is disabled. This PR does not add Cloud sandbox isolation.

The session-file hash identifies the file path, not content integrity; it cannot detect arbitrary partial history corruption retaining messages. Recovery preserves a Session and permits a later Turn; it does not claim the interrupted Turn succeeded or automatically replay external effects.

## Checklist

- [x] Full repository verification, CI and real Pi acceptance pass.
- [x] Canonical English docs and existing Chinese mirrors are synchronized.


Gandy2025 (changes requested):
Requesting changes for five validated blockers/gaps on the fixed review object `71e1d172f628a2df5c2b00bec12af8cef0ccb1be...b8e6e44ecab05e7542b772ae0c613b5c12edf7f7`.

## Spec findings

1. **[P1] Preserve the Provider-readiness v1 wire vocabulary** — `packages/shared/src/agent.ts:23` adds `pi` to `AGENT_RUNTIME_PROVIDERS`, and `packages/shared/src/runtime-protocol.ts:162-199` derives both the observation and negotiation schemas from that enum while the readiness version remains `1`. The Server consequently advertises Pi in a version-1 welcome (`packages/server/src/runtime/runtime-session.ts:322-334`) to every existing Client, because clients from the base revision unconditionally send `x-opentag-provider-readiness: 1`. Those clients have a strict `codex | claude-code` enum; they reject this welcome as malformed, and `runtime-connection.ts:587-613` treats it as fatal with no fallback. This breaks the documented Server-first rolling upgrade and disconnects deployed daemons until they upgrade. Keep readiness v1 limited to its frozen two-provider vocabulary, negotiate Pi through a new readiness version/capability, and add an old readiness-aware Client compatibility test.

2. **[P1] Do not materialize the durable Pi binding before the session file exists** — `packages/client/src/providers/pi/agent-runtime.ts:187-209` persists `sessionFileHash` immediately after the pre-prompt `get_state`, before Pi has written an assistant message. Pi 0.84.2 does not create the JSONL until the first assistant message; if the process crashes, the provider fails, or the Turn is cancelled in that window, the durable binding points to a path that never existed. On restart, Pi cannot find the session ID and creates a new timestamped path; OpenTag then rejects it at lines 202-203 as “another session file,” permanently stranding the Session. Keep the binding unmaterialized while `messageCount === 0`, persist the path only after the first successful materialization, and cover restart after interruption before the first assistant message. Upstream behavior is visible in Pi 0.84.2's [`main.ts`](https://github.com/earendil-works/pi/blob/914cf1472e715297caa30db4b9535d534a9eb718/packages/coding-agent/src/main.ts#L438-L450) and [`session-manager.ts`](https://github.com/earendil-works/pi/blob/914cf1472e715297caa30db4b9535d534a9eb718/packages/coding-agent/src/core/session-manager.ts#L1015-L1038).

## Standards findings

3. **[P1] Restore the required Agent Runtime coverage gates** — the exact-head `Checks` job runs all 401 Agent Runtime tests but fails the repository's mandatory 100% threshold because the default `PiAgentRuntimeFactory` construction at `packages/client/src/runtime/client-runtime-composition.ts:1081` is uncovered (99.97% statements/lines, 99.96% branches, 99.86% functions). Separately, `Enforce Patch Coverage` fails because the deliberately explicit manifest at `scripts/unit-coverage.mjs:66-85` omits the new `providers/pi/runtime-policy.ts`. Exercise the non-injected default Pi factory path, add the policy source to the authoritative manifest, and rerun both gates; the PR description's 100% claim does not match the current head.

4. **[P2] Extend the authoritative cross-layer onboarding matrix to Pi** — `scripts/fixtures/onboarding-preparation-contract.json:15,43` still declares only Codex and Claude Code, while its own description says it is the CLI/Server/Web contract matrix. The CLI and Web fixture types likewise hard-code those two providers, and `packages/server/src/services/agents/__tests__/agent-setup-contract.test.ts:354-369` explicitly pins “both” Runtime identities. As a result, the suite intended to catch readiness/projection/gating drift has no Pi success path. Add a `success-pi` scenario and extend the CLI, Server, and Web assertions, preferably using the shared provider type instead of another local union.

5. **[P2] Update the normative production-support contracts** — `docs/design/agent-runtime-contract.md:76-85` still says Pi is “not registered” and intentionally excluded from production composition pending the exact end-to-end boundary review this PR performs. `docs/design/agent-runtime-test-plan.md:196-205` remains Codex-only, and `docs/design/context-tree-integration.md:390-413` still defers every provider beyond Codex/Claude Code. CLI help and the English/Chinese setup documentation also omit Pi. Update the normative support/security boundary, binding lifecycle, test plan, CLI help, and synchronized mirrors in this PR so they no longer prescribe the opposite behavior from production.

## Verification

This was a high-risk Standards + Spec review because the change crosses the shared provider schema, persisted binding lifecycle, wire negotiation, Server admission, Client process ownership, and a PostgreSQL enum migration. The migration journal/snapshot/count and the existing Codex/Claude paths were also checked; no additional blocker was found there. Current GitHub checks independently confirm the two coverage failures above.

Gandy2025 (changes requested):
Re-review of the fix delta `b8e6e44ecab05e7542b772ae0c613b5c12edf7f7...b4b31d8f26c2264f937c957ad0b421458cb5f413` against base `71e1d172f628a2df5c2b00bec12af8cef0ccb1be`.

The previous coverage blocker is resolved: the default Pi factory path now has behavioral coverage, the explicit source manifest owns `pi/runtime-policy.ts`, and the exact-head `Checks` and `Enforce Patch Coverage` jobs pass. The CLI help and PR validation evidence were also corrected. Four findings remain.

## Spec findings

1. **[P1] Preserve the Provider-readiness v1 wire vocabulary** — this fix commit does not change the readiness boundary. `packages/server/src/api/runtime.ts:95-100` still returns the complete admitted provider set to every Client sending the v1 header; `packages/server/src/runtime/runtime-session.ts:312-334` still encodes that set as `providerReadiness.version: 1`; and the strict enum derived in `packages/shared/src/agent.ts:23-24` / `runtime-protocol.ts:193-199` now includes Pi. A Client built from the base revision sends that header but accepts only Codex and Claude Code, so it fatally rejects the welcome. The documented Server-first rolling upgrade remains broken. Keep readiness v1 frozen to its original vocabulary, negotiate Pi through a new readiness version/capability, and add an old readiness-aware Client compatibility test.

2. **[P1] Keep a first-Turn interruption recoverable instead of permanently rejecting the Session** — `packages/client/src/providers/pi/agent-runtime.ts:192-198,228-250` still persists `sessionFileHash` before the prompt while Pi reports `messageCount === 0`. Pi 0.84.2 does not create the JSONL until an assistant message exists. If the first process dies before that point, the durable binding therefore names a file that never existed. On restart, Pi selects a new timestamped path and fails the hash check at lines 240-241; even if the path happens to match, the new check at lines 243-245 only turns the same state into an explicit protocol failure. `piBindingRequiresUnmaterializedReplacement` at lines 1070-1072 still makes only a binding without a hash replaceable, so every later Turn resumes the same unusable binding. The new tests prove fail-closed behavior, not recovery. Delay materialization until Pi has actual history (or model an explicitly replaceable pre-materialized state), then add a cross-manager restart test covering interruption before the first assistant message.

## Standards findings

3. **[P2] Extend the authoritative cross-layer onboarding matrix to Pi** — `scripts/fixtures/onboarding-preparation-contract.json:15,43` still lists only Codex and Claude Code even though the fixture declares itself the CLI/Server/Web contract matrix. The CLI/Web fixture types retain the same two-provider union, and `packages/server/src/services/agents/__tests__/agent-setup-contract.test.ts:354-369` still pins only “both” identities. Add a `success-pi` scenario and extend all three surface assertions so this contract can detect Pi readiness/projection/gating drift.

4. **[P2] Update the normative production-support contracts** — `docs/design/agent-runtime-contract.md:76-85` still says Pi is “not registered” and intentionally excluded from production pending this exact boundary review; `docs/design/agent-runtime-test-plan.md:196-205` still describes production as Codex-only; and `docs/design/context-tree-integration.md:389-413` still defers every provider beyond Codex/Claude Code. The new E1 recovery notes and CLI help are synchronized, but they do not repair these contradictory source-of-truth contracts or the public README/CLI README support guidance. Update the affected canonical documents and mirrors to record the admitted Pi boundary and its actual lifecycle/security constraints.

No new regression was found in the delta's cleanup-failure handling, terminal mapping, default skill arguments, or existing Codex/Claude paths. The remaining re-review scope is the four boundaries above.

liuchao-001 (commented):
Independent review of exact head `b4b31d8f26c2264f937c957ad0b421458cb5f413` (base `71e1d172`). I re-derived the open findings from the code rather than relying on the earlier review. All four remaining items reproduce, so I'm not approving yet. Two details below affect how the fixes should be shaped.

**1. Readiness v1 vocabulary (P1): confirmed.** `docs/runtime-protocol.md:7` says the v1 dialect and its Provider-readiness extension stay *frozen for rolling upgrades*. `api/runtime.ts:99` still sends the full `SERVER_ADMITTED_AGENT_RUNTIME_PROVIDERS` to any `x-opentag-provider-readiness: 1` Client. On a base-revision Client, `ServerRuntimeFrameSchema.safeParse` fails on `"pi"` during the welcome. That path falls through to `failProtocol` (base `runtime-connection.ts:587-613`), so the Client never reaches `registered`. One thing to cover in the fix: once v1 is frozen, `projectComputerProviderReadiness` (`services/computers/provider-readiness.ts:86-95`) will report Pi as `checking` forever for a Computer whose Client only negotiated v1, because no snapshot ever arrives. A v1-only connection should project Pi as unavailable/unsupported, not as an indefinite probe.

**2. First-Turn materialization (P1): confirmed against Pi 0.84.2 source.** `SessionManager._persist` does not write the JSONL until the first assistant entry exists. `main.ts` handles `--session-id` with no existing file by calling `SessionManager.create(cwd, sessionDir, { id })`, which picks a new timestamped path. `#restoreSessionState` stores `sessionFileHash` and sets `#sessionExists = true` while `messageCount === 0`. After that, every later Provider Run (each one spawns a fresh Pi process) either fails `Pi opened another session file` or `Pi session has no conversation history`. `piBindingRequiresUnmaterializedReplacement` never gives that binding a way out. The stop case the E1 harness exercises (Client shutdown mid-Turn) is exactly the trigger when it lands on a Session's first Turn. The window is also wider than a Client restart: the next Run in the *same* Client process hits the same mismatch. So the regression test should cover both a same-manager retry and a cross-manager restart.

**3. Onboarding contract matrix (P2): confirmed.** `scripts/fixtures/onboarding-preparation-contract.json:15,43` still declares two providers.

**4. Normative docs (P2): confirmed.** `docs/design/agent-runtime-contract.md:76-85` is already stale for Claude Code too: its table still says "not registered". When updating it for Pi, please correct the whole production-composition column in one pass.

Nothing new found in the migration (`ADD VALUE 'pi'` only), Pi usage normalization (`input + cacheWrite` as input, `cacheRead` as cached, which matches the `provider !== "codex"` Server normalization), the Pi policy rejection path, or the Context Tree `--skill` path. Those look correct.

Gandy2025 (changes requested):
Re-review of exact head `25df76eaa41c16058e272aca2d0452b58031b4f5` against the current base `fb607ce6756833e50d77155c8044b45a3ec16ead`.

The four previously open implementation gaps are substantially resolved: readiness v1 is frozen at the handshake, first-Turn Pi bindings remain recoverable, the cross-layer onboarding matrix covers Pi, and the primary production contracts plus coverage gates are repaired. Two adjacent compatibility blockers and one remaining documentation gap still prevent approval.

## Spec findings

1. **[P1] Do not let preparation widen the negotiated readiness provider set** — `packages/server/src/runtime/connection-registry.ts:237-257` treats `providerReadinessProviders` as mutable and unconditionally appends the provider passed to `beginPreparation()`. Its caller at `packages/server/src/runtime/provider-cli-reconcile-owner.ts:242-291` checks only the independent `providerCliPrewarm` capability, not whether the Runtime provider was negotiated. A protocol-v2/readiness-v1 Client therefore starts with Pi correctly unavailable, but creating or binding a Pi Agent changes Pi to `checking` and sends an unversioned prewarm frame containing `runtimeProvider: "pi"`. A base-revision Client's strict business-frame schema accepts only Codex/Claude Code, so `runtime-connection.ts:587-613` rejects that frame and closes the connection. If a tolerant Client returns a result, the Server can instead mark the unnegotiated Pi provider ready and admit it through `packages/server/src/index.ts:232-246`, bypassing the v2 opt-in. Keep the negotiated set immutable, reject preparation for a provider outside it, and cover Pi placement/preparation on a v1-only connection (no Pi frame, Pi remains unavailable, no runtime admission).

2. **[P1] Stage the Agent API and durable Pi writes so old binaries remain readable and rollback-safe** — `packages/shared/src/agent.ts:23-24,74-85,277-281` expands the existing unversioned Agent response schemas to Pi, while `packages/server/src/api/account.ts:126-129` returns every Agent without a Client-version projection. The base Client uses the old strict `codex | claude-code` schema to parse the entire list (`packages/client/src/api.ts:291-299`), so after an Account creates one Pi Agent an old CLI/Client cannot list even that Account's existing Codex/Claude Agents. Rolling the Server image back is also unsafe: the old Server encounters the persisted Pi row and fails its own response parse. This is the durable-data activation boundary for which `docs/runtime-protocol.md:95-99` requires an expand/contract and rollback plan, and it contradicts the PR's no-breaking-change claim. Introduce a legacy-safe/versioned Agent API projection before admitting Pi writes, or otherwise split activation into a release sequence with an explicit rollback gate; add old-Client list/read and old-Server rollback coverage around a persisted Pi row.

## Standards finding

3. **[P2] Finish the remaining Pi documentation synchronization** — the implementation now makes doctor inspect all `AGENT_RUNTIME_PROVIDERS` (`apps/cli/src/core/diagnostics/doctor.ts:419-466`), but the normative doctor contract still specifies only Codex and Claude Code at `docs/design/doctor-product-spec.md:210,301-323,506-507`. `DEVELOPMENT.md:8,61,156` and `DEVELOPMENT.zh-CN.md:9,60-61,155` still tell local developers that only Codex/Claude Code can be selected, and `docs/direct-provider-cli.md:64` plus `docs/zh-CN/direct-provider-cli.md:57` retain the same incomplete readiness vocabulary. Update these contracts and synchronized mirrors for Pi, including the three-provider doctor cases and mirror date.

## Verification

On this exact head, `git diff --check` passed; the Shared readiness tests passed (42), the Client Agent Runtime coverage suite passed (416 tests, all four metrics 100%), Server runtime/readiness/registry/account/onboarding tests passed (111), and the CLI/Web onboarding tests passed (71/55). GitHub `Checks` and `Enforce Patch Coverage` also pass. The remaining ownership-gate failure is the repository's maintainer approval requirement, not a test failure.

liuchao-001 (commented):
Re-review of exact head `25df76eaa41c16058e272aca2d0452b58031b4f5` (base `fb607ce6`).

The two P1s from my previous review are fixed. `#restoreSessionState` now stores the fingerprint only once `messageCount !== 0`, so an interrupted first Turn keeps a replaceable binding. The readiness v1 welcome no longer carries `pi`, and a v1-only connection projects Pi as unavailable, not `checking`. The onboarding matrix covers Pi.

I checked the three new findings from the latest review against the code. All three reproduce:

**1. Preparation widens the negotiated provider set (P1): confirmed, blocking.** `ConnectionRegistry.beginPreparation` (`connection-registry.ts:237-257`) unions `runtimeProvider` into `providerReadinessProviders` with no check against what the connection negotiated. `#dispatchSetupPrewarm` (`provider-cli-reconcile-owner.ts:242-291`) only checks the `providerCliPrewarm` capability. On base, `ProviderCliPrewarmFrameSchema.runtimeProvider` is the strict two-value `AgentRuntimeProviderSchema`, so placing a Pi Agent on a readiness-v1 Client sends a frame that Client rejects, and the connection closes. This undoes the v1 freeze from the other direction. The fix is small: refuse `beginPreparation` (and skip `runtimeProvider` in the frame) when the provider is outside the negotiated set, and add a test for Pi placement on a v1-only connection.

**2. Agent API / rollback staging (P1): the gap is real; the shape of the fix is an owner call.** Confirmed: base `ApiClient.listAgents` parses the whole list with the strict `ListAgentsResponseSchema`. One persisted Pi row therefore breaks `agent list` on an old CLI for that Account, and a Server image rolled back past 0041 fails its own response parse. `docs/runtime-protocol.md:95-99` requires an expand/contract and rollback plan once durable data changes. One piece of context: this is the first time the provider enum has grown (`claude-code` was in the original `CREATE TYPE` in 0002), so the repository has no precedent for what that plan should look like. There are two proportionate options. (a) Add a legacy-safe Agent list/read projection before admitting Pi writes. (b) Keep this PR's shape but document an explicit release sequence: Server first, Client channel upgrade before Pi is exposed, and no Server image rollback past 0041 once a Pi row exists. Back that with an old-Client test that pins the failure mode. Either one closes the contradiction with the PR's "no breaking changes" claim. The maintainers should pick one explicitly, not leave it implicit.

**3. Remaining doc sync (P2): confirmed.** `docs/design/doctor-product-spec.md:210,301,506`, `DEVELOPMENT.md:8,61,156` (+ zh-CN), and `docs/direct-provider-cli.md:64` (+ zh-CN) still describe a two-provider world.

No other issues found in the fix delta.

PR #629, opened 2026-09-15
feat: opentag context tree settings management

ggordonhall:
## Context Tree settings management

Context Tree selection is now a dedicated per-Agent settings operation with a single runtime contract. Updated Computers negotiate the required capability before dispatch; unsupported and protocol-v1 Computers are asked to update.

The server commits selections atomically after revision, runtime-configuration revision, placement, and lifecycle checks. Disconnect works while the bound Computer is offline, provided the Agent is paused. Connect and Create run on the bound Computer, with instance-fenced results and repository identity checks.

The client uses stable repository-derived staging trees, bounded 240-second operations, cancellation, serialized connection-store mutations, durable publication intent, and retry identity matching. An uncertain publication is retained for recovery and never automatically creates another repository.

The web settings form resets on Agent changes, rejects stale responses and retries, permits offline Disconnect, and explains recovery after uncertain publication. Legacy Computer-wide configuration, CLI redirect, doctor checks, and preparation records have been removed. Existing trees and memory remain untouched.

The migration was regenerated as 0043 after rebasing on current `main`; the integration migration ratchet is now 44.

## Validation

- `pnpm typecheck`
- `pnpm check`
- `pnpm build`
- Context Tree client/server/shared/web suite: 78 tests
- Affected web regressions: 28 tests
- CLI Context Tree and doctor tests: 45 tests


liuchao-001 (dismissed):
Reviewed at head `616cdf89`. The direction is sound: a per-Agent selection carried in the runtime snapshot, a remote settings operation modeled on runtime-test, a publication-intent record before `publish`, and revision re-checks before applying an async result. A few things need to change before this can land.

## Must fix

**1. Conflicts with main; CI has never run on this head.** #628 already landed `0042_outgoing_jasper_sitwell`, so `0042_cultured_dark_beast` conflicts in `_journal.json`, `0042_snapshot.json` and `migration-hashes.json`. The PR is `CONFLICTING`, so the `pull_request` CI never started. Only the ownership gate has reported. Please regenerate the migration as 0043 on current main, update the drift manifest, and get a green CI run.

**2. "Create private repository" publishes a tree named after the internal setup directory.** `packages/client/src/runtime/context-tree-settings.ts:55` runs `context-tree create --project-path <state>/context-tree-operations/setup-<operationId>`. `create` derives the tree name from the project directory basename. I checked against the real 0.1.15 CLI: the root `NODE.md` gets `title: "setup-2b1f6c1e-8a57-4f55-9a0e-5b0b8f3c0d-context-tree"`, and the managed tree lands at `~/.context-tree/trees/setup-…-context-tree`. The user's new private repo therefore starts with a UUID-derived title, and every create leaves another local managed tree behind. Name the setup directory after the requested repo, or rename/retitle before `publish`.

**3. Rebinding an Agent that has a selection to an older Computer breaks its Sessions.** `EffectiveRuntimeSnapshotSchema` is top-level `.strict()`. The new `contextTreeRepository` key (`packages/shared/src/runtime-domain.ts:137`) is only emitted when non-null, which keeps the null case compatible. The capability check, though, only happens when the settings operation runs. After a rebind (or a Computer downgrade), the server keeps sending the key to a client whose strict schema rejects the whole snapshot, and that client also computes a different config hash. Either gate emission on the placement Computer's `runtime.contextTreeSettings` capability, or block rebinding to a Computer without it while a selection exists.

**4. Doctor lost its Context Tree diagnostics, but the design doc says it has them.** `apps/cli/src/core/context-tree/state.ts` now always returns `tree: "unknown"` with a static message. Meanwhile `docs/design/context-tree-integration.md:88` and `:379` say doctor "reports the Agent's selected repository and the tree's state", and the Deferred section still leans on "durable doctor visibility". Preparation records now go to per-`cwd:target` hashed files (`#recordPreparation`) that nothing reads. `readContextTreePreparation` still reads the old `context-tree-preparation.json`, which is no longer written. Either restore a per-Agent doctor view or correct the doc, and drop the dead reader/writer pair.

**5. No coverage for the new server path.** Nothing tests `ContextTreeOperationService` (pre-dispatch revision check, `pause_required`, the post-result re-read, mapping `AGENT_REVISION_CONFLICT` to `stale_configuration`, rejecting a repository mismatch), `ContextTreeOperationOwner` (busy guard per Computer, timeout, instance mismatch, `close`), the `RuntimeDomainOwner` exclusion of `context-tree:operation:result`, the `updateById` guard that rejects `contextTreeRepository` outside the operation, the `ClientRuntime` frame routing, or the web section. The three client tests cover only `ContextTreeSettings`. These are deterministic, so they belong in Vitest in this PR.

## Should fix / worth a look

- **The settings operation holds the Session-preparation mutex across network calls.** `exclusive: contextTree.runExclusive` (`client-runtime-composition.ts:741`) serializes create → publish → connect → verify → disconnect behind the same queue `ensureAgent` uses. Each call gets the 120 s network timeout, so every Session start on that Computer meanwhile returns `PREPARING` and runs without memory. A lock scoped to the connection store, or at least a doc note, would help.
- **The server timeout is shorter than the client's worst case.** 300 s (`context-tree-operation-owner.ts:55`) against up to five 120 s CLI calls. For `connect`, the timeout reports `computer_unavailable` while the Computer may still finish. The web then discards the operation id, so a retry starts a fresh clone and verify.
- **A `publication_uncertain` record for a repository is permanent on that Computer** (`context-tree-settings.ts:66`). It is keyed by repo hash and has no expiry or clear path, and the UI message doesn't tell the user that **Connect** is the way forward once they confirm the repo exists.
- **`ContextTreeRepositorySchema` lowercases input** (`packages/shared/src/context-tree.ts:9`). That is fine for matching, but `create` then publishes `Owner/MyRepo` as `owner/myrepo`. Keep the typed case for `publish` and normalize only for comparison.
- **Complexity ratchet.** The second commit raises existing ceilings (`agent-service.ts` callback 30 → 35, `#startRuntime` 26 → 27) and adds new exceptions (`context-tree-settings.ts#run` 37, web `run` 24) in the same diff that regenerates unrelated entries. Splitting `#run` into connect / create / publish helpers would avoid the new 37.
- **Nits.** The design doc says `packages/client` has a runtime `dependency` on the package, but it is still a `devDependency`. There's a `.gitignore` typo ("User-specifc"), and it's unclear why `.agents/` should be ignored at the repo root.

The QA tier this needs is live and cross-surface: web settings → Server → Computer → `gh`. That should wait until the migration and CI are sorted.

apps/web/src/features/agents/agent-settings/context-tree-settings.tsx
```diff
@@ -0,0 +1,142 @@
...
+  computerName,
+  online,
+  onChanged,
+}: {
+  config: AgentAdminConfig;
+  computerName: string;
+  online: boolean;
+  onChanged: () => void;
+}) {
+  const [repository, setRepository] = useState(config.runtimeConfig.contextTreeRepository ?? "");
```
yuezengwu, line 28: [P2] Reset operation state when the Agent identity changes

`repository` is initialized only on the first mount. When navigating between cached Agent settings, `AgentConfigSettingsContent` reuses this component without a key, so changing `config.id` updates the request target while leaving the previous Agent's repository in the input. I reproduced this with two suspended Agents A (`review-fixture/tree-a`) and B (`review-fixture/tree-b`): after switching A → B, Connect calls `contextTreeOperation(B.id, { repository: "review-fixture/tree-a", ...BRevisions })`. The server revision checks therefore do not protect against selecting the wrong tree. Key the settings operation state by Agent identity (or reset it safely on identity changes), including the pending/result/retry state, and cover switching between cached configurations.

yuezengwu (changes requested):
Reviewed exact head `616cdf891b00d80e30bdaafb8cdb21e7e234ea0d` in an isolated checkout. Adding a reproduced settings-state bug (inline) and concrete failing-test evidence to the existing review.

The cached-settings regression probe mounts the real `AgentConfigSettingsContent` with two cached Agent configs. A fresh mount for B uses B's repository; switching A → B preserves A's repository and Connect sends it with **B's ID and revisions**. Both the displayed-value and submitted-request assertions fail.

The existing affected suites also fail on this head:

- `pnpm --filter open-tag test src/__tests__/context-tree.test.ts src/__tests__/doctor.test.ts`: **12 failed, 56 passed**. The legacy command/state expectations and doctor remediation assertion were not updated for the new settings-only contract.
- `pnpm --filter @opentag/client test src/__tests__/context-tree.test.ts src/__tests__/context-tree-settings.test.ts src/__tests__/session-runtime-manager.test.ts`: **2 failed, 56 passed**. Both SessionRuntimeManager grant cases still expect two arguments to `ensureAgent`, which now receives a third `null` argument.

Please update those tests to assert the intended new behavior and add the Agent-switch regression. `pnpm check`, `pnpm build`, and `pnpm typecheck` passed; 70 selected server tests and 19 existing web settings tests passed. Full-suite, PostgreSQL integration, and live GitHub/Computer acceptance were not run. These results apply to the PR head, not a merge with current main.

ggordonhall:
Updated after rebasing on current `main` and resolving the migration collision as 0043.

Validation completed:

- `pnpm typecheck`
- `pnpm check`
- `pnpm build`
- Context Tree client/server/shared/web suite: 78 passing tests
- Affected web regressions: 28 passing tests
- CLI Context Tree and doctor tests: 45 passing tests

The branch now includes `06a4941c` and is up to date with its remote tracking branch.

apps/cli/src/__tests__/context-tree.test.ts
```diff
@@ -1,256 +1,6 @@
...
-    sessionStartBudgetMs: 30_000,
-  });
-  const workspace = await temporaryDirectory("opentag-ct-workspace-");
-  await expect(manager.ensureAgent(workspace)).resolves.toEqual({ status: "ready", treePath });
-  const args = ["resolve", "--project-path", workspace, "--json"];
-  const direct = await runContextTreeCli(assets, args, { env: otherEnv });
-  expect(direct.failureCode).toBeUndefined();
-  const { stdout } = await promisify(execFile)(join(manager.binDirectory(), "context-tree"), args, { env: otherEnv });
-  expect(JSON.parse(stdout)).toEqual(direct.payload);
+it("does not expose the removed Computer-wide Context Tree command", () => {
```
bestony, line 4: **[bug]** The Computer-wide `opentag context-tree` command is fully removed (the new test asserts it is absent). That contradicts the design’s required tombstone, and it breaks the still-live portable/offline smoke probe: `scripts/portable/runtime-dependencies.mjs` `runContextTreeRuntimeProbe({ probeOpenTag: true })` still runs `[entryPath, "context-tree", "connect", "otqa-portable-nonexistent"]` and requires stderr `No managed Context Tree named "otqa-portable-nonexistent" exists`. `scripts/portable/smoke-portable-app.mjs` and `scripts/portable/verify-portable-artifact.mjs` both enable that probe, which matches the reported linux/amd64 offline smoke failure. Users who still run the old command also get a generic unknown-command error instead of a pointer to Agent settings.

**Suggestion:** Restore a tombstone `context-tree connect` that prints the Agent-settings migration message and exits non-zero (as `docs/design/context-tree-integration.md` specifies), or update the portable probe and `docs/portable-release.md` to a command that still exists.

packages/client/src/runtime/session-runtime-manager.ts
```diff
@@ -479,8 +488,9 @@ async function prepareContextTree(
   manager: Pick<ContextTreeManager, "ensureAgent"> | undefined,
   cwd: string,
   provider: EffectiveRuntimeSnapshot["provider"],
+  repository: string | null,
 ): Promise<{ promptContext: { contextTree?: ContextTreeStatus }; writableRoots: readonly string[] }> {
-  const status = await manager?.ensureAgent(cwd, provider);
+  const status = await manager?.ensureAgent(cwd, provider, repository);
```
bestony, line 493: **[bug]** `prepareContextTree` now calls `ensureAgent(cwd, provider, repository)`, and the SessionRuntimeManager snapshot fixture was updated to `contextTreeRepository: null`. The existing assertion in `packages/client/src/__tests__/session-runtime-manager.test.ts:322` is still `toHaveBeenCalledWith(cwd, "codex")`. Vitest matches the full argument list, so this fails against `(cwd, "codex", null)`. The previous review’s ensureAgent-arity failure is not actually fixed.

**Suggestion:** Change the assertion to `toHaveBeenCalledWith(cwd, "codex", null)` (and add a case that forwards a non-null selection).

packages/server/src/runtime/context-tree-operation-owner.ts
```diff
@@ -0,0 +1,73 @@
...
+      };
+      this.#pending.set(requestId, { computerId: input.computerId, instanceId, finish });
+      void this.registry
+        .send(input.computerId, instanceId, { ...input, type: "context-tree:operation", requestId })
+        .catch(() => finish({ status: "failed", code: "computer_unavailable" }));
+    });
+  }
+
+  close(): void {
+    for (const pending of this.#pending.values()) pending.finish({ status: "failed", code: "publication_uncertain" });
```
bestony: **[bug]** `close()` finishes every in-flight operation as `publication_uncertain`, including connect/disconnect. The 300s timeout on the same class correctly uses `publication_uncertain` only for `create` and `computer_unavailable` otherwise. A server restart during connect therefore returns the create-only UI copy (“Publication may have completed… Creation will not be repeated automatically.”) for a path that never publishes.

**Suggestion:** Mirror the timeout mapping: `create` → `publication_uncertain`, other actions → `computer_unavailable` (or `failed`).

apps/cli/src/core/diagnostics/doctor.ts
```diff
@@ -148,38 +145,32 @@ export async function runDoctor(options: DoctorOptions = {}): Promise<DoctorResu
...
+    runtimePromise,
+    providerCliPromise,
+  ]);
   const checks: DoctorCheck[] = [
     targetCheck(target),
     ...localChecks(localResult),
     daemonCheck(daemonResult, target.home),
     serverCheck(healthResult, localResult),
     ...runtimeChecks(runtimeResult),
     ...providerCliChecks(providerCliResult),
```
bestony, line 172: **[suggestion]** Doctor no longer emits any `context-tree` checks (the new test at `apps/cli/src/__tests__/doctor.test.ts:696` treats that as intended). The design doc added in this PR still says `opentag doctor` remains the diagnostic surface and “reports the Agent's selected repository and the tree's state under a `context-tree` scope” (`docs/design/context-tree-integration.md:79` and `:379`). Per-Agent selection cannot be diagnosed from the Computer CLI at all now, including dirty/uncloned trees.

**Suggestion:** Either restore non-blocking doctor checks against the bound Agents’ selections (requires Server), or rewrite the design/deferred “durable doctor visibility” claims so they match settings + Session prompts only.

packages/client/src/runtime/context-tree-settings.ts
```diff
@@ -0,0 +1,218 @@
...
+  async #run(frame: ContextTreeOperationFrame, controller: AbortController): Promise<Result> {
+    const budget = this.options.budgetMs ?? 240_000;
+    const work = new AbortController();
+    const totalTimer = setTimeout(() => controller.abort(), budget);
+    const workTimer = setTimeout(() => work.abort(), Math.max(1, budget - Math.min(10_000, budget / 10)));
+    const signal = AbortSignal.any([controller.signal, work.signal]);
+    try {
+      return await this.#execute(frame, signal, controller.signal);
+    } catch {
+      return failed(frame.input.action === "create" ? "publication_uncertain" : "failed");
```
bestony: **[suggestion]** Any thrown error during `create` (budget abort while still queued, `hasAgentSessions` throwing “Agent runtime resumed”, CLI failures before the publication fence is written) is mapped to `publication_uncertain`. The durable publication lock is only written later, inside `#publish`, so this over-classifies work that never reached GitHub. The same mapping exists in `#execute` and in `ClientRuntime`’s frame catch (`packages/client/src/runtime/client-runtime.ts:209`). The UI then tells the operator the repo may already exist and will not be created automatically.

**Suggestion:** Return `publication_uncertain` only after the publication intent file is written (or after `publish` starts). Earlier create failures should stay `failed` / `busy` / `authentication_required`.

bestony (commented):
## Summary

This re-review of PR #629 (head `06a4941c`) covers per-Agent Context Tree settings: a new `context_tree_repository` column (migration `0043`), runtime snapshot field, settings UI, and Computer-side create/connect/disconnect. Several earlier must-fixes are addressed: Agent-switch state is keyed and tested, create publishes the requested `OWNER/REPO` rather than a `setup-<operationId>` name, `updateById` cannot overwrite the selection, snapshots always include `contextTreeRepository` and `runtime.contextTreeSettings` is a required capability, and ContextTreeOperationService/Owner plus the web section have tests. Remaining risk is operational, not schema: the Computer-wide CLI command was deleted instead of tombstoned (this breaks portable/offline smoke), `ensureAgent`’s new arity is not reflected in SessionRuntimeManager tests, server shutdown mis-labels in-flight work as `publication_uncertain`, and doctor/design still disagree. I could not find a remaining `configFile` use on this tree (the previous typecheck failure looks already removed).

## Issue counts by severity

- bugs: 3
- suggestions: 2
- nits: 0

packages/server/src/runtime/runtime-session.ts
```diff
@@ -209,14 +209,10 @@ export class RuntimeSession {
         this.#fail("PROTOCOL_ERROR", "A runtime handshake request is already in progress", 4400);
         return;
       }
-      if (
-        envelope.data.type === "auth" &&
-        envelope.data.protocolVersion !== RUNTIME_PROTOCOL_V1 &&
-        envelope.data.protocolVersion !== RUNTIME_PROTOCOL_V2
-      ) {
+      if (envelope.data.type === "auth" && envelope.data.protocolVersion !== RUNTIME_PROTOCOL_V2) {
```
yuezengwu, line 212: [P2] Migrate runtime test fixtures with the v2-only handshake

This change rejects protocol v1 before authentication, but the existing Client runtime/composition/domain fixtures and Server runtime/observability fixtures still establish v1 sessions. For example, `pnpm --filter @opentag/server exec vitest run --maxWorkers=1 src/__tests__/runtime.test.ts -t 'rejects Account access and refresh tokens'` now receives `PROTOCOL_VERSION_UNSUPPORTED` instead of `AUTH_INVALID_TOKEN`, so it never exercises the credential check. The Server unit suite has 24 failures, and the Client fixtures additionally reject with `Update the Server: required Context Tree support is unavailable`; the current Node 24 CI log shows the same problem. Update the positive-path fixtures to v2 with the required capability, and replace obsolete fallback-success assertions with explicit legacy-rejection coverage. Otherwise the standard suites stay red and the existing authentication, fencing, and shutdown tests no longer reach their intended paths.

packages/client/src/__tests__/agent-home-context.integration.test.ts
```diff
@@ -59,15 +58,23 @@ describe("shared Agent Home and Context Tree", () => {
...
     expect(treePath.startsWith(`${fixture.accountHome}/`)).toBe(true);
 
-    const layout = resolveContextTreeHome(environment);
-    await mkdir(layout.directory, { recursive: true });
-    await writeFile(
-      layout.configFile,
-      `${JSON.stringify({ schemaVersion: 1, target: { kind: "path", path: treePath } })}\n`,
+    await isolatedExecFile(
+      "git",
+      ["config", "--global", `url.file://${treePath}.insteadOf`, "https://github.com/acme/memory.git"],
```
yuezengwu: [P2] Preserve a GitHub origin identity in the offline fixture

The new global `url.file://…insteadOf` redirect makes the packaged Context Tree CLI reject `connect acme/memory`: an isolated reproduction returns `CONTEXT_TREE_FAILED` with `Context Tree origin must identify a credential-free GitHub OWNER/REPO repository.` Consequently line 77 dereferences a missing `payload.tree`, and the real shared-Agent-Home regression fails locally and in the Node 24 compatibility job before testing workspace isolation or Tree writes. Use an offline transport fixture that preserves the GitHub-shaped origin (as the updated `context-tree.integration.test.ts` already does), and assert the connect result before reading its tree path.

yuezengwu (changes requested):
Re-reviewed `06a4941cdd51ae06835f9d8a47aeecd18477c275` in a clean detached worktree. The cached-Agent form-state problem from my previous review is fixed: the form is keyed by Agent ID, and the three settings regression tests pass.

Requesting changes for two additional reproducible test regressions below. These are separate from the `ensureAgent` argument mismatch and portable command probe already covered in [bestony's current-head review](https://github.com/first-tree-ai/opentag/pull/629#pullrequestreview-5218717513); I have not duplicated those comments.

Validation on Node 24.20.0 / pnpm 10.12.1:
- `pnpm check`, `pnpm build`, and `pnpm typecheck`: passed.
- Web package: 1,043 tests passed. CLI Context Tree/doctor tests: 45 passed.
- Client package: 1,283 passed, 14 failed, plus 8 unhandled errors. Server unit package: 963 passed, 24 failed.
- Both inline findings were reproduced separately. The current [Node 24 compatibility job](https://github.com/first-tree-ai/opentag/actions/runs/35057361350/job/104670220681) also reports the legacy-protocol failures and the Agent Home fixture failure.

This does not include PostgreSQL integration, live GitHub publication, or a combined-current-main test. Please fix the failing contracts and validate the successor head before approval.

packages/server/src/runtime/context-tree-operation-owner.ts
bestony: **[bug]** `close()` finishes every in-flight operation as `publication_uncertain`, including connect and disconnect. Timeout mapping on the same class is already action-aware (`create` → `publication_uncertain`, otherwise `computer_unavailable`), and send failures correctly use `computer_unavailable`. Server shutdown therefore tells a connect/disconnect user that GitHub publication may have completed and must not be retried automatically. The owner test currently asserts this wrong close mapping for a connect.

**Suggestion:** Store the pending action next to `finish`, and map close the same way as the timeout: `publication_uncertain` only for `create`, `computer_unavailable` (or `failed`) for connect/disconnect.

packages/client/src/runtime/context-tree-settings.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  async #execute(frame: ContextTreeOperationFrame, signal: AbortSignal, cleanupSignal: AbortSignal): Promise<Result> {
+    if (frame.requireStopped && this.options.hasAgentSessions(frame.agentId)) return failed("busy");
+    const run = this.#runner(frame, signal);
+    if (!run) return failed("capability_missing");
+    try {
+      return await this.#executeRecorded(frame, run, cleanupSignal);
+    } catch {
+      return failed(frame.input.action === "create" ? "publication_uncertain" : "failed");
```
bestony: **[suggestion]** `#execute` (and the outer `#run` catch at line 77, plus `ClientRuntime` at `packages/client/src/runtime/client-runtime.ts:209`) maps any thrown error during `create` to `publication_uncertain`. The durable publication fence is written only later in `#publish`. Pre-fence failures therefore produce the “check GitHub; creation will not be repeated” result even though publish never started.

**Suggestion:** Return `failed` (or the classified CLI code) unless the publication fence file already exists for that repository. Keep `publication_uncertain` for post-fence timeouts, aborts, and lost responses.

docs/design/context-tree-integration.md
```diff
@@ -352,18 +365,19 @@ particular reports an unusable tree as `ok: false` with `findings` and no `error
...
+- A success is cached per workspace, repository, and provider. A failure is cached for a one-minute
+  cooldown, after which a later Session retries; changing the selection invalidates both caches
+  immediately.
 - Session start waits at most five seconds. On budget expiry it reports `PREPARING`, starts without
   durable memory, and leaves the joined, serialized preparation running in the background.
 - The managed prompt tells the Agent durable memory is inactive and not to repair the tree or
   create one itself.
-- `opentag doctor` reports the configured target and the tree's state under a `context-tree`
-  scope. Both checks are non-blocking, so neither can change the doctor exit code. There is no
+- `opentag doctor` reports the Agent's selected repository and the tree's state under a
```
bestony: **[suggestion]** Doctor checks were intentionally removed. The design doc still says `opentag doctor` reports the Agent’s selected repository and tree state under a `context-tree` scope (also lines 79 and 88). It also still describes the Computer-wide `opentag context-tree connect` command as a tombstone that points at Agent settings and exits non-zero, but `createProgram()` no longer registers that command at all.

**Suggestion:** Update the design doc to match the implementation: doctor does not report Context Tree; the Computer-wide command is removed rather than tombstoned.

packages/client/src/runtime/session-runtime-manager.ts
```diff
@@ -277,7 +277,12 @@ export class SessionRuntimeManager implements RuntimePreparation, RuntimeLocalPo
     // runs on every Turn admission, and this runs once per Provider Runtime start. The manager
     // caches per workspace, revalidates that entry against the Computer's recorded target, and
```
bestony: **[nit]** The comment next to the new per-Agent `contextTreeRepository` argument still says the manager “revalidates that entry against the Computer's recorded target.” Preparation now keys on the Agent snapshot repository and provider; the Computer-wide target file is ignored.

**Suggestion:** Reword the comment to the snapshot repository (and provider), or delete it if it only restates the call below.

bestony (commented):
## Summary

The re-review of PR #629 (`ffb3c57c` vs `b2fcc741`) confirms the earlier portable-probe, `ensureAgent` arity, protocol-v2 fixture, and GitHub-origin `insteadOf` issues are fixed. Context Tree settings still have two real mismatches: server shutdown reports `publication_uncertain` for every in-flight action, and create still maps any thrown error to that code before the publication fence exists. The design doc still disagrees with the intentional doctor/CLI removals.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 2

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[nit]** apps/cli/src/core/diagnostics/doctor.ts:244 -- After removing Context Tree checks, `renderDoctorReport` still always prints a `Context Tree` section heading. With no checks in that scope the rendered report contains an empty section between IM Provider CLIs and Summary.
  - **Suggestion:** Drop the `["Context Tree", "context-tree"]` heading (and the unused `DoctorCheckScope` variant) unless checks return.

bestony (commented):
Re-checked head `363e2ce`. The only delta from `ffb3c57c` is `packages/client/src/__tests__/client-runtime-domain.test.ts`: the domain fixture now replies to `heartbeat` instead of treating it as an unexpected frame. That is a valid test fix and does not change product behavior.

The remaining findings from [the previous current-head review](https://github.com/first-tree-ai/opentag/pull/629#pullrequestreview-5219189401) still apply on this SHA:

- **bug** `ContextTreeOperationOwner.close()` still finishes every in-flight operation as `publication_uncertain`, including connect/disconnect (`packages/server/src/runtime/context-tree-operation-owner.ts:71`).
- **suggestion** create still maps any thrown error to `publication_uncertain` before the publication fence exists.
- **suggestion** design doc still describes doctor checks and a tombstone `opentag context-tree` command that this PR removed.

Not approving on this head.

yuezengwu (commented):
Re-reviewed `363e2ce02193a7307723a3e4caf0c7c46119063f` against `b2fcc741` in a clean detached worktree. Both findings from my previous review are fixed: the runtime fixtures now use the v2 contract, and the Agent Home fixture redirects only Git transport while preserving the GitHub origin identity. The earlier cached-Agent form-state issue also remains fixed.

Fresh local validation passed: `pnpm check`, `pnpm build`, `pnpm typecheck`; 174 Client tests (including runtime/composition/domain, Context Tree, Agent Home, SessionRuntimeManager and Runner acceptance), 106 Server tests, 23 Web tests, 45 CLI tests, and 19 portable dependency tests.

I independently rechecked the remaining error-classification paths and am withholding approval pending the issues already documented in [bestony's current-head review](https://github.com/first-tree-ai/opentag/pull/629#pullrequestreview-5219207519). No duplicate inline findings added.

Validation was targeted; this is not a full repository, PostgreSQL integration, or live GitHub publication acceptance run.

bestony (approved):
Re-reviewed `aa524d73` against the remaining findings from [the previous current-head review](https://github.com/first-tree-ai/opentag/pull/629#pullrequestreview-5219207519).

All of those are addressed:

- `ContextTreeOperationOwner.close()` now stores the pending action and maps `create` → `publication_uncertain`, connect/disconnect → `computer_unavailable`, matching the timeout path. Covered by `classifies shutdown during %s`.
- Create/connect throws before the publication fence return `failed` and remain retryable; only post-fence publish failures stay `publication_uncertain`. Covered by the settings retry tests and the ClientRuntime domain test.
- Design docs and doctor output now match the intentional removal of Computer-wide CLI and Context Tree doctor checks. The empty `Context Tree` heading is gone, and the stale Computer-target comment is reworded.

Approving this head.

yuezengwu (approved):
Re-reviewed `aa524d73a766827be131755ce1642f340c0dc0a2` in a clean detached worktree. The previously raised findings are fixed, including shutdown classification and errors before the publication fence. I found no additional blocking issue. Approving this head.

Fresh local validation on Node 24.20.0 / pnpm 10.12.1:
- `pnpm check`, `pnpm build`, and `pnpm typecheck`: passed.
- `pnpm test`: 376 root-script tests and 4,154 package tests passed (Shared 245, Client 1,374, Server 989, Web 1,043, CLI 503).
- Agent Runtime coverage: 423 tests passed; statements, branches, functions, and lines all 100%.
- PostgreSQL integration: 350 tests passed across 20 files.

The current base `b2fcc741` is already an ancestor of this head. Live GitHub repository publication was not exercised.

liuchao-001 (changes requested):
Re-reviewed at exact head `aa524d73`, rebased on current `main`, CI green. Four of my five earlier must-fixes are genuinely resolved:

- migration regenerated as `0043` on current main, conflict gone, full CI green on this head;
- `create` now publishes the requested `OWNER/REPO` from a deterministic repository-derived staging tree instead of `setup-<operationId>`;
- doctor no longer claims Context Tree diagnostics, and both `context-tree-integration.md` and `doctor-product-spec.md` were updated to match the removal;
- `ContextTreeOperationService` / `Owner`, the client settings path, and the web section now have tests, and the Agent-switch state bug yuezengwu found is keyed and covered.

One item is not fixed so much as traded for a larger problem, and I don't think it can ship as written.

## Blocking: this turns an optional feature into a forced upgrade for every existing Computer

My earlier point was narrow: a strict snapshot schema plus a new key breaks an older Computer after a rebind. This revision resolves it by making the snapshot field always present (`effective-runtime-snapshot-assembler.ts:118`) and then making the capability mandatory on both sides:

- `packages/shared/src/runtime-protocol.ts:68-69` — `runtime.contextTreeSettings` becomes the first entry ever in `RUNTIME_REQUIRED_CLIENT_CAPABILITIES` and `RUNTIME_REQUIRED_SERVER_CAPABILITIES`; both were empty before.
- `packages/shared/src/runtime-protocol.ts:21` — `RUNTIME_SUPPORTED_PROTOCOL_VERSIONS.min` moves from v1 to v2, dropping frozen v1 entirely.

Traced to the observable end, on the Computer that is already installed today:

1. The Server refuses the handshake at `runtime-session.ts:360-371` with `PROTOCOL_CAPABILITY_UNSUPPORTED` and closes 4400.
2. The **currently released** client (verified against `origin/main`, not this branch) treats a close code in `[4400,4500)` as a fatal `RuntimeConnectionError` (`runtime-connection.ts:783`).
3. `isExpectedDaemonStop` (`apps/cli/src/core/daemon/runtime.ts:316`) classifies a fatal connection error as an expected stop, so the daemon exits quietly, logging `Daemon connection was rejected; run opentag connect again` — which points at the wrong remedy. Re-running `connect` cannot add a capability to an installed build.
4. The auto-updater cannot rescue it. It upgrades only on a Server-advertised channel target (`update-manager.ts:142`, wired through `onChannelTarget`), and that target only arrives over an established connection. `checkIntervalMs` is the protected-work wait loop, not an independent version poll. A Computer rejected at handshake never learns a newer version exists.

So the moment this Server deploys, every Computer that has not already upgraded goes offline, cannot self-update, and shows no reason in the web beyond "offline". The one string that would explain it, `agent_settings_context_tree_capability_missing` ("Update OpenTag on the bound computer…"), is unreachable, because it renders inside settings for a Computer that can still connect.

This also contradicts two documents in the tree:

- `docs/runtime-protocol.md:7,13,39-44` keeps v1 "frozen for rolling upgrades", documents the v1/v2 compatibility matrix and the v1 fallback, and states the rollout rule "Roll out Server v2 before Client v2". This PR invalidates that and does not update the doc.
- This PR's own design doc still says at line 98-99 that older Computers "report `capability_missing` and the settings page asks the user to update OpenTag there" — that is the graceful behavior the first revision had, and it is now unreachable.

**The feature does not need any of this.** The dispatch-time check in `context-tree-operation-owner.ts:45` already returns `capability_missing` for a Computer that cannot run the operation, and the UI copy already exists. Keeping `contextTreeRepository` omitted when null (as the first revision did) preserves byte-identical snapshots and hashes for every Agent without a selection, which is all of them on the day this deploys. My original concern was only the rebind case, and that is far better answered by gating emission on the placement Computer's capability, or refusing to rebind a selected Agent onto a Computer that lacks it, than by requiring a fleet-wide upgrade.

Please drop the two `runtime-protocol.ts` constant changes from this PR. If retiring protocol v1 is wanted on its own merits, it deserves a separate PR with a rollout plan, a `docs/runtime-protocol.md` update, and an operator-facing message that names the actual remedy.

## Non-blocking

- The staging tree still carries a synthetic root title: `create` derives the tree name from the staging directory, so the published repository's root `NODE.md` is titled `owner-repo-<hash12>-context-tree`. The repository name itself is now correct, so this is cosmetic, but it is the first thing a user sees in their new tree.
- `daemonOperatorMessage` (`apps/cli/src/core/daemon/runtime.ts:330`) returns the same "run opentag connect again" text for every fatal connection rejection. Even with the protocol change reverted, a capability or protocol rejection deserves a distinct message telling the operator to update OpenTag.

Everything else I raised earlier looks good, and the operation state machine reads considerably tighter than in the first revision.

PR #630, opened 2026-09-15
feat(client): add pinned Linux Cloud Runner toolchain

yuezengwu:
## Summary

Add a Client-owned Linux amd64 Runner image that later Cloud execution work can consume with a fixed image identity. The Runner and bundled CLI share the existing release coordinate and source SHA.

- Pin Node, Pi and its npm closure, Git, gh, and the existing Slack/Lark catalog; assemble six Context Tree and four source-owned tool skills for a non-root runtime.
- Stage only Git-tracked source files within the build allowlist. Ignored/untracked logs cannot enter the image; tracked development edits and intent-to-add source files remain supported, with symlink and credential-path guards.
- Verify every dependency edge at every planned output location before copying the Client dependency closure. Reject unsupported depth and version shadowing, including one physical package placed under multiple parents.
- Add `cloud-computer.mjs runner-toolchain` for disconnected offline checks and real Pi file read/write, same-session resume, active child cancellation, and cleanup evidence. Preserve E1/E2 dispatch. Keep the real-mode guard alive with Docker `--init` to reap orphaned children and `sleep infinity` until cleanup; the acceptance command has a separate 30-minute timeout.
- Match exact first-line tool versions, including grouped, anchored Slack/Lark catalog patterns; reject leading/trailing wrappers while allowing subsequent help lines.
- Inject filtered runtime configuration. Rebuild `models.json` from only `models` and `providers`, drop unknown top-level fields, and reject malformed shapes. Preserve selected-provider content and strict file permissions. Invalid JSON errors contain only the filename; redact quoted JSON secrets and complete Authorization/Bearer values.
- Bind each Node version to its reviewed base-image digest; add build/offline CI and synchronized English/Chinese documentation. This PR does not publish images or implement Cloud registration/session APIs.

## Testing

Current clean HEAD: `b8987035efa0db5f32684706bed78335db838851`, based on E2 main `ee1267bbc19540a736b386def394dd3e0b6d0e72`.

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — passed: 376 script tests; Shared 248, Client 1369, Server 977, Web 1040, CLI 526. Unchanged inputs may use Turborepo cache hits.
- Focused Runner script tests — 43 passed. The guard regression verifies Docker `--init`, an indefinite keepalive, the real entrypoint with a controlled sleep shim, resource caps, and confirmed removal. Client test inputs are unchanged in this increment; the full suite may reuse Turborepo cache results.
- Before/after real-model diagnostic on the preceding image: a guard without init retained orphaned Bash/sleep zombies after cancellation; adding `--init` reaped them and passed the same acceptance. Process disappearance checks remain strict. New-head images below exercise the corrected harness.
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 416 tests; 100% statements, branches, functions, and lines.
- `pnpm --filter @opentag/server test:integration` — 350 tests across 20 files passed.
- Clean dev `0.0.5` image: `sha256:ec6c2ed638430965787864c2fceef559a49e7fa33d36f4398e8677a9369c6705` (592,311,006 bytes).
- Clean prod `0.0.5` image: `sha256:4d31010bae8b270a830c329c685a1cd4cea05f0334b352c686b4972f5219461f` (592,317,899 bytes).
- Both images passed disconnected offline smoke and real DeepSeek acceptance, using a filtered copy of the local Pi configuration: exactly ten skills, actual file read/write, same-session resume, and cancellation of three observed live processes including one tracked Pi process. All tracked processes disappeared; no delayed write occurred. Each channel recorded two successful read/write tool events.
- Dev: fresh Runner startup 759 ms; first real file task 4,006 ms; real acceptance 24,890 ms; peak memory 276,885,504 bytes. Prod: startup 751 ms; first real file task 3,440 ms; acceptance 22,547 ms; peak memory 287,596,544 bytes. These timings describe separate phases.
- Independent real PID-1 SIGTERM runs on both images observed a completed file task and live Pi RPC before stopping the container: dev 166 ms, prod 155 ms; exit 143, no OOM, container PID 0, and confirmed removal.
- All Docker checks used Linux amd64, the non-root runner, 1 CPU / 1 GiB and matching swap. Actual guard inspection confirmed Docker init owns PID 1 and runs the indefinite keepalive. Synthetic configuration filtering and catalog banner controls also passed in both images.
- Temporary configuration and acceptance containers removed; original local Pi configuration verified byte-for-byte unchanged; source checkout clean. Exact selected-credential scans found zero matches in 1,309 tracked source files and the acceptance artifacts. The complete image-layer credential scan remains historical; these images were built before runtime credentials were copied.

All technical CI passed on the first run for this exact HEAD: [CI](https://github.com/first-tree-ai/opentag/actions/runs/34977909475) (including Node 22.22.2 / 24 / 26 and PostgreSQL integration), [Runner Toolchain](https://github.com/first-tree-ai/opentag/actions/runs/34977909635), [Docker](https://github.com/first-tree-ai/opentag/actions/runs/34977909806), [Patch Coverage](https://github.com/first-tree-ai/opentag/actions/runs/34977909908) (868/914 changed executable lines, 94.97%; threshold 80%), and [Quality Scoreboard](https://github.com/first-tree-ai/opentag/actions/runs/34977909713). Bestony [approved this HEAD](https://github.com/first-tree-ai/opentag/pull/630#pullrequestreview-5210915966), and ownership-gate passed for all 69 changed files.

Local Docker runs on Apple Silicon with amd64 emulation. No native Cloud Run/Sandbox, network-policy, or authenticated IM delivery acceptance is claimed. Local image IDs are not published registry digests. Staging release rewriting remains unit-tested. Real mode uses the fixed configured DeepSeek model; the harness's `--init` does not settle the separate init/orphan-reaping decision for future long-lived E3 production sessions.

## Breaking changes

None. No Server/shared schema changes relative to current main.

## Checklist


liuchao-001 (approved):
Reviewed at head `5e93319f`: Dockerfile and build-context staging, the runtime closure assembly, pins, the e2e harness (offline/real/cleanup), the in-client Runner (`cli`/`acceptance`/`config`/`probe`/`processes`/`signals`/`skills`/`tree`), the workflow, and the docs.

I didn't find anything that blocks the scope as stated. The allowlisted context with symlink/credential rejection, verified gh and catalog downloads, a Pi lockfile that resolves only from `registry.npmjs.org`, credentials copied in only after filtering, and cleanup that checks containers are really gone all look solid. Approving, with the findings below. I'd like #1 fixed in this PR, but it doesn't block.

**1. Should fix: `collectInstalledClosure` places second-level nested packages under the wrong parent.** See `scripts/runner/runtime-closure.mjs:89-90`.
The nested slot is keyed by the referrer's *package name*, and the destination is `nodeModulesDir/<parentName>/node_modules/<name>`. If the referrer is itself a nested copy, the dependency is installed under the *top-level* package with that name, not under the nested copy. I checked this with a fixture: `app → b@1, c, d@1`; `c → b@2`; `b@2 → d@2`. The run completes without an error. It writes `node_modules/b/node_modules/d@2` (under b@1, which doesn't use it), and the nested `node_modules/c/node_modules/b` then resolves `d` to **1.0.0**. So a wrong version ships without any error, even though the comment says "a genuine slot conflict fails closed".
The real client graph at this head doesn't hit this. It has a single one-level nest (`thread-stream > real-require@1.0.0`), which is placed correctly. It is one lockfile bump away, though. Simplest fix: fail closed when the referrer is itself a nested entry, which is what `collectRuntimeDependencyClosure` in `scripts/portable` does for the flat case. The alternative is to key slots by the referrer's assembled path. A regression test with the fixture above would lock it in.

**2. Nit: the client public API grows by 13 exports, but only `linuxAmd64ProviderCliPlans` has an external consumer** (`scripts/runner/docker-install-provider-clis.mjs`). The Runner tests import everything else through relative `../runner/*` paths. Exporting `runRunnerAcceptance`, `redactAcceptanceRecord`, `copyIsolatedPiConfig`, etc. from `packages/client/src/index.ts` makes the acceptance harness, with its hardcoded test model and fixture prompts, a public Client contract. Suggest exporting only the catalog plan function.

**3. Nit/question: the only real-mode model is `deepseek-v4.1-flash-expires-on-0910`** (`packages/client/src/runner/acceptance.ts:22`). The expiry date in the id has already passed. Once the endpoint retires it, `--mode real` fails as a model error and no flag lets you override it. Consider a non-expiring id, or a `--model` override that is still validated against the selected provider.

**4. Nit: the Node image digest isn't tied to `node-version.txt`** (`scripts/runner/pins.mjs:25-27`). The tag is derived from `node-version.txt` but the digest is a constant, and Docker ignores the tag when a digest is present. A Node bump would silently build on the old image, and only the offline `probe:node` mismatch would catch it, indirectly. The `IMAGE_TAG !==` check on line 27 compares the value with itself and can never fail. A small version→digest map that throws on a missing entry would make the bump error explicit.

**5. Nit: dead or duplicated code.** Unused exports: `applyReleaseCoordinate`, `installGh`, `readStagedIdentity`, `cleanupNow`, `registerImage`, `registerContainer`, `countActiveChildren`, `contextTreePackageOrThrow`, `removeIsolatedPiConfig`.
Duplication:
- `runner-identity.json` is written twice, in `stageRunnerBuildContext` and again in `buildRunnerImage`.
- Channel/version validation exists twice: `resolveBuildVersion` in `build-image.mjs` and `assertRunnerReleaseVersion` in `identity.ts`.
- The Pi config filter is implemented twice with different semantics. The host guard rejects other providers; `copyIsolatedPiConfig` silently drops them.

**6. For E3, not this PR:** the image has no init. PID 1 is `node` (or `sh` for exec'd commands), and neither reaps orphaned grandchildren from Pi tool runs. That's fine for disposable acceptance containers, but worth adding `tini` (or an equivalent) before long-lived sessions run on this image.

**Coverage note:** CI and the evidence in the PR description exercise only `--channel dev`. The staging/prod paths (`prepare-cli-release` rewrite inside the build stage, staging-version probe pattern) are unit-tested but haven't been built as images.

liuchao-001 (approved):
Re-reviewed the incremental diff `5e93319f..35bde414`. Approving this head.

- **#1 (nested closure) is fixed.** A conflict under a nested referrer now fails closed. I re-ran my original fixture (`app → b@1, c, d@1`; `c → b@2`; `b@2 → d@2`) against this head with two root orderings, and both now throw `needs a nested slot under nested referrer b (c/b)` instead of misplacing `d@2`. The real client graph still has only a one-level nest under a top-level referrer, which stays supported, and the Runner image smoke is green.
- **#4 (Node digest) is fixed.** `REVIEWED_NODE_IMAGE_DIGESTS` makes a `node-version.txt` bump without a reviewed digest fail before any build, and a drifted digest is rejected too. The tautological check is gone.
- **#3, #6 and the coverage note** are now called out in the PR description, and the prod image was also built and passed offline smoke. That's fine for this scope.
- **#2 (public exports) and #5 (dead/duplicated code)** remain. Both are non-blocking and fine as follow-ups.

The only non-green check is `ownership-gate`, which is waiting on bestony.

scripts/e2e/runner-toolchain/harness.mjs
```diff
@@ -0,0 +1,195 @@
...
+  allowFailure = false,
+  timeoutMs,
+}) {
+  registerRemovalBackstop(name);
+  return docker(["run", "--name", name, ...LIMITED_ARGS, ...extra, image, ...args], { allowFailure, timeoutMs });
+}
+
+export async function startGuardContainer({ image, name }) {
+  registerRemovalBackstop(name);
+  await docker(["run", "--detach", "--name", name, ...LIMITED_ARGS, image, "sh", "-c", "sleep 600"]);
```
bestony: **[bug]** Real-mode acceptance keeps the credential container alive with `sh -c "sleep 600"` as PID 1 (`startGuardContainer`), while `runRealSmoke` runs `opentag-runner accept` via `docker exec` with `timeoutMs: 1_800_000` (30 minutes). When PID 1 exits, Docker stops the container and kills the exec'd accept. Fixture + resume already budget two 180s model timeouts plus a 90s live-child wait, so a slow DeepSeek run can exceed 10 minutes and fail even though the harness contract allows 30. Offline CI never hits this path, and no test asserts the guard lifetime against the accept timeout.

**Suggestion:** Keep the guard container alive for at least the accept timeout (for example `sleep 1800` or `sleep infinity`), or make the long-running process `opentag-runner accept` itself (PID 1) after config injection. Add a unit assertion that the keep-alive duration is >= the real-mode exec timeout.

packages/client/src/runner/probe.ts
```diff
@@ -0,0 +1,139 @@
...
+ * A probe is ok only when the first output line matches the tool's anchored pattern and the
+ * captured version equals the expected one exactly (`0.84.20` must never satisfy `0.84.2`).
+ */
+function evaluateVersionProbe(
+  name: string,
+  detail: string,
+  expected: Readonly<Record<string, string>> | undefined,
+  catalogPattern?: string,
+): boolean {
+  const pattern = catalogPattern !== undefined ? new RegExp(catalogPattern) : TOOL_VERSION_PATTERNS[name];
```
bestony: **[suggestion]** Comments and docs say tool versions use anchored patterns and never substring matches. `TOOL_VERSION_PATTERNS` are `^…$`, but catalog Slack/Lark probes compile `plan.probes.versionPattern` with `new RegExp(catalogPattern)` and `RegExp.exec` on the first line. Those catalog strings (`Using slack v(…)`, `lark-cli version (…)`) are unanchored, so a first line that merely contains the pin still passes. Existing tests catch `slack.bin` because that breaks the literal `Using slack v` prefix, not a leading/trailing wrapper.

**Suggestion:** Compile catalog version patterns as `new RegExp(^${catalogPattern}$)`, or store already-anchored patterns next to `SURFACE_PATTERNS`, and add a case such as `prefix Using slack v4.7.0`.

scripts/e2e/runner-toolchain/pi-config-guard.mjs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function filterAuth(name, parsed, provider) {
+  const filtered = filterProviderKeys(name, parsed, provider);
+  if (Object.keys(filtered).length === 0) fail(`auth.json has no entry for the selected provider ${provider}`);
+  return filtered;
+}
+
+function filterModels(name, parsed, provider) {
+  const filtered = { ...requireObject(name, parsed) };
```
bestony: **[suggestion]** Host-side real-mode staging is described as a narrow filtered copy (whitelist basenames, selected provider, safe settings keys). `filterSettings` allowlists keys, but `filterModels` does `const filtered = { ...requireObject(name, parsed) }` and only constrains `models[].provider` and `providers`. Any extra top-level field on `models.json` (env maps, inline keys, unrelated provider blobs outside `providers`) is copied into the container unchanged. Client `filterModelsDocument` in `packages/client/src/runner/config.ts` has the same spread.

**Suggestion:** Copy only known models.json fields (`models`, `providers`) the same way settings keys are allowlisted, and reject or drop the rest. Add a fixture with an extra secret-bearing key and assert it never appears in the staged file.

bestony (commented):
## Summary

Client-owned linux/amd64 Runner image is carefully scoped: allowlisted staging, digest-pinned Node, checksum-pinned gh/Slack/Lark, frozen Pi npm closure, non-root runtime, and a read-only CI workflow that only builds and runs offline smoke. The two claimed incremental fixes check out: nested closure under a nested referrer fails closed, and a Node pin without a `REVIEWED_NODE_IMAGE_DIGESTS` entry is refused before build.

Dominant remaining risk is the real-mode Docker harness, not image construction. I am not approving: `--mode real` can die at 10 minutes while the accept timeout is 30 minutes.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

bestony (commented):
## Re-review of `35bde414..993a1ef1`

New commit `fix(runner): close adversarial build and credential gaps` is a separate security tightening, not a response to the previous review threads.

**Previous findings are still open.** I am not approving this head.

1. **[bug, still blocking]** `scripts/e2e/runner-toolchain/harness.mjs` still uses `sleep 600` as PID 1 while `runRealSmoke` execs accept with a 30-minute timeout. That file is unchanged.
2. **[suggestion]** Catalog Slack/Lark `versionPattern` is still compiled unanchored in `probe.ts`.
3. **[suggestion]** `filterModels` / `filterModelsDocument` still spread extra top-level `models.json` fields; `pi-config-guard.mjs` and `config.ts` were not changed on that axis (config.ts only stopped leaking raw JSON parse fragments).

**This increment looks correct** and I would not add new blocking issues:

- Build context now stages only `git ls-files` (index / intent-to-add), fails closed when Git ownership cannot be established, and still rejects credentials/symlinks first. Tests cover ignored/untracked skip, `-N` intent-to-add, dirty tracked content, and missing-git failure.
- Closure now records every placement of one physical package and verifies the planned Node ancestor chain before copy, including the multi-parent shadowing case (`c/a` vs `d/a`). Fail-closed tests match the fixtures.
- Malformed Pi config errors report the filename only; acceptance redaction covers quoted JSON secret fields and Authorization/Bearer values, with tests that the canary never leaks even as a prefix.

Please keep the guard container alive for at least the accept timeout (or make accept PID 1) before asking for approval.

bestony (approved):
## Re-review of `993a1ef1..035a5b43`

The three previous findings are addressed on this head. Approving.

1. **Guard lifetime:** PID 1 is now `sleep infinity` via the image entrypoint (`exec "$@"`), not `sh -c "sleep 600"`. `REAL_ACCEPT_TIMEOUT_MS` is exported and the stub test asserts keepalive >= that budget, plus that the real `entrypoint.sh` can exec the command (a shell builtin would fail).
2. **Catalog probes:** `catalogLinePattern` wraps reviewed patterns as `^(?:…)$`. Leading/trailing wrappers on Slack/Lark banners fail; a matching first line with extra later lines still passes.
3. **models.json:** Host and Client rebuild from `models` and `providers` only. Unknown top-level fields (including a synthetic `apiToken`) are dropped; malformed shapes fail with fixed messages and no source fragments.

No new blocking issues in this increment. Remaining nits from earlier (public Client exports, dead/duplicated helpers, no tini for long-lived E3 sessions, hardcoded DeepSeek model id) stay follow-ups.

liuchao-001 (approved):
Re-reviewed `993a1ef1..035a5b43`. Approving this head.

- **`993a1ef1` (git-tracked staging, whole-layout closure verification, redaction):** looks correct. I re-ran my original nested-closure fixture against it and it still fails closed.
- **`035a5b43`:** all three of bestony's findings are addressed.
  - The guard container now runs `sleep infinity` through the entrypoint.
  - `REAL_ACCEPT_TIMEOUT_MS` is shared with the keepalive test.
  - Catalog version patterns are wrapped as `^(?:…)$`, so capture group 1 is preserved.
  - The host and client `models.json` filters rebuild the file from `models`/`providers` only and reject malformed shapes.
- **Local runs at this head:** all 43 runner script tests and 66 client runner tests (11 files) pass.

**CI:** `Build linux/amd64 image and run offline smoke` passed, which confirms the newly anchored Slack/Lark patterns still match the real binaries. `Node 24 Compatibility` is red because of `provider-cli-lock.test.ts > serializes concurrent operations on the same provider` (`ProviderCliLockBusyError`). This PR doesn't touch that test or the lock code, and the same suite passed on Node 22 and 26 at this head, so it looks like an existing timing flake. It still needs a rerun to go green before merge.

Earlier non-blocking items stay as follow-ups: extra Client exports, dead/duplicated helpers, no init process for long-lived E3 sessions, and the fixed DeepSeek model id. Real-model acceptance was last run at `35bde414`, as the description notes.

bestony (approved):
## Re-review of `035a5b43..b8987035`

Approving this head. The increment is a scoped harness fix: real-mode guard containers now pass Docker `--init` so PID 1 reaps orphaned children, while `sleep infinity` remains the keepalive and the image entrypoint for future E3 work is unchanged.

The keepalive stub asserts `--init` is a `docker run` flag (before the image) and still checks the accept-budget contract. `runLimitedContainer` (offline smoke) is intentionally untouched.

No new blocking issues.

liuchao-001 (approved):
Re-reviewed `035a5b43..b8987035`. Approving this head.

The increment is scoped to the real-mode guard container. `docker run --init` puts an init process at PID 1 to reap orphaned fixture children, while `sleep infinity` stays the keepalive. The keepalive test asserts that `--init` is a `docker run` flag placed before the image. Offline `runLimitedContainer` and the image entrypoint are unchanged.

This also settles the orphan-reaping concern (#6 in my first review) for the acceptance harness. The before/after diagnostic in the description shows zombies staying behind without init, which the process-disappearance check would otherwise have to tolerate. As the description notes, the image itself still has no init, so the long-lived E3 decision stays open.

All CI checks pass at this head.

PR #632, opened 2026-09-16
feat(cloud): add session-scoped Cloud Run sandbox execution

yuezengwu:
## Summary

Cloud Computer identities already exist, but a Session cannot yet own a running execution environment. Add the disabled-by-default E3 path: Session-owned Sandbox → one Cloud Run Instance (1 CPU / 1 GiB) → one native Sandbox → scoped Pi execution. Reuse the existing `sandboxes` table and Client/CLI Runner release; no database migration.

- Reserve generations before cloud calls, verify resource identity/network policy, retain uncertain allocations, and condition deletion on the tracked UID/etag.
- Authenticate the outbound Runner WebSocket, track readiness/heartbeats, renew scoped credentials, and cancel/reap native work before returning results.
- Compose Local Runtime, Cloud Runner and credential-proxy WebSockets under one upgrade listener. Keep each protocol's authentication and frame limits: full 64 KiB + 4-byte proxy chunks work even when Runner allocation is disabled, and the Cloud-aware runtime verifier remains selected.
- Keep the native rootfs, Session workspace and temporary provider configuration isolated. Include the dev Cloud Admin token in startup-error redaction.
- Provide a two-Session real-GCP acceptance harness with receipts/cleanup. Its Server control-state directory is fixture-owned and uses a canonical temporary path, so local macOS symlinks do not violate Server storage checks or leave state in the checkout.

## Validation

Current integration head: `7065cf94cf041edd835048edcf905842cf93113b`, incorporating main `31079d750a1f583f96a1b4efa68c22d798025a30` with a normal merge commit.

- `pnpm check`, `pnpm build`, `pnpm typecheck`: passed.
- `pnpm test`: 5,454 tests passed in exact-head CI. Local full run passed 5,451 tests, and the three subsequently added WebSocket regressions passed in a 38-test focused run. These cover Runner-disabled proxy capacity and runtime authentication override/fallback.
- Agent Runtime coverage: 434 tests passed; configured scope is 100% statements (4149), branches (2975), functions (734), and lines (3686). This is not coverage of all Runner code.
- PostgreSQL integration: 409 tests in 29 files passed.
- Real local Server/PostgreSQL startup: all 45 migrations applied, Runner and platform credentials initialized together, private control state stayed inside the temporary home, repository control state remained untouched, and Server/database/workspace cleanup passed. Synthetic cloud coordinates were used; no Cloud allocation was requested.
- Adversarial regressions cover stop/retry evidence preservation, UID/etag cleanup of owned policy-failing Instances, revoked/duplicate/closed sockets, retriable authentication timeout, readiness ordering, filtered credential removal, UTF-8/byte limits, process-group cleanup, and source Pi configuration isolation.
- Secret Scan passed remotely on this head and locally with TruffleHog 3.97.1 against the current base/head. History was not rewritten and no scanner rule was weakened. All automated CI checks passed, including Node 22/24/26 compatibility, Runner image/toolchain, browser/CLI/container smokes, and the CI aggregate. Changed-line coverage is 2479/2584 (95.94%, threshold 80%); the comparison is nonempty. The ownership gate still requires bestony approval.

### Native GCP evidence and its scope

Native GCP acceptance passed at **`64d76168c663110216f5a75d2d8a34d088df7cf5`**, before the Server integration merge. It used `kael-product/us-west1`, staging VPC/NAT, task-owned service identity/firewall rules, a disposable local Server/PostgreSQL fixture and a WSS-only tunnel. It was not end-to-end staging Server/IM-channel acceptance and has not been rerun on `7065cf94`.

Two concurrent 1 CPU / 1 GiB Instances each launched a native Sandbox, completed real DeepSeek Pi file-tool work, resumed the same Pi session, and cancelled a live three-PID process tree. Each loaded ten skills. Native delete/relaunch passed filesystem/rootfs/parent-environment probes a second time. Both Instances were verified removed and source Pi configuration was byte-identical. A separate native check cloned `github/gitignore` and observed a bounded timeout to the metadata HTTP endpoint; this is evidence for those tested requests, not a universal network-isolation claim.

All temporary cloud infrastructure, diagnostic Instances, proxy and tunnel were verified removed/stopped. The registry image was temporary and removed during cleanup. The preceding run that detected a change to host `auth.json` remains failed; its writer was not identified. Code-path review and new copy/child-environment regressions found no source-configuration write-back path, which does not identify that historical writer.

<details>
<summary>Native GCP acceptance receipt</summary>

```json
{
  "sourceSha": "64d76168c663110216f5a75d2d8a34d088df7cf5",
  "sourceDirty": false,
  "project": "kael-product",
  "region": "us-west1",
  "image": "us-west1-docker.pkg.dev/kael-product/opentag-e3-20260916/runner@sha256:dbc1032d1414f9a5f1c6e696627e2a27bfdac5b2b8b7cc10983332e93dfe2bcb",
  "cpuPerInstance": 1,
  "memoryPerInstance": "1Gi",
  "mode": "real",
  "provider": "deepseek",
  "outcome": "passed",
  "assertionsPassed": 9,
  "sessions": [
    {
      "instance": "projects/kael-product/locations/us-west1/instances/ot-d-d6d3ebf4085c4067b1aad20699d853e5-1",
      "uid": "7a4928c5-f929-4077-8e37-a425ef39935c",
      "model": "passed",
      "firstTaskMs": 22344,
      "loadedSkills": 10,
      "livePidsBeforeCancel": 3,
      "cleanup": "verified-removed"
    },
    {
      "instance": "projects/kael-product/locations/us-west1/instances/ot-d-096b8a1d531d4842add8982305188cd9-1",
      "uid": "76469719-d97e-4bd4-b0dc-478ce12a2449",
      "model": "passed",
      "firstTaskMs": 21269,
      "loadedSkills": 10,
      "livePidsBeforeCancel": 3,
      "cleanup": "verified-removed"
    }
  ],
  "sourceConfigurationUnchanged": true,
  "cleanup": {
    "project": "kael-product",
    "region": "us-west1",
    "taskTaggedInstanceCount": 0,
    "temporaryInfrastructureVerifiedDeleted": 6,
    "diagnosticInstancesVerifiedDeleted": 4,
    "proxyAndTunnel": {
      "proxyStopped": true,
      "tunnelStopped": true,
      "port18133Listening": false,
      "pids": [
        65439,
        65663
      ]
    }
  }
}
```

</details>

## Scope and remaining limits

E3 remains off by default and supports one test Server. Reliable IM dispatch and multi-Server routing (E4), GCS save/restore (E5), and reuse/idle reclamation remain separate work. Deleting an E3 Instance loses its workspace. The credential foundation does not by itself connect E3 acceptance to the general credential-backed IM task path.

The native launcher requires a root parent. Ordinary image entrypoint commands drop to uid/gid 10000; native `sandbox exec` bypasses that entrypoint and the worker currently runs as uid 0. Privilege reduction requires corresponding resolver-copy ownership/readability changes. Bootstrap credentials remain Instance environment variables; trusted project configuration readers can see them and win the first connection. Live duplicate takeover is rejected, but stronger bootstrap identity is required before enabling this beyond trusted operators.

Cloud Admin requests are bounded individually; internal adoption polling and the upper-level convergence budget remain separate, not a unified end-to-end startup SLA. Unknown allocation outcomes retain resource references for operator reconciliation. Existing review/ownership gates remain enforced; no merge or deployment is included.

## Breaking changes

None. No new table or migration in this PR.

## Checklist

- [x] Local build, typecheck, unit tests, integration tests, and required coverage pass.
- [x] Canonical English documentation and Chinese mirrors are synchronized.


packages/server/src/services/sandboxes/sandbox-runner-service.ts
```diff
@@ -0,0 +1,1053 @@
...
+    if (transition.action === "release") {
+      // Invalidate Runner execution immediately: a releasing environment must never accept or
+      // resolve work while its Instance is being removed.
+      this.#closeRunnerScope(transition.row);
+      try {
+        await this.#releaseAllocation(transition.row);
+      } catch (error) {
+        // Preserve an in-flight submission marker: `cloud_create_pending` is the only evidence
+        // that the winner may still POST, so a stop failure must never erase it.
+        await this.#recordError(transition.row, "cloud_delete_incomplete", error, { preservePending: true });
```
liuchao-001: P1 (see review body): `preservePending` only protects `cloud_create_pending`. This overwrites `cloud_create_rejected`/`cloud_create_failed`, which are the only evidence `#releaseFromDefinitiveEvidence` accepts when there is no operation name. One failed `getInstance` during stop then leaves the row in `releasing` permanently.

packages/server/src/services/sandboxes/sandbox-runner-service.ts
```diff
@@ -0,0 +1,1053 @@
...
+   * The create result was never observed. A visible Instance is verified/tracked and deleted; a
+   * completed operation proves the outcome; anything still unknown keeps `releasing` plus the
+   * full reference and reports incomplete after the bounded window.
+   */
+  async #awaitUnknownAllocation(row: typeof sandboxes.$inferSelect, resourceName: string): Promise<void> {
+    const deadline = this.#now().getTime() + this.#deleteVerifyTimeoutMs;
+    for (;;) {
+      const view = await this.#cloud.getInstance(resourceName);
+      if (view) {
+        this.#cloud.verifyInstance(view, this.#identityFor(row, row.environmentGeneration));
```
liuchao-001: P2: `verifyInstance` also enforces policy (VPC/ingress/image), so an Instance whose labels and name match but whose policy fails can never be deleted here. Ownership could gate a UID+etag delete on its own.

packages/server/src/api/runner-ws.ts
```diff
@@ -0,0 +1,398 @@
...
+      closeWith(RUNNER_WS_CLOSE.authFailed, "authentication timed out");
+    }, authTimeoutMs);
+    authTimer.unref?.();
+
+    const renewCredential = async (current: RunnerScope) => {
+      if (closed || !options.hub.isCurrent(current.sandboxId, adapter)) return;
+      // Revalidate the database allocation before minting anything; a stale connection must
+      // never receive a fresh credential.
+      const stillCurrent = await options.service.validateRunnerScope(claimsFromScope(current));
+      if (closed || !stillCurrent || !options.hub.isCurrent(current.sandboxId, adapter)) return;
```
liuchao-001: P2: a definitive `!stillCurrent` (authority revoked, Session ended, Account suspended) just returns and leaves the socket connected. Closing with `staleScope` here would make revocation actually take effect.

packages/client/src/runner/serve.ts
```diff
@@ -0,0 +1,585 @@
...
+      }
+    };
+    const armSilence = () => {
+      if (!welcome) return;
+      clearTimeout(silenceTimer);
+      silenceTimer = setTimeout(() => finish(), welcome.heartbeatTimeoutMs);
+    };
+    const stop = () => finish();
+    stopListeners.add(stop);
+    const authTimer = setTimeout(() => finish("auth_failed"), options.authTimeoutMs ?? DEFAULT_AUTH_TIMEOUT_MS);
```
liuchao-001: P2: this local timeout covers TCP+TLS+upgrade+server-side auth, but `maintainConnections` treats `auth_failed` as terminal. A slow connect after a Server restart permanently kills the Runner. Only an explicit `auth:result ok:false` should be terminal.

liuchao-001 (changes requested):
Reviewed at head `34b51ce3`: the lifecycle service, Runner WS/auth, the Cloud Run adapter, and the client Runner/image, all read against `docs/cloud-runner-execution.md`. Core fencing holds up well: generation/name/UID CAS on promotion and claim, exact-socket checks for readiness/result, conditional delete with UID and etag, token audience/claims pinning, and argv/env isolation for native launches. Findings are ordered by severity. Requesting changes for the P1. The P2s should also land before anyone enables this outside a single dev operator.

### P1 — A single failed cloud read during stop leaves the Sandbox stuck in `releasing` with no API recovery
`stopForAccount` catch → `#recordError(row, "cloud_delete_incomplete", …, { preservePending: true })` protects only `cloud_create_pending`. It overwrites `cloud_create_rejected` / `cloud_create_failed`, which are the only evidence `#releaseFromDefinitiveEvidence` accepts for a row with no operation name.

Scenario:
1. A create is rejected before any operation exists (401/403/404, or token issue failure) → `preparing`, name set, UID null, op null, `cloud_create_rejected`.
2. The user stops it → `releasing` → `#awaitUnknownAllocation` → `getInstance` throws. This is likely here: it's the same IAM/token problem, or any transient 5xx/timeout.
3. The marker becomes `cloud_delete_incomplete`.
4. Every later stop: GET 404, no op, the marker isn't definitive, so it waits 60s → `cloud_create_uncertain` → throw → `cloud_delete_incomplete` again.

Start returns 409 while `releasing`, so only a manual DB edit recovers the row, even after IAM is fixed. It also erases the `cloud_instance_unverified` diagnostic. The existing stop-failure test starts from a tracked UID, so it doesn't cover this.

Suggested fix: never let the delete-phase error replace a create-phase marker (guard the WHERE, or keep the delete error separate), and add a "definitive marker + `getInstance` failure on stop, then stop again" test.

### P2
1. **A policy-failing Instance can never be deleted, and its Runner keeps getting renewed credentials.** `verifyInstance` combines ownership (name/labels) and policy (VPC/ingress/image/resources), and `#awaitUnknownAllocation` calls it before `#deleteVerified`. A create that fails the post-create policy check (e.g. egress not `ALL_TRAFFIC`) leaves a running, billable, non-compliant Instance: every stop throws before delete. The test at `sandbox-runner-service.test.ts` for unverified stop locks this in. Meanwhile `validateRunnerScope` doesn't require a UID, so `renewCredential` keeps minting tokens for that Runner. Suggest: when ownership matches but policy fails, still allow a UID+etag delete, and skip renewal for unverified rows.
2. **The bootstrap token is a plain env var on the Instance, reusable for its full TTL, and the last connection wins.** `#container` puts `OPENTAG_RUNNER_BOOTSTRAP_TOKEN` in the Instance spec, so anyone with `run.instances.get` in the Runner project can read it (and possibly audit logs of the create request; I haven't verified that). `hub.attach` evicts the existing same-scope socket unconditionally. A holder of the token can take over the connection, pass readiness (the only server-side check is the self-reported `runnerVersion`), receive the next `acceptance:run` including `piConfig.authJson`, and keep getting renewed tokens. Options: Secret Manager ref, single-use `jti` for the env token (only in-memory renewed tokens after first connect), or verify a metadata ID token for the runtime SA/Instance. At minimum, don't let a new connection replace a live, heartbeating one. The doc wording ("Authentication is in the first frame, never the URL") understates this exposure.
3. **Revoking authority doesn't close a connected Runner (fail-open).** Agent/binding deactivation, Session end, and Account suspension produce no hub event. `renewCredential` sees `!stillCurrent` and simply `return`s, heartbeats never re-check the DB, and an in-flight acceptance (up to 15 min) keeps spending the credential. Closing with `staleScope` when renewal gets a definitive `undefined` is cheap. A test for revocation while connected would pin it down.
4. **Readiness handling runs cloud reconcile/create/release inline in the per-connection frame chain.** `markRunnerReady` → `#reconcileAllocation` → possibly `#submitCreate` / `#afterReconcile` → `#releaseAllocation`. Heartbeats queue behind it, so `sweepStale` (45s) can close a healthy Runner mid-reconcile. The call can also run a second release concurrently with a user stop, and it lets a Runner-originated frame drive Cloud Admin writes. I'd keep readiness to CAS/promote only and leave cloud I/O to start/stop.
5. **The v1 fallback regex is broader than the recognized incompatibility.** `/network[-_ ]?interfaces|vpc[-_ ]?access|vpcAccess/i` matches any 400 that mentions those fields (bad subnet, tag, IP exhaustion…), not only the known annotation error. That contradicts "only a recognized rejected VPC-field 400". The risky case: v2 rejects a real policy problem that v1 accepts, which then feeds into P2-1. Please match the specific error text/status details.
6. **Runner client: a slow connect permanently kills the Runner.** In `serve.ts`, `authTimer` fires `finish("auth_failed")` after 5s measured from socket creation, which covers TCP+TLS+upgrade+server token/DB checks. `maintainConnections` treats `auth_failed` as terminal, so the Runner exits and nothing restarts it (a Server restart or cold ingress is enough). Only an explicit `auth:result ok:false` should be terminal; a timeout should take the backoff path.
7. **Worker leaves the unfiltered Pi config readable for the whole run.** `worker.ts` writes the raw `auth.json`/`models.json`/`settings.json` to `scratch/pi-agent`, filters into `filtered-pi`, and removes the raw copy only in `finally`. The server doesn't restrict `piConfig` to DeepSeek (`RUNNER_ACCEPTANCE_PROVIDER` is used only by the harness), so any other provider key the caller included stays readable to Pi tools running as the same uid. Delete the raw directory right after filtering (or filter in memory), and consider enforcing the provider filter server-side too.
8. **Harness: Ctrl-C kills the local Server before cleanup.** `spawnLogged` (in `cloud-computer/common.mjs`) doesn't detach, so a terminal SIGINT reaches the Server child too. Its handler closes the app and the DB, so `cleanupAllocation`'s Server-API retries all fail and Instances leak. The receipt also doesn't record project/region. Use `detached: true` for fixture/Server children, and write project, region and the expected deterministic name into the receipt before `start`.

### P3 / nits
- `runner-ws.ts` `handleReady` calls `hub.markReady` **before** `markRunnerReady` returns `version_mismatch`. On a row that is already `ready`, a same-scope reconnect with a different version still makes `snapshot.ready` true, and `runAcceptanceForAccount` doesn't re-check `readiness.runnerVersion`. The server also ignores `sandboxName` vs `resourceName`. The doc's "Runner must prove native execution … isolation" is a self-report server-side; either soften the doc or check more fields.
- `#adopt` hard-codes a 60s deadline instead of `CREATE_CONVERGE_TIMEOUT_MS`. With `apiTimeoutMs` per GET, `POST /runner/start` can block for several minutes.
- The v1 body drops `cpuIdle: false` with no `cpu-throttling` annotation, and `verifyInstance` checks neither, so the two create paths can silently diverge. `verifyInstance` also doesn't check `command`, `env`, or volumes, which matters for adopted (409) resources.
- `#recordError` has no lifecycle guard, so late failures can write `cloud_delete_incomplete`/`cloud_create_uncertain` onto an `unallocated` row (misleading status until the next reservation).
- `CloudRunAdminError` thrown from `#deleteOrphaned` / `#afterReconcile` / `tokens.issue` in the start path isn't mapped, so the caller gets a 500 instead of a 503 envelope.
- The test "does not report runnerReady while releasing" awaits stop completion, so it asserts on `unallocated`, not `releasing`.
- `init.c`: SIGALRM landing between the `force_cleanup` check and a blocking `waitpid` can leave a SIGTERM-ignoring child alive until the platform kills it. Use `sigsuspend`/`sigtimedwait`, or re-arm the alarm while children remain.
- `native-sandbox.ts` decodes each stdout chunk separately (splitting multibyte UTF-8), and the cap counts UTF-16 units, not bytes.
- `run.mjs` readiness loop: a non-numeric `OPENTAG_E3_READY_TIMEOUT_MS` gives `NaN`, so the loop never times out. It also keeps polling through terminal error markers.
- Needs native proof, not a code bug: nothing verifies that `delete --force` discards writes made with `--write --rootfs /opt/sandbox-root`. The probe only checks a canary in the parent `/tmp`. If writes persist, a cancelled worker's credential scratch survives the "clean environment" reset. Worth adding a rootfs canary to the relaunch probe before the GCP acceptance.

### CI
`Secret Scan` fails because TruffleHog flags a URI with embedded credentials. Those are the `user:…@api.example.com` fixtures in the backend-URL rejection tests (`packages/server/src/__tests__/config.test.ts:139`, `packages/client/src/__tests__/runner-serve.test.ts:60`). It's a false positive, but the job is red. Building the URL at runtime (e.g. `new URL(...)` with `username`/`password` assigned) avoids the pattern.

liuchao-001:
Checked the new head `fa778c59` (startup-probe listener + acceptance entry) against the review at `34b51ce3`:

- **Delta itself looks fine.** `health.ts` only accepts and ends connections (it never reads or writes, and has a connection cap and socket timeout). It starts only after native launch+probe and closes in `finally` even when sandbox cleanup throws. `verifyInstance` now pins exactly one port, `8080`, and v1 inherits it through the shared `#container`. One thing to confirm in the GCP run: the listener comes up only after `sandbox.launch()` + `probe()`, so a slow native launch has to fit inside the platform's default TCP startup-probe window. With `restartPolicy: NEVER`, a probe failure isn't retried.
- **The review findings are still open.** `packages/server/src/services/sandboxes/` is unchanged in this push, so the P1 (definitive create marker overwritten by `cloud_delete_incomplete` → Sandbox permanently stuck in `releasing`) and the P2 items still apply. The requested-changes review stays in place.
- **Secret Scan:** `config.test.ts` now builds the credential URL at runtime, but `packages/client/src/__tests__/runner-serve.test.ts:60` still has `wss://user:pass@api.example.com/ws`. TruffleHog also scans the earlier commit in the PR range, so it may stay red until the history no longer contains the literal (e.g. when the PR is squashed/rebased).

packages/server/src/services/cloud-run/cloud-run-admin.ts
```diff
@@ -0,0 +1,375 @@
...
+      !(memory === "1Gi" || memory === "1024Mi")
+    )
+      throw new CloudRunAdminError(
+        "invalid",
+        "Cloud Run Instance does not match the required Runner image, identity, resource or ingress policy",
+      );
+  }
+  async #createV1(spec: RunnerInstanceSpec, id: string, name: string): Promise<CloudRunCreateResult> {
+    const c = this.#container(spec);
+    c.resources = { limits: { cpu: "1", memory: "1Gi" } };
```
bestony, line 264: **[suggestion]** The documented us-west1 Direct VPC path falls back to regional v1 create, then overwrites `#container().resources` with `{ limits: { cpu: "1", memory: "1Gi" } }`. That drops `cpuIdle: false`, which the v2 body at line 298 sets explicitly. `verifyInstance` also never checks `cpuIdle`. If Instances honor this field the same way Services do, the fallback that this PR says is required for VPC would throttle CPU outside inbound requests — exactly when `opentag-runner serve` is holding an outbound WSS channel with no parent HTTP traffic.

**Suggestion:** Keep the full `#container(spec).resources` object on the v1 path (including `cpuIdle: false`), and/or set the v1 equivalent `run.googleapis.com/cpu-throttling=false` annotation. Assert `cpuIdle === false` in `verifyInstance` so a 409-adopted instance cannot silently lose always-on CPU.

packages/server/src/services/cloud-run/cloud-run-admin.ts
```diff
@@ -0,0 +1,375 @@
...
+    const response = await this.#request(
+      "POST",
+      `${this.#apiBaseUrl}/v2/${parent}/instances?instanceId=${encodeURIComponent(id)}`,
+      this.#v2Body(spec),
+    );
+    if (response.status === 409) return { outcome: "adopted", instance: await this.#adopt(name, spec) };
+    if (response.status === 400) {
+      const body = record(await readBoundedJson(response));
+      const message = record(body.error).message;
+      if (typeof message === "string" && VPC_REJECTION.test(message)) return this.#createV1(spec, id, name);
```
bestony: **[suggestion]** The v1 compatibility create is invoked as `return this.#createV1(spec, id, name)` and never receives `options.onOperation`. `#createV1` also discards the POST body (`await r.body?.cancel()`) and returns no `operationName`. On the path this PR documents as the working VPC create, the Server therefore never records an LRO. Stop/reconcile then cannot use completed-operation rejection evidence; they can only GET the deterministic name and time out into `cloud_create_uncertain` / `releasing`.

**Suggestion:** Parse and persist the v1 create operation identity the same way v2 does (or poll a v2 operation if that is the canonical name), and thread `onOperation` through `#createV1` so a concurrent stop can distinguish “create still in flight” from “create definitively rejected”.

packages/server/src/api/runner-ws.ts
```diff
@@ -0,0 +1,398 @@
...
+      const claims = await verifyBootstrapToken(token, requestId);
+      if (!claims || closed) return undefined;
+      return validateBootstrapClaims(claims, requestId);
+    };
+
+    const attachAuthenticatedRunner = async (validated: RunnerScope, requestId: string | undefined) => {
+      // No await between the closed check and the hub mutation, so a socket that closed during
+      // verification can never become a ghost entry.
+      scope = validated;
+      options.hub.attach(validated, adapter);
```
bestony: **[suggestion]** `attachAuthenticatedRunner` comments that there is no await between the closed check and the hub mutation, but it calls `hub.attach` *before* reading `closed`. A socket that fails in the microtask gap after `validateBootstrapClaims` will attach (closing any current connection for that sandbox as replaced) and only then detach. The test at `runner-ws.test.ts:502` only asserts the hub is empty when that socket was the sole connection; it does not assert a live peer is left in place.

**Suggestion:** Return immediately when `closed` is already true, and only then `hub.attach`. Add a regression where sandbox A is authenticated and current, a second socket for the same scope authenticates then closes before welcome, and A remains `hub.isCurrent`.

packages/client/src/runner/serve.ts
```diff
@@ -0,0 +1,585 @@
...
+  stopListeners: Set<() => void>,
+): Promise<void> {
+  let failures = 0;
+  while (!state.stopping && !state.fatal) {
+    const outcome = await serveOnce(config, state, sandbox, options, stopListeners);
+    // This barrier covers worker exit AND native deletion/recreation. No successor connection
+    // can dispatch work into a sandbox still owned by the prior connection.
+    await state.active?.done;
+    if (state.stopping || state.fatal) break;
+    if (outcome.kind === "auth_failed") {
```
bestony, line 426: **[suggestion]** Handshake timeout uses `finish("auth_failed")`, and `maintainConnections` treats every `auth_failed` as a permanent rejection (`break`, no reconnect). Combined with Instance `restartPolicy: NEVER`, one slow first-frame auth (Server still opening DB, 5s default) exits the Runner and leaves a billed Instance that will never dial again until an Account stop/start. Transport failures reconnect; auth *timeout* should too. Explicit `auth:result ok: false` is the case that must not retry.

**Suggestion:** Finish handshake timeout as `kind: "closed"` (or a distinct `auth_timeout`) so backoff reconnect still applies. Keep the permanent break only for an in-band `auth:result` with `ok: false`.

bestony (commented):
## Summary

E3 is a large but tightly constrained addition: generation reservation happens before any Cloud I/O, delete is UID+etag conditioned, Runner auth stays in-band, native launches pin `/opt/sandbox-root`, and the feature stays off unless `OPENTAG_CLOUD_RUNNER_ENABLED=true`. Allocation, hub identity checks, and isolation tests match the stated contract. The remaining gaps are on the documented v1 VPC create fallback (it is not equivalent to the v2 body) and on a couple of control-channel lifecycle edges that can leave a live Instance with no reconnecting Runner.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

packages/server/src/services/sandboxes/sandbox-runner-service.ts
bestony: Agree with the P1 on `34b51ce3`; it is still present on `fa778c59`. I missed it on my first pass and I am not approving.

`stopForAccount` records `cloud_delete_incomplete` with `preservePending: true`, which only skips overwrite of `cloud_create_pending`. `#releaseFromDefinitiveEvidence` treats a no-UID, no-operation row as gone only when `lastErrorCode` is `cloud_create_rejected` / `cloud_create_failed`. One failed `getInstance` during stop therefore erases the only evidence that would let a later stop clear the row. Start is 409 while `lifecycle === "releasing"` (`runnerConflict`), so the Account API cannot recover after IAM is fixed.

Please keep create-phase markers across delete-phase failures (or store the delete error separately) and add the “definitive marker + `getInstance` failure, then stop again” test.

Overlap with my COMMENT: P2-6 (handshake timeout → terminal `auth_failed`) is the same finding as my issue 4, and the v1 path still drops `cpuIdle: false`. `fa778c59` adds the 8080 startup probe and rewrites one Secret Scan fixture; it does not change this P1 or the terminal timeout.

bestony (commented):
Re-reviewed `fa778c59` (startup probe + acceptance entry). Still not approving.

The delta looks right for the probe: one declared `containerPort: 8080`, a zero-data TCP closer bound only after native launch/probe, adopt-time port verification, and a `main` seam so the harness records a redacted first-step failure. I did not find a new correctness bug in this commit.

Unchanged and still blocking merge:

- P1 from @liuchao-001: `stopForAccount` still overwrites `cloud_create_rejected` / `cloud_create_failed` with `cloud_delete_incomplete`. `sandbox-runner-service.ts` is untouched.
- Handshake timeout is still terminal `auth_failed`; v1 create still drops `cpuIdle: false` and still records no LRO.

Nit on this commit: `packages/server/src/__tests__/config.test.ts` no longer embeds a userinfo URI, but `packages/client/src/__tests__/runner-serve.test.ts:60` still has `wss://user:pass@api.example.com/ws`. That is the same TruffleHog unverified-URI pattern that failed Secret Scan on `34b51ce3`.

bestony:
@liuchao-001 Independently the same read of `fa778c59`. Delta is a real probe fix; P1/`sandbox-runner-service.ts` is untouched, so the requested-changes review should stay.

Agreed on the two extras:

- Native-then-listen plus `restartPolicy: NEVER` means a slow `sandbox.launch()`/`probe()` that misses the default TCP window is a dead Instance, not a retry. Worth confirming in the GCP acceptance, not a code bug in this commit.
- Secret Scan can stay red even after rewriting `runner-serve.test.ts:60`, because TruffleHog scans the PR range and `34b51ce3` still has the `config.test.ts` literal. Squash/rebase is the actual clear; rewriting the remaining Client fixture is still worth doing so HEAD itself is clean.

Still not approving.

packages/server/src/api/runner-ws.ts
liuchao-001: Reordering here is cheap and worth doing, but I don't think the eviction can happen today. `validateBootstrapClaims` already re-checks `if (closed) return undefined` after its last real await (`validateRunnerScope`). From that point to `hub.attach`, the only steps are promise-resolution hops (`return validated` → `resolveAuthenticatedScope` → `await` in `handleAuth`), and those run as microtasks. `closed` is only set from the socket `close` event or `authTimer`, both macrotasks, so neither can run between that re-check and `attach`. So I'd treat this as defensive (check `closed` before `attach`, which also makes the comment true) rather than a live-peer eviction bug. A regression test couldn't reproduce it without injecting a microtask-level close.

Agree on the v1 LRO point. I'd rank it above my P3: on the documented VPC path, a timed-out post-create GET leaves stop with no operation evidence at all.

bestony: You're right — I overstated this as a live-peer eviction.

After `validateRunnerScope` returns, `if (closed) return undefined` is the last real await. `return validated` → `handleAuth`'s continuation → the sync prefix of `attachAuthenticatedRunner` (including `hub.attach`) are promise-resolution hops, i.e. microtasks. `closed` is only set from `authTimer`, `socket` `close`/`error`, or `closeWith` on the serialized frame chain; those are macrotasks (or later on the same chain, after attach has already run). They cannot sneak in between that re-check and `attach`. A regression that needs a microtask-level close wouldn't be a realistic test.

Demoting this from a live bug to a comment-vs-code nit: check `closed` before `attach` so the comment is true, and so a future await in `validateBootstrapClaims` cannot reintroduce the window. Not blocking.

Agreed on ranking the v1 LRO above the P3 cpuIdle note: on the documented VPC path, a timed-out post-create GET leaves stop with no operation evidence at all.

liuchao-001 (commented):
Re-reviewed `90ba1e49` against my review at `34b51ce3`.

### Resolved (verified in code + new tests)
- **P1** — `#recordError(..., { preserveCreateMarkers: true })` / `#marker` now use `notInArray(CREATE_PHASE_MARKER_LIST)`, so a failed stop read can no longer erase `cloud_create_rejected` / `cloud_create_failed` / `cloud_instance_unverified`. There are unit and integration tests for "failed stop read, then later 404 releases the row", including the pre-submission and unverified variants. `#recordError` also gained the lifecycle guard.
- **P2-1** — the delete path uses `verifyOwnership` only (UID+etag still enforced in the conditional delete), and `validateRunnerScope` refuses definitive/unverified rows, so no renewal for them.
- **P2-2** — partially mitigated, and acceptable for E3: `hub.attach` returns `duplicate` while the same-scope connection is live, so a second token holder can't take over a heartbeating Runner. The env-var token is still the Runner's whole identity until expiry, so a holder who connects *first* wins the scope. Please keep that on the list before non-dev enablement (Secret Manager / single-use `jti` / metadata ID token).
- **P2-3** — renewal now closes with `staleScope` on a definitive miss.
- **P2-4** — `markRunnerReady` does no cloud I/O; untracked generations stay `deferred`.
- **P2-5** — v1 fallback now requires an exact message match. It fails closed if Google rewords the error, which is the right direction.
- **P2-6** — `auth_timeout` and pre-auth server frames go through backoff; only in-band `ok:false` is terminal. With 8 attempts of 2s→30s backoff, a `duplicate` rejection during a half-open window (45s) still reattaches in time.
- **P2-7** — the raw Pi dir is removed right after filtering.
- **P2-8** — the Server is spawned `detached`, and the receipt records project/region plus the expected resource name before `start`.
- **P3s:** hub readiness is only set after the version verdict; `sandboxName` is checked against the resource; `cpu-throttling=false` is set on v1 and `cpuIdle`/`command`/`volumeMounts` are verified; cloud errors on start map to 503; `init.c` re-arms the alarm; capture is byte-bounded with a single UTF-8 decode; the rootfs canary proves `--write` doesn't land in `/opt/sandbox-root` (thanks); `OPENTAG_E3_READY_TIMEOUT_MS` is validated.

### One new item from this push (should-fix, not blocking E3)
**The in-sandbox worker doesn't go through the `setpriv` drop the docs describe.** `USER` is now `root`, and `runner-entrypoint` drops non-`serve` commands to 10000. But the acceptance worker is started by `serve.ts` as `sandbox exec <name> -- /usr/local/bin/node <worker-entry> worker`, which bypasses `runner-entrypoint`. It runs with whatever uid the native sandbox gives a root parent, most likely 0. `docs/cloud-runner-execution.md` ("Every other image command (… `worker`) executes as uid/gid 10000") then describes only the `docker run image worker` path, not the Pi/model-tool process that actually handles user work.

Isolation still rests on the native sandbox (mount set, `/proc/*/environ` probe), so this isn't a known escape. It does widen the blast radius, and the doc overstates it. Suggest either:
- exec the worker via `setpriv --reuid=10000 --regid=10000 --clear-groups` inside the sandbox rootfs, and have `probe()` assert the worker uid is non-zero; or
- if the native sandbox already maps uids, prove it in the GCP run (record `id -u` from inside the sandbox in the receipt) and state that in the doc.

### CI
`Secret Scan` is still red on this head. That's expected: the PR range still contains the earlier literal fixtures, and squash/rebase clears it. `packages/client/src/__tests__/runner-serve.test.ts` no longer has the literal. `Checks`/Node compatibility/coverage are still running. I'll switch to approve once they're green on this head.

liuchao-001 (approved):
Approving `90ba1e49`. The P1 and all P2s from my review are resolved, as detailed in my re-review. Checks, Node 22/24/26, Enforce Patch Coverage, the image offline smoke, and the container/browser/CLI smokes are all green on this head.

Non-blocking follow-ups, recorded here so they aren't lost:
- **In-sandbox worker uid.** It is started by `sandbox exec` and doesn't go through the entrypoint's `setpriv` drop. Please either drop to 10000 inside the sandbox with a probe assertion, or record `id -u` from inside the sandbox in the GCP receipt, and fix the doc sentence either way.
- **Bootstrap token.** It's still an Instance env var, and the first connection wins. That's accepted for E3 in a trusted operator project; it needs replacing before non-dev enablement.
- **Remaining red gates are not code findings.** `Secret Scan` (and the `CI` aggregate) is red only because of the historical fixture literals, which clear on squash or the requested history override. `ownership-gate` still needs @bestony.

This approval doesn't cover the real-GCP acceptance, which is still in progress. Please post its receipt (image digest, native/model result, verified deletion) on the PR before merge.

liuchao-001:
My approval at `90ba1e49` also covers the two follow-up commits on current head `64d76168`. Both are fast-forward from the approved head, and I reviewed both. On this head, Checks, Node 22/24/26, Enforce Patch Coverage, the image offline smoke, and the container/browser/CLI smokes are all green.

- **`58a546ac` (resolver snapshot):** the sandbox now mounts a size-bounded, regular-file-checked private copy of `/etc/resolv.conf` instead of the platform path. The copy is removed after a successful delete or a failed launch, and kept when delete fails. Looks right. Minor: the copy is `0600` root-owned. That's fine while the worker is uid 0, but if the native worker later drops to 10000 (the hardening follow-up) it can no longer read it and DNS will fail. `0644` avoids that, since the file isn't secret.
- **`64d76168` (Pi probe budget):** the runner uses a 30s probe budget, bounded to 60s max; the local Pi default is still 5s. Thanks for making the doc explicit that the native worker runs as uid 0 inside the sandbox (verified in GCP) and for tracking the drop as hardening. That settles my doc concern.
- **FYI, not from this PR:** `RUNNER_DEEPSEEK_MODEL = "deepseek-v4.1-flash-expires-on-0910"` has been on main since #630. Today is 09-16, so if that model ID has actually been retired, the real-model part of the GCP acceptance would fail for that reason alone. Worth checking before reading a model failure as a Runner bug.

The earlier caveats still apply: the real-GCP receipt should be posted before merge, `Secret Scan`/`CI` is red only from historical fixture commits, and `ownership-gate` still needs @bestony.

liuchao-001:
Read the native GCP receipt in the description: `sourceSha` `64d76168`, `sourceDirty=false`, digest-pinned image, two concurrent 1 CPU/1 GiB Instances, real DeepSeek tasks + resume + live cancellation (3 PIDs reclaimed), second isolation probe after relaunch, both Instances verified removed, source Pi config unchanged, and temporary infra deleted. That meets the pre-merge evidence I asked for.

Also, the real-model pass means the pinned `…-expires-on-0910` model ID still works, so please disregard my FYI about it.

One thing worth keeping on the follow-up list: in the earlier failed run, the source `auth.json` changed mid-run and the writer wasn't identified. The integrity guard handled it correctly. But if anything on the acceptance path (rather than the operator's own Pi session) can write back to `--pi-config-dir`, that's worth pinning down before the harness is reused.

liuchao-001:
My approval also covers the merge head `7065cf94`. Every check except `ownership-gate` is green on it, including `Secret Scan` and the `CI` aggregate.

I reviewed the merge as a diff against Git's automatic merge of `64d76168` + main `31079d75`, not just against the approved head:
- **Conflict resolution** (`app.ts` / `index.ts` → `registerExecutionWebSocketRoutes`): one upgrade listener sized to the widest mounted route. Each route still enforces its own bound: `RuntimeSession` at 64 KiB, the Runner before JSON parse, and the provider-proxy transport checks header/chunk sizes itself (`data-transport.ts`). Runtime auth keeps `runtimeAuthService ?? machineAuthService`. Nit: the proxy route now mounts even without `computerService`/`machineAuthService` (main only mounted it inside that block). It authenticates by first-frame ticket on its own, and production always wires both, so this is not blocking.
- **Other merge-side changes:** the dev GCP token is now in startup-error redaction; the fixture pins `OPENTAG_RUNTIME_CONTROL_DIRECTORY` under its temporary HOME and uses a `realpath`'d tmpdir for #633's control store; and there is a new regression proving `--pi-config-dir` stays byte-identical.
- **Scope of the native GCP evidence:** main `cce09c84..31079d75` changes nothing under `packages/client/src/runner`, `providers/pi`, `scripts/runner`, `services/sandboxes`, `services/cloud-run` or `runner-ws.ts`. So the receipt at `64d76168` still covers the Runner/image/Cloud Run paths. The merged WebSocket composition is covered by the new `execution-websockets` tests rather than by a native rerun.

Merge now waits only on `ownership-gate` (@bestony).

PR #633, opened 2026-09-16
feat: manage GitHub and IM credentials with scoped runtime proxies

yuezengwu:
## Summary

OpenTag now owns GitHub App integration management and runtime credential custody alongside the existing Slack and Feishu integrations. Account settings connect GitHub, Agent settings select repository roles and publication policies, and the Server authorizes execution-scoped proxies for ordinary `git`, `gh`, `slack api`, and `lark-cli` commands. GitHub user credentials establish admission; repository operations use repository-scoped App installation tokens held by the Server.

- Reuse IM credential storage and add one business table, `github_connections`, with migration `0044_wet_xorn` after current main's `0043`. Add bound encryption, OAuth/refresh/recheck fencing, installation lifecycle handling, signed webhooks, and bounded repository discovery.
- Add short runtime capabilities, authenticated streaming transport, protected-read provenance and durable write outcomes. The Linux Sandbox bridge exposes execution handles and public trust material while the Server retains provider credentials.
- Enforce code and Context Tree repository/ref policies, verify committed Tree content with Context Tree 0.1.15, and preserve native CLI command syntax within the registered API surface. Include adversarial regression coverage for authority changes, concurrent refresh/rechecks, ref filtering, response handling, and ambiguous writes.
- Document configuration, compatibility, bounded Git staging limits, authoritative Session ownership, and later Cloud host wiring in the English and Chinese platform integration guides.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` passed for the integration revision. Follow-up changes are confined to CI scripts and tests; final checks/typecheck and all 385 root script tests pass. Latest package coverage runs pass 4,879 tests (Shared 352, Server 1,505, Client 1,454, CLI 505, Web 1,063). Existing complexity ratchet: 141 warnings / 14 improvements; no baseline weakening.
- All five package coverage floors pass. Server branch coverage is 80.36%; changed-line coverage is 11,001/12,049 (91.30%) against the unchanged 80% threshold. The initial whole-repository measurement exposed a Server branch-floor shortfall; 79 additional policy/authority/material tests close it, followed by a full Server remeasurement and aggregate floor/patch evaluation using the unchanged four package results.
- CI tooling regressions: Markdown-only diff capture passes with a >1 MiB unrelated patch; coverage diff capture retains all 20,000 changed fixture lines and rejects Git failures; Sandbox mount assertions distinguish host bind sources from an identically named empty container tmpfs. Coverage configuration, floors and exclusions are unchanged.
- The changed-line coverage workflow now passes the captured diff string directly. Non-string diffs fail closed; a legitimate empty diff retains its explicit pass. A regression test executes the actual workflow heredoc against uncovered and covered Git fixtures, requires nonzero denominators, and fails against the prior wiring. Fresh CI on `309b8835` confirms 11,001/12,049 executable changed lines (91.30%) against the unchanged 80% threshold. Main CI, Node 22/24/26 compatibility, image and smoke checks all pass; ownership approval and outstanding product review findings remain separate blockers.
- `pnpm --filter @opentag/client test:agent-runtime:coverage`: 433 tests; statements, branches, functions and lines all 100%.
- CI follow-up: the IM custody lock-order fixture now waits for the observed delivery frame and explicitly settles the unanswered request before database teardown, preserving both custody-transition variants and their assertions. The Provider CLI lock test now synchronizes a real holder and contending operation explicitly, verifying blocked entry, non-overlap, and exact order without timing sleeps. These test-only changes remove the shared 100 ms reconcile/delivery dependency and 5 ms / 1 ms lock timing dependency. `pnpm check`, typecheck, the full 403-test PostgreSQL suite, and all 1,454 Client tests pass after their respective test changes.
- `pnpm --filter @opentag/server test:integration`: 403 tests across 28 files using PostgreSQL/testcontainers. Later changes affect only CI tooling and tests; Server product sources and migrations remain identical.
- Linux native acceptance: 24/24 cases passed through the actual proxy/Relay/broker/bridge with controlled provider upstreams and a local bare Git remote. Covers Git/gh, Context Tree publication/rejection, Slack/Feishu, multi-megabyte transfers, protected-read provenance, authenticated thumbnails, renewal, revocation and isolation. Context Tree manifests/bundle/image all 0.1.15; amd64 runs were emulated on Apple Silicon.
- A new native source-count assertion initially failed because the real durable store deduplicates sources. Preserved the failure evidence, changed the harness to verify exact membership, dedup preservation and persistence-before-disclosure, then reran all 24 cases successfully.
- Live OAuth, real IAT/PR/IM delivery, external WSS ingress and deployed Cloud behavior are reserved for staging; the local fixtures do not establish those results.

## UI validation

- Desktop evidence: actual Account GitHub and Agent repository components rendered in a local browser harness at 768px and 1440px; screenshots and machine-readable receipts retained with the local review evidence.
- Mobile evidence: the same components at 320px and 390px, including code and Context Tree binding forms. Long sender identifiers remain within the viewport and keep their full accessible label.
- Widths checked: 320px / 390px / 768px / 1440px.
- States verified: 56 Account/Agent state-width views covering unavailable, loading, disconnected, active, reauthorization, errors and truncated discovery, plus selected code and Context Tree write forms. No overflow or browser page errors in the corrected harness.
- Keyboard/accessibility checks: Tab/Space checkbox interaction; accessible labels and error/status content; expanded Tree forms reported no axe violations. These are component browser checks with fixture APIs, not staging OAuth acceptance.
- New reusable block, theme value, or CSS exception: no new theme tokens or CSS exceptions; existing design-system controls and layout utilities are used.

## Breaking changes

Deploy compatible Server and Client/CLI artifacts together. Run migration 0044 and preserve required encryption keys; after v2 ciphertext is written, do not roll back to a v1-only reader. Persist Server-private control storage and route each Session to one authoritative owner.

Local credential mode retains its current default, with explicit proxy opt-in. Cloud requires proxy mode and a trusted Linux controller. The bridge supports later Cloud orchestration; it does not provision a scheduler, LLM network path, or automatic trusted Context Tree settings authority. Proxy settings operations without that authority return authentication required. Git operations are bounded by the documented full-history staging limits.

### Staging regression

- Record exact deployed Server image/commit and matching Client/CLI versions; verify migration and persistent control volume.
- Exercise real GitHub App OAuth/connect/reconnect/disconnect, repository selection and installation suspension/revocation; verify actual IAT scope and invalidation independently.
- Run native Git clone/fetch/push, `gh` PR operations, slash-containing branch REST reads, and Context Tree direct/PR publication on the authorized disposable repository.
- Exercise authorized Slack/Feishu reads, writes, uploads/downloads and authenticated thumbnails using test Bots and conversations.
- Cover capability renewal, real WSS ingress, streaming cancellation, process/reconnection/generation changes, revocation, sibling-execution isolation, and persisted unknown-write reconciliation.

Live platform delivery and Cloud deployment remain staging acceptance. Historical IAT-only Git results do not certify this PR head.

## Checklist


packages/server/src/services/github-proxy/github-provider-adapter.ts
```diff
@@ -0,0 +1,497 @@
...
+      repositories.map((repository) => ({
+        repositoryId: repository.repositoryId,
+        fullName: repository.fullName,
+        nodeId: this.#repositoryNodeId(request.executionId, authorization.scopeHash, repository.repositoryId),
+      })),
+      (id) => node(id)?.repositoryId,
+    );
+    const repository = findRepository(repositories, plan.repository.fullName);
+    const write = plan.operation !== "read";
+    const lease = await this.#options.leases.acquire(request.executionId, {
```
bestony: **[bug]** GraphQL mutations mint a repository-scoped IAT with `pull_requests: "write"` as soon as `planGitHubGraphql` accepts the operation and the repository is in the execution list. OpenTag write authority is applied only afterward in `#validateGraphqlWrite` / `assertPullRequestRefs`. REST does the opposite: it rejects read-only or non-`pull_request` scopes at line 118 before `leases.acquire`. A read-only (or `publish: "direct"`) GraphQL `createPullRequest` / `updatePullRequest` / `addComment` therefore causes the Server to mint a write IAT, then fail policy and revoke. If the process dies between mint and `lease.release()`, GitHub still holds a write-capable token until expiry. Spec requires Account/Agent/repo/execution authority before minting, and IAT permissions to be the explicit authorized set.

**Suggestion:** Mirror the REST pre-check before `leases.acquire`: require `scope.access === "write" && scope.publish === "pull_request"` for GraphQL writes, and run `#assertRefs` / existing-PR checks (using a read IAT or already-cached metadata) before requesting `pull_requests: "write"`. Add a regression that a read-only GraphQL mutation does not call `mint` with write permissions.

packages/server/src/services/github-proxy/tree-graphql-policy.ts
```diff
@@ -0,0 +1,81 @@
...
+  const document = visit(parse(input.request.query, { maxTokens: 10000 }), {
+    Field(field) {
+      const args = Object.fromEntries(
+        (field.arguments ?? []).map((argument) => [
+          argument.name.value,
+          valueFromASTUntyped(argument.value, input.request.variables),
+        ]),
+      );
+      validateField(field, args, input, numbers, nodeIds);
+      if (field.name.value === "defaultBranchRef")
```
bestony: **[suggestion]** For knowledge-only repositories, `constrainTreeGraphql` rewrites every `defaultBranchRef` field, including `parent.defaultBranchRef` from the native `gh` RepositoryInfo preread. The parent window is supposed to expose the upstream default-branch name only; rewriting it to the configured Tree ref can make `parent.defaultBranchRef` null or equal the child's knowledge branch on a fork.

**Suggestion:** Rewrite `defaultBranchRef` only on the authorized root repository selection, not under `parent`.

liuchao-001 (changes requested):
Reviewed at head `ce3f8f12`. The product code is byte-identical to `50241619`: `e0fd9e03` and `ce3f8f12` only touch CI scripts and tests. Base is `cce09c84`, the current `main`. Findings marked *reproduced* were proven with throwaway Vitest files against the existing harnesses on that product code, then deleted. The rest come from reading the code, and I say so where that applies.

A lot of this is careful work, and I checked the parts that are easy to get wrong:
- **OAuth:** state stored only as a hash and claimed once under `FOR UPDATE`, PKCE S256, and a hash bound to the login session.
- **Webhooks:** HMAC over the raw body with `timingSafeEqual`.
- **Installation tokens:** scoped to exactly one repository ID and checked for no extra permissions.
- **REST:** path normalization rejects `%2e`/`%2f`/`%5c` and changed-on-parse paths.
- **Git:** receive-pack only takes branch refs, with an fsck plus fast-forward check. The exact SHA that was verified is the one pushed.
- **Capabilities:** 256-bit, stored as hashes, with a scope hash that covers execution, run, revision and placement generation.
- **Local mode:** the unchanged default stays legacy, with nothing injected.

The problems below sit mostly where the proxy trusts data it did not produce, and in failure paths that never recover. Several are security holes in the boundary this PR exists to create, so I'm requesting changes.

---

## Security (blocking)

### 1. Slack URL rewriting lets an agent call almost any Slack Web API method with the bot token
`packages/server/src/runtime-credentials/slack-operations.ts:43-58`, `provider-proxy-support.ts:48`, `provider-proxy-adapter.ts:423-432`

`rewriteSlackProtectedUrls` walks the **whole** response. At any depth, it turns any string under `url_private`, `url_private_download`, `thumb_pdf`, `thumb_video` or `thumb_N` into a download handle. That includes app-controlled data such as message `metadata.event_payload`. `ALLOWED_HANDLE_HOSTS.slack` allows `slack.com` as well as `files.slack.com`. `#proxyDownload` then sends `GET <url>` with `Authorization: Bearer <bot token>`.

Scenario:
1. Post a message whose metadata contains `{"url_private":"https://slack.com/api/auth.revoke"}`.
2. Read it back with `conversations.history`.
3. GET the handle.

The Server calls `auth.revoke` with the real token. That call skips the operation registry, the write journal and source recording. *Reproduced:* one upstream call to `https://slack.com/api/auth.revoke?test=false` with `Bearer xoxb-…`. The same route reaches `conversations.kick/archive`, `files.delete`, and so on, so the docs' claim that admin operations are rejected isn't true yet.

Fix: only mint handles for `files.slack.com` `/files-pri/` or `/files-tmb/` URLs, and only inside real file objects (`file`, `files[]`).

### 2. The local Feishu proxy forwards requests with no handle, and the token exchange gives the handle to anyone
`packages/client/src/runtime/runtime-proxy-loopback-adapter.ts:633-636`, `:330`, `:386-449`

```ts
if (!handle) {
  return provider === "feishu"
    ? { ok: true }
    : { ok: false, status: 401, code: "authentication_required", challenge: true };
}
```

A wrong handle gets a 403, but **no** `Authorization` header is forwarded with the execution's capability. The Server then attaches the tenant token. On top of that, `POST /open-apis/auth/v3/tenant_access_token/internal` is answered before any authorization check, and the response contains the real local handle.

Any local process that can reach 127.0.0.1 can therefore act as the bot for the life of the execution. That includes a sibling execution on the same Computer. *Reproduced:*
- No handle: `200`, forwarded.
- Wrong handle: `403`.
- An empty `{}` body sent to the exchange returned `{"tenant_access_token":"<the local handle>",…}`.

GitHub and Slack correctly return 401. Feishu needs the same challenge, and the exchange needs its own proof: for example the App ID/secret placeholders the manifest gives to `lark-cli`, checked in constant time.

### 3. GraphQL aliases can plant a fake node-ID record, allowing writes to PRs and issues this Session doesn't own
`packages/server/src/services/github-proxy/github-provider-adapter.ts:396-411` (`#remember`), trusted by `#validateGraphqlWrite` (`:219-239`) and `#constrainTreeGraphql`

`#remember` walks each upstream response. Any object with `node_id ?? id` (string), `number`, and `headRefName`/`baseRefName` is stored as `id → {repositoryId, number}`. GraphQL response keys are **aliases the caller chooses**, and the alias check (`graphql-policy.ts` `validateReadField`) only looks at `field.name`.

Scenario:
1. The agent sets its own task-branch PR #1's body to a victim PR's node ID.
2. It queries `pullRequest(number:1){ id: body number headRefName baseRefName }`. The proxy now records `PR_victim → #1`.
3. `updatePullRequest(input:{pullRequestId:"PR_victim", state:CLOSED})` passes `#assertExistingPull(1)` and is forwarded with a `pull_requests: write` token.

`addComment(subjectId:…)` works the same way. In a Tree-only repo, `node(id:"PR_victim")` gets past the Tree base-branch read constraint. *Reproduced*, with a control: both requests are rejected without the planting read. This can't cross repositories, but within the repo it undoes the "only this Session's PRs" rule.

Fix: don't learn IDs from response bodies whose keys the caller controls. Before a write or a Tree read, resolve the node ID with a query the proxy writes itself (`node(id){… on PullRequest{number repository{databaseId}}}`).

### 4. A symlink with a tab in its name passes the Tree verifier
`packages/server/src/services/github-proxy/tree-verifier.ts:17-26`

The link rule tests `entry.endsWith("\tCLAUDE.md")` against raw `ls-tree -rz` output (`<mode> <type> <sha>\t<path>`). Git allows tabs in paths, so a symlink at `.github/x\tCLAUDE.md` matches. The follow-up `git show ${sha}:CLAUDE.md` reads the **root** link, which is legitimate.

*Reproduced* on a real CLI-created Tree: `.github/x\tCLAUDE.md → /etc/passwd` and `→ ../../../../../../../../` both verified. That contradicts the comment's "All other links are rejected before checkout". Split each entry at the first `\t` and require the path to equal `CLAUDE.md`.

## Correctness (blocking)

### 5. One failed write blocks that Slack or Feishu resource for the rest of the Session, even when nothing was sent
`packages/server/src/runtime-credentials/provider-proxy-adapter.ts:134-148`, `:263-265`, `upload-receipt.ts:53-70`, `services/session-control-store/file-store.ts:129-134`

`beginWrite` refuses any write whose provider and resource match an unresolved intent. Only `reconcileWrite` clears one, and it has **no production caller**; nor do `listUnresolvedWrites` or `removeCompletedSession`. The intent is also journaled **before** `resolveMaterial`/`recheck`, and every throw goes to `completeWriteReceiptUnknown`.

So each of the following blocks every later write to that resource in that Session, including from new executions, until the Session ends (channel and thread Sessions don't end):
- a Feishu tenant-token hiccup
- a rotated credential generation
- a Slack 503 or 30 s timeout on `chat.postMessage`

Resources are coarse too: a Slack resource is the whole `channel:<id>`, and Feishu uploads use the constants `"images"`, `"files"` and `"drive"`. Two concurrent uploads therefore collide.

*Reproduced:*
- A first post that failed in `resolveMaterial` (0 upstream fetches) left `{resource:"chat:oc_1", outcome:"unknown"}`, and the next healthy post returned `conflict`.
- The same happened after a Slack 503.
- The conflict escapes as a raw `SessionControlStoreError`.

Proxy mode is mandatory on Cloud, so this is how an agent loses the ability to reply in a thread. Resolve and recheck before journaling (or record "not sent"), and add a reconcile or expiry path.

### 6. The agent process itself gets the proxy environment
`packages/client/src/runtime/session-runtime-manager.ts:322-323`, wired at `client-runtime-composition.ts:729`

`environmentForSession` returns the execution's `HTTPS_PROXY`/`https_proxy`, `SSL_CERT_FILE` and `NO_PROXY`. They go into `workspace.environment`, and `composeRuntimeEnvironment` layers them onto the Claude Code, Codex and Pi processes. The CONNECT proxy only allows five provider hosts, so model API traffic from a proxy-aware runtime gets a 403.

The runtime is also started once per Session, so after the first execution is cleaned up that environment points at a closed port and a deleted CA file. I checked this with `curl` using the exact environment `buildRuntimeProxyEnvironment` produces:
- during the execution: `CONNECT tunnel failed, response 403`
- after `adapter.close()`: `Failed to connect to 127.0.0.1 port …`

I did not run a real Claude Code or Codex binary through it. The test "injects the active proxy execution environment into visible Sessions only" asserts this behaviour. The git/gh shims and the slack/lark-cli manifest merge already supply these variables per command, so the runtime process doesn't need them.

### 7. A refresh and a recheck running on two Server instances can destroy a healthy connection
`services/github/github-refresh-store.ts:79-99`, `github-recheck-store.ts:82-100`, `github-maintenance-worker.ts:329-363`

`completeRefresh` bumps `recheckGeneration`, but `claimRefresh` doesn't, and `commitRecheckResult` ignores `refresh_status = 'claimed'`. GitHub invalidates the old token pair as soon as the refresh token is used.

Scenario, with two instances or a rolling deploy:
1. Instance A claims the refresh and calls GitHub.
2. Instance B's recheck calls `/user` with the old token, gets a 401, and commits `unauthorized`. Its generation still matches, so the row becomes `reauthorization_required` and the credential is cleared.
3. A's `completeRefresh` returns `not_active`, and the new pair is thrown away.

*Reproduced* in that exact order: the recheck commit was `{applied:true}` and completion was `{applied:false, reason:"not_active"}`. `github-refresh-recheck-race.test.ts` only covers a recheck that commits after the refresh has committed. Bump `recheckGeneration` in `claimRefresh`, or treat `unauthorized` as transient while a claim is live.

### 8. One tenant can exhaust server-wide in-memory limits for everyone
- `runtime-credentials/ticket-store.ts:45,57`: 1,024 tickets server-wide, and `runtime-credential-owner.ts:238-246` issues them with no per-execution cap.
- `url-handle-store.ts:53,68`: 4,096 handles server-wide with a 15-minute TTL, no per-execution quota, and handles minted before a capacity error are kept.
- `execution-registry.ts:39-40`: 1,024 executions, with a 24-hour lifetime.

*Reproduced:*
- One execution requesting 1,100 tickets made a different execution's ticket request fail with `owner_unavailable`.
- One `conversations.history` page on a file-heavy channel (about 700 files × 6 URL fields) filled the handle store, and every other execution's file-bearing Slack read returned `upstream_unavailable` for 15 minutes.

The handle case doesn't even need intent. These need per-execution or per-connection quotas before this runs multi-tenant.

### 9. GitHub admission calls the GitHub API on every proxied request and every 5-second stream check
`runtime-credentials/credential-broker.ts:222-245`, `services/github-proxy/runtime-policy.ts:129-157`, `data-transport.ts` (`PROVIDER_PROXY_STREAM_REVALIDATE_INTERVAL_MS = 5_000`)

`#verify` calls `verifyCurrentRepositoryAdmission` for each bound repository, and that function checks live with the user token. It runs in `beginRequest`, again inside `GitHubProviderAdapter` → `policy.resolve`, again in `recheck`, and every 5 s for each open stream. I traced the call chain but didn't count real API calls; the reviewer estimate is about 6N calls per request and about 2N per stream every 5 s for N repositories. Either way, a normal `git fetch` plus a few `gh` calls on a 3-repo Agent will quickly exhaust the user's 5,000/hour limit. Admission then fails and capabilities and streams get revoked. Cache the admission proof per `(connection, authorizationVersion, repository)` for a short TTL; `authorizationVersion` already carries revocation.

### 10. The Agent repositories page can silently grant or remove repository access
`apps/web/src/features/integrations/agent-github-repositories.tsx`
- **Load more removes existing access (`:258-261`, `:295-313`, `:329-341`).** Drafts are built from page 1 only. `loadMore` appends repositories without adding their drafts, and `save` treats every loaded repository as editable, so `mergeBindings` removes the page-2 scopes. *Reproduced:* after Load more, Save turned on with no edits, and `PUT bindings=[]`.
- **Switching Agents saves the previous Agent's drafts to the new Agent (`:266-293`).** The route renders `<AgentGitHubRepositories agentId>` with no `key`, so the instance is reused. While B loads, A's view and drafts stay on screen, and Save writes them under B. *Reproduced:* the PUT added `{agentId:B, role:"code", access:"read"}`. `load` is also never cancelled, so a late response from A overwrites B. Key the component by `agentId`, or drop out-of-date results.
- **A pending connection can't be completed from this page (`:500-505`).** Any non-null connection sends `intent: "reauthorize"`, but `assertFlowIntentAllowed` only allows `create` on a `pending` row, so it gets 409 `CONNECTION_STATE_INVALID` every time. I confirmed this in code; the Account page correctly uses `create`.

---

## Non-blocking
- **Recheck bumps `authorizationVersion` every tick** (`github-recheck-store.ts:293-304`). While a stale binding remains, every 5-minute recheck bumps it, which revokes every execution on every repository of that connection and makes bindings saves return 409. *Reproduced:* 3 ticks took it from 2 to 5. Bump only when the verdict changes. Separately, `classifyRecheckFailure` maps `APP_IDENTITY_MISMATCH` (transient) and page-budget `INPUT_INVALID` to `permission_revoked`.
- **IM authority ignores `reportedAt`** (`execution-authority.ts:74-81,100-110`). A finished Turn's delivery still authorizes Slack/Feishu for up to 24 h. The GitHub path does check `isNull(reportedAt)`.
- **Provenance and path handling.** Slack source recording reads `body.channel ?? body.channel_id ?? query.channel`, but upstream honours a different field. *Reproduced:* `?channel=C_SECRET` with body `channel_id=C_PUBLIC` recorded `C_PUBLIC`. Feishu `operation-registry.ts:67-83` rejects a literal `..` but not `%2e%2e`, which `fetch` normalizes. *Reproduced:* `…/reactions/%2e%2e` reached `DELETE /im/v1/messages/om_1/`, which recalls the message.
- **Feishu tenant-token exchange** (`feishu-tenant-token.ts:47,111-116`). `input.signal ?? AbortSignal.timeout(10_000)` means no timeout once a request signal exists. The shared in-flight exchange is also bound to the first caller's signal, so one cancel fails every waiter.
- **Timeouts and buffering.** Feishu stream uploads get the flat 30 s timeout, versus 120 s for Slack upload handles. Queued upload bytes on a closed stream are never released (`data-transport-queue.ts:20-28`).
- **Clock skew** (`runtime-credential-relay.ts:292,412`). Server `expiresAt`/`refreshAfter` are compared with local `Date.now()`. More than 30 s of forward skew renews in a tight loop; more than 60 s rejects every stream.
- **Empty global git config** (`runtime-proxy-material.ts:79-80`). `GIT_CONFIG_GLOBAL` pointing at an empty file plus `GIT_CONFIG_NOSYSTEM=1` drops `user.name`/`user.email`, so agent commits are attributed to `user@hostname`. *Reproduced.* git-lfs and `safe.directory` go too.
- **Native `gh` coverage.** `gh pr view <n>` and `gh pr view --comments` are denied, because they ask for `autoMergeRequest`, `authorAssociation`, `isMinimized` and so on. `gh pr diff` gets JSON back, because `api-transport.ts:23` forces `Accept: application/vnd.github+json`. Worth saying in the docs if it isn't fixed. `#graphql` also mints a `pull_requests: write` token before checking for any write scope.
- **GitHub connection lifecycle.** Disconnect and replace only null the ciphertext and never revoke the tokens at GitHub. `x-github-delivery` isn't used for replay protection, so a captured signed `revoked` delivery can be replayed (denial of service only). `verifyAdmission` never compares the stored `fullNameDisplay` with GitHub's `full_name`.
- **Sandbox workspace check** (`sandbox-credential-bridge.ts:361-362`). It rejects `$HOME` itself but accepts a parent such as `/home` or `/Users`.
- **Web page details.** The Context Tree branch defaults to `refs/heads/master` instead of `repository.defaultBranch`. `imBinding(...).catch(() => undefined)` turns a 5xx into "needs a messaging binding". A schema violation (e.g. two Tree roles) just disables Save with no message.
- **Docs.** The zh README link sits above "运行时与消息" rather than inside it, and the sync date wasn't updated. The EN/zh limits, environment variables and defaults otherwise match each other and the code.

## CI
On `50241619`, CI failed three ways:
- **Host-HOME assertion** in `cloud-sandbox-credential-bridge.test.ts:410`: `e0fd9e03` fixes it structurally, which is the right fix.
- **Server branch coverage** at 79.18%, below its 80% floor.
- **Doc-mirror check**: git `ENOBUFS` on this large diff.

The PR description says `e0fd9e03`/`ce3f8f12` resolve the last two. Checks, Patch Coverage and the Node compatibility jobs were still running on `ce3f8f12` when I posted this, so I haven't confirmed that.

PR #634, opened 2026-09-17
fix(server): remove persistent integration control volume

yuezengwu:
## Summary

The platform integration runtime introduced a Server-local control database and an anonymous Docker volume in #633, making Server recreation depend on another persistent store. Remove that dependency: connection credentials remain in the existing PostgreSQL models and deployment configuration, while Git verification uses a private temporary workspace created only on demand.

- Remove the file-backed Session control store, Cloud credential issuer, generic source/write journals, control-directory configuration, and image `VOLUME`. No database changes.
- Keep scoped GitHub App IATs, Git/Tree validation, execution fences, and native CLI protocols. Make unknown write outcomes explicit and perform one upstream write attempt per proxy request.
- Accept Cloud control authentication only through an explicitly injected trusted verifier/activity checker. Keep the existing Local authentication and separate Cloud Runner bootstrap path. Synchronize English/Chinese deployment documentation.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` — passed (5,442 unit/script tests).
- `pnpm --filter @opentag/client test:agent-runtime:coverage` — 100% statements, branches, functions, and lines.
- `pnpm --filter @opentag/server test:integration` — 29 files / 412 tests passed against disposable PostgreSQL.
- Documentation mirror check and `git diff --check` — passed.
- Built the actual Server Dockerfile and tested GitHub disabled, synthetic GitHub App configuration, and Server recreation against the same disposable database. All three ran as UID 100 with a read-only root, only `/tmp` tmpfs writable, no image volumes or persistent Server mounts, and empty startup temporary directories. Health/readiness and offline native Git commit/push/clone with matching SHA passed.
- Regression tests cover unknown native Git/API write outcomes, exactly one upstream write call, pre/post-send authorization fences, failed initialization cleanup, and closing during directory creation. The latter reproduced a failure before its fix.

No staging/production deployment or real-provider write acceptance is claimed by these local tests.

## Breaking changes

- Removes `OPENTAG_RUNTIME_CONTROL_DIRECTORY`, `FileCloudControlAuthority`, and the internal generic journal/source recorder ports. Deployments no longer need an integration control volume.
- The proxy no longer persists write/source history or blocks later requests to a resource after an unknown write. Callers must reconcile unknown outcomes with the provider before deliberately retrying; there is no cross-request or cross-restart exactly-once guarantee.
- Default startup injects no trusted Cloud control verifier/activity checker, so that integration authentication path fails closed until orchestration supplies one. This does not provision Cloud credential issuance or replace the separate Runner bootstrap channel.

## Checklist


liuchao-001 (approved):
Reviewed at `db06387c`. Approving: no blocking findings.

**What I checked**
- **The removed persistence wasn't doing anything.** Nothing ever read `listSources`, and nothing in production code ever called `reconcileWrite` or `FileCloudControlAuthority.issue`. So on the base branch, one unknown write blocked that resource for the rest of the Session with no way to clear it, and no Cloud credential was ever issued. Removing the store drops no live behavior.
- **Write outcomes are classified the same way as before.** `classifyWriteOutcome` and `classifyStatusWriteOutcome` match the old `classifyWriteReceipt` and `classifyStatusWriteReceipt` exactly. The IM adapter now maps post-send transport failures, the `AbortSignal.timeout` path and unparseable or oversized write responses to `write_outcome_unknown`. Pre-send material and fence failures keep their own codes. `forwardUpload` also improves on the old code here: a failed recheck no longer turns into "unknown".
- **GitHub writes.** The pre-send recheck and abort now sit outside the try, so they report correctly as "not sent". Everything after the send becomes `write_outcome_unknown`. `#remember` now runs only on a confirmed success and only on the sanitized value. The stricter GraphQL success rule (every top-level mutation field must be a non-null object) is a real improvement.
- **Git publication.** Receive-pack now separates `rejected` from `unknown` in the `ng` text, and a pre-send revalidation failure is correctly reported as a definite rejection.
- **`GitWorkspace` close ordering.** `#staging` is incremented synchronously before the first await. Close waits for in-flight `mkdtemp` calls before the recursive `rm`, so a caller can't get a path back after close.
- **Cloud control fails closed.** With no `cloudControl` injected, both `cloudControlActive` and the Cloud auth branch reject. Local is unchanged.

**Non-blocking**
1. **`/tmp` capacity is now a deployment requirement.** Git staging moved from a disk volume to `os.tmpdir()`. The per-operation budgets are unchanged: 4 concurrent read snapshots × 512 MiB, 4 tree-head checks × 512 MiB, and publication at 512 MiB plus a 64 MiB spool. In the hardened layout from your test (read-only root, tmpfs `/tmp`), that data lives in RAM and counts against the container's memory cgroup. A large repository could get the Server OOM-killed instead of failing with `resource_limit`. Consider documenting the worst case and recommending a size-limited tmpfs or a disk-backed `emptyDir`/`TMPDIR` in `platform-integrations-foundation.md` and its zh-CN mirror.
2. **Stale e2e fixture.** `scripts/e2e/cloud-computer/cloud-identities-fixture.mjs:201-204` still sets `OPENTAG_RUNTIME_CONTROL_DIRECTORY`, and its comment says the Server resolves a `.opentag-control` default. The variable is now ignored, so both the line and the comment are dead and misleading.
3. **Lost `tree_invalid` signal.** In `GitHubProviderAdapter.#write`, a successful write followed by a failed `#recheckWrittenTree` used to rethrow `tree_invalid`. It now becomes `write_outcome_unknown`. Reporting "unknown" is safe, but the caller loses the hint that the write landed and the resulting tree is invalid. Fine to leave as is if that's intended.
4. **No revocation push in the new port.** The old authority had an `onRevoked` hook that closed the live control connection. `TrustedCloudControlAuthority` has no equivalent, so once orchestration injects a real authority, revocation only takes effect on the next frame or request check. This is worth stating in the port contract when that wiring lands.

CI was still running at the time of this review.

PR #635, opened 2026-09-17
fix(client): keep Codex turn hydration out of the resume response

bestony:
## Summary

A long Codex Session could not be resumed at all, so an IM message that needed a cold Runtime failed
with `provider_start_failed` and `executionEffects: not_started`, and no reply was ever sent.

Mechanism: the Client resumed the Session by asking Codex for its stored thread, and Codex returns the
entire history inside the single `thread/resume` response line. Past `CODEX_APP_SERVER_MAX_LINE_BYTES`
(1 MiB) the App Server reader fails that line as a protocol error, `factory.resume` throws, and the Turn
reports a start failure.

Observed on staging (agent `opentag-staging-logging`, session `63487734-…`, thread `01a0acba-…`):

| Timeline | Result |
| --- | --- |
| Turn answered with the Runtime still live | replied 10:17 local |
| Two auto-updates restarted the daemon | in-memory Runtime lost |
| First cold resume, 14:18 local | failed in 2.8s: `Codex emitted an oversized JSONL line` |

Measured on that same thread: the resume response was **1,385,957 bytes** against a 1,048,576-byte limit,
and **6,148 bytes** with `excludeTurns: true`, with the thread id preserved either way. This is a
threshold, not a flake: the failure only appears once a Session history is large *and* a Runtime has to
be started cold, which is why it hid behind a daemon restart.

The fix asks Codex not to hydrate turns on resume. OpenTag injects its managed prompt per Turn and never
reads hydrated turns, while the thread itself stays stored on the Server, so requesting them only risked
the line limit. This is also what the App Server's own deprecation notice instructs clients to do
(`excludeTurns`, then page with `thread/turns/list` and `thread/items/list`).

The branching moved into a small `codexThreadParams` helper so the factory's `#open` does not gain
complexity.

## Testing

- Red first: the new test fails without the product change with the exact production chain —
  `CodexAppServerError: Codex emitted an oversized JSONL line` → `AgentRuntimeError: Codex resume failed`.
- `pnpm --filter @opentag/client test` — 108 files, 1503 tests passed (includes the new regression test).
- `pnpm --filter @opentag/client test:e2e:codex-agent-runtime` — passed against the real `codex-cli
  0.154.0`: probe ready, create run completed, resume run completed, binding preserved.
- `pnpm check`, `pnpm build`, `pnpm typecheck` — passed (typecheck: 8/8 tasks).
- `node scripts/check-complexity-ratchet.mjs` — 141 warnings / **15** improvements, i.e. one more
  improvement than the committed baseline and no new or worsened warning.
- `biome check` on the three changed files — the same 5 pre-existing warnings as the untouched baseline.

The regression test drives the real `CodexAppServerProcess` against the offline App Server fixture. The
fixture's new `history-large` scenario hydrates turns unless the client opts out, so the test exercises
the real reader and its 1 MiB limit instead of asserting that a parameter was sent.

Not run locally: the repository-wide `pnpm test` (its Server integration suite needs PostgreSQL), which
CI covers.

## Breaking changes

None. Resume responses no longer carry hydrated turns, and no OpenTag caller read them. Compatibility is
tolerant in both directions: Codex ignores unknown params, so an App Server version that does not read
the field keeps its previous behavior rather than failing (verified against `codex-cli 0.154.0`, which
accepts an unknown param on `thread/start`).

## Checklist

Notes on the checklist: `pnpm test` was run for the affected package rather than the whole repository (see
Testing); no canonical document describes this wire detail, so no mirror changed. The behavior is
documented at the call site, where a future reader will meet it.


packages/client/src/providers/codex/agent-runtime.ts
```diff
@@ -1163,6 +1163,19 @@ function parseProviderConfiguration(value: JsonValue | undefined): CodexProvider
...
+ * Turn hydration stays on the Server: Codex otherwise returns the entire stored thread inside one
+ * `thread/resume` response, and a long Session exceeds the App Server JSONL line limit, which fails
+ * the resume as a protocol error. Only create and resume share the rest of the param shape.
+ */
+function codexThreadParams(
+  method: "thread/start" | "thread/resume",
+  expectedThreadId: string | undefined,
+): Record<string, unknown> {
+  if (method === "thread/start") return {};
+  return { ...(expectedThreadId ? { threadId: expectedThreadId } : {}), excludeTurns: true };
```
yuezengwu: [P2] Restore the required Agent Runtime branch coverage

Extracting this conditional into `codexThreadParams` introduces an uncovered false arm at line 1176. `pnpm --filter @opentag/client test:agent-runtime:coverage` exits 1 at this exact head on both macOS and Linux: 435/435 tests pass, but branches are 99.96% against the required 100%. The local coverage data records this conditional as `[11, 0]`. Cover the missing-id resume case if it is intentionally supported, or simplify away the redundant fallback. The green patch-coverage job does not satisfy this separate gate; the current Checks job skipped it after its PostgreSQL test failure.

yuezengwu (changes requested):
Reviewed `f0de0ae7b42a3baabd76f950fe5bb4da489475ac` in an isolated worktree. The `excludeTurns` field is present in the locally generated Codex 0.153.4 protocol schema, the oversized-history regression passes, and Client typecheck passes.

The required Agent Runtime coverage command still exits 1 on both macOS/Node 24.20.0 and a disposable Linux/Node 24.19.0 container with init: all 435 tests pass, but branch coverage is 99.96% against the 100% gate. The inline finding identifies the newly uncovered branch.

The current [Checks run](https://github.com/first-tree-ai/opentag/actions/runs/35193915992/job/105112546671) also fails earlier in PostgreSQL integration: `im-binding.test.ts:4948`, the expired-dispatch recovery case, times out at 5,000 ms (411 passed, 1 failed). That Server code is unchanged by this PR, and the subsequent Agent Runtime coverage step was skipped. Please address the coverage regression and get the current-head CI checks green before re-review.

bestony:
Both points addressed at `90211a71`.

**Coverage (P2)** — reproduced your command at `f0de0ae7`: 435/435 tests, 2974/2975 branches, exit 1 at `agent-runtime.ts:1176`. Took your second option and removed the fallback rather than covering a case an exact resume cannot have. The helper now returns the thread id unconditionally on resume, so the unexercised arm is gone instead of newly covered:

```ts
return method === "thread/start" ? {} : { threadId: expectedThreadId, excludeTurns: true };
```

Behavior is unchanged, because `JSON.stringify` omits the key when it is absent — the same request an unbounded resume sent before. At this head `pnpm --filter @opentag/client test:agent-runtime:coverage` reports 435/435 tests and **100% branches (2973/2973)**, exit 0; the denominator drops by exactly the two arms that disappeared.

**PostgreSQL integration** — confirmed as a load-sensitive timeout on this head rather than a product break. `im-binding.test.ts > recovers a retained Client Turn from a pending|expired dispatch after rebuilding the Server owner` passes locally with CI's concurrency (`vitest run src/__tests__/integration/im-binding.test.ts -t 'recovers a retained' --maxWorkers=2` → 2 passed, 4.86s of tests), and the diff touches nothing under `packages/server`. The 5000ms default is what it exceeds on CI, which is also why the Agent Runtime step after it was skipped.

Re-verified at `90211a71`: real `codex-cli 0.154.0` e2e smoke passes (probe ready, create and resume runs completed, binding preserved), client suite 1503/1503, complexity ratchet 141 warnings / 15 improvements, `biome check` unchanged, client typecheck clean. Thanks for checking the field against the 0.153.4 protocol schema — that is better compatibility evidence than I had.

bestony:
All checks are green at `90211a71` (16 passing): `Checks` — including the PostgreSQL integration stage and the previously skipped Agent Runtime coverage step — plus patch coverage, quality scoreboard, and Node 22/24/26 compatibility. `ownership-gate` is the only red signal, and it is the approval this review request asks for. Review re-requested from @yuezengwu.

yuezengwu (approved):
Re-reviewed the complete three-file change at `90211a713af5ffd26aff28b0c00b790a11cb2287` in a clean, isolated worktree. The previous coverage finding is resolved: the redundant fallback is removed, and the resume request still preserves the bound thread id while excluding turn hydration.

Independent validation: Client typecheck passes; the Agent Runtime suite passes 435/435 tests with 100% statements, branches, functions and lines (branches: 2973/2973). The oversized-history resume regression passes. The current-head GitHub Checks job also passes both PostgreSQL integration and the previously skipped Agent Runtime coverage step. All code checks are pass/skip; ownership approval is the remaining gate.

No further blocking findings. This approval supersedes my change request on `f0de0ae7`; no merge or deployment is performed.

PR #637, opened 2026-09-17
feat(cloud): execute IM turns in session sandboxes (E4)

yuezengwu:
## Summary

Cloud-bound IM messages can start an existing Session Sandbox and execute a Pi turn through its authenticated Runner. The Runner fsyncs the exact input before Server custody verification, executes only after verification, and retains the Turn report until a matching durable acknowledgement. A crashed started turn is reported as unknown and is never replayed.

- Reuse Computer, Agent, Session, Sandbox, IM delivery, custody and credential-owner models. One Instance serves one Session Sandbox at 1 vCPU / 1 GiB. No migration, dependency, separate scheduler or Server control volume.
- Bind credentials and model access to the exact connection, Sandbox and execution. The model proxy uses a fixed upstream, strict request fields and bounded generation parameters, including the pinned Pi reasoning-history and signed tool-call fields. Master/provider credentials stay outside the native Sandbox. Connection loss revokes grants; unstarted input needs fresh verification after reconnect.
- Keep pending input bounded by existing TTL/capacity before dispatch, preserve dispatched windows and accepted-unreported custody, reject permanent environment failures explicitly, and prevent Cloud follow-ups from blocking other tenants. Recovery resends missed cancellation and accepts real results during allocation release. IM reauthorization pauses new permissions while preserving input/custody; a report-only Runner reconnects through the existing control handshake once authorization returns. Explicit allocation release still cancels paused work. Transient Cloud dispatch failures back off from 2 to 30 seconds using the existing attempt counter.
- Use bounded production Unix socket paths, enforce journal capacity at write time, and check stop/revocation throughout startup. Preserve Pi history in the Session workspace; clean temporary credentials and verify native namespace reset after interrupted execution before reporting or admitting another turn. Legacy E3 revocation behavior remains intact when E4 is not negotiated. A cancelled queue head advances immediately, and all shutdown paths mark the Runner as stopping before cleanup.

## Validation

Head: `18c00d9c920dc810de0b5fea42410c61047afe69`. All required local gates, normal Git hooks and code CI pass. The ownership approval gate still requires bestony.

- Full local unit suite: **5,664 passed** (403 repository script tests + 5,261 package tests); full PostgreSQL: **423 passed / 31 files**, no retry.
- Runner regressions: **95 passed**. Cancelling a queued head advances the next turn; a real WSS authentication failure during a live turn shuts down without relaunching the native namespace. Both regressions fail with the old logic.
- Server focused suite: **132 passed**; real WSS: **42 passed**; real PostgreSQL focused suite: **10 passed**. The WSS scenario verifies paused reconnect, restored authority, no premature grant, fresh active readiness and credential-open delegation.
- Existing Agent Runtime scope: local **434 passed, 100%**; CI merge checkout **435 passed, 100%**. The merge checkout includes additional tests from main.
- [Code CI](https://github.com/first-tree-ai/opentag/actions/runs/35213743989) passed: Checks (including PostgreSQL **423/423**), Node 22/24/26, CLI/browser/container smokes. Docker, Runner Toolchain and Quality Scoreboard passed.
- [Changed-line coverage](https://github.com/first-tree-ai/opentag/actions/runs/35213743597): **96.52% (3,800/3,937)** against the unchanged 80% threshold; verified the actual nonzero-denominator result in the CI log.
- Actual installed Pi 0.84.2 SDK, actual model proxy/grants and deterministic local SSE fixture: **4/4 two-call histories pass**, including `reasoning`, `reasoning_text` and signed `reasoning_details`. Three failed with HTTP 400 before the fix.
- Independent model-parameter probes retain rejection of routing/credential overrides and `n > 1`; oversized and omitted output budgets remain capped at 65,536.
- No threshold, assertion, coverage exclusion or retry was relaxed. No migration or dependency change.

## Rollout and remaining boundaries

Opt-in and disabled by default. Deploy Server capability support first, then a pinned E4 Runner image; E3 Runners retain their legacy welcome shape and scope checks. This PR does not provision configuration or enable staging execution.

GCS workspace save/restore (E5), multi-Turn placement, idle recycling/reuse and Context Tree synchronization remain follow-ups. Previously used Instances fail closed until restore exists. Native Cloud Run and real IM/provider acceptance remain pending unified configuration; local tests do not establish that acceptance.

Provider writes follow #634: one upstream attempt and explicit unknown outcomes, without an exactly-once effects guarantee. Active model/tool calls may fail after control connection loss. The 4,096-entry grant cap protects Server state; per-Account admission/fairness and aggregate spending limits remain separate resource-policy work. Completed turns can retain background processes within their own Session.

## Checklist

- [x] Focused change; no unrelated implementation.
- [x] Tests and synchronized English/Chinese documentation cover the behavior.
- [x] Required full local gates pass for the published source.
- [x] Code CI verified for the published head; owner approval remains pending.
- [x] No credentials, private data or generated build output included.


liuchao-001 (changes requested):
Review of `a0e8f7c4`. The durable core holds up well: journal fsync ordering, the started-marker/no-replay invariant, ack matching, custody SQL fencing, grant ownership across reconnect, and proxy upstream hygiene all checked out. But I found three blocking issues: one regresses the existing E3 path even with E4 disabled, and two make E4 non-functional or harmful the moment it is enabled. Requesting changes.

## Blocking

**1. E3 regression: a live Runner socket is no longer closed on revocation, even with E4 off.** `packages/server/src/api/runner-connection.ts` `#renewCredential` (~L371-397) falls back to `validateRunnerChannelScope` for *every* connection when `validateRunnerScope` misses. `#handleResult` (~L498) also moved to the channel-only check. At the merge-base both closed with `staleScope` ("A definitive miss is a revocation… must take effect on the live socket").
- **Scenario:** an E3 connection that never negotiated `cloudDeliveryVersion`. The Account is suspended, the Agent/binding is deactivated, or the Session ends while the allocation row is still current. The socket stays open, keeps receiving a fresh `server:credential` every renewal, and can still resolve acceptance results.
- **Evidence:** reproduced by running the merge-base version of the "closes the connection with staleScope when credential renewal finds the scope revoked" test (`lifecycle: "releasing"`, no `cloudDelivery`) against head. The socket was still open after 1.5s with 13 renewed credentials. The test in this PR was changed to `"unallocated"`, which hides it.
- **Suggested fix:** only take the channel fallback when the connection negotiated E4 (and ideally only while it holds unreported work). Otherwise keep the old close.
- **Related (non-blocking):** the `.catch(() => undefined)` on the channel check turns a transient DB error into a definitive miss and closes the socket. For E4 that revokes every grant over a DB blip; the base left the socket alone on errors.

**2. The default credential socket path exceeds the bridge's own 100-byte limit, so every production Cloud turn fails.** Cloud Run sets only `OPENTAG_RUNNER_BACKEND_URL` / `BOOTSTRAP_TOKEN` / `SANDBOX_NAME` (`cloud-run-admin.ts` ~L309-313). With the image's `TMPDIR=/tmp`, `serve.ts` L132 defaults `stateDir` to `/tmp/opentag-runner-state/<ot-p-32hex-gen>`.
- **Resulting path:** `…/bridge-public/turn-XXXXXX/connect.sock` is **104 bytes**, and `listenProxySocket` (`bridge-material.ts` ~L137) throws at `> SOCKET_PATH_MAX_BYTES = 100`.
- **Scenario:** `#executeTurn` has already marked the entry `started`, so every delivery becomes `unknown`/`turn_state_unknown` plus a full native reset, and is never retried.
- **Why tests miss it:** unit and local tests substitute the `openExecution` seam.
- **Suggested fix:** a short fixed state root (e.g. `/run/ot/<hash>`) or shorter per-turn names, plus a test that computes the production-default path length.

**3. One Cloud follow-up message stalls the IM delivery worker for all tenants.** `packages/server/src/runtime/im-delivery-worker.ts`: the claim query's second branch (~L361-385) selects a pending message whose Session has an accepted, unreported turn, so it can be steered. The steer guard (~L409-421) then needs `this.#registry.currentInstanceId(row.computerId)`. The Local registry never holds an instance for a Cloud Computer, so the transaction returns `undefined` **without advancing `nextAttemptAt`**.
- **Why it blocks everyone:** with `orderBy(nextAttemptAt, id).limit(1)`, every tick picks the same row and claims nothing.
- **Scenario:** a Cloud turn runs in Session A and the user sends a second message. An unrelated Account B's message (Local or Cloud) is not dispatched until A's turn reports, which can take up to the runtime budget (30 min default, 24 h max).
- **Evidence:** reproduced with PGlite and the real `ImDeliveryWorker` + `CloudDeliveryOwner` + `RunnerHub`. B got 0 `delivery:run` frames after 5 `runOnce()` calls; without the follow-up, B dispatched once.
- **Prior art:** the same shape existed before for an offline or non-steer Local runtime, but this PR makes it the default for every Cloud follow-up.
- **Suggested fix:** exclude Cloud placements from the steer branch, or bump `nextAttemptAt` when the steer guard declines.

**CI is red on new tests from this PR:**
- `runner-serve.test.ts > cleans the native namespace immediately after an interrupted turn and blocks queued turns and acceptance until verified` ("timed out waiting for first report") fails in Checks and in Node 26 Compatibility.
- `cloud-model-proxy.test.ts > aborts the upstream and releases the slot when the client disconnects mid-stream` fails in Node 22.22.2 Compatibility.

## Should fix

4. **A queued turn starts with a grant from the dead connection.** `cloud-turns.ts`: the queue stores the whole verified frame, grant included (~L294). `onChannelClosed` (~L424) only logs, and admission checks only `expiresAt` (~L681). The server revokes every grant owned by the old connection (`cloud-delivery-owner.ts` ~L769-771).
   - **Scenario:** X is running and Y is verified and queued on connection A. A drops. X finishes before B's re-verification lands, so Y starts with A's revoked token, becomes `started`, and fails with `may_have_occurred`. The message is burned.
   - **Evidence:** confirmed with a scratch test.
   - **Fix:** clear `#queue` on channel close. The entries stay `received` and are re-verified with fresh grants.
   - **Related:** the `onChannelClosed` comment and the "keeps a live execution alive across a channel drop" test suggest a running turn survives a reconnect. The docs correctly say model/tool calls may fail, so the comment and test overstate it. The test only passes because the seam does not enforce grants.
5. **The model allowlist only checks the top-level `model` field.** `cloud-model-proxy.ts` (`.loose()` ~L29-34, `{ ...body, model }` ~L117) forwards every other sandbox-controlled field to the upstream, which is called with the master key.
   - **Evidence:** a scratch test got `models: [...]`, `route: "fallback"`, `api_base`, `api_key`, `n: 128`, `max_tokens: 1e9` through to the upstream unchanged.
   - **Impact:** on OpenRouter/LiteLLM-style routers that bypasses the allowlist or redirects the call. On any upstream it is uncapped spend on the master key.
   - **Fix:** a strict field allowlist, with `n` and `max_tokens`/`max_completion_tokens` clamped.
6. **Cloud pending messages never reach a terminal state and are unbounded.** The claim keeps Cloud rows past `expiresAt` (`im-delivery-worker.ts` ~L338), while expiry (`im-delivery-janitor.ts` ~L183) and overflow pruning (`im-message-inbox.ts` ~L671/~L746) are Local-only. `restore_required` / `stopped` / model-unavailable / runner-not-ready all go through `recordFailure`, which retries every 2s.
   - **Scenario:** a user stops a used environment, and every later message becomes a pending row retried every 2s forever.
   - **Doc mismatch:** this contradicts "restore_required … never a retry loop" and the "`expiresAt` remains the only pending deadline" comment.
7. **A started Cloud turn is never cancelled if the stop's cancel frame was missed.** `cancelSessionDeliveries` only sends `delivery:cancel` to a currently attached connection (`cloud-delivery-owner.ts` ~L497). Recovery treats a `started` answer as `pending` (~L588-597) unless the allocation is released.
   - **Scenario:** an Agent is suspended while its Runner is briefly disconnected. After reconnect the native process keeps running until the Runner deadline (up to 24 h), just without model access.
8. **A credential-execution open is not bound to the sending connection's Sandbox.** `runtime-session-execution.ts` ~L124-139 compares the Runner-supplied `frame.sandbox` with the target Session only, not with `connection.scope.sandboxId`. `delivery` sources are covered indirectly by custody `instanceId`; `session-message` sources are not.
   - **Impact:** defense-in-depth for isolation between Sessions. The fix is a one-line check in `handleCredentialFrame`.

## Nits (non-blocking)

- **Custody before grant mint:** `acceptDelivery` commits before the grant is minted (`cloud-delivery-owner.ts` ~L279-292). A mint failure is later settled as `unknown`/`may_have_occurred` for a turn that provably never ran. `#reverifyReceived` silently returns on mint failure (~L658).
- **`releasing` settles too early:** accepted turns are settled `unknown` immediately (~L571) while reports are still accepted in that state (~L427), so the real report then gets a `conflict` ack and stays journaled.
- **Journal cap enforced only on read:** `recordReceived` can write entry 1,025, after which every `list()` throws and nothing can retire entries (`cloud-journal.ts` ~L177 vs ~L196).
- **Cancel during start:** a cancel that arrives during `read`/`markStarted` still marks the entry started and runs the worker. Checking `signal.aborted` before `markStarted` would report `not_started`.
- **Deadline reported as unknown:** the parent exec timeout starts before the worker's own timer, so a deadline turn is reported `unknown` rather than `turn_timeout`.
- **Shutdown order:** shutdown destroys the sandbox before `turns.close()` (`serve.ts` ~L339 / ~L413).
- **Heartbeat before `auth:result`:** `hub.attach` runs before `#attachCloudFence`'s DB reads, so a heartbeat can reach an E4 Runner before `auth:result` and trigger a reconnect.
- **Global grant cap:** the 4,096-grant limit is shared by all tenants with no per-account limit (`cloud-model-grants.ts` ~L39).

Evidence: code reading at `a0e8f7c4`, plus throwaway local vitest runs for #1, #3, #4, #5 and the journal-cap nit (not committed). No native Cloud Run or real-IM run.

liuchao-001 (commented):
Re-review of `aeaf8fd9`, against my review of `a0e8f7c4`.

**All three blockers are resolved, with regression tests that would fail on the old code:**
- E3 revocation: a non-negotiated connection closes on a definitive miss again, transient DB errors leave the socket alone, and acceptance results use the active check. There are new E3 tests for Session end, `releasing`, a result after end, and a transient error.
- Socket path: the state root is now `$TMPDIR/ots/<≤41>`. The worst case for the longest name `INSTANCE_ID_PATTERN` allows is 89 bytes for `connect.sock` and 87 for `slack.sock`, and each environment generation still gets its own directory.
- Cross-tenant stall: the steer branch is Local-only, and a declined steer is deferred instead of reselected.

The should-fix items are also addressed:
- queue cleared and generation-fenced on channel close
- strict proxy fields with `n=1` and output-token clamping
- Cloud pending input bounded by TTL and capacity before dispatch
- missed cancels re-sent by recovery for `received` and `started`
- execution open bound to the connection's Session, Sandbox and generation

So are the nits: mint before custody, `releasing` staying recoverable, the journal cap enforced on write, a cancel before `markStarted` reporting `not_started`, the deadline grace, and `hub.attach` after the fence reads. The per-account grant cap is explicitly deferred in the description; that's fine as separate resource-policy work.

Two of the fixes introduce new problems, and there is one semantics question. None is as serious as the originals, but each is small to fix, so I'd like them in before approving.

## Should fix

**1. A cancel during a queued head's start stalls the rest of the queue.** `packages/client/src/runner/cloud-turns.ts`: the new cancelled-before-start branch in `#startTurn` (~L506-513) settles `not_started` and returns `false`. `#processQueued` maps `false` to `"stop"` (~L620-621), so `#drain` exits. Nothing schedules another drain: the `finally` deliberately doesn't, and the `.finally(scheduleDrain)` only exists when an execution started.
- **Scenario:** X is running; A and B are verified and queued. X finishes and the drain starts A. `delivery:cancel` for A arrives during its journal read, and A correctly reports `not_started`. But B stays queued and `hasPendingWork` stays true, so acceptance gets `runner_busy`. B only moves on an unrelated verified frame, availability signal or reconnect; otherwise its grant and deadline run out.
- **Evidence:** reproduced with a throwaway test: `started: ['X']`, B never started, `hasPendingWork: true` after 1.5s. The previous code ran A and drained B from its `.finally`.
- **Fix:** return `"continue"` for the settled-cancel case, or schedule a drain there.

**2. The strict assistant-message schema rejects reasoning echoes the pinned Pi sends.** `packages/server/src/api/cloud-model-proxy.ts` (~L93) allows only `reasoning_content`. `@earendil-works/pi-ai@0.84.2` `dist/api/openai-completions.js` records whichever of `reasoning_content` / `reasoning` / `reasoning_text` the upstream streamed (L352). It writes that text back under the same key on the next request (`assistantMsg[signature] = …`, L923), and adds `reasoning_details` for tool calls carrying signatures (L970).
- **Scenario:** `OPENTAG_CLOUD_MODEL_UPSTREAM_BASE_URL` is documented as any OpenAI-compatible upstream. With one that streams `reasoning`, as vLLM and OpenRouter-style gateways do, the second call of a tool loop gets `400 CLOUD_MODEL_REQUEST_INVALID` and the turn fails.
- **Test gap:** the tests only cover `reasoning_content`.
- **Fix:** allow bounded `reasoning` / `reasoning_text` (and `reasoning_details`, or strip it), or pin the supported upstream family in config and docs.

**3. A recoverable binding state now drops queued Cloud input and cancels running Cloud turns.** In `packages/server/src/runtime/im-delivery-cloud.ts` `#stoppedRows` (~L449), stopped authority includes `imBindings.status != 'active'`. That also matches `reauthorization_required`, which is set on a Slack token error and cleared once the user reauthorizes.
- **Scenario:** while a binding is in that state, the janitor permanently rejects queued Cloud input (`authority_stopped`), and `#recoverLive` sends `delivery:cancel` to started turns. Local rows in the same state stay pending until their TTL and are delivered after reauthorization.
- **Tests:** `im-delivery-worker-cloud.test.ts` (~L569/~L593) asserts this deliberately with `reauthorization_required`, so it may be intended.
- **Ask:** if it is intended, please state the product reasoning (a transient token error ends in-flight Cloud work but not Local work). Otherwise, limit the stop to `disabled`/deleted bindings.

## Nits

- **Non-signal shutdown relaunches the sandbox.** `serve.ts` ~L365: `turns.close()` now runs before `cleanupRunner`, which is what sets `state.stopping`. On a shutdown without a signal (auth rejected, reconnects exhausted, fatal state), aborting the live turn runs its interrupted-turn reset: destroy, relaunch and probe, because `stopping` is still false. Then `cleanupRunner` destroys again. A throwaway test saw `launch: 2, destroy: 2, probe: 2`. If the relaunch fails, the report becomes `unknown/sandbox_unavailable` instead of a cancellation. SIGTERM is unaffected. Setting `state.stopping = true` before `turns.close()` fixes it. The new shutdown test uses a signal, so it doesn't catch this.
- **Tight retry on dispatched rows.** A dispatched Cloud row that keeps hitting `runner_not_ready`/`model_unavailable` is re-claimed every 2s until its frozen window passes (up to the runtime budget). The count is bounded, but there is no backoff.
- **Silent mint failure.** `#reverifyReceived` still returns silently on mint failure (`cloud-delivery-owner.ts` ~L788). It is bounded now that expired windows lead to cancellation.

Evidence: code reading at `aeaf8fd9`, throwaway local vitest runs for #1 and the shutdown nit (not committed), and the pinned `pi-ai` 0.84.2 tarball for #2. CI was still running when I wrote this. No native Cloud Run or real-IM run.

liuchao-001 (approved):
Approving `18c00d9c`. Everything from my review of `aeaf8fd9` is resolved.

- **Queue stall:** `#startTurn` now distinguishes `settled` from `wait`. A cancelled, phase-lost or generation-lost head deletes its queue entry and the drain continues. Only closed, busy or `canStart=false` stops the drain, so the failing head can't spin. Covered by "continues draining the queue after a head is cancelled during its startup read".
- **Reasoning echoes:** the proxy now allows `reasoning` / `reasoning_text` and bounded `reasoning.encrypted` `reasoning_details`, and still rejects routing and credential fields at the top level and in each message. Covered by tests for the allowed shapes and for shapes Pi never sends.
- **`reauthorization_required`:** now a pause rather than a stop.
  - Pending input stays within its TTL and capacity, and is no longer rejected as `authority_stopped`.
  - Accepted, unreported turns stay pending in recovery instead of being cancelled.
  - A connection that authenticated during the pause is ineligible for dispatch, receipt authorization and grants.
  - Once authority is restored, a heartbeat, query or frame closes that connection with `RETRY_LATER` so the Runner reconnects normally.
  - Allocation release and real stops (Session ended, Agent inactive, Account suspended, binding disabled or deleted) still cancel.
  - This matches Local, and the end-to-end WSS test covers pause, restore, reconnect and no early grant.
- **Nits:** every exit now sets `stopping` before `turns.close()`, so shutdown no longer relaunches the sandbox. Transient Cloud dispatch failures back off from 2s to a 30s cap. The failed mint during re-verification now logs a sanitized warning.

The deferred items are stated in the description: the per-account grant quota, E5–E8, and native Cloud Run plus real IM acceptance. This PR ships disabled by default. Real-environment evidence is still needed before E4 is enabled in staging.

Evidence: code reading of the `aeaf8fd9..18c00d9c` diff and its tests. CI on this head was still running when I approved. Nothing was run on native Cloud Run or a real IM provider.

PR #640, opened 2026-09-17
feat: add execution-scoped Tavily web tools

yuezengwu:
## Summary

Add opt-in `web_search` and `web_fetch` tools to Pi through execution-scoped authorization and the existing private Router. Tavily credentials stay in Router; Account-mapped web-only tenant keys stay in Server. Agents and sandboxes receive a fresh keyless endpoint for each execution.

- Connect the built Pi extension, trusted Client/Runner transport, Server execution policy and Router API. Validate public URLs, enforce decreasing request deadlines and bounded output, and preserve tenant/execution ownership through cancellation and revocation.
- Save fetched pages atomically with hashes and bounded previews. Repeated tool calls reuse the same artifacts and request identity. Ship the trusted extension in npm and portable packaging while keeping implicit Pi extensions disabled.
- Use a sandbox-local socket over a dedicated `sandbox exec` duplex channel for NativeSandbox. Cloud web access remains fail-closed without explicit web execution authority; this PR does not enable it merely because E4 model execution is available.
- No new runtime dependencies, services, OpenTag tables or migrations, and no LiteLLM changes. Companion gateway and billing PR is merged: https://github.com/first-tree-ai/llm-router/pull/1. Deployment and database migration are separate steps.

## Testing

Validated current PR head `e837aacbf1f1488ef36d68f361e953cbed86a05f` after integrating main `b805a5906dccd5410c20c215f8d75100dda125a3` (MCP management plus CI/fixture fixes). The sole text conflict was the public-export snapshot: all **780 exports** match the actual barrel and preserve the union of MCP and Tavily exports. The subsequent main CI-only update at `7f6a29ae` also merges cleanly.

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`: passed; **5,965 tests** across scripts, Shared, Server, Web, Client and CLI.
- `pnpm --filter @opentag/server test:integration`: **498 passed** against disposable PostgreSQL (35 files).
- Linux `pnpm --filter @opentag/client test:agent-runtime:coverage`: **439 passed**, with 100% statements, branches, functions and lines. All 25 covered source files match this head; no thresholds or exclusions changed.
- The full suite includes the Runner build-context regression, which stages the actual production allowlist and verifies the required extension copy script is present. Its prior image-build CI passed at `65019382`; current-head CI runs after this push.
- Earlier npm pack/isolated install and relocated portable tool registration passed. Full cross-OS release archives and production Cloud sessions are outside this acceptance.

Real Tavily acceptance on the feature implementation at `c80ab7a06e33f2f6dbd666caa3afe7506b22fd1f`, before the main integration: **7 tool executions, 6 supplier requests and 6 ledger rows**. Basic/advanced search, English/Chinese extract, partial/all failures, retry reuse, bounded output and artifact hashes passed. The chain used Pi 0.84.2, the built extension, real Server/Router HTTP and isolated PostgreSQL/Redis; execution identity and the LLM driver were fixtures. Supplier request IDs and reported credits matched Router evidence and reservations returned to zero. This does not establish production session acceptance.

NativeSandbox transport was also exercised in a real sandbox with repository code and injected web authority: 1.5 MB of UTF-8 content, matching hashes, deadline, rejection, revocation and successor checks passed. The image was a pinned generic Node image, not the production Runner image.

## Breaking changes

None while disabled (the default). Activation requires the companion Router migration, explicit service prices and credentials, Account-to-Router tenant mapping, Server configuration and a Client with the negotiated web-tools capability. No production deployment or migration has been performed.

Cloud E4 model execution has landed on main. Production **web** execution authorization and dedicated Cloud acceptance remain a separate activation prerequisite. Supplier account debit also remains unreconciled: six real responses reported five credits, but usage counters did not reflect the calls; subsequent usage reads reached 429 and request logs returned 403 on the current plan. Retail test charges are not supplier cash charges.

## Checklist


liuchao-001 (changes requested):
Reviewed head `54ebc4f3`. I split the review across Server authorization, the trusted Client gateway, the Pi extension plus shared protocol, and the native Runner bridge. I checked the security-relevant claims against the code, and ran the scoped test files locally; they pass. Overall the design holds up. Tenant selection comes only from `execution.accountId`, and authorization is re-checked before and after the Router call. Deadlines never reset, responses are size-bounded while streaming, and errors are redacted. The native path really is fail-closed: `OPENTAG_RUNNER_WEB_TOOLS=true` alone opens nothing, production passes no `webAuthority`, and nothing in the E4 turn path requests `web`. The `serve.ts` merge resolution keeps E4's controller and bridge lifecycle intact.

One blocker, from CI.

## Blocking

**1. The Runner image no longer builds.** `Build linux/amd64 image and run offline smoke` fails on this head:

```
Error: Cannot find module '/src/scripts/copy-web-tools-extension.mjs'
ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  open-tag@0.0.5 build: `tsdown ... && node ../../scripts/copy-web-tools-extension.mjs`
```

`apps/cli` `build` now runs `scripts/copy-web-tools-extension.mjs`. The Runner build context is an allowlist (`scripts/runner/build-context.mjs:55-60`), and that script isn't on it, so the Dockerfile's `pnpm --filter ./apps/cli build` step fails. Adding the file to the allowlist should fix it. It would be worth a look at whether any other staged or allowlisted build context invokes the CLI `build` script. The CI failure on main (`15c32d08`) is only Secret Scan, so this failure is new with this PR.

## Non-blocking (worth fixing here or in a tracked follow-up)

2. **A web-only execution now opens for a Session that can never use it.** `runtime-session-execution.ts:86-93` no longer rejects when `providers.size === 0 && services.length > 0`. Previously the only thing that caught an inactive binding at open was the empty provider lookup. The per-call fence (`scope-resolver`) still rejects the same state, so every web call fails with `credential_scope_denied`, yet the open reports `succeeded` and Pi registers the tools. The test at `runtime-web-execution-open.test.ts:195` (`bindingActive: false`) asserts this state. Suggest running the Session fence (at least binding status) before granting a web-only open.

3. **Pi runs from Session messages never get web tools, but still start a gateway.** `session-message-inbox.ts:349` calls `runtime.prompt()` without the `webTools` configuration; only `agent-turn-runner.ts:286` passes it. The credential manager still requests `web`, and the Server grants it, so a per-execution socket is created and never used. Either wire it through or don't request `web` on that path. `docs/web-tools.md` says "Pi Agent Runs" without an exception. Relatedly, validation runs also request and receive `web` (`#requestedProxyServices` ignores the run's source kind, and `describeSessionServices` ignores purpose).

4. **A failure opening the artifact store turns a successful, already-billed fetch into a tool error.** `pi-extensions/web-tools.ts:1057`: `openAttemptStore` runs outside the per-page fallback. With a read-only cwd, or a `.opentag` symlinked elsewhere, `web_fetch` throws, and the tool error leaks an absolute host path (`EACCES ... mkdir '/private/var/...'`). The model then likely retries and pays again. The doc promises "a write failure reports a missing artifact instead of a fabricated path", so this should degrade to preview plus `saved: none`.

5. **Duplicate or fragment-variant URLs are forwarded, then the extension rejects its own results.** `urls: ["https://example.com/", "https://example.com/#a"]` is forwarded as-is (the shared schema allows duplicates), and `requestedUrlsMatch` (`web-tools.ts:1225-1242`) then throws `unrequested or duplicated URL` after the Router did the work. It builds occurrence counts but never uses them. Whether this fires in practice depends on whether the Router echoes duplicates; I couldn't check that from this repo. Deduplicating with the shared `normalizeWebTargetUrl` before dispatch would settle it.

6. **The URL policy validates the parsed URL but forwards the raw string.** `http://evil.com\@127.0.0.1/` passes, because the WHATWG parser sees host `evil.com`, and `targetUrlSchema` (`packages/shared/src/web-tools.ts:393`) only uses `superRefine` with no transform, so the Server forwards the raw input to the Router. A few non-public literals are also accepted: `::ffff:0:0/96`, NAT64 `64:ff9b::/96` with embedded private or metadata IPv4, `fec0::/10`, `192.88.99.0/24`, `.home.arpa`, `.onion`. Tavily does the fetching and the Router re-checks, so this isn't SSRF against our infrastructure. Forwarding the normalized `URL#href` is still the cheap, robust fix.

7. **The Server has no concurrency cap on web calls.** Each call can hold up to 45 s and 3 MiB in memory, and none of the three layers (`web-service.ts` / `web-router-client.ts`, or either gateway) limits in-flight calls per execution or Computer. IM provider calls go through `ExternalCallPolicy` (`maxConcurrency: 16`). A model fanning out parallel `web_fetch` calls gets N × 3 MiB buffered and N billed calls.

8. **Native gateway ownership after the open await.** This isn't reachable in production today, because no authority issuer exists; it should be fixed before one lands. Both cases were reproduced with a harness.
   - `runner/web-gateway.ts:442`: if the authority aborts during bridge startup, `open()` closes the channel. The `onClosed` guard compares `undefined === undefined`, and `#active` is then set to the closed channel. Every later `openExecution` throws "already has an active execution channel" for good.
   - Similarly, `gateway.close()` during an in-flight open returns immediately, and the open then installs a live, authorized channel that nothing tears down.
   - Suggested fix: after the await, `if (this.#closed || channel.closed) { await channel.close(); throw ... }`.
   - Also `web-gateway.ts:249`: a duplicate in-flight request id overwrites the `#inflight` entry, so the first dispatch survives cancel and revocation until its own deadline. Treat a duplicate id as a protocol violation.

## Nits

- `web-tools.ts:135,137`: `.update("", "utf8")` adds no separator to the identity hash, so `("/x/a","bc")` and `("/x/ab","c")` collide. Not exploitable given the descriptor formats; probably meant `"\0"`.
- The Router key env vars aren't added to `collectKnownSecrets` for startup-error redaction. E4 added its model master key there.
- The mapping's `tenantId` is never sent or checked; only `routerKey` is used. A mis-pointed `keyEnv` silently bills another tenant.
- The Router HTTPS gate uses `isHostedEnvironment(OPENTAG_ENV)`, so plain HTTP is allowed when `OPENTAG_ENV` is unset. The stricter pattern `!OPENTAG_ENV_EXPLICIT || ...` would be safer.
- Setting `OPENTAG_WEB_TOOLS_ENABLED=true` with legacy credential mode, or without a machine token, silently disables web tools with no log line.
- `RuntimeCredentialEnvironmentManager.webToolsForSession()` is unused.
- `opentag-web-*` temp directories under the OS temp folder aren't swept after a daemon crash (mode 0700, just litter).
- The 48 KiB final truncation can cut off the `metadata:` line and the artifact-failure note when three pages sit at their preview cap.

Not covered: Pi actually honoring `--extension` alongside `--no-extensions`/`--offline` (the tests only assert argv; there's no `pi` binary here), real Cloud `sandbox exec` stdin streaming, and the companion Router PR.

liuchao-001 (approved):
Re-reviewed head `65019382`. The blocker is fixed: `scripts/copy-web-tools-extension.mjs` is now on the Runner build-context allowlist and copied in the doctor Docker build. The new regression test stages the real production context rather than a fixture. I checked the other places that build the CLI: `ci.yml`, `publish-npm-package.yml` and `build-portable.mjs` all run on a full checkout, so no other context needs the script. `Build linux/amd64 image and run offline smoke` and the rest of CI are green on this head.

Compared with `54ebc4f3`, this push touches only build contexts, one test and a doc comment in `runner/web-gateway.ts`, so my earlier code review still applies to this head. Approving.

Non-blocking items 2–8 from my previous review are still open. The ones I'd fix before activation: the web-only open on an inactive binding (2), web requested but never wired for Session-message runs (3), and an artifact-store failure turning an already-billed fetch into a tool error (4). A tracked follow-up is fine for these, since the feature ships disabled.

liuchao-001 (approved):
Re-reviewed head `e837aacb` (merge of main `b805a590`, MCP management). Approving again. The one red check is not this PR's.

**The merge is sound.** The only conflict was `public-exports.snapshot.json`, and the resolution keeps both sides (`normalizeExtraHeaders` / `probedServerDescription` from main, the web entries from here). I verified that against the real barrel rather than the file alone: `public-exports.test.ts` passes on this head, 780 exports. The other files both sides touched (`app.ts`, `config.ts`, `server/index.ts`, `shared/index.ts`, both `docs/README.md`) carry only additive web content over main; nothing from the MCP change was dropped. Locally on this head: 45 server web tests, 86 client web tests and 26 shared tests pass.

**`Enforce Patch Coverage` is inherited from main, not caused by this PR.** It fails here with `cli` branches 78.02% and `server` branches 78.29% against the 80% floor. I measured the same two packages on the merge base `b805a590` with no PR code in the tree:

| package | main `b805a590` | PR head `e837aacb` | floor |
| --- | --- | --- | --- |
| cli branches | 78.05% | 78.02% | 80.00% |
| server branches | 78.23% | 78.29% | 80.00% |

So main is already below the floor on both, and this PR is within noise of it (server is marginally better). The gate only runs on pull requests, and #641 was merged at 00:13 while its own `Enforce Patch Coverage` run was still going; that run finished `failure` at 00:16. Every PR opened against main now inherits the red.

That's a main-branch fix, not something to pile onto this PR: either raise MCP branch coverage or consciously re-baseline the floors, in its own change. Worth a heads-up to whoever owns #641, and this PR probably shouldn't be merged past a required red check until main is clean. My earlier non-blocking items 2–8 are still open and still fine as follow-ups.

PR #644, opened 2026-09-18
ci: stop commit messages from failing the secret scan

bestony:
## Summary

Fixes the failing `Secret Scan` job in [run 35292871181](https://github.com/first-tree-ai/opentag/actions/runs/35292871181/job/105439280979), and the class of failure it belongs to.

TruffleHog scans two kinds of chunk: the content of a file at a commit, and the commit's own metadata. The finding that failed the run has no `file` field — it is the **commit message** of a6be9fb9 (#643), whose body quoted the very placeholder DSN it was removing from `.github/trufflehog-exclude-paths.txt`. A metadata chunk carries no path, so `--exclude-paths` can never reach it, and the message of a pushed commit cannot be rewritten. The job had entered a state that could not be fixed forward.

It is not a one-off. A full-history scan of `main` reports **41** such findings across two commit messages: one from #643 and forty Cloudflare-shaped ones from an earlier commit. Every squash merge turns a PR body into a commit message, so any PR that discusses a credential shape can red the scan permanently.

The job now invokes the pinned TruffleHog image directly and passes its JSON results through `scripts/secret-scan-gate.mjs`:

- **File findings are unchanged** — reported whether or not TruffleHog verified them, with the existing path exclusions already applied by the scanner.
- **Commit metadata findings are reported only when verified.** A live token pasted into a message still fails the build; prose describing a credential shape does not. An unverifiable placeholder is, by construction, not a leak.
- **Unrecognized result shapes fail closed** — anything the gate cannot positively identify as a commit metadata chunk is reported.

The `trufflesecurity/trufflehog` action always runs `trufflehog git --fail` and offers no seam for that filter, which is why the job runs the image itself. The image is pinned by tag *and* digest, the two the action resolved for the failing run. Dependabot cannot see an image referenced this way, so the comment above it says to bump both by hand — same as the pinned `actionlint` and `ShellCheck` versions in the job above.

The `--github-actions` output format is replaced by `::error::` annotations the gate emits itself, so failures still surface in the run UI.

## Testing

Reproduced and verified against the exact failing invocation, by extracting the new `run:` block from `ci.yml` and executing it verbatim with the run's own environment:

- **Fails before:** `unverified_secrets: 1`, exit `183`, no `file` field on the finding — commit message line 16 of a6be9fb9.
- **Passes after:** `0 reported, 1 suppressed`, exit `0`.
- **Not over-broad:** a credential-shaped DSN planted in `packages/server/src/` in a scratch clone is still reported by path and line, exit `1`. Removed afterwards.
- **Branches exercised:** `push` (range from `github.event.before`), `pull_request` (base/head shas, clean range), and an unreachable base (`0000…`) falling back to a full-history scan with the message it logs.
- **Full history:** 43 raw results → 41 suppressed, 2 reported (see Non-goals).

Required commands, all passing:

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test` — 416 script tests (403 + 13 new), 43 CLI files / 505 tests, 94 web files / 1081 tests
- `actionlint` 1.7.12 over the workflows (includes ShellCheck over the new `run:` block)

No migration is added, so `CURRENT_MIGRATION_COUNT` is unchanged and the PostgreSQL integration suite is not required.

## Breaking changes

None. The gate is read only by the `Secret Scan` job.

## Non-goals

- **Two historical file findings are left alone.** A full-history scan still reports a container-hostname DSN in `scripts/e2e/doctor-docker.mjs` and `.github/workflows/ci.yml` as they stood at two old commits; neither string is in the tree today. They surface only on `workflow_dispatch`, where the scan has no base and walks all of history — never on `push` or `pull_request`. Silencing them would mean excluding workflow and script paths wholesale, which is exactly where a leak would matter, so that is a separate decision.
- **The exclusion file's breadth over the integration and e2e suites is unchanged**, as documented in #643.
- **Detector selection and entropy filtering are untouched.** Filtering placeholders by Shannon entropy was measured and rejected: the threshold that drops a `user:pw` placeholder also drops real secrets shorter than roughly 20 characters.

## Checklist

- [x] Canonical English docs and Chinese mirrors are synchronized when applicable — N/A, no documentation file changes.


scripts/secret-scan-gate.mjs
```diff
@@ -0,0 +1,155 @@
...
+    commit: isGitChunk && typeof git.commit === "string" ? git.commit : "",
+    file,
+    line: isGitChunk && Number.isInteger(git.line) ? git.line : 0,
+    fromCommitMetadata: isGitChunk && file.length === 0,
+  };
+}
+
+/** A finding is suppressed only when it sits in commit metadata and TruffleHog could not verify it. */
+export function isSuppressed(finding) {
+  return finding.fromCommitMetadata && !finding.verified;
```
yuezengwu, line 74: [P1] Keep unverifiable credentials blocking by default

Verified=false does not establish that a credential is a placeholder. A real database password on a private/unreachable host, or a provider verification timeout, also produces an unverified result; the pinned Postgres detector records verification errors separately (https://github.com/trufflesecurity/trufflehog/blob/v3.97.1/pkg/detectors/postgres/postgres.go#L178). An offline CLI probe with valid Git metadata, Verified=false and a VerificationError exits 0 with the finding suppressed here. Thus a real credential in a commit message can pass this gate. Restrict suppression to explicitly reviewed synthetic findings, and keep verification errors and unknown verification states blocking.

scripts/secret-scan-gate.mjs
```diff
@@ -0,0 +1,155 @@
...
+ * describes a credential shape therefore fails the scan, and because the message of a pushed
+ * commit cannot be rewritten, that failure can never be fixed forward.
+ *
+ * The gate closes that hole without weakening the scan of the tree:
+ *
+ *   - File findings are reported exactly as TruffleHog emits them, verified or not. Path
+ *     exclusions have already been applied by the scanner.
+ *   - Commit metadata findings are reported only when TruffleHog verified the credential
+ *     against its provider. A real token pasted into a commit message still fails the build;
+ *     a `postgres://user:pw@host` placeholder in prose does not.
```
yuezengwu: [P1] Remove the credential-shaped example from scanned source

This comment is a file finding, so the new metadata filter correctly leaves it blocking. The current head's Secret Scan job reports Postgres (unverified) at scripts/secret-scan-gate.mjs:18 and exits 1: https://github.com/first-tree-ai/opentag/actions/runs/35297980390/job/105454369654. Please replace the literal with a prose description and resolve its occurrence in the PR's scanned commit range while preserving script coverage; this head does not yet restore the failing gate.

scripts/secret-scan-gate.mjs
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Normalizes a raw result into the fields the gate reasons about. `fromCommitMetadata` is true
+ * only for a recognized git chunk that carries no path, which is how TruffleHog reports a finding
+ * in a commit message or author line.
+ */
+export function describeResult(result) {
+  const git = result?.SourceMetadata?.Data?.Git;
+  const isGitChunk = Boolean(git) && typeof git === "object";
+  const file = isGitChunk && typeof git.file === "string" ? git.file : "";
```
yuezengwu, line 60: [P2] Reject malformed metadata instead of treating it as pathless

The object check accepts Git: [], and a present non-string file value is converted to an empty string. Both are then labeled commit metadata and suppressed. Offline CLI probes with Git: [] and with file: 42 each exit 0, contradicting the stated fail-closed behavior for unrecognized results. Validate the Git metadata shape and verification field, distinguish an absent/empty path from an invalid path type, and report malformed cases instead of suppressing them.

scripts/secret-scan-gate.mjs
```diff
@@ -0,0 +1,155 @@
...
+  const results = [];
+  const lines = input.split("\n");
+  for (let index = 0; index < lines.length; index += 1) {
+    const line = lines[index].trim();
+    if (line.length === 0) continue;
+    let parsed;
+    try {
+      parsed = JSON.parse(line);
+    } catch {
+      throw new Error(`line ${index + 1} of the TruffleHog output is not JSON: ${line.slice(0, 120)}`);
```
yuezengwu, line 42: [P2] Omit scanner payloads from parse-error messages

The first 120 characters of a malformed JSON line are copied into the thrown error and then console.error. A truncated line beginning with a Raw field reproduces the issue: the CLI exits 1 but prints the synthetic secret marker to stderr. This bypasses formatFinding's redaction on the error path. Report the line number and a generic parse failure without including the input bytes, and cover that error path with the existing no-secret-output guarantee.

yuezengwu (changes requested):
The new gate needs changes before it can be relied on: an unsuccessful verification is not evidence of a synthetic credential, malformed Git metadata is currently suppressed, and malformed output can be copied into the logs. The current Secret Scan also fails on the example in the new script itself.

Validation on this head: pnpm check, the 13 existing gate tests, and git diff --check pass. Six offline CLI probes reproduce the verification-error and malformed-input cases, with verified metadata and unverified file findings as blocking controls. The failed CI job independently identifies the new script comment. No live credential verification or deployment was performed.

bestony:
Two checks are red for reasons outside this change:

- **Enforce Patch Coverage** — `Coverage floor breach for package "cli": branches 78.13% ... below floor 80.00%`. Pre-existing on `main`: the identical breach (same 78.13%, plus `server` at 78.23%) fails on [the dependabot actions-bump PR](https://github.com/first-tree-ai/opentag/actions/runs/35291204090), which touches only action pins, and on #643 before it. This PR adds no `apps/cli` code.
- **ownership-gate** — awaiting review, by design.

`Secret Scan` itself passes in 18s.

PR #648, opened 2026-09-18
feat(cloud): persist sandbox workspaces across instances

yuezengwu:
## Summary

Cloud Sessions previously lost their workspace and Pi history when their Cloud Run Instance was released. E5 saves the latest complete workspace to `<storage_uri>/state.tar.gz` and verifies restoration before admitting work on a replacement Instance.

The implementation reuses the Sandbox row, allocation generation, lifecycle, bootstrap credential, and Server delivery custody. It adds no table, migration, snapshot catalogue, or scheduler. The Server streams GCS transfers for the trusted Runner; Google credentials stay outside the Sandbox.

- Quiesce native writers before saving; fence stale uploads with GCS generation/metageneration preconditions. Reconnects never restore over unsaved local work.
- Normal release closes admission, settles reports with durable acknowledgments, seals and verifies the current archive, then deletes the UID-bound Instance. Transient save failures preserve the allocation for retry.
- Deterministic archive failures now publish an honest `workspace_failed` report, preserve execution-effects evidence, and keep the control channel available without retrying unchanged files. The Cloud prompt explains the size, entry, and file-type limits.
- Explicit generation-bound `discardUnsavedChanges` releases an unsaveable or unavailable allocation while retaining the last good archive. Stale discard requests fail; delete retries retain the discard intent. Provider-confirmed pre-persistence Instances retain their legacy stop behavior.
- Previously allocated Sessions with no archive fail before creating another Instance. The seal deadline covers drain and both transfers. PostgreSQL regressions cover initial creation failure and late callbacks overwriting release evidence.
- Validate archive paths, links, checksums, and resource limits. Preserve Git and Pi history while excluding trusted parent state. Web-tool startup follows restoration; namespace replacement does not renew an old execution's authority.

## Testing

Passed locally: frozen dependency installation, `pnpm check`, `pnpm build`, `pnpm typecheck`, the full unit suite (6,139 tests including repository scripts), `pnpm test:coverage`, Agent Runtime coverage, and PostgreSQL integration. Runtime coverage passes all four 100% gates; PostgreSQL integration passes 502 tests. Repository line coverage is 90.85%; all coverage floors pass without modification. Changed-line coverage is 1,845 / 1,848 (99.84%), above the 80% gate.

Coverage investigation found CLI and Server MCP management paths from current main were not covered by unit tests. This PR adds focused authority, configuration-inheritance, and credential-revocation unit tests without changing MCP production code or lowering coverage thresholds.

Wire tests use real loopback HTTP/WebSocket with deterministic native/GCS doubles, including a real hard-link save failure, report replay, refusal to seal, and unchanged last-good storage. Other tests use real PostgreSQL. These are not live GCP acceptance.

## Breaking changes

Workspace-enabled allocations require the matching E5 Runner capability. No GCP or staging configuration changed. Live acceptance requires a matching Runner image, owned GCS prefix and Server permissions, and ingress supporting the 128 MiB / 120-second transfer contract. Archive limits are 128 MiB compressed, 256 MiB payload, and 50,000 entries; hard links and special files are rejected. Explicit discard is an Account API operation in this phase.

Existing E3/E4 workspaces are not automatically migrated. OOM or machine loss restores only the last successful saved boundary, never process memory or unsaved progress. A crash during saving before the terminal report is journaled leaves unknown execution effects; external actions are never automatically replayed. See `docs/cloud-workspace-persistence.md` and its Chinese mirror for live acceptance and cleanup.

## Checklist

- [x] No new database tables, migrations, or application snapshot versions.
- [x] Failure, recovery, and authority boundaries have regression tests.
- [x] Canonical English documentation and Chinese mirrors are synchronized.


liuchao-001 (changes requested):
Reviewed at head `224b027a`. The archive codec is careful work: symlinks are installed only after the digests verify and the full link graph is resolved, there is no-follow staging, and restores go into an empty destination. GCS fencing is also sound: generation plus metageneration CAS, and owner-bump on claim, verified by read-back. The targeted client and server suites pass locally (90 and 87 tests). One gap blocks this: **no path ends a workspace that can never be saved.** Today that failure mode is easy to hit.

### Blocking

**1. A deterministic save failure wedges the Session. The Turn report is withheld, and the Runner retries forever.**
`createWorkspaceArchive` fails closed on more than 50,000 entries, more than 256 MiB, `nlink > 1`, sockets or FIFOs, or non-portable names. One `npm install` in the workspace crosses the entry and size limits for most real projects. pnpm may also place a same-device store under the workspace mount, and its hard links then trip the `nlink` rule. The failure is not transient:
- `cloud-turns.ts:782`: the checkpoint throws after `recordReport`, so the terminal report stays in the journal and is never sent. `persistenceFailure` → `finish()`.
- On reconnect, `prepareRunnerConnection` → `prepare()` → `initialize()` → `cloud-workspace.ts:110` retries `#save(false)`, which fails again, and the connection closes. `reconcile()` sits behind `prepare()`, so the report is never replayed.
- `serve.ts:913`: persistence Runners never hit the reconnect limit, so this loops at 30 s backoff forever.

The user sees a Session that stops answering, with no error. The agent also doesn't know the limits: the system prompt only says "unsaved changes can be lost."

**2. `stop` can no longer release such an allocation. The Instance can't be stopped at all.**
A `ready` → `releasing` transition stamps `workspace_save_required` (`sandbox-runner-service.ts:379`). A present, UID-matched Instance then requires a proven seal (`:978`). With finding 1, the seal fails every time. It also fails when no live Runner holds the channel (`:1015`), which covers:
- a crash-looping Runner;
- an **E3/E4 Runner still live at deploy**, rejected by `runner-connection.ts:277`. Its row was `ready`, so it is stamped save-required, and nothing can ever connect to seal it while that Instance stays present.

`stopForAccount` is the only release entry, and I found no idle reaper, so the Instance runs and bills indefinitely. Before this PR, stop always released. Fail-closed on *transient* errors is right, but the account needs a terminal exit.

Suggested direction (details are yours):
- (a) Classify archive-level failures (`too-many-entries`, `workspace-too-large`, `unsafe-entry`, `unsupported-entry`) as non-retryable. Publish the held report with an honest `workspace_save_failed` outcome instead of looping.
- (b) Give stop an explicit "discard unsaved changes" release that keeps the last good archive and deletes the Instance, and a bounded fallback when no Runner can attach. Also exempt allocations created without the persistence flag, e.g. stamp save-required only when the allocation was created with `workspacePersistence`.
- (c) Put the limits in the Cloud system prompt so the agent can keep dependencies out of the workspace.
- Add a test for a deterministic failure end to end. The current failed-save test (`serve-workspace.test.ts:87`) only covers a transient failure that succeeds on retry.

### Non-blocking

3. **Legacy released Sessions now allocate an Instance that can never become ready.** With persistence on, `ensureIngressAllocation` (`:325`) no longer returns `restore_required` for `unallocated` rows with generation > 0. Each IM message on a pre-E5 Session reserves a new generation and creates an Instance. The Runner claims, gets 404 on the missing object, and loops forever in `preparing`. The docs say these sessions "fail restoration", but the concrete effect is a live Instance per touched Session and a permanent `pending`. Consider keeping `restore_required` when `head()` reports no object for generation > 0.
4. **The seal deadline is shorter than the work it waits for.** The Server waits `sealTimeoutMs` = 120 s. The Runner-side seal runs `drainForRelease(120 s)`, then possibly the aborted Turn's own unsealed checkpoint (archive + upload), then the sealed archive + upload (each upload up to 120 s). A large but legitimate workspace will therefore fail the first stop with 503 and succeed on retry via `#workspaceSealProven`. That is correct but confusing. Either size the Server deadline to cover the sum, or return a "releasing, retry" status rather than a save failure.
5. `#insertMultipart`'s "framing violation provably cannot commit → rethrow `invalid_input`" branch likely never fires. undici rejects `fetch` with its own `TypeError` when the request body stream errors, and `#send` maps that to `unavailable` → `unknown_result` → read-back. The outcome is still safe; only the classification/comment is off.

— reviewed by liuchao-staff (agent) on behalf of @liuchao-001

liuchao-001 (approved):
Re-reviewed at head `28a33a79`. My earlier blocking findings are resolved:

- **Deterministic save failure**: archive-level violations now set `terminalFailure`. The Turn publishes an honest `workspace_failed` report instead of looping. `reconcile()` now runs before `prepare()`, so journaled reports are replayed even while restore/save keeps failing. The real hard-link wire test covers the path end to end.
- **Unreleasable Instance**: the generation-bound `discardUnsavedChanges` stop skips the seal and keeps the last good archive. Stale generations are rejected. Provider-confirmed pre-persistence Instances (`workspacePersistence === false`) keep legacy stop.
- **Legacy released Sessions**: `head()` runs before reservation, so a missing archive returns `restore_required` instead of creating an Instance.
- **Seal deadline**: now 480 s, covering drain plus both transfers.

Targeted suites pass locally at this head: client 101/101, server 135/135, including the new MCP authority tests.

Non-blocking follow-ups:
1. `workspaceSaveFailure` uses one message for every checkpoint failure. The Turn text says "Further execution is blocked … explicitly discard unsaved changes" even for a *transient* storage failure that recovers on reconnect. Consider a separate wording for `retryable` errors, so users don't discard work that would have saved.
2. That message asks users to discard, but discard is Account-API-only in this phase. The web stop UI needs a follow-up before this is reachable for normal users.
3. Ordinary stop can now hold the HTTP request for up to 480 s. Worth confirming the web client and ingress timeouts during live acceptance.

— reviewed by liuchao-staff (agent) on behalf of @liuchao-001

PR #650, opened 2026-09-18
feat(cloud): run concurrent sessions with reliable sandbox recovery (E6)

yuezengwu:
## Summary

A long Cloud Session currently occupies its whole Agent's delivery lane. This change lets Sessions of the same Cloud Agent run independently while retaining one execution at a time within each Session, durable input custody, and Local Agent exclusivity/steering.

- Scope Cloud scheduling, custody and pre-dispatch checks to the Agent Session. Preserve FIFO ordering through retry, row contention, Worker loss and late provider events.
- Retain a persistent Runner's unsaved files after auth rejection. Signed expired bootstrap credentials can obtain a fresh token only after current allocation and physical UID/ownership checks, then require a new handshake. Ordinary HTTP access still rejects expired tokens.
- When a disconnected, previously ready Instance is confirmed absent, conditionally clear its exact binding and restore the same storage URI into the next allocation. Provider uncertainty never authorizes replacement.
- Scope credential proxy routing/CA material to provider CLI processes and GitHub's Git host. Ordinary HTTPS and public GitLab access retain public routing and system trust; the credential proxy allowlist is unchanged.
- Drain received inputs through the existing receipt protocol: reject unaccepted input, cancel/report accepted input, and retry unanswered acknowledgments within the release deadline. Expired unaccepted dispatches remain retryable within message TTL; accepted work is never blindly replayed.

No new tables, migrations, configuration flags or scheduling service. Sandbox topology, fixed storage URI and 1 vCPU / 1 GiB policy remain unchanged. English/Chinese documentation is synchronized.

## Validation

- `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` passed.
- Server integration: 37 files / 511 tests with actual PostgreSQL.
- Agent Runtime coverage: 439 tests; all four enforced metrics at 100%.
- Focused recovery, expired-token authority, stale-query CAS, dispatch-expiry, stopped-receipt and delayed-ack regressions.
- Real loopback HTTP/WebSocket Runner seal/restore tests and local curl/Git TLS/proxy fixtures; native execution and GCS use explicit doubles.

Native GCP, real IM and model acceptance remain pending unified staging configuration. This PR does not deploy or change Cloud resources or real IM messages.

## Compatibility and rollout

Deploy Server before its matching pinned Runner image: persistent Runners now send the opt-in `renewExpired` auth field, which older strict-schema Servers do not accept. Existing Clients remain compatible with the new Server. Renewal evidence is valid only for the lifetime of the current allocation and is revoked by allocation replacement/removal.

Idle reclamation/reuse remains E7. Account budgets, broader Context/Session capabilities and product completion remain separate work.

## Checklist

- [x] Changes address concurrency and the full-scope acceptance audit.
- [x] Tests and English/Chinese documentation cover changed behavior.
- [x] Required local checks pass; no credentials or generated builds are committed.
- [x] Shared custody and database models remain unchanged.


liuchao-001 (approved):
Approving at `7e84f434`. I found no blocking issues.

What I checked:
- **Occupancy rule.** The claim subquery and the last-boundary recheck (`findOtherCustody`) both use `occupancyConflict`, so they can't drift apart. A Cloud candidate is fenced only by its own Session, or by a Local occupant of the same Agent. A Local candidate keeps the Agent-wide fence. `session_placements.session_id` is the PK, so the new left joins can't fan out. Rebind is already blocked while custody is uncertain, so mixed Local/Cloud custody for one Agent shouldn't happen.
- **Claim race.** The advisory lock is still Agent-scoped and still taken before the re-query. Under READ COMMITTED, the post-lock `findOtherCustody` sees a competing claim that just committed in the same Session. That claim is then deferred with `IM_DELIVERY_CUSTODY_DEFERRED` instead of being re-selected every tick.
- **Ordering guard.** It reads committed `pending` rows no matter who holds the row lock, so an input can't overtake an earlier one that is in backoff or skip-locked. Rows that already carry a dispatch or claim marker are exempt. That avoids the mutual fence where a late earlier input and an expired claim each wait on the other. Allocation stays per Session (`sandboxes_session_id_unique`).

Non-blocking notes:
1. **Per-Agent fairness.** Cloud lanes are now per Session. One Agent with many active Sessions can take all of the Worker-global `maxConcurrent`, and `maxQueuedPerAgent` is now effectively per Session. The PR scopes budgets to E9, which is fine. Maybe rename the option or add a follow-up issue, since the name no longer matches what it does.
2. **TTL-lapsed head row.** The guard counts an earlier undispatched `pending` row even after its `expiresAt` has passed. The outer claim won't pick that row up anymore, so the later input waits until the janitor expires it. The janitor bounds the wait, so this is acceptable. Adding `expiresAt > now()` to the guard's earlier-row predicate (for undispatched rows) would remove the wait.
3. **Nit.** When a Cloud Session is fenced at the last boundary, `CloudDeliveryCoordinator` still records `IM_DELIVERY_AGENT_CUSTODY_FENCED`. That code now also means a Session-scoped fence, which could confuse anyone reading the metrics.

liuchao-001 (approved):
Re-reviewed at `a686d62e`. The E6 commit `7e84f434` is unchanged (range-diff `=`), so this review covers only the new commit, `fix(cloud): harden sandbox recovery and receipt settlement`. Approving again. I found no blocking issues. My earlier approval was bound to `7e84f434`, and this ruleset doesn't dismiss stale reviews, so this one explicitly covers the new head.

What I checked:
- **Expired-token renewal.** JOSE checks alg, signature, issuer, audience and `nbf` before it raises `JWTExpired`, and only `claim === "exp"` is treated as renewable. `renewExpiredBootstrap` requires a valid current channel scope, a matching resource name and UID, and ownership. It re-checks the UID and scope after the provider and signing awaits, so a release or replacement that happens mid-renewal revokes it. The `auth:renewed` reply attaches no channel, and the Runner has to do a fresh handshake. Workspace HTTP still rejects expired tokens.
- **Loss repair.** `#releaseMissingReadyAllocation` clears the binding only when the Runner is disconnected, the provider confirms the exact UID is absent, and the check re-runs after the GET. The CAS covers generation, name, UID and `ready`. A failed GET throws and never authorizes replacement.
- **Receipt settlement.** Splitting `expired` / `expired_unaccepted` is correct: an expired receipt that was never accepted is now refused instead of getting fabricated custody, so the worker's stale-window release can retry within TTL. Re-announcing a `received` entry during drain, instead of synthesizing `cancelledBeforeStart`, avoids the report-without-custody conflict.

Non-blocking notes:
1. **Renewal lifetime equals allocation lifetime.** The initial bootstrap token reaches the Runner through `OPENTAG_RUNNER_BOOTSTRAP_TOKEN`, so it also sits in the Instance spec. Before this change, anyone who could read Instances held something that expired in 30 minutes. Now it can be exchanged for fresh tokens for as long as the allocation lives. The docs call this out and recommend restricting Instance-read permissions, which is fine as a tradeoff. A cheap backstop would be a maximum renewal age (for example, reject renewal when `now - iat` exceeds some bound, or only renew from a token that was itself renewed). That would keep the reachable window finite if Instance-read permissions ever get broader.
2. **Cloud API GET on every ingress attempt.** `ensureIngressAllocation` now falls through to `startForAccount` for any `ready` sandbox with no connected Runner. `#releaseMissingReadyAllocation` then does a Cloud Admin `getInstance` on every claim attempt. The per-delivery retry cadence is about 2s, so a short Runner reconnect gap turns into a steady stream of Admin reads, and a Cloud API error now fails the claim instead of returning `ready`. Consider a short per-sandbox cache or throttle for the "present" verdict.
3. **Local behavior change.** Provider routing now reaches only provider CLI children on Local Computers too, so a raw `curl` to a Slack `upload_url` / `url_private` from the Agent only works after sourcing `$OPENTAG_PROVIDER_ENV_FILE`. The guidance text covers this, but it's a model-behavior dependency. A QA case for Slack file upload/download on both Local and Cloud would lock it in.
4. **Scope.** This commit is a second feature set on top of E6: the auth protocol, loss repair, proxy scoping and drain. The updated description reflects that. Splitting it would have made bisecting and rollback cleaner. Not asking for a split now.

Independent exact-head QA is still outstanding. The description says native GCP/IM acceptance is pending staging, and the note on server-before-Runner rollout ordering matters for deploy.

liuchao-001 (approved):
Re-approving at `4c1d3a34`. The first two commits are unchanged (range-diff `=`), so my review of `a686d62e` still applies. The new commit only changes `runtime-proxy-network-scope.test.ts`:
- It runs `git ls-remote` outside the checkout, so the Authorization extraheader that actions/checkout adds can't leak into the fixture.
- The fixture now sends a valid upload-pack advertisement.
- It replaces the swallowed `.catch(() => undefined)` with an assertion that the command succeeds with empty stdout.

This makes the test stricter and doesn't change any production code. The non-blocking notes from my previous review still stand.

liuchao-001 (approved):
Re-approving at `e984be0b`. The first three commits are unchanged (range-diff `=`). The new commit only changes `runner-serve.test.ts`. The test now waits for a `delivery:query:result` that is queued behind the initial reconnect reconcile before it injects the journal fault, so the corrupt directory can no longer race reconciliation. It waits on an observable protocol event and uses no sleep or retry, and it asserts the request ID and the `none` phase. No production code changed, and my earlier notes still stand.

PR #651, opened 2026-09-18
feat(cloud): reclaim idle sandboxes and reuse instances (E7)

yuezengwu:
Cloud Sessions currently retain their physical Instance after work finishes, and every new Session cold-allocates another one. E7 adds one business-idle budget (default 120 seconds) and a bounded sweep in the existing Server. A quiescent Instance can be reused on demand by another Session in the same Account: save and verify the original workspace, atomically transfer the exact provider name/UID, clear previous native execution state, and restore the new Session before readiness. Otherwise the sweep saves and deletes the idle Instance.

The only database addition is nullable `sandboxes.idle_reclaim_at` (migration 0046; 47 total). Existing Agent/Session identities, latest-state storage URI, fixed 1 vCPU / 1 GiB, and Client-owned Runner releases remain unchanged. There is no pool/lease/Instance table or separate scheduler. Physical control credentials use a separate JWT audience; ordinary Session credentials cannot follow a transfer. RunnerHub ownership is process-local, so E7 requires one Server process.

Pending/accepted work and live acceptance block reclamation. Reuse requires negotiated capability and current deployment policy; cross-account reuse is excluded. Save failures preserve unsaved local state, interrupted deletion retries, and stale startup recovery permits both creation and workspace restoration time. Runner serve defaults to the declared TCP health port 8080 when `PORT` is absent. Only Pi's offline probe gets the existing 30-second native allowance; other tools keep 10 seconds.

Review repairs in `26a4315f`:
- Check persistence and deployment policy before claiming; legacy/non-persistent Instances remain usable. Recheck generation/name/UID under the row lock after provider I/O.
- Rotate attempted ready rows by `updated_at` while idle eligibility still uses only business activity; a full batch of failed seals no longer starves other Sessions.
- Immediately release a proven sealed origin when its transfer cannot commit. Blindly clearing the claim would not reopen a sealed Runner; unproven saves retain local state and retry.
- Persist an assignment before adopting it in memory; native deletion failure terminates reconnects. Regression tests fail against the previous buggy implementations.
- Drain an active sweep before database shutdown, and explicitly document the single-Server boundary in English and Chinese.

Validation on `26a4315f2e887577b8d86e85201ff1ed697d64c9`:
- All seven required local gates passed, including 518 real PostgreSQL integration tests and 439 Agent Runtime tests at 100% coverage. Two initial Docker port-binding setup timeouts passed on a full unchanged rerun.
- Lifecycle/worker tests cover legacy ingress, 25 failed seals plus a healthy candidate, incompatible-policy skips, lost concurrent borrower reservation, slow/never-ready restore and graceful shutdown. A real PostgreSQL gated test verifies the same-borrower race.
- Exact clean linux/amd64 Runner built and offline toolchain passed; containers removed.
- All technical GitHub CI passed on this head, including Node 22/24/26, Runner/Docker builds, smokes and Secret Scan. Changed-line coverage passed at 1042/1070 (97.38%, threshold 80%).
- Fresh GCP acceptance is blocked before registry publication: local gcloud credentials require reauthentication. No Cloud resources were created by this repair attempt. The previous head (`42c97499`) passed 21 real GCP/GCS checks, but that receipt is historical and is not acceptance of this head.

No shared-staging configuration/deployment or production rollout was performed, and no real Slack/Feishu destination was contacted. liuchao-001 re-reviewed and approved this exact head, confirming the original findings are addressed. The remaining ownership gate requires bestony approval; this PR is not merged.


liuchao-001 (changes requested):
Reviewed at head `42c97499`: `sandbox-runner-service.ts` (claim / seal / sweep / borrow / transfer), `idle-reclaimer.ts`, the control-token audience split in `runner-bootstrap-token.ts` and `runner-connection.ts`, `runner-hub.ts`, custody and dispatch fencing, the shared protocol, and the Runner side (`serve.ts`, `assignment-state.ts`, `probe.ts`, `acceptance.ts`). I read the code and did not run tests.

Most of this holds up well. Stop, borrow, sweep and dispatch are fenced by conditional updates on lifecycle, generation, name, UID and the claim marker. Cross-account reuse is excluded at three independent points. The Session and control JWT audiences can't stand in for each other on any path. The Runner only discards the old assignment after a proven seal.

One liveness bug is serious enough that I'd like it fixed before merge.

**1. Blocking: a claim whose seal can never succeed wedges the Session permanently.**
`#claimIdle` (`sandbox-runner-service.ts:1808`) sets `idleReclaimAt` without first checking that the Instance can be sealed. For an Instance with `workspacePersistence === false`, `#sealWorkspaceIfOwed` (`:1309-1312`) throws on every attempt once the claim exists. `#reclaimIdleCandidate` (`:1786-1797`) records `workspace_save_failed` and returns, and the row stays `ready` with the claim set. Nothing clears `idleReclaimAt` on that path: it is only cleared by explicit stop (`:517`), `#clearReleased` (`:1525`) and the transfer (`:2052`, `:2078`).

While the claim is set, the Session can't do anything:
- ingress returns `pending` (`:439`)
- `startForAccount` returns `reclaiming` (`:410`)
- channel validation, acceptance and dispatch custody all refuse it.

The only way out is an explicit stop, and the user gets no signal that one is needed.

Scenario: E7 rolls out while older Instances without persistence are `ready`. 120 s after its last activity, each of them is claimed and never serves again. The borrow path can reach the same state, because `#claimAndSealSibling` claims first and hits the same throw.

The test "keeps a legacy Instance's sole local copy…" (`sandbox-idle-instance-reuse.test.ts:437`) checks `lifecycle: "ready"` and zero deletes, but not `idleReclaimAt` or ingress, so it passes over the wedge.

Suggested fix: refuse to claim (or skip in the sweep query) an Instance that can't persist, or roll the claim back on a permanent `WorkspaceSaveError`. Then extend that test to assert that the Session still accepts ingress.

**2. Non-blocking, but it amplifies #1: wedged rows can starve stage 1 of the sweep.**
`#idleReadyCandidates` (`:1625`) includes rows that are already claimed and orders them by `lastActivityAt ASC` with `LIMIT 25`. A failed seal never advances `lastActivityAt`, so wedged rows stay at the head of the queue. With 25 of them, no other idle Instance in any account is reclaimed by stage 1, and each pass repeats 25 provider GETs plus seal attempts. Rotating by `updatedAt` for claimed rows, or backing off after `workspace_save_failed`, would fix it.

**3. Non-blocking: a borrow that doesn't commit strands a recently active sibling.**
A borrow claims with `automatic: false`, so any quiescent sibling qualifies no matter how recent its activity. `#claimAndSealSibling` (`:1924-1960`) only runs `verifyTrackedInstance` and the persistence check *after* the claim and seal. Its catch returns `undefined` and keeps the claim. The same happens when the claimant's own transaction later finds it is no longer `unallocated`, for example with concurrent ingress and UI starts.

In each case the sibling is blocked (`pending`) until its original idle budget passes, and then the sweep deletes it. After a deployment policy change, every new-Session start in an account evicts one warm sibling this way. Running the policy and persistence checks before `#claimIdle`, and releasing the claim when the transfer doesn't commit, would fix both.

**4. Non-blocking: the Runner updates its assignment marker in memory before writing it to disk** (`serve.ts:642-650`).
If `writeRunnerAssignment` throws, the rebind fails and the Runner reconnects. On retry, `assignmentsMatch` compares against the in-memory value (`:632`) and returns early, so the write is never attempted again. The disk still names the previous assignment as `sealed: true`. If the process then restarts with its state directory intact, the next rebind treats the new Session's unsaved workspace as a sealed assignment and discards it. Writing to disk first and then assigning `current.assignment` closes this.

**5. Nits**
- `discardSealedAssignment` (`serve.ts:~562`): if `sandbox.destroy()` fails, the Runner reconnects but `state.fatal` is never set. That creates a retry loop. `ServeWorkspace#quiesce` treats the same failure as fatal.
- `#recoverStalePreparingAllocations` measures from `lastActivityAt`, which is set at track/transfer time, before boot and restore. The restore budget (`RUNNER_WORKSPACE_TIMEOUT_MS`) is also 120 s, and the converge timeout can be as low as 10 s. So a slow restore of a large workspace can be deleted mid-restore and livelock.
- `SandboxIdleReclaimer.stop()` only clears the interval and doesn't wait for a pass that is already running. It converges on restart, so this just means shutdown log noise.
- The client-side probe timeout of 30 s now applies to every offline tool probe, not only Pi. One hung binary costs 3× what it did before.
- As the description says, busy and readiness checks go through the per-process `RunnerHub`. Worth an explicit note in the docs that E7 assumes a single Server process until control ownership is shared.

liuchao-001 (approved):
Re-reviewed the delta `42c97499..26a4315f`. Everything I raised is addressed, so I'm approving.

- **#1 wedge (was blocking):** `#claimIdle` now preflights the provider view before it takes the claim. It refuses a non-persistent Instance and runs the policy check (full policy for a borrow, ownership only for the sweep). The row lock then re-checks generation, name and UID, so the preflight can't be applied to a different binding. The legacy test now asserts `idleReclaimAt: null`, that ingress returns `ready`, and that runner scope validation still passes, which is exactly the check that was missing.
- **#2 starvation:** every attempted candidate gets an `updatedAt` bump, and the order is now by `updatedAt`. Eligibility still uses only `lastActivityAt`. The new test with 25 failed seals plus one healthy row covers it.
- **#3 stranded sibling:** policy and persistence are checked before the claim. If a borrow seals but doesn't commit, it now goes through verified release (guarded by `isNull(lastErrorCode)`). A seal that hasn't been proven keeps its claim. I agree that "unclaiming" a sealed Runner isn't safe, so immediate verified release is the right fallback. The concurrent-start race test pins it down.
- **#4 Runner marker:** the marker is now written to disk before memory is updated.
- **Nits:** a failed destroy is now fatal, the reclaimer drains on `stop()`, the startup deadline now covers the restore budget, and the 30 s timeout applies to the Pi probe only. All good.

One optional nit: in the `.catch` added to `startForAccount`, if `#releaseUnusedBorrow`'s own `UPDATE` throws, that error replaces the original reservation error. Wrapping it in `.catch(() => undefined)` would keep the root cause visible.

This review covers code only; I didn't run tests. CI was still running on this head when I checked.

PR #652, opened 2026-09-19
feat(mcp): deliver MCP tools to a running Agent through an execution-scoped gateway

bestony:
## Why

The MCP management plane is shipped: an Account registers Servers, mounts them per Agent, authorizes
each mount (anonymous / Bearer / OAuth), and watches the probe discover a tool list. **And the Agent
still cannot call one of those tools.** Nothing under `packages/client/src/runtime/`, `src/runner/`,
or `src/cloud-runtime/` mentioned MCP. The management plane was a promise the runtime did not keep.

## What

One inbound Streamable HTTP MCP endpoint on the Server — `POST /api/v1/mcp` — that a provider CLI
mounts like any other remote MCP Server. It aggregates everything the calling Agent has bound and
proxies `tools/call` upstream using the credential already sealed in the database.

```
Claude Code (per run, local Computer)
   │  Authorization: Bearer <execution token>
   ▼
POST /api/v1/mcp
   │  the bearer resolves to one executionId; the fence yields accountId + agentId
   ▼
resolveActiveCredential → buildHeadersFor → McpTransport → McpOutboundFetcher (runtime instance)
   ▼
the Agent's bound MCP Servers
```

**No upstream credential ever leaves the Server.** The earlier design note proposed the opposite —
push credentials to the Client behind a `127.0.0.1` proxy. This inverts it: the Client holds only a
token that is worthless outside one live execution.

### Decisions worth reviewing

- **The bearer proves which execution is calling, and nothing else.** The token record holds an
  execution id and no identity; Account and Agent are read from the live execution record on every
  request, so a token cannot assert whose tools it wants.
- **Execution-scoped by construction.** 256-bit random, retained only as its SHA-256 digest, one live
  token per execution, revoked from the single point every execution ends at.
- **Fetched through its own frame**, as capability tokens and proxy tickets already are, so the
  execution-open result keeps its documented property of carrying no secrets.
- **Both protocol eras are answered.** Claude Code's own bridge speaks `2025-03-26`; a modern-only
  gateway would be unusable by the client this exists for.
- **Transparent aggregation** as `<serverName>__<toolName>`, resolved by recomputing composed names
  over the Agent's own snapshots. A name two mounts both produce is published by neither.
- **A tool call is never replayed.** Retry only on `-32022 UnsupportedProtocolVersion`, which proves
  nothing ran; a 5xx, a `-32603`, a `-32601` that may come from the tool, and an unrecognized
  400/404 all propagate. The runtime path never re-detects a protocol era.
- **Runtime calls do not share the probe's budget.** Their own fetcher instance: 120 s deadline,
  8 concurrent per Account. The URL gate, redirect refusal and response bound are unchanged.
- **An upstream failure is an `isError` result, not a transport error**, carrying the upstream's own
  bounded message so a model can retry with better arguments.

## ⚠️ Prerequisite: proxy mode

**The gateway does nothing on a default install.** `resolveRuntimeCredentialMode` defaults to
`legacy`, and in legacy mode the credential relay never opens an execution — so there is no execution
record, no service grant, no token, and no gateway entry in `mcp.json`. The fence requires a live
execution record, so this is an architectural dependency, not a missing wire.

To use it: set `OPENTAG_RUNTIME_CREDENTIAL_MODE=proxy` in the daemon launch environment.

Tracked in **#653**, which also asks the design question — when does proxy mode become the default,
and is a services-only execution open worth adding so features like this do not depend on it.

When the gateway is withheld, the open path now records a reason at debug
(`capability_not_negotiated`, `not_requested`, `policy_unavailable`, `no_usable_mount`); a default
install with MCP bound reports `not_requested`.

## Scope

| | Status |
| --- | --- |
| Claude Code, local Computer, **proxy mode only** | Delivered |
| Codex | Not yet — its app-server is spawned once per Session runtime from a frozen argument vector, so a per-execution bearer cannot be injected. `http_headers_helper` is the likely mechanism; it needs a spike |
| Cloud sandbox | Not applicable — the Turn worker constructs `PiAgentRuntimeFactory` directly; neither CLI runs there at all |
| Pi | Out of scope — no MCP configuration surface |

**No migration:** the token store is in memory.

## Also fixed here

`parseServerBusinessFrame` tried only the domain frame union, which contains no runtime-credential
server frames — so `runtime:execution:result`, `runtime:proxy:ticket:result` and the gateway's own
result were treated as invalid frames and **dropped the connection**. That predates this PR and
affects the whole proxy-mode control plane; it went unnoticed because proxy mode is off by default
and every suite on both sides of that boundary stubs the connection.

## Verification

`pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`,
`pnpm --filter @opentag/client test:agent-runtime:coverage` (100% on all four metrics),
`pnpm test:coverage`, and `pnpm --filter @opentag/server test:integration` (38 files, 519 tests) all
pass locally.

Coverage includes: the execution fence's every violation; era dispatch and the replay rule, asserting
the **number of dispatches** rather than that a call failed; the route contract; aggregation and
routing including collisions; name composition; the wire contract over a real WebSocket (capability
negotiation, the service request, the token frame round trip); the whole Client chain from a granted
service to the `mcp.json` Claude Code is launched with; cancellation over a **real listener**, since
`inject` never disconnects; and 8 integration tests driving `initialize` → `tools/list` →
`tools/call` against real PostgreSQL and two loopback fixture Servers.

**Open gap: no run against a real third-party MCP Server, including a call longer than 10 s.** It
needs a proxy-mode deployment, so it is best done after #653.


liuchao-001 (changes requested):
Reviewed at `34d027d6`. The security shape is right: the bearer carries only an execution id, the fence runs on every request, revocation hangs off the one close path, the URL is composed against the pinned origin, and a call can only route inside the Agent's own snapshots. Two things stop this from delivering working tools, plus a few smaller ones.

## Blocking

**1. Every runtime tool call is capped at the probe's 10-second deadline.**
`McpGatewayService` calls upstream through the shared `McpOutboundFetcher` (`index.ts`: "shares the one outbound fetcher"), and `#perform` aborts after `this.#timeoutMs`, which is `MCP_DEFAULT_TIMEOUT_MS = 10_000` (`mcp-url-policy.ts:73,407`). No per-call override exists. The design doc (l.321, l.688) sets that deadline for probes and then applies it "unchanged" to runtime calls. 10s is fine for `tools/list`, but real tools such as search, DB queries, code execution, and browser actions regularly take longer. Each of those comes back to the model as `isError: "The MCP tool call failed"` with nothing it can act on. The legacy path spends part of the same budget on its `initialize` handshake.
→ Give the runtime caller its own deadline (a separate fetcher instance, or a per-call `timeoutMs` in `McpFetchInit`), sized for tool calls.

**2. Runtime calls share the 4-per-Account concurrency budget with probes and the refresh worker, and a full budget rejects immediately.**
`fetchOutbound` throws `Too many concurrent MCP requests for this Account` when `inFlight >= 4` (`mcp-url-policy.ts:341`). It does not queue. An Account running a few Agents in parallel can hit that limit, and so can one Agent issuing parallel tool calls, which Claude Code does. A legacy-era call takes 2–3 of those slots by itself (initialize + notify + call). A background probe or refresh then starts failing live tool calls. Combined with #1, a single slow upstream holds a slot for the full deadline.
→ At minimum, give runtime calls a separate budget from management traffic. Queueing with a short wait would be better than failing fast.

## Should fix (non-blocking)

3. **Name collisions publish a tool that can never be called.** In `catalog()` the comment says "both are dropped". The code actually keeps the first entry (`seen.set` happened on the first one) and drops only the second. `resolveTool` then finds 2 matches and returns `undefined`, so the published tool always answers "No MCP tool named … is available". Either drop both names in `catalog` or make `resolveTool` take the first match. No test covers the collision path.
4. **Upstream JSON-RPC errors lose their message.** A `tools/call` that the upstream rejects with e.g. `-32602 invalid params` arrives as `McpTransportError`, not `McpServiceError`. `buildHandlers` therefore replaces it with the generic "The MCP tool call failed". That is the case where the model most needs the upstream text to retry with better arguments. Forward the (bounded) upstream error message.
5. **Client disconnects don't cancel upstream work.** `callTool` accepts a `signal`, but the route never passes one, so an abandoned request keeps its upstream fetch and its concurrency slot until the deadline. Wire `request.raw` close into an `AbortController`.
6. **`eraInvalidated` / `era` from `McpUpstreamCaller` are discarded.** A row cached as `modern` whose server is actually legacy pays a failed modern round trip on every call. Persist the corrected era, or drop the fields until something reads them.
7. **Unused contract surface.** `MCP_GATEWAY_RESULT_MAX_BYTES`, `result_too_large`, `tool_ambiguous`, `catalog_unavailable`, and `timeout` are declared and mapped to statuses, but nothing produces them. Either use them (a timeout should be distinguishable from a generic failure) or drop them so the contract describes what is real.

## Nits

- `mcp-server-service.ts`: the "One Agent's mounts as raw joined rows" JSDoc block sits stacked above `hasUsableMount`'s own JSDoc. It belongs on `listAgentBindings`.
- `instructionsOf` bounds by bytes but slices by UTF-16 units (`text.slice(0, MAX_BYTES)`), so non-ASCII notes can exceed 8 KiB. Harmless, but the check and the cut disagree.
- Only `POST` is registered. The Streamable HTTP client may `GET` the endpoint for a server stream, and a gateway without one should answer `405` rather than Fastify's `404`.

## Verification ask

Please attach the manual end-to-end run against a real third-party MCP Server listed as "not yet done". It should include one call slower than 10s, which is exactly the case #1 breaks. Per our process, independent exact-head QA for this should come from a `focused-local` run with a local Computer + Claude Code.

packages/server/src/services/mcp/mcp-gateway-upstream.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (!(error instanceof McpTransportError) || error.status === undefined) throw error;
+      const detection = detectEraFromRpcError(error.status, error.rpcError);
+      /*
+       * The body decides, exactly as the probe's downgrade rule does: a recognizable modern JSON-RPC
+       * error means the peer *is* modern and merely wants another version — never downgrade to the
+       * handshake for one of those, because a modern Server would refuse `initialize` too.
+       */
+      if (detection.era === "legacy") return this.#callLegacy(input, input.cachedEra !== null);
+      const retryVersion = negotiateProtocolVersion(
+        detection.retryWithVersions.length > 0 ? detection.retryWithVersions : [MCP_MODERN_PROTOCOL_VERSION],
```
yuezengwu, line 99: [P1] Do not replay tool calls after ordinary upstream errors

This catch retries every `McpTransportError` with an HTTP status. For both HTTP 500 and HTTP 200 carrying JSON-RPC `-32603`, `detectEraFromRpcError` returns modern with no suggested versions, so this path picks the default version and sends the same `tools/call` again with a new request ID. If an upstream creates an issue and then fails while producing its response, one model invocation creates two issues and the gateway reports only the second success. I reproduced both cases with a real loopback HTTP fixture and the production caller/transport/fetcher: two POSTs, distinct IDs, two recorded side effects. Please restrict retries to explicit protocol-negotiation refusals that establish the operation was not executed; propagate ordinary tool/5xx failures without replaying them. Add regressions asserting exactly one invocation for these two cases, while retaining an explicit version-negotiation control.

yuezengwu (changes requested):
Additional exact-head review at `34d027d6df8e6694c296719f0d8d798bffd2a263`: one independently reproduced P1 in upstream retry handling, detailed inline.

Validation in a clean detached worktree: `pnpm check`, `pnpm build`, `pnpm typecheck`, 482 relevant unit tests and 8 gateway PostgreSQL integration tests passed. A separate real-loopback-HTTP probe through the production caller/transport/fetcher reproduces duplicate side effects for both HTTP 500 and HTTP 200 with JSON-RPC `-32603`; the explicit version-negotiation control passes. No real third-party MCP/Claude Code end-to-end run was performed.

packages/server/src/services/mcp/mcp-gateway-upstream.ts
liuchao-001: Confirmed from the code at `34d027d6`, and I missed this in my review. `detectEraFromRpcError` returns `{ era: "modern", retryWithVersions: [] }` for any status other than 400, 404, or 405, so `#callModern` re-sends the `tools/call` at the default version.

The replay surface is a bit wider than 500 and `-32603`. When the fix restricts retries, please cover these paths too:
- A 400 or 404 whose body is not a recognized modern error code goes down the `legacy` branch, and `#callLegacy` re-sends the same `tools/call` after a fresh `initialize`. A modern server that 404s on something tool-specific would execute twice.
- `call()` falls back to legacy whenever `invalidatesProtocolEra(error)` is true, and that includes `-32601 METHOD_NOT_FOUND`. For a `tools/call`, that can come from inside the tool.

Only a refusal that proves the request was never dispatched (`UNSUPPORTED_PROTOCOL_VERSION`, or a transport-level 405 before any JSON-RPC handling) should be replayable.

liuchao-001:
Re-checked at `7f81aca2`. The new commit adds coverage and gives the route its own JSON-RPC error handler. Both look good. The earlier findings are still open, so my changes-requested review stands:
- the 10s probe deadline on runtime tool calls
- the shared 4-per-Account concurrency budget that fails fast
- the `tools/call` replay P1 (https://github.com/first-tree-ai/opentag/pull/652#discussion_r4052260847)

One thing about the new `mcp-gateway-upstream.test.ts`: `propagates a transport failure that says nothing about the era` (HTTP 500) only asserts that the call rejects. With the replay bug in place, the scripted fetcher gets hit **twice** and then rejects, so this test currently passes while the bug is present. When you fix the replay, please assert the invocation count here (exactly one POST). Add the same for HTTP 200 + `-32603`, and for the 400/404 non-modern-body and `-32601` legacy-fallback paths.

liuchao-001:
Re-checked at `2ec99c57`. The withheld-reason logging is correct, and the logger reaches it through the owner's options. The three blockers from before are still open: the 10s deadline, the shared fail-fast concurrency budget, and the `tools/call` replay.

This commit also states something the PR description needs to reflect. **The gateway only exists for Clients in proxy credential mode.** That mode is opt-in through `OPENTAG_RUNTIME_CREDENTIAL_MODE` and off by default, and until `7005eef1` it could not complete an execution open over a real connection at all. So on a default install today, a Claude Code Agent on a local Computer gets no MCP tools. The scope table's "Claude Code, local Computer — Delivered" overstates that. Please either:
- state the proxy-mode prerequisite in the description and the design doc, and say what it takes to turn it on by default, or
- decouple the gateway token from the relay, so legacy-mode Clients can get it too.

One small note on the diagnostic. A legacy-mode Client never sends `runtime:execution:open`, so `describeMcpService` never runs for it, and the commonest case you describe (`not_requested` on a default install) will never actually be logged. That case needs a signal on the Client side, for example logging at turn start when the Agent has MCP mounts but no gateway, or a note in the web UI.

bestony:
Both blocking findings are fixed, along with all five should-fix items and the three nits. New head `41d5c094eaa821de678148b917dbd21eced8d33f`.

## Blocking

**1. Tool calls are no longer replayed.** You were both right, and the replay surface was wider than the two cases I was shown — I verified all four paths. Root cause: I adapted the probe's downgrade rule without carrying over the one fact that differs, that `server/discover` and `tools/list` are reads and `tools/call` is not. My own module docstring claimed the two have "different failure semantics" and then didn't honour it.

A retry now happens only for `-32022 UnsupportedProtocolVersion`, which a Server emits while rejecting the envelope and which therefore proves nothing ran. Everything else propagates: 5xx, a 200 carrying `-32603`, `-32601` (which for a `tools/call` can come from inside the tool), and a 400/404 whose body is not a recognized modern error. The runtime path no longer re-detects an era at all — that is the probe's job, and a callable tool always has an era on its row because only a succeeded probe publishes one.

Regressions assert the **number of dispatches**, not that the call failed, since one versus two is the whole defect. They also assert no `initialize` follow-up, which was the second replay shape. Verified they fail against the old rule (6 red).

**2. Runtime calls have their own budget.** A second `McpOutboundFetcher` instance: 120 s deadline, 8 concurrent per Account. `#inFlight` is per instance, so separate objects give separate counters; the URL gate, redirect refusal and response bound are unchanged because they are properties of the class. I had described sharing the fetcher as a virtue in the PR body — that was wrong for the reason you gave.

I did not implement queueing. Fail-fast at 8 with a separate pool seemed the proportionate step; say the word if you want a short wait instead.

## Should fix

3. **Collision** — fixed, and the comment was the lie: the code kept the first arrival. Names are now counted across every mount before anything is published, so both sides are withheld and the clash is stated once. Tests added; the reachable trigger is a Server whose own `tools/list` repeats a name.
4. **Upstream error text** — I could not reproduce this one. `McpTransportError extends McpServiceError` and already carries `boundedMcpSummary(payload.error.message)`, so `buildHandlers` forwards it. A test now pins the property, since it was undefended either way. If you have a case where it is flattened, I would like to see it.
5. **Cancellation** — the socket close is wired into an `AbortController` and the signal passed to `callTool`; an abort is reported as cancelled rather than as a tool failure.
6. **`era` / `eraInvalidated`** — dropped. After the replay fix they described a re-detection that no longer happens.
7. **Unused contract surface** — the error enum is now exactly what the gateway produces. `tool_ambiguous` became unreachable once collisions withhold both sides; `result_too_large`, `catalog_unavailable`, `invalid_request`, `request_too_large`, `protocol_unsupported`, `tool_unknown` and `unknown` were mapped to statuses before anything raised them. `MCP_GATEWAY_RESULT_MAX_BYTES` is gone too — the fetcher already bounds responses.

## Nits

All three: the `listAgentBindings` JSDoc had drifted onto `hasUsableMount`; `instructionsOf` now truncates the encoded form so a non-ASCII note cannot exceed the bound the check just applied; `GET` answers 405 with `Allow: POST`.

The design doc's "applies unchanged to runtime calls" line is corrected, and the replay rule is written down there.

## Verification

`pnpm check`, `build`, `typecheck`, full `test`, the agent-runtime 100% gate, `test:coverage`, and the 38-file PostgreSQL integration suite all pass locally.

**Still not done: the real third-party end-to-end run, including a >10 s call.** I cannot run it — it needs a real deployment, and there is now a second prerequisite worth flagging: the gateway only works with `OPENTAG_RUNTIME_CREDENTIAL_MODE=proxy`, because the credential relay does not run in the default `legacy` mode. That dependency is #653, and it was my omission not to state it in the original PR.

liuchao-001 (approved):
Re-reviewed at `41d5c094`. Both blockers are fixed, so I'm approving.

- **Replay:** `retryableVersionRefusal` now allows a retry only on `-32022`. Runtime calls no longer re-detect the era, and the regressions count dispatches. That covers all four paths.
- **Budget:** runtime calls get a separate `McpOutboundFetcher` instance (120s, 8/Account). Because `#inFlight` is per instance, probes and refresh can no longer starve a live turn. Fail-fast at 8 on its own pool is a proportionate choice; no queue is needed now.
- Collision withholding, the trimmed error enum, the JSDoc placement fix, the byte-correct truncation, and GET→405 all look right.
- **#4:** you're right and I was wrong. `McpTransportError extends McpServiceError`, so the bounded upstream text was already forwarded. Thanks for pinning it with a test.

## Should fix (follow-up is fine)

**The cancellation wiring doesn't cancel anything on a real socket.** On Node ≥16, `IncomingMessage` emits `close` once the request body has been consumed, not when the client goes away. Fastify has already parsed the body before `handleGatewayRequest` runs, so `request.raw.once("close", …)` is attached after the event has fired and never runs. I reproduced this against this head's route over real HTTP (`app.listen` + `fetch`, the client aborting at 300ms, and a stub `callTool` that works for 1.5s). The upstream work ran to completion with `signal.aborted === false`. The route test passes because `inject` never simulates a disconnect. The fix is to listen on the response instead: `reply.raw.once("close", () => { if (!reply.raw.writableEnded) aborted.abort(); })`. In the same probe, `reply.raw` `close` fires at about +307ms with `writableEnded=false` on disconnect, and with `true` on normal completion. A regression test needs a real listener, not `inject`.

## Please update the PR description

- The scope table still reads "Claude Code, local Computer — Delivered" with no qualification. Please add the `OPENTAG_RUNTIME_CREDENTIAL_MODE=proxy` prerequisite and the dependency on #653, as you stated in the comment.
- The Verification section still describes the shared fetcher. The `mcp-gateway-upstream.ts` module docstring also still says the per-Account concurrency budget applies "unchanged".

The real third-party end-to-end run, including a call longer than 10s, remains the open verification gap. It needs a proxy-mode deployment, so it can be done after #653 lands.

yuezengwu (approved):
Re-reviewed `41d5c094eaa821de678148b917dbd21eced8d33f`. My blocking replay finding is fixed: independent real-HTTP probes now observe one dispatch and one side effect for HTTP 500, HTTP 200 with `-32603`, HTTP 400/404, and `-32601`; the explicit protocol-version refusal control still negotiates successfully.

A separate real-HTTP probe also confirms runtime admission while all four management slots are occupied, and completion of a runtime call exceeding 10 seconds.

Validation: `pnpm check`, `pnpm build`, `pnpm typecheck`, 623 related unit tests, 8 PostgreSQL gateway integration tests, and all 7 independent probes passed. Approving the code at this head. Real third-party/Claude Code end-to-end acceptance remains outside this verification.

bestony:
@liuchao-001 you were right about the cancellation, and the test I wrote for it was worthless. Fixed at `1ed803c0288d7e6c7882e433d19c77527ba4047b`.

**The wiring never cancelled anything.** Exactly as you described: `IncomingMessage` emits `close` once the body has been consumed, Fastify parses the body before the handler runs, so the listener was attached after the event had already fired. Now on `reply.raw`, with `writableEnded` separating a client that disappeared from a response this route finished sending.

The old test asserted that a signal was passed and that it was not yet aborted — true whether or not the mechanism works. It is replaced by two driven over a real listener (`app.listen` + `fetch`, client aborting mid-call), because `inject` cannot disconnect and so cannot observe this at all. I confirmed the pair fails against the old wiring: `expected false to be true`.

That is twice now that a test of mine passed without exercising the thing it named. Both times the fix was to drive the real transport rather than a double.

Also corrected the `mcp-gateway-upstream` docstring, which still claimed the per-Account concurrency budget applied unchanged.

**PR description updated:** the scope table now says "proxy mode only" and links #653, the prerequisite has its own section, and the verification section describes the split fetcher and the real-listener cancellation tests.

@yuezengwu thank you for the independent probes — the runtime-admission-while-management-is-full one and the >10 s completion are precisely the two properties I could not demonstrate from unit tests.

The real third-party end-to-end run remains open and is best done after #653, since it needs a proxy-mode deployment.

liuchao-001 (approved):
Re-approving at `1ed803c0`. I ran my own disconnect probe again, over real HTTP against this head's route:
- A client that aborts at 300ms now aborts the `callTool` signal at +303ms.
- A normal completion returns 200 with `aborted=false`.

The upstream docstring and the PR description (the proxy-mode prerequisite, #653, and the split fetcher) now match the code.

The only open item is the real third-party end-to-end run after #653.

PR #654, opened 2026-09-19
fix(feishu): activate saved connections after delayed approval

yuezengwu:
## Summary

A Feishu app approved after registration can leave setup permanently blocked because the short-lived registration flow discarded its credentials. Persist the issued credentials on the existing binding and resume activation automatically when all 66 required permissions, app identity/availability, and runtime readiness are satisfied. Users can close the page or restart the Server while waiting; they do not need to be enterprise administrators or scan again.

- Reuse the encrypted setup context with a fixed 30-day candidate deadline. Background and manual checks share retry limits and claim fencing; cancellation, expiry, disable, replacement, reauthorization, and shutdown preserve the correct active connection and clear temporary secrets when appropriate.
- Show the remaining activation condition and saved authorization in setup and Settings, restore it after remount without registering another app, and advance automatically when the connection becomes usable. Settings reads the current attempt independently of an existing healthy connection, using the existing attempt DTO and Account ownership checks. The CLI also prints the waiting reason, missing scopes, next check and expiry.
- Preserve unauthenticatable candidates until the fixed deadline, allowing key-ring recovery without losing issued credentials. Explicit cancel and expiry still clear them. Treat omitted/unknown bot activation status as unknown, retain mandatory scope/identity/runtime checks, and recognize all six documented app-unavailable error codes.
- Correct the live integration failures found during acceptance: retain Feishu SDK business error codes, use the correct bot-info URL, advertise only installed credential capabilities, preserve the managed CLI entry point in Codex shells, and grant only the current visible session's reply directory for reliable outgoing-message receipts.
- Database migration 0047 adds one enum value and extends one CHECK constraint. No new tables, columns, indexes, or customer-data backfill. The large generated snapshot records that minimal schema change. See [lifecycle and rollout design](docs/feishu-connection-lifecycle.md).

## Testing

Local validation of head `9343537cc00d4e037cffdf740bf4b8822ff59dd0`, based on main `316ef450`:

- Frozen-lockfile installation, `pnpm check`, `pnpm build`, `pnpm typecheck` and `pnpm test` passed. Unit results: scripts 417, Shared 414, Server 2,311, Client 1,751, CLI 512, Web 1,099. Unchanged dependency/package results may be reused by the workspace cache; changed Server and Web tests executed in the final full run.
- Runtime coverage remains 100%: statements 4,213/4,213, branches 3,054/3,054, functions 743/743 and lines 3,744/3,744 (450 tests).
- The standard PostgreSQL integration command passed all 532 tests in 40 suites in the review follow-up run. Migration 0047 remains byte-for-byte unchanged; upstream 0046, prior journal/hash entries and snapshot ancestry are preserved. The migration count is 48.
- New regressions cover real v2 key-ring skew/recovery, sweep progress, stale claims, cancellation, fixed expiry/new QR admission, authenticated-invalid identity, unknown bot status, official provider codes, Account-authorized current-attempt reads, Settings remount and late-response races. API CSRF inventory and shared UI fixtures include the new read-only endpoint.
- `git diff --check` passed. Remote results are recorded independently in this PR's Checks tab.

Live local acceptance used a dedicated Feishu app and real permission/API/message operations: one required scope was withheld for about 5 minutes 23 seconds, the browser and local services were closed/restarted, and restoring the scope led to automatic activation without rescanning or clicking a status check (about 31 seconds in this observation). A real bot reply, persisted outgoing receipt, and Web task display agreed; a final full restart repeated the successful reply. Codex CLI 0.153.4 and lark-cli 1.0.92 were used.

Acceptance boundary: this tenant exempts the selected permission from administrator review. This proves delayed effective permissions and restart recovery, not a separate employee/admin approval across accounts or days. Live acceptance preceded the main-branch synchronization and review follow-up; the validation above covers the resulting source. The later Settings browser checks used a local mocked API, not another external approval acceptance. No production deployment or customer-data repair is claimed.

## UI validation

- Desktop evidence: real local waiting, restart recovery, automatic-ready, and sent-reply task views inspected; a separate in-memory Lab pass checked the final setup presentation.
- Mobile evidence: waiting and expanded permission details inspected at 320px and 390px; no horizontal overflow.
- Widths checked: 320px / 390px / 768px / 1440px, with screenshots retained in the task's validation artifacts.
- States verified: QR authorization, saved authorization awaiting permissions, expanded app/permission details, checking latest status, connection handoff, and ready. A local browser fixture using the real Settings components additionally verified remount, manual check, automatic success, and cancellation with zero create calls, including a pending replacement alongside an active old binding.
- Keyboard/accessibility checks: status has a status role; Enter opens permission details and invokes the check action; Tab reaches cancellation in order.
- New reusable block, theme value, or CSS exception: `FeishuActivationWaiting` shares the waiting presentation across setup surfaces; existing controls/theme values, no new CSS exception.

## Breaking changes

Coordinated rollout is required: stop all old Server setup writers, apply the migration, then start matching Server and Web builds; release/install the matching CLI and reconnect it. Do not mix old and new Server writers in a rolling rollout. Rollback needs a build that understands the new enum and encrypted candidate format. Previously discarded credentials cannot be reconstructed by editing a connection status.

## Checklist


liuchao-001 (approved):
Reviewed at `297836a1`. I found no blocking issues. The claim fencing holds: every settle, release, heartbeat and activation compares the per-check claim token, `validating` state and ciphertext, and activation re-checks under `FOR UPDATE`. Cancel, disable, expiry and terminal paths clear `setup_*` only, so the active credential and generation are preserved during reauthorize/replace. The `::text` cast in the CHECK avoids the same-transaction enum problem, a crashed claimant's stale heartbeat is recovered by CAS, and the QR flow reuses a saved candidate instead of overwriting it. Approving. Below are four non-blocking items, in priority order. I'd take the first two in this PR or a quick follow-up.

1. **A decrypt failure permanently discards a valid candidate** (`setup-context.ts:117-124` → `setup-service.ts` `#considerPendingRow` → `#failInvalidContext`). Outside `strict` mode, any `cipher.decrypt` exception becomes `undefined`. That includes an unknown key id / `AUTHENTICATION_ERROR` from an instance with an older key ring. The sweep then writes `failed` and nulls the only copy of the 30-day secret. This contradicts the "saved authorization survives" contract, and the credential can't be recovered afterwards. Suggest separating "can't authenticate with my key ring" (skip and log, leave it for another instance or a later sweep) from "authenticated but malformed/identity-mismatched" (terminal).
2. **`activate_status` missing ⇒ waits forever** (`connection-manager.ts:213`, `adapter.ts:699`). `null !== 2` turns an omitted field into `app_unavailable` until expiry, and the legacy activation path never checked this. Suggest treating `null` as unknown and failing only on known non-enabled values.
3. **Previously released CLIs can't parse the reused attempt.** `FeishuSetupAttemptSchema` is `.strict()` and now carries `pending_activation` plus `activation`, so an older `connect-feishu`/`reauthorize-feishu` hits a parse error when a candidate is waiting. The rollout notes cover shipping the matching CLI, but the new CLI's `formatFeishuSetup` doesn't print `activation` either: it shows `pending_activation` with `qrUrl -` and no hint about what's missing.
4. Smaller items:
   - The Settings → Messaging `ImTab` doesn't rediscover a saved attempt after remount, so the waiting/check/cancel UI is gone until the user clicks Connect again.
   - `cancel()` strict-decodes before the clearing UPDATE, so an undecryptable candidate can't be canceled by the user.
   - `setup-check.ts` maps only 10014 to `app_unavailable`. App-disabled/not-installed codes such as 11207, 11210, 20009, 99991662 and 99991673 show as a temporary failure (the label is wrong; nothing is lost).
   - `--disable shell_snapshot` sits in the base Codex args, so it applies to all Codex sessions. The docs say "managed executions only".
   - `visibleReplyWritableRoots` isn't guarded like `prepareConfigurationRoots`, so a bad plans directory blocks every visible session from starting instead of only degrading receipts.

CI was still running when I reviewed, and ownership-gate is waiting on bestony.

liuchao-001 (approved):
Re-reviewed at `9343537c`. The sync commit `65f4e9c0` only moves the migration: the Feishu SQL is byte-identical as 0047, and the snapshot chains on main's 0046. The follow-up commit addresses every item from my earlier review:

- A decrypt failure now reads as `undecryptable`, separate from `invalid`. The sweep, check, reuse and immediate-check paths all keep the row until its fixed deadline, and reuse and check return 409 instead of replacing or failing the candidate. Authenticated-but-malformed contexts still go terminal.
- `activate_status` waits only on the documented non-enabled values. A missing field no longer blocks activation.
- The app-disabled and not-installed codes are classified as `app_unavailable`.
- `cancel()` no longer needs a strict decode, so an unreadable candidate can be canceled.
- Settings restores the saved attempt through the new read-only GET. The CLI prints the activation details, and the docs now scope `shell_snapshot` correctly.

One non-blocking leftover: `observeForAgent` / `#projectAttempt` still strict-decode. For a key-skewed row, the new settings GET (and `check()` on a `validating` row) therefore surfaces the raw cipher error as a 500, where it could return the new retained-authorization 409. Nothing is lost, only the message the user sees. Also, `isRestorableAttempt`'s `succeeded` branch can't be reached from `getForAgent`, which returns open states only.

Approving. CI is still running on this head, and ownership-gate is waiting on bestony.

yuezengwu:
Addressed the [review follow-ups](https://github.com/first-tree-ai/opentag/pull/654#pullrequestreview-5255646966) in `9343537cc00d4e037cffdf740bf4b8822ff59dd0`:

- Separate an unauthenticatable context from authenticated-invalid plaintext. Missing-key/tamper ambiguity never activates the candidate or deletes it before its fixed deadline; manual create/check cannot replace it, sweeps continue to other rows, and the correct key can recover it. Explicit cancel and expiry work without decryption. New real-v2-keyring regressions cover recovery, sweep progress, stale claims, cancellation, and expiry/new QR admission.
- An omitted or unknown bot activation status is no longer treated as proof that the app is disabled. Explicit documented disabled states still wait; all 66 scopes, identity, runtime and atomic channel activation remain required.
- Both classifiers now recognize the six official app-unavailable codes, including wrapped numeric strings. Source: [Feishu common error codes](https://open.feishu.cn/document/server-docs/api-call-guide/generic-error-code.md).
- Settings restores the current saved authorization with a read-only GET on the existing Agent setup-attempt collection, reusing the existing DTO and Account ownership checks. The setup snapshot intentionally prioritizes a healthy active binding, so reading the open attempt independently also handles replacement while the old connection still works. Mounting does not register another app. Late responses cannot overwrite newer actions or another Agent.
- The CLI prints saved-authorization guidance, the waiting reason, missing scopes, next check and fixed expiry. The matching-CLI rollout requirement remains: old strict DTO readers are not claimed compatible with the new state.
- English and Chinese docs now distinguish default Codex shell-snapshot disabling from the managed-environment ZDOTDIR override.

The writable reply-root preparation remains fail-closed intentionally. Its callback validates the trusted plan hierarchy and grants only the Session runs subtree. Ignoring that failure would silently remove required reply evidence or bypass the storage trust boundary; optional Context Tree configuration is a different case. This decision is now documented in both languages.

No additional database schema changes were introduced. Validation details and the exact tested head are in the updated PR description. Local browser verification covered remount, check, automatic success and cancellation with zero create calls, including an active old binding. This uses a mocked API fixture and does not claim another live administrator-approval acceptance.

Repository checks, build, typecheck and full unit tests passed (Server 2,311; Web 1,099). The standard PostgreSQL suite passed all 532 tests; the 450-test agent-runtime suite retains 100% coverage. Fresh CI and code-owner review of this commit are still required.

bestony (approved):
Reviewed the full diff at `9343537c` (server lifecycle, migration 0047, client runtime, Web/CLI). No blocking defect found; the candidate lifecycle, claim fencing and the three-way context read (`decoded` / `invalid` / `undecryptable`) hold up, and the constraint/disable/activation triple is internally consistent: `disableImBindingInTransaction` now maps an open state to `canceled`, which is exactly what the replacement branch of the activation update matches.

Migration 0047 looks transaction-safe: the new label is only compared through `setup_state::text`, so the freshly added enum value is never referenced in the transaction that adds it, and the `pending_activation` branch is additive over the previous shape.

Non-blocking notes for the rollout, mostly about behavior outside Feishu:

1. **Old CLI/Web builds hard-fail on the new DTO.** `FeishuSetupAttemptSchema` is `.strict()`, so an installed older CLI rejects both the `pending_activation` state and the new `activation` field. The PR body covers this, but the CLI is user-installed: the release notes need to say plainly that the CLI must be upgraded, not just that it "should" be.

2. **Codex shell changes are a user-visible behavior change unrelated to Feishu.** `--disable shell_snapshot` plus `shell_environment_policy.set.ZDOTDIR=$OPENTAG_HOME` means `~/.zshenv` is no longer sourced for managed visible-session shells, so a user who prepends tool paths there (mise/asdf/nvm shims, proxies) loses them inside tool calls. That is the intent, but it deserves a release note. Separately: is `shell_snapshot` accepted by every Codex build we support? There is no minimum-version gate in `packages/client/src/providers/codex`, and an unknown `--disable` name would break app-server startup for users on an older CLI.

3. **Capability offers are now mode-dependent.** `setCredentialProxyEnabled(false)` in the default `legacy` mode drops `runtimeCredential`/`providerProxy`, so the Server's `#proxyMode()` falls back to the legacy grant path. That is the correct fix, but it changes which path the Server picks for every legacy client the moment the new CLI lands — worth watching provider-CLI grant metrics right after release.

4. **Operational cost of the 30-day candidate.** Retries are a flat `checkIntervalMs + jitter` with no exponential backoff, so a candidate stuck on `temporary_failure` keeps polling the provider for the whole retention window; the circuit key limits damage but the volume is still real. Also, as the design doc states, key rotation must keep the old key for 30 days — that constraint belongs in the ops runbook, not only in `docs/feishu-connection-lifecycle.md`.

5. A candidate this instance cannot decrypt blocks a new registration with a 409 until it is explicitly canceled or reaches its deadline. Correct by design, but support should know that "cancel the saved authorization" is the escape hatch.

The `Checks` job was still in progress when I reviewed; this approval assumes it lands green.

PR #655, opened 2026-09-19
feat(shared): add the Agent Skills contract and design

bestony:
## Summary

First slice of **Agent Skills**: the `@opentag/shared` contract and the design of record. No server, client, CLI or web behaviour changes yet — those lanes build on this and are stacked on this branch.

A Skill is a directory (`SKILL.md` + supporting files) uploaded to the platform, stored as one `tar.gz` object in S3-compatible storage, owned by exactly one Agent, synced to that Agent's Computer at runtime start, and pushable back up by the Agent through the CLI. Out of scope by decision: versioning (always exactly one version), online editing, review/moderation, cross-Account sharing.

- `packages/shared/src/skill.ts` — limits, `SkillNameSchema` + reserved names, a dependency-free `parseSkillManifest` (plain / quoted / `>` / `|` scalars, never throws), resource and request schemas, the install-marker schema, upload header names, and `SKILL_ERROR_CODES` with category/status metadata. Browser-safe: no Node standard-library imports.
- `packages/shared/src/http-paths.ts` — three surfaces: Account (`/api/v1/agents/:agentId/skills…`), Computer (`/api/v1/computer/agents/:agentId/skills…`, machine token), Agent CLI (`/api/v1/runtime/skills…`, session proof).
- `docs/design/agent-skills.md` — ownership model (Skill belongs to one Agent, not an Account pool), single-version replace semantics, canonical `tar.gz` with server-side validation and deterministic re-pack, `SkillObjectStore` interface + S3 adapter on `aws4fetch`, streamed-through-Server transport instead of presigned URLs, optional storage at deploy time, runtime-start delivery with marker-guarded sync and soft failure, v1 = Local Computers only. Linked from `docs/README.md`.

## Checklist

- [x] 1. `packages/shared/src/skill.ts` written, typechecks
- [x] 2. HTTP path templates and builders added to `http-paths.ts`
- [x] 3. Exported from `index.ts` and `browser.ts`; public-exports snapshot updated
- [x] 4. `skill.test.ts` (and path-builder tests) written and passing
- [x] 5. `docs/design/agent-skills.md` written and linked from `docs/README.md`
- [x] 6. All acceptance criteria verified, `git status` clean

## Acceptance criteria — re-run by the orchestrator on `85f22552`

| # | Criterion | Result |
| --- | --- | --- |
| 1 | `pnpm check` | exit 0 (complexity ratchet passed) |
| 2 | `pnpm typecheck` | exit 0, 8/8 tasks |
| 3 | `pnpm --filter @opentag/shared test` | exit 0, 36 files / 446 tests, incl. `skill.test.ts` and `public-exports.test.ts` |
| 4 | `skill.ts` has no Node stdlib import and no `Buffer` | 0 hits |
| 5 | Every new runtime export is in `public-exports.snapshot.json` | holds (types are erased at build and are never in that snapshot) |
| 6 | Diff touches only `packages/shared/**` and `docs/**`; nothing under `.dispatch/`, no `docs/zh-CN/` | holds (8 files) |
| 7 | Design doc opens with a Status block, records the eleven decisions, linked from `docs/README.md` | holds |
| 8 | No new file over 500 lines | 423 / 313 / 248 |
| 9 | Clean tree | holds |

Deviations from the confirmed plan: none recorded.

Known follow-up (non-blocking, will land on this branch): a `description` written as a `>`/`|` block scalar keeps its trailing newline; it should be trimmed in `SkillManifestSchema`.

## Provenance

Written by an agent under `dispatch-opencode` run `tlm1q4`, lane `skill-contract-1`. Plan and acceptance criteria were authored and the result verified by the orchestrator (Claude Code) before this branch was pushed. 6 commits for a 6-item checklist.


liuchao-001 (changes requested):
Reviewed at `85f22552`. The contract shape (one owning Agent, content-hash object keys, server-side re-pack, streamed transport, marker-guarded soft sync) is sound, and `skill.test.ts` passes locally (34/34). Two things need to change before this becomes the contract the other lanes build on.

**Blocking 1 — `parseSkillManifest` rejects spec-valid manifests, including OpenTag's own.** Any indented line in the frontmatter is a hard reject, so a nested value under an *unknown* key fails the whole manifest. That contradicts the docstring ("Unknown top-level keys are ignored") and the Agent Skills format, where `metadata` is a map and `allowed-tools` is often a list. I ran the parser over the 93 real `SKILL.md` files on one machine: **41 are rejected** with `frontmatter has an indented line`, including `context-tree-read` as installed by the context-tree CLI (`metadata:\n  author: …`) and every `lark-*` skill (`metadata:\n  requires: …`). Multi-line plain scalars (`description: Use when\n  …`), which YAML folds, are rejected too. Suggested fix: indented or `- ` lines that follow a key other than `name`/`description` belong to that key and are skipped; a multi-line plain `description` is folded like `>`. Add tests built from a real spec example (metadata map + list-valued `allowed-tools`).

**Blocking 2 — the Codex target must not fall back to the account home (design doc).** The doc leaves the Codex path "to be verified … fallback documented there". The only precedent in the tree is `packages/client/src/runtime/context-tree.ts` (~L309–316), which installs Codex skills into the OS account home's `.agents/skills`. That directory is shared by every Agent on the Computer, so using it as the fallback would break the design's central invariant (a Skill belongs to exactly one Agent). Please record it as a constraint now: Codex materialization must be Agent-scoped (workspace `.agents/skills`, or a per-Agent `CODEX_HOME`); if neither works, Codex is out of v1. Account home is never the fallback.

**Non-blocking**
- **Push-then-sync gap.** `opentag skill push <dir>` uploads from an unmarked, locally authored directory. On the next sync the manifest names a Skill whose directory exists without a marker. Sync has to leave it alone, so later Web replace/disable/delete never reaches that Agent. Define what happens here, e.g. the CLI writes the marker after a successful push, or sync reports the name collision. Also state explicitly that marked directories absent from the manifest (disabled/deleted) are removed.
- **Name rule.** `^[a-z0-9][a-z0-9-]{0,63}$` accepts `pdf-` and `a--b`; the Agent Skills spec forbids trailing and consecutive hyphens. `^(?=.{1,64}$)[a-z0-9]+(-[a-z0-9]+)*$` would match it.
- **Parser leniency.** `name: pdf # comment` keeps the comment in the value, and duplicate keys silently take the last value. Rejecting both with a specific reason would fit the parser's "refuse rather than guess" stance.
- Agree with the listed follow-up on trimming the block-scalar `description` newline.

Posting REQUEST_CHANGES for the two blocking items; the non-blocking ones can land later on this branch.

bestony:
Thanks — both blocking items were right, and both trace back to the brief the lane was given rather than to how it executed it. Addressed at exact head `dedaa26e` (base `main` @ `316ef450`):

**Blocking 1 — parser rejected spec-valid manifests** → `88d8c787`. A value under any key other than `name`/`description` is now skipped whole (blank, indented and `- ` lines, at any depth), and a multi-line plain `name`/`description` folds like `>`. An indented line with no preceding key is still rejected. Tests use a realistic spec shape (`metadata` map, list-valued `allowed-tools`, `license`) and the exact `metadata:\n  author: x\n  requires:\n    - y` shape.
I re-ran your experiment rather than take the lane's word for it: **1313 real `SKILL.md` files on this machine → 1304 parse, 0 rejected for indentation.** The 9 rejections are all spec violations: 4× description over 1024 chars (e.g. `claude-api`, 1077), 3× capitalised names (`Zotero`, `Presentations`, `Spreadsheets`), 2× no frontmatter.

**Blocking 2 — Codex target** → `6803099d`. Recorded as a constraint: Codex materialization is Agent-scoped (`<workspace>/.agents/skills`, or a per-Agent `CODEX_HOME`); the OS account home's `.agents/skills` is never a target and never a fallback; if neither works, Codex is out of v1. The client lane's brief carried the account-home fallback you anticipated — it has been removed there too, before that lane reached the step.

**Non-blocking, all taken:**
- Push-then-sync → `a5bcbb5a`: after a successful push of a directory that is the Skill's own materialization target, the CLI writes the install marker so the directory becomes platform-managed; an unmarked name collision is skipped and reported; marked directories absent from the manifest are removed.
- Name rule → `6d31a769`: `^(?=.{1,64}$)[a-z0-9]+(-[a-z0-9]+)*$`; `pdf-` and `a--b` rejected. The server lane's `agent_skills` CHECK is being aligned (two POSIX checks, no lookahead in SQL).
- Parser strictness → `dedaa26e`: an inline ` #` comment on a plain `name`/`description` and a duplicate `name`/`description` key are each rejected with a specific reason.
- Block-scalar description trim → `29a12c75` (landed before your review posted).

Validation run on `dedaa26e`: `pnpm check` 0, `pnpm typecheck` 0 (8/8), `pnpm --filter @opentag/shared test` 0 (36 files / 455 tests), pre-push hooks green. Diff is still only `packages/shared/**` and `docs/**`.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-contract-1`._

liuchao-001 (approved):
Re-reviewed at `dedaa26e`. Both blocking items are resolved.

- **Parser.** I re-ran the same probe over the same 93 real `SKILL.md` files: 91 parse now, 0 fail on indentation (previously 41 failed). The 2 still rejected are genuine spec violations: descriptions over 1024 characters. My edge cases behave as intended: a `metadata` map, a list-valued `allowed-tools` and a multi-line plain description are accepted; `name: pdf # c`, a duplicate `name`, and `pdf-` are rejected with specific reasons. `skill.test.ts` passes locally (43/43).
- **Codex target.** The constraint is recorded the way I asked: Agent-scoped only, account home is never a target or fallback, otherwise out of v1.
- **Non-blocking items** (push-then-sync, name rule, comments/duplicates) are all addressed.

Nit, take it or leave it: in the push-then-sync section, the paragraph starting "Push also reports name collisions" describes what *sync* does with an unmarked directory, so "Sync skips and reports…" would read more accurately.

The remaining red checks are outside this review: `ownership-gate` is waiting on yuezengwu, and `Documentation Mirror Advisory` is advisory.

packages/shared/src/skill.ts
```diff
@@ -0,0 +1,487 @@
...
+ * A plain scalar, optionally continued on following indented lines. Continuations are folded the way
+ * `>` folds them, so a long description written over several lines keeps its paragraph breaks.
+ */
+function readPlainScalar(lines: string[], index: number, key: string, rawValue: string): ManifestFieldValue {
+  const collected = collectBlockLines(lines, index + 1);
+  const continuation = stripBlockIndent(collected.lines);
+  if (hasInlineComment(rawValue) || continuation.some(hasInlineComment)) {
+    return { ok: false, reason: `Skill manifest has an inline comment on the ${key} value` };
+  }
+  return { ok: true, value: foldLines([rawValue, ...continuation]), nextIndex: collected.nextIndex };
```
yuezengwu: [P2] Reject collection-valued manifest fields before folding them

`description:\n  - read\n  - write` currently returns `ok: true` with `description: "- read - write"`; a block mapping and `[read, write]` / `{purpose: read files}` are also accepted. These are YAML collections, not description strings (the [reference validator](https://github.com/agentskills/agentskills/blob/main/skills-ref/src/skills_ref/validator.py#L70-L76) rejects them). Folding before `SkillManifestSchema.safeParse` erases the type distinction, so this shared upload-validation contract accepts a bundle whose unchanged `SKILL.md` is not a valid Skill manifest. Please reject collection syntax for the fields the platform reads, while retaining multiline plain strings and skipping collections under unknown metadata keys; add negative cases for block and flow collections.

yuezengwu (changes requested):
Reviewed at `dedaa26ed16d19c91148d42855ccfac467d20a27`. One additional P2 contract issue is detailed inline: collection-valued `description` fields are accepted as strings. This is separate from the compatibility issues already addressed in the earlier review.

Validation on this head: all 455 shared tests, `pnpm check`, shared build/typecheck, and `git diff --check` passed. An independent probe reproduced all four block/flow collection cases being accepted; normal multiline text and ignored metadata controls passed. No Server/Client delivery behavior is implemented by this PR or claimed validated here.

packages/shared/src/skill.ts
liuchao-001: Confirmed. I reproduced all four cases on `dedaa26e`: a block sequence, flow `[…]`, flow `{…}` and a block mapping under `description` each return `ok: true` with the collection flattened into a string. My re-review missed this, because the probe only checked that collections under unknown keys are skipped, not that the fields we read reject them. Agree with the fix as scoped. Multi-line plain strings should keep working, and collections under unknown keys should still be skipped. I'll re-review the new head once the fix and the negative tests are in.

bestony:
Fixed at exact head `16e34d4d` — and thank you both: this was a regression introduced by round 1's fix. Folding multi-line plain scalars *before* `safeParse` erased the string/collection distinction, and my round-1 verification could not have caught it — a corpus of real `SKILL.md` files proves valid manifests are not rejected, never that invalid ones are not accepted.

**What changed (`16e34d4d`).** For the two fields the platform reads, a collection value is now rejected with `Skill manifest <key> must be a string, not a list or map`: flow `[…]` / `{…}`, a block sequence, and a block mapping under an empty first line. Unchanged: multi-line plain strings, quoted scalars, `>`/`|` block scalars, and collections under unknown keys (still skipped whole). The parser moved to `packages/shared/src/skill-manifest.ts` (375 lines; `skill.ts` was at 487 of the 500-line ratchet) and is re-exported from `skill.ts`, so `public-exports.snapshot.json` is byte-identical.

**Verification I ran myself on `16e34d4d`, independent of the lane's own tests:**

| Case | Result |
| --- | --- |
| `description:` + `- read` / `- write` (your case 1) | rejected |
| `description:` + `purpose: read files` (block map) | rejected |
| `description: [read, write]` | rejected |
| `description: {purpose: read files}` | rejected |
| `name:` block sequence, `name: [a]` | rejected |
| sequence at column 0 under `description:` | rejected |
| plain first line **then** a sequence (`description: Tools` + `- read`) | rejected |
| multi-line plain; empty first line + plain text | accepted |
| `\|` block whose text contains `- item` lines | accepted, text preserved |
| quoted `"[beta] does things"` | accepted |
| `metadata` map + nested list, `allowed-tools` list, `tags: [a, b]` under unknown keys | accepted, skipped |
| `A well-known - tool`, `Use when: the user asks` | accepted |

Corpus regression: the same 1313 real `SKILL.md` files → 1304 parse, identical to round 1 (the 9 rejections are still 4× description > 1024, 3× capitalised name, 2× no frontmatter), so the stricter rule rejects nothing real.

`pnpm check` 0 · `pnpm typecheck` 0 (8/8) · `pnpm --filter @opentag/shared test` 0 (36 files / 460 tests) · pre-push hooks green. Diff is still only `packages/shared/**` and `docs/**`.

One note on history: the extraction and the fix landed as a single commit; the lane should have split them. I have not asked it to rewrite a pushed branch.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-contract-1`._

liuchao-001 (approved):
Re-reviewed at `16e34d4d`. The collection-value P2 is fixed.

- **Negative cases:** all four collection cases I reproduced on `dedaa26e` (block sequence, flow `[…]`, flow `{…}`, block mapping under an empty first line) are now rejected with `description must be a string, not a list or map`.
- **Positive controls, each in a manifest that also has a `metadata` map and a list-valued `allowed-tools`:** multi-line plain text, an empty first line followed by plain text, a `|` block whose text starts with `- `, a quoted scalar, and a plain value containing `: ` all still parse to the expected string.
- **Real-file regression:** the same 93 real `SKILL.md` files give 91 parsed, the same result as the previous head. The 2 rejections are still descriptions over 1024 characters.
- **Tests:** the full `@opentag/shared` suite passes locally (36 files / 460 tests).
- **Extraction:** moving the parser into `skill-manifest.ts` and re-exporting it keeps the public surface unchanged.

No further findings from me.

liuchao-001 (approved):
Re-approving at `fd334d22`. The only change since `16e34d4d` is one line in `docs/zh-CN/README.md`, a link to `../design/agent-skills.md` matching the English index. No code changed, so my verification at `16e34d4d` carries over.

bestony:
New exact head `fd334d22` — a one-line docs change, no code touched since `16e34d4d`:

- `docs/zh-CN/README.md`: added `- [Agent Skills](../design/agent-skills.md)` under 设计文档（英文）. The **Documentation Mirror** check was failing because `docs/README.md` gained the index link while its mirror did not. That was my brief's mistake, not the lane's — it said "`docs/design/**` is exempt, do not create a zh-CN copy" and forgot the README index is itself mirrored. `node scripts/check-doc-mirrors.mjs --base origin/main --head HEAD` now passes (27 pairs); `pnpm check` 0; pre-push hooks green.

`git diff 16e34d4d..fd334d22` is exactly that one added line, so the parser verification posted above still describes this head. The red `ownership-gate` status reflects the outstanding change request on `dedaa26e`, not a CI failure.

@liuchao-001 sorry to void your approval of `16e34d4d` with this — requesting a fresh exact-head review from you and @yuezengwu.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-contract-1`._

bestony:
New exact head `403aff84` — one more contract fix, found by integration rather than review:

`ErrorCodeSchema` spread `MCP_ERROR_CODES` but not `SKILL_ERROR_CODES`, so `ErrorEnvelopeSchema` rejected every `SKILL_*` failure. The stacked server lane hit it first (it could not emit a Skill error through the shared envelope), and it was worse downstream: the web app and the client both `safeParse` `ErrorEnvelopeSchema`, so a `SKILL_NAME_CONFLICT` from the server would have been unreadable and the web "replace on conflict" flow dead end to end — invisible to each lane's own tests because they mock the error.

`403aff84`: `errors.ts` now spreads `...Object.values(SKILL_ERROR_CODES)` (no import cycle — `skill.ts` imports only `zod`), plus tests that every Skill code parses and that a full `SKILL_NAME_CONFLICT` envelope round-trips with its metadata category. `git diff fd334d22..403aff84` is 3 lines of source + 16 of test.

Validation on `403aff84`: `pnpm check` 0 · `pnpm typecheck` 0 · `pnpm --filter @opentag/shared test` 0 (36 files / 462 tests) · doc-mirror check passes · pre-push hooks green.

Apologies for the churn on exact-head reviews — this is the cost of building the consumers in parallel with the contract; the upside is that the contract is now exercised by a real server, client and web implementation (server lane: 148 unit files / 2263 tests and 40 integration files / 528 tests green against it). Requesting a fresh review of `403aff84`.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-contract-1`._

liuchao-001 (approved):
Re-approving at `403aff84`. Good catch from integration, and one my earlier reviews missed too.

- `ErrorCodeSchema` now spreads `SKILL_ERROR_CODES` the same way it spreads `MCP_ERROR_CODES`, so a `SKILL_*` envelope parses.
- No import cycle: `skill.ts` imports only `zod` and `./skill-manifest.js`, and `skill-manifest.ts` imports only `zod`.
- Checked locally at this head: the full `@opentag/shared` suite passes (36 files / 462 tests), including the new envelope round-trip, and repo-wide `pnpm typecheck` exits 0.
- No other file changed since `fd334d22`, so my earlier parser verification still applies.

packages/shared/src/skill-manifest.ts
```diff
@@ -0,0 +1,375 @@
...
+ * sequence (`- read`) or, after an empty first line, a block mapping (`purpose: read files`). Folding
+ * first would erase the type distinction and accept a `SKILL.md` the reference validator rejects.
+ */
+function readPlainScalar(lines: string[], index: number, key: string, rawValue: string): ManifestFieldValue {
+  const collected = collectBlockLines(lines, index + 1);
+  const continuation = stripBlockIndent(collected.lines);
+  const nextContent = firstNonBlankLine(lines, index + 1);
+  if (nextContent !== undefined && isSequenceLine(nextContent)) return collectionRejection(key);
+  const firstContent = firstNonBlankLine(continuation, 0);
+  if (rawValue.length === 0 && firstContent !== undefined && isMappingLine(firstContent)) {
```
yuezengwu: [P2] Check collection syntax on the first actual value line

The [earlier collection-type finding](https://github.com/first-tree-ai/opentag/pull/655#discussion_r4053215416) still reproduces when the value moves to the following line: `description:\n  [read, write]` and `description:\n  {purpose: read files}` both return `ok: true`. `description:\n  "purpose": read files` also passes because `isMappingLine` only recognizes unquoted keys. These are the same YAML list/map types as the cases already fixed, and both built public entrypoints still flatten them into description text. The flow check only examines `rawValue`, which is empty in these examples. Please validate the actual first value line before folding (including quoted mapping keys), while retaining the valid multiline, quoted and literal-block controls.

yuezengwu (changes requested):
Re-reviewed at `403aff8485adf2ff37dc95e9780fc0b40e989934`. The original four collection cases now reject correctly, and the Skill error envelopes work through both built public exports. The same P2 remains open for collection values starting on the following line and block mappings with quoted keys; the remaining cases are detailed inline.

Validation: 462 shared tests, `pnpm check`, repo-wide `pnpm typecheck` (8/8 tasks, including dependency builds), and `git diff --check` passed. Eleven independent parser cases were run through both built root/browser entrypoints: eight passed, three remaining invalid collections were accepted. A separate YAML parser confirmed those three inputs are arrays/maps. All nine Skill error envelopes parsed through both entrypoints.

packages/shared/src/skill-manifest.ts
liuchao-001: Confirmed at `403aff84`. All three cases reproduce. Two more variants of the same class also return `ok: true`: a single-quoted mapping key (`description:\n  'purpose': read files`) and an explicit-key mapping (`description:\n  ? purpose\n  : read`).

The fix has now missed variants twice, so a structural rule would hold up better than another enumeration. Run the collection checks on the **effective first value line**: `rawValue` if it is non-empty, otherwise the first non-blank continuation line. Then apply YAML's plain-scalar start rule to that line. The line must not start with `[` `{` `]` `}` `,` `#` `&` `*` `!` `|` `>` `'` `"` `%` `@` or a backtick. It must not start with `-`, `?` or `:` followed by whitespace. It must not match `<plain-or-quoted key>:` followed by whitespace or end of line. The quoted case needs care: `description: "quoted"` on the key line is a valid string and is already handled before this point, so the quote rule only applies to the continuation-line path. Negative tests should cover each branch of that rule, not just the five cases reported so far.

I missed this class in my own re-review too. I'm replacing my approval with a request for changes until the fix lands.

liuchao-001 (changes requested):
Superseding my approval of `403aff84`. yuezengwu's follow-up P2 reproduces: collection values that start on the line after `description:` still parse as strings. That covers flow lists and maps, mappings with quoted keys, and explicit `?` keys. Details and a suggested structural rule are in the [inline thread](https://github.com/first-tree-ai/opentag/pull/655#discussion_r4053302372). I'll re-review the head that fixes it.

bestony:
New exact head `0ab6b1c6`. You were both right that the fix had now missed variants twice — so this round replaces the enumeration with the structural rule, and I changed how it is verified.

**What changed**
- `644cab96` — **structural classification.** For `name`/`description` the parser takes the *effective first value line* (`rawValue`, else the first non-blank continuation line) and applies YAML's plain-scalar start rule: not `[ { ] } , # & * ! | > % @` or a backtick; not `-`/`?`/`:` followed by whitespace; not `<plain-or-quoted key>:` followed by whitespace or EOL. A plain continuation line may not contain `: ` either. Block scalars are exempt (a `|` block may contain `- item` / `key: v` lines).
- One deliberate refinement of the suggested rule: "a continuation line must not start with a quote" would reject valid YAML that exists in the wild — Anthropic's own `math-olympiad` skill is `description:\n  "Solve … 'IMO' …"` spanning five lines. So a value starting with a quote is accepted **only if the folded text is one complete quoted scalar**; `"purpose": read files`, trailing text after the closing quote, and an unterminated quote are rejected.
- `41220775` — **same class, scalar side** (added proactively): a plain value that YAML resolves to a non-string — null, boolean, number, date — is rejected (`name: 123`, `description: true`, `description: 2024-01-01`). The reference validator rejects these; we used to accept `"123"`.
- `0fc2a663` — a quoted scalar may span lines on the key-line path too (`description: "two\n  lines"` was being wrongly **rejected**).
- `0b268a68` — merge of `main` `22a85448`; the only conflict was `packages/shared/src/index.ts` (both sides added export lines; both kept).

**How it was verified — differential test against PyYAML, not more hand-picked cases.** 1466 cases = 153 generated (every value shape × key-line/continuation-line × both fields) + 1313 real `SKILL.md` files. The oracle accepts iff the frontmatter loads, `name` and `description` are `str`, and they meet the platform's own limits.

| | `403aff84` (previous head) | `0ab6b1c6` |
| --- | --- | --- |
| accepts what YAML says is not a valid string manifest | **82** | **0** |
| value differs from YAML's | 7 (incl. the real `math-olympiad` file, parsed with literal quote characters in its description) | 0 |
| rejects what YAML accepts | 3 | 10 — all generated, all deliberate (below) |
| real files disagreeing with PyYAML on accept / reject / value | 1 | **0 of 1313** |

Your five cases were 5 of those 82. The 10 remaining divergences are intentional strictness, none occurring in a real file: values starting with `&` (anchors — we do not implement them), and `0o17` / `1e3` / `y`, which PyYAML (YAML 1.1) reads as strings but YAML 1.2 parsers read as number/boolean. A manifest whose type depends on which parser reads it is not portable, so it is refused rather than guessed. Inline ` #` comments and duplicate `name`/`description` keys stay rejected as before.

`pnpm check` 0 · `pnpm typecheck` 0 · `pnpm --filter @opentag/shared test` 0 (37 files / 475 tests) · doc-mirror check passes · pre-push hooks green. `skill-manifest.ts` is 492 lines (ratchet is 500); the scalar resolver lives in its own `skill-manifest-scalar.ts`.

Requesting a fresh exact-head review of `0ab6b1c6`.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-contract-1`._

liuchao-001 (approved):
Re-reviewed at `0ab6b1c6`. The structural rule closes the class. The differential test against PyYAML is the right way to verify it, rather than more hand-picked cases.

Checked independently at this head:
- **Negative cases:** all 14 are rejected. That includes all 9 collection cases from both earlier rounds (block and flow, key line and next line, quoted and single-quoted keys, the explicit `?` key), plus `&anchor`, `*alias`, `!tag`, `true` and `42`.
- **Positive controls:** each is in a manifest that also has a `metadata` map and a list-valued `allowed-tools`. Multi-line plain text, an empty first line followed by text, a `|` block starting with `- `, a quoted value containing `: `, a quoted scalar spanning lines, `'it''s fine'`, a URL, `12:30`, `C# and F#` and `true story` all parse to the expected text.
- **My own mistake:** I had listed `Use (see: docs)` as a positive control. The parser now rejects it with the "quote the value" reason, and PyYAML raises a `ScannerError` on it too, so the rejection is correct.
- **Independent differential:** I ran my 93-file local corpus against PyYAML 6.0.3 as the oracle (the frontmatter loads, both fields are `str`, and the platform's limits hold). The parser agrees on **93/93** files, on accept/reject and on the exact description text.
- **Tests:** the full `@opentag/shared` suite passes locally (37 files / 475 tests).

The deliberate YAML 1.1 / 1.2 strictness (`0o17`, `1e3`, `y`, anchors) is a reasonable portability call. No further findings.

yuezengwu (approved):
Re-reviewed exact head `0ab6b1c6133475d0531714396ce425d8b3f22a77`. My previous collection-valued manifest findings are resolved: the parser now classifies the effective value before folding, rejects the reported flow/block and quoted-key mapping cases, and preserves the supported quoted/plain/block string forms. No remaining blocking finding in this contract PR.

Independently verified on this head: all 475 shared tests passed; 39 manifest regression/control cases passed through each of the root and browser exports, and all 9 Skill error codes passed both error envelopes. Shared build, repository typecheck (8/8 tasks), `pnpm check`, and `git diff --check` passed. The remaining ownership gate is awaiting this code-owner review; this approval is scoped to #655.

PR #656, opened 2026-09-19
feat(client): sync Agent Skills to the Computer and add the opentag skill CLI

bestony:
> **Stacked on #655** (`feat/skill-contract`). Base is that branch so this diff shows only its own work; it will be retargeted to `main` once #655 merges. Design of record: `docs/design/agent-skills.md` on the base branch.

## Summary

Computer-side half of **Agent Skills**: at provider runtime start the Computer syncs its Agent's Skills from the platform, and the `opentag skill` CLI lets a human or the Agent itself push a local skill directory up.

- **`packages/client/src/api.ts`** — skill methods on three surfaces: account token, machine token (`getComputerSkillManifest`, `openComputerSkillBundle`, streamed like `openImResource`), session proof (`listRuntimeSkills`, `pushRuntimeSkill`, `openRuntimeSkillBundle`).
- **`src/skills/skill-archive.ts`** — `packSkillDirectory` (limits, exclusions, local manifest validation) and `extractSkillArchive` (byte-limited; rejects traversal, absolute paths, links, devices; `0o600`/`0o700`).
- **`src/skills/skill-sync.ts`** — `SkillSyncManager.ensureAgent`: fetch manifest → compare each `.opentag-skill.json` marker → download, verify size + sha256, extract to staging, swap in. Marked directories absent from the manifest are removed. **Directories without a marker are never read, modified or removed** (skills the Agent is authoring, Context Tree skills); a manifest name that collides with one is skipped and reported. A locally edited managed copy that the platform also changed is quarantined under `.opentag/skill-conflicts/`.
- **Soft failure is the contract**: the whole sync is bounded (15 s) and any error — network, 404 from an older server, 503 storage unavailable, bad archive — logs `skill_sync_unavailable` and resolves with what is already on disk. Losing Skills never costs the Agent its turn.
- **Runtime wiring** — injected into `SessionRuntimeManager` as a narrow `Pick<SkillSyncManager, "ensureAgent">` next to Context Tree preparation; Pi receives the paths as `--skill` arguments.
- **`opentag skill push|list|pull|remove|enable|disable`** — Agent mode when `OPENTAG_SESSION_PROOF_FILE` is set (agent comes from the proof; `--agent` rejected; lifecycle commands refused), otherwise account mode with `--agent`. After a successful push of a directory that is the Skill's own materialization target, the CLI marks it managed (`markSkillDirectoryManaged`) — the push-then-sync rule from #655's review.
- Managed prompt tells the Agent how to save a routine as a Skill.

### Codex target — verified empirically, per #655 review
Materialization is Agent-scoped only: Claude Code `<workspace>/.claude/skills`, **Codex `<workspace>/.agents/skills`**, Pi explicit paths. The lane verified against the installed Codex CLI `0.153.4` with a temp `HOME`/`CODEX_HOME`: `codex debug prompt-input` lists a cwd-level `.agents/skills` skill (and not a `.claude/skills` one). So no fallback is needed and none exists — the OS account home's `.agents/skills`, shared by every Agent on the Computer, is never a target.

## Checklist

- [x] 1. `OpenTagApi` skill methods (account, machine, session-proof)
- [x] 2. `skill-archive.ts` pack + extract
- [x] 3. `SkillSyncManager` (Codex target verified and recorded)
- [x] 4. Runtime wiring in `session-runtime-manager.ts` + `client-runtime-composition.ts`
- [x] 5. Managed prompt paragraph
- [x] 6. `opentag skill` CLI commands
- [x] 7. Tests, including the 100% agent-runtime coverage gate
- [x] 8. All acceptance criteria verified, `git status` clean

## Acceptance criteria — re-run by the orchestrator on `ae506c8a`

| # | Criterion | Result |
| --- | --- | --- |
| 1 | `pnpm check` | exit 0 |
| 2 | `pnpm typecheck` | exit 0 |
| 3 | `pnpm --filter @opentag/client test` | **129 files / 1783 tests green with `PORT=18081`.** On the verification host the default run has 1 failure, `runner-bin-signals` → `listen EADDRINUSE 0.0.0.0:8080`: OrbStack holds 8080 there (confirmed with `lsof`). This branch touches no `runner/serve` file or test; CI is unaffected. |
| 4 | `test:agent-runtime:coverage` (100% stmts/branches/funcs/lines) | exit 0 |
| 5 | `pnpm --filter open-tag test` | exit 0, 45 files / 520 tests |
| 6 | `pnpm build` | exit 0 |
| 7 | A rejecting sync never fails runtime start; an unmanaged dir is byte-identical after sync | both asserted by tests |
| 8 | No file > 500 lines, no new dependency, lockfile unchanged | holds (largest new file 390) |
| 9 | Diff only under `packages/client/` and `apps/cli/` | holds |
| 10 | Codex finding recorded | holds (above) |

Deviation from the confirmed plan, recorded: the brief's original Codex step allowed an account-home fallback; it was revised mid-flight after #655's review and the lane implemented the revised rule.

Reviewer notes (non-blocking, from the orchestrator's read): `installBundle` is remove-then-rename rather than a single atomic swap — a crash in between leaves the Skill missing until the next sync reinstalls it; and the bundle download is buffered and size-checked after the fact rather than capped by `content-length`.

## Provenance

Written by an agent under `dispatch-opencode` run `tlm1q4`, lane `skill-client-cli-3`. Plan and acceptance criteria authored, and the result verified, by the orchestrator (Claude Code) before push. 13 commits for an 8-item checklist.


bestony:
New head `1dad7141` — the branch was DIRTY against its base and is merged up.

`feat/skill-contract` gained a merge of `main` `22a85448` (#654, Feishu activation), which conflicted here in two files, both under the 100% agent-runtime coverage gate. Resolved by keeping **both** sides — the changes are independent:
- `client-runtime-composition.ts` — import block: the `SkillSyncManager` import and main's `ensurePrivateDirectory, RuntimeStorageError` import.
- `session-runtime-manager.ts` — `#startRuntime`: the Skill sync boundary (`prepareAgentSkills`, spread as `…skills`) and main's `visibleReplyWritableRoots(…)` (spread into `writableRoots`).

The orchestrator checked mechanically that every line #654 added to those two files is still present after the merge (7/7 and 13/13) and that no conflict marker remains.

`1dad7141` also removes a workaround: before #655 `403aff84` the shared error envelope rejected `SKILL_*` codes, so the CLI detected an upload conflict by HTTP 409. It is back to an exact `SKILL_ERROR_CODES.NAME_CONFLICT` check, and `skill-api.test.ts` asserts `SKILL_NOT_FOUND` from a parsed envelope again.

Re-run by the orchestrator on `1dad7141`: `pnpm check` 0 · `pnpm typecheck` 0 · client tests 129 files / 1788 tests (with `PORT=18081`; OrbStack holds 8080 on the verification host) · **agent-runtime coverage gate still 100%** · CLI 45 files / 520 tests · `pnpm build` 0 · diff vs base still only `packages/client/` and `apps/cli/`.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-client-cli-3`._

liuchao-001 (commented):
Reviewed at head `1dad7141`: the diff against `feat/skill-contract`, 25 files. I read the sync, archive, CLI and runtime wiring code and probed two of the behaviours below with a throwaway vitest file (not committed).

The overall shape is good. Sync failures stay soft, the runtime-start guard is a real hard boundary, unmarked directories are never touched, and the Codex target is Agent-scoped as #655 required. I also checked the Pi wiring: Pi's docs say explicit `--skill` paths still load under `--no-skills`, so putting `#skillArgs` after `PI_RESOURCE_DISABLE_ARGUMENTS` is correct.

I'm leaving this as a comment, not an approval. The first two items below should be fixed before merge.

### Should fix

**1. Removing a managed Skill deletes local edits without quarantine** (`packages/client/src/skills/skill-sync.ts:352`)
The update path protects local edits: `#prepareReplacement` compares the content sidecar and moves an edited copy into `.opentag/skill-conflicts/`. The removal loop does no such check and runs `rm -rf` on every marked directory that is missing from the manifest.
I probed this: I adopted a directory with `markSkillDirectoryManaged`, added `notes.md`, and synced against an empty manifest. The result was `status=synced`, `notes.md` gone, and nothing quarantined.
The realistic case: an Agent pushes `.claude/skills/foo` (so it gets adopted), keeps editing it, and a human then disables or deletes the Skill in Web. The Agent's unpushed work is destroyed on the next runtime start. The design says "the platform is the source of truth", but the update path already treats a locally edited managed copy as worth keeping. Running the same edited-check before removal and quarantining keeps the two paths consistent.

**2. Account-mode `push` adopts directories under the human's shell cwd** (`apps/cli/src/core/skill/operations.ts:86-87`)
`isSkillMaterializationTarget(..., dependencies.cwd ?? process.cwd())` runs in both modes. A human in their own project who runs `opentag skill push .claude/skills/foo --agent X` therefore gets `.opentag-skill.json` and `.opentag-skill.content` written into their repo. That directory is not an Agent workspace, so no sync will ever manage it, and the files are likely to be committed. The design describes adoption as the Agent's own push, so I'd limit it to `authority.mode === "agent"`.
Agent mode relies on `process.cwd()` being the workspace root. If the Agent `cd`s into the skill directory first, adoption silently doesn't happen and you get exactly the shadowing this rule exists to prevent. Deriving the workspace some other way, or at least reporting "not adopted" in the CLI output, would make that visible.

### Worth fixing (non-blocking)

**3. The content sidecar is packed into re-pushed archives** (`packages/client/src/skills/skill-archive.ts:84`)
`EXCLUDED_NAMES` has `SKILL_MARKER_FILE` but not `SKILL_CONTENT_SIDECAR_FILE`. Probe result: re-packing an adopted directory gives `[".opentag-skill.content","SKILL.md"]`, and its sha256 differs from the clean pack of the same content. A stale local-bookkeeping file ends up in the platform's canonical archive, shows up in `skill pull` output, and makes identical content hash differently depending on where it was pushed from. Moving the constant into `@opentag/shared` next to the marker and excluding both would fix it.

**4. If the pushed Skill comes back `enabled: false`, adoption leads to deletion on the next sync** (`operations.ts:87`)
One example is `--replace` on a Skill a human had disabled, if the server keeps that state. The Agent's authored directory then gets removed at the next runtime start. Skipping adoption when `uploaded.enabled === false`, or saying so in the output, avoids that surprise. I couldn't check the server default because it isn't in this stack.

**5. Staging lives inside the discovered skill root** (`skill-sync.ts:237`)
`.opentag-skill-XXXX` is created inside `.claude/skills`/`.agents/skills`. Two consequences:
- If the process crashes before the marker is written, an unmarked directory containing a `SKILL.md` is left behind forever. Claude Code and Codex will discover it as a duplicate Skill, and sync will never clean it up.
- Concurrent session starts for the same Agent share the same cwd. One session's `listManagedDirectories` can see another session's marked staging directory as "not in manifest" and delete it mid-install. That failure is soft, but it's noisy.

A staging directory under `.opentag/` on the same filesystem avoids both.

**6. `skill pull` doesn't verify the bundle** (`operations.ts:145`)
It extracts without checking `skill.archiveBytes`/`archiveSha256`, while sync does check them. It's cheap to reuse the same check here.

The PR description already records the other two known gaps as accepted: remove-then-rename isn't atomic, and the bundle is buffered rather than capped by `content-length`.

packages/client/src/skills/skill-sync.ts
```diff
@@ -0,0 +1,379 @@
...
+
+async function downloadedBundle(
+  api: SkillSyncApi,
+  token: string,
+  agentId: string,
+  entry: RuntimeSkillManifest["skills"][number],
+  signal: AbortSignal,
+): Promise<Uint8Array> {
+  const response = await api.openComputerSkillBundle(token, agentId, entry.id, { signal });
+  const bytes = new Uint8Array(await response.arrayBuffer());
```
yuezengwu: [P1] Keep the sync deadline active while consuming the response body

With the real `OpenTagApi`, a server that sends 200 headers plus one bundle byte and then stalls leaves `response.arrayBuffer()` pending beyond the sync budget. `#fetchResponse` calls `prepared.cleanup()` as soon as headers arrive, removing the caller-abort listener and request timer. I reproduced this on this head with a local HTTP server: a 150 ms budget was still pending at 475 ms and returned `synced` only after I released the remaining body. `prepareAgentSkills` directly awaits this promise, so a stalled storage response can hold runtime startup until some unrelated cancellation/transport timeout instead of degrading after 15 s. Preserve cancellation through body consumption (and cancel the stream on expiry), with a regression test that stalls after headers.

packages/client/src/skills/skill-sync.ts
```diff
@@ -0,0 +1,379 @@
...
+  } catch {
+    return undefined;
+  }
+}
+
+async function listManagedDirectories(layout: SkillInstallLayout): Promise<Map<string, ManagedDirectory>> {
+  const managed = new Map<string, ManagedDirectory>();
+  let names: string[];
+  try {
+    names = await readdir(layout.root);
```
yuezengwu, line 145: [P2] Validate the workspace skill root before discovering managed directories

`readdir` and `readMarker` follow an existing symlinked root, but the directory safety check only runs later in `installBundle`. I installed a managed Skill in workspace B, symlinked workspace A/.agents/skills to B/.agents/skills, and synced A against an empty manifest: it returned `synced` and deleted B's Skill. The removal-only path never calls `ensurePrivateDirectory`, and the replacement path also touches the old copy before that guard. Validate the provider root and its parent path against the Agent workspace before listing, reading or reconciling it, rejecting symlinked roots rather than deleting through them. Cover both the skills-root and provider-parent symlink cases.

packages/client/src/skills/skill-sync.ts
```diff
@@ -0,0 +1,379 @@
...
+        this.#logger.warn(
+          { code: "skill_name_shadowed_locally", skill: entry.name },
+          "An unmanaged directory shadows this Skill name; leaving it untouched",
+        );
+        continue;
+      }
+      // Download and verify before touching the old copy: a corrupt or truncated bundle must
+      // never cost the Agent the Skill it already had.
+      const bytes = await downloadedBundle(this.#api, token, input.agentId, entry, signal);
+      if (current) await this.#prepareReplacement(input, entry.name, current);
```
yuezengwu: [P2] Finish staging the replacement before removing the usable copy

`#prepareReplacement` removes an unedited current directory before `installBundle` even starts extracting the replacement. A handled extraction or staging-write failure therefore leaves no old Skill for the soft-failure fallback. I reproduced this by installing a valid Skill, then returning an unextractable bundle whose size and SHA exactly matched the manifest: sync returned `unavailable`, `skillPaths` was empty, and the old SKILL.md was gone. The same ordering exposes valid downloads to disk/extraction failures. Extract and validate the complete staging directory first, then quarantine/swap the current directory, and retain or restore the current copy if installation fails.

yuezengwu (changes requested):
Reviewed exact head `1dad714187b3611d1ce37bc756b88bd12560cbd4`. Requesting changes for the three additional, independently reproduced sync failures inline. Existing review points are not repeated.

Validation on this head: 144 affected client tests and 8 Skill CLI tests passed; `pnpm check`, scoped shared/client/CLI builds, scoped typechecks, and `git diff --check` passed. The separate filesystem/real-local-HTTP probes reproduced all three failures. No author branch changes or external services were used. Full repository suites and live provider acceptance were not run.

bestony:
All six addressed at exact head `882191ff`, one commit each. I re-ran your two probes myself against the new head rather than rely on the lane's tests.

| # | Fix | Commit |
| --- | --- | --- |
| 1 | **Removal now quarantines local edits.** Update and removal share one edited-check helper, so they cannot drift apart again: an edited managed copy is moved to `.opentag/skill-conflicts/<name>-<ts>/`, an unedited one is removed. | `3f4bce55` |
| 2 | **Adoption only in Agent mode**, and the outcome is always reported (`adopted: true\|false` + reason, in text and `--json`). The materialization-target check no longer depends on `process.cwd()`: it matches the pushed path's own shape (`…/.claude/skills/<name>`, `…/.agents/skills/<name>`, `…/.opentag/skills/<name>` with `<name>` = manifest name), so `cd`-ing into the skill dir first still adopts. | `0b24ee3f` |
| 3 | **Marker and content sidecar are excluded from packed archives and ignored on extract.** The sidecar constant stays in `packages/client` (exported from `src/skills/`) rather than moving to shared, to avoid another contract head on #655. | `adb65be5` |
| 4 | **`enabled: false` is not adopted**, with the reason in the output (implemented in #2's adoption decision; this commit is its test). The server does keep `enabled` across a replace, so your scenario is real. | `77ca62eb` |
| 5 | **Staging moved to `<cwd>/.opentag/skill-staging/`** (same filesystem, `0o700`), with a sweep of stale staging dirs at the start of each sync — no more orphan `SKILL.md` dirs for providers to discover, and concurrent starts cannot delete each other's staging. | `71530540` |
| 6 | **`skill pull` verifies `archiveBytes` / `archiveSha256`** through the same exported check sync uses. | `882191ff` |

**Probes re-run by the orchestrator on `882191ff`:**
- #1, your exact sequence (adopt with `markSkillDirectoryManaged`, add `notes.md`, sync against an empty manifest): `status=synced`, original dir removed, `notes.md` **preserved** under `.opentag/skill-conflicts/foo-<ts>/`. Control: an unedited adopted dir is removed and no quarantine dir is created.
- #3: a clean dir and an adopted copy of the same content (the adopted one holding `.opentag-skill.content` + `.opentag-skill.json`) both pack to `fileCount=1` with the **same sha256**.

`pnpm check` 0 · `pnpm typecheck` 0 · client tests 129 files / 1793 tests (`PORT=18081`) · **agent-runtime coverage gate still 100%** · CLI 45 files / 524 tests · `pnpm build` 0 · diff still only `packages/client/` and `apps/cli/`.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-client-cli-3`._

liuchao-001 (commented):
Re-reviewed at exact head `882191ff`. I read the fix diff since `1dad7141` and re-ran both of my probes on this head with a throwaway vitest file (not committed).

**All six items from my earlier review are resolved:**
1. **Removing an edited Skill now quarantines it.** I probed this: a disabled Skill with an unpushed `notes.md` gives `status=synced`, the original directory removed, and `notes.md` preserved under `.opentag/skill-conflicts/demo-<ts>/`. Update and removal now share `#retireManaged`, which closes this properly.
2. **Adoption is now Agent-only**, and the result is always reported.
3. **The marker and sidecar are excluded from packing and ignored on extract.** Probe: an adopted directory re-packs to `fileCount=1` with the same sha256 as a clean pack.
4. **A disabled upload is not adopted.**
5. **Staging now lives in `.opentag/skill-staging/`**, with a sweep of stale entries.
6. **`pull` now runs the shared `verifySkillBundle` check.**

**Non-blocking follow-ups** (not a reason to hold the PR):
- **Adoption can still write markers into a nested repo.** `resolveMaterializationWorkspace` matches the path shape anywhere. If an Agent pushes `<workspace>/<some-checkout>/.claude/skills/foo`, the directory is adopted: markers get written into that checkout, and the checkout is not the workspace that sync manages. It's a narrower form of my old #2. One option is to adopt only when the resolved workspace is the Session's workspace, for example when it is an ancestor of the proof file's workspace or of the runtime cwd.
- **The staging sweep can race a concurrent install.** It treats anything whose mtime is older than `budgetMs` as stale. Install and extract are not bound by the sync signal, so a slow extraction in a concurrent session can outlive 15 s and have its staging directory removed from under it. The failure is soft. A larger margin, or a pid/lease marker, would avoid it.

**Not approving yet:** yuezengwu's three findings are still open at this head:
- **P1:** The body consumed after the headers arrive is not bound by the sync deadline (`downloadedBundle` is unchanged).
- **P2:** The skills root is not validated before `listManagedDirectories` or removal, so a symlinked root can still delete through the link.
- **P2:** `#retireManaged` still runs before `installBundle` has staged anything, so a failed install leaves no copy.

I'll re-review once those land.

bestony:
@yuezengwu all three fixed at exact head `8dd19787` — and apologies: your review of `1dad7141` landed while I was verifying the previous round, and I pushed `882191ff` without re-checking this PR for new reviews first.

All three were right, and the P1 is the one I should have caught: my own read filed "the bundle is buffered rather than capped by content-length" as a non-blocking note and missed that the deadline itself stopped applying. Each was reproduced as a failing test before the fix.

| | Fix | Commit |
| --- | --- | --- |
| **P1** deadline ends at headers | Bundles are no longer read with `arrayBuffer()`. `readBundleBody(response, { signal, maxBytes })` reads the stream in a loop, races every `read()` against the abort signal, caps at the manifest's `archiveBytes`, and `cancel()`s the reader on expiry or overrun. Used by sync and by `skill pull` (which gets its own deadline). `OpenTagApi.#fetchResponse`'s cleanup is unchanged for other callers. | `eec0e6fc` |
| **P2** symlinked root | `assertSafeSkillRoot` runs **before** anything is listed, read or reconciled: every path segment from the workspace down to the provider root is `lstat`ed (absent or real directory only), and `realpath(root)` must sit inside `realpath(cwd)`. Same check for the staging and conflicts roots. A refusal is soft: `skill_root_unsafe`, nothing touched, `status: "unavailable"`. | `fa662899` |
| **P2** copy removed before staging | Order is now download+verify → extract and validate the whole staging dir → retire the current copy (quarantine if edited, otherwise move **aside**) → `rename` in → delete the aside copy. A failed `rename` moves the aside copy back. | `4b62be56` |

@liuchao-001's two non-blocking points are in too: adoption requires the resolved workspace to carry `.opentag/skill-staging` (the sentinel sync creates at every runtime start), so a nested checkout is no longer marked (`6e2156ae`); and the staging sweep uses a 10-minute threshold unrelated to the sync budget (`ac5ec43d`).

**Your three probes, re-run by me against both heads** — real `OpenTagApi` + a local `node:http` server for P1, real filesystems for the P2s — so the probe is known to detect each bug:

| Probe | `882191ff` | `8dd19787` |
| --- | --- | --- |
| 200 + one byte, then stall; budget 150 ms | **still pending at 3000 ms**, socket open | `unavailable` after 154 ms, server saw the socket close |
| `A/.agents/skills` → `B/.agents/skills`, sync A against an empty manifest | `synced`, **B's Skill deleted** | `unavailable` (`skill_root_unsafe`), B's Skill intact |
| `A/.agents` → `B/.agents` (provider-parent symlink) | `synced`, **B's Skill deleted** | `unavailable`, B's Skill intact |
| installed Skill, then a bundle whose size+sha match but which does not extract | `skillPaths=0`, **old SKILL.md gone** | `skillPaths=1`, old copy intact, no staging leftovers |

`pnpm check` 0 · `pnpm typecheck` 0 · client tests 130 files / 1799 tests (`PORT=18081`) · **agent-runtime coverage gate still 100%** · CLI 45 files / 526 tests · `pnpm build` 0 · no new file over 500 lines (the sync module is split into `skill-sync`, `skill-install`, `skill-roots`, `skill-bundle-body`) · diff still only `packages/client/` and `apps/cli/`.

Requesting a fresh exact-head review from you both.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-client-cli-3`._

liuchao-001 (approved):
Approving at exact head `8dd19787`.

I read the fix diff since `882191ff` (`skill-bundle-body`, `skill-roots`, `skill-install`, and the reworked `skill-sync`/CLI adoption) and ran the five Skill-related client test files: 71 tests pass. I also ran my own probe for yuezengwu's symlink finding, using a throwaway vitest file (not committed):
- `A/.agents/skills` → `B/.agents/skills`: sync returns `status=unavailable` and B's Skill is intact.
- `A/.agents` → `B/.agents`: sync returns `status=unavailable` and B's Skill is intact.

**yuezengwu's three findings, as I read them at this head:**
- **Deadline covers the body.** `downloadedBundle` now reads through `readBundleBody`, which races every `read()` against the sync signal, caps the body at `archiveBytes`, and cancels the reader on abort. The P1 is closed.
- **Roots are checked first.** `unsafeSkillRootReason` checks the provider root, the staging root, and the conflicts root before anything is listed or removed, and failing that check is soft.
- **Staging before replacing.** The new bundle is extracted, validated (root `SKILL.md`), marked, and digested before the current copy is touched. The unedited copy is moved aside and restored if the rename fails.

My two earlier non-blocking points are also addressed: adoption requires the workspace to contain the staging-root sentinel, and the staging sweep now uses a 10-minute threshold.

**One tiny follow-up (non-blocking):** `moveAside` renames the current copy into `skill-staging/retired-<uuid>`. `rename` does not update the moved directory's own mtime, so a Skill installed more than 10 minutes ago immediately looks stale to a concurrent session's `sweepStaleStaging`. That session could delete the aside copy during the swap window. Only the restore-on-rename-failure path is affected, and the fallback is soft: the next sync reinstalls the Skill. `utimes` on the aside copy, or skipping `retired-*` in the sweep until it is older than the threshold by ctime, would close it.

CI was still pending when I reviewed this head, and `ownership-gate` still needs yuezengwu.

yuezengwu (commented):
Re-reviewed exact head `8dd197872b081a1e25ed05f53a4d98d20f63e1ff`. All three findings from my previous review are now resolved in independent real-filesystem / local-HTTP regressions: a stalled bundle returns `unavailable` before the held body is released; a symlinked root leaves the other workspace intact; and an unextractable hash-matching replacement retains the old Skill and its returned path. The existing rename-rollback regression also passes.

155 affected client tests and 14 Skill CLI tests passed, along with scoped shared/client/CLI builds, scoped typechecks, `pnpm check`, and `git diff --check`.

Approval is waiting on this head's failing [Node 26 Compatibility job](https://github.com/first-tree-ai/opentag/actions/runs/35451628295/job/105919680353). Its actual failure is `provider-cli-turn-runner.test.ts:427`, `forwards SIGTERM and SIGINT to the target`: stderr contains only `ready\n`, not `got SIGINT`. The fixture `provider-cli-turn-target.mjs` writes `ready` at line 7 before installing its signal handlers at lines 12-13, exposing a readiness race consistent with this failure. Please register the handlers before announcing readiness and restore current-head CI. These test/fixture files are unchanged by the Skill diff; I have not treated the failure as a proven harmless flake or rerun the remote workflow.

yuezengwu (approved):
Approving exact head `8dd197872b081a1e25ed05f53a4d98d20f63e1ff`. The three fixes and independent regression results from my previous review remain applicable to this unchanged head. I have now verified that [CI attempt 2](https://github.com/first-tree-ai/opentag/actions/runs/35451628295/attempts/2) passed, including Node 26 and the aggregate CI gate; the other technical checks also pass. The remaining ownership gate is waiting for this review.

The successful rerun clears the observed CI blocker; it does not establish that the signal-handler readiness race was fixed. That unchanged test-fixture hardening remains a follow-up. This approval covers #656 only.

packages/client/src/skills/skill-archive.ts
```diff
@@ -0,0 +1,403 @@
...
+  const tap = new PassThrough();
+  tap.on("data", (chunk: Buffer) => {
+    state.bytes += chunk.length;
+    if (state.bytes > SKILL_UNPACKED_MAX_BYTES) {
+      tap.destroy(new SkillArchiveError("unpacked_too_large", "Archive expands beyond the unpacked ceiling"));
+    }
+  });
+  await mkdir(dirname(dest), { recursive: true, mode: 0o700 });
+  await pipeline(stream, tap, createWriteStream(dest, { flags: "wx", mode: 0o600 }));
+  await chmod(dest, 0o600);
```
yuezengwu: [P2] Preserve owner execute permission for executable Skill files. Both the writer and this chmod unconditionally use 0600, so an uploaded executable script or binary becomes non-executable after sync/pull. Independent reproduction at this exact head: pack a Skill containing `scripts/run.sh` at 0755, then extract it; direct execution succeeds before packing and fails with EACCES afterward (mode 0600). Keep the private-file policy while carrying the normalized owner execute bit, e.g. 0700 for executable members and 0600 otherwise. The existing "modes intact" test asserts 0600 and misses this behavior.

yuezengwu (changes requested):
Superseding my approval of this unchanged head with one additional, independently reproduced finding from the subsequent archive round-trip review. The original three fixes remain verified and CI attempt 2 passed; neither covers executable-file preservation.

I reproduced this using only this PR's `packSkillDirectory` and `extractSkillArchive` on a real temporary filesystem: a 0755 shell script runs successfully before the round trip, then becomes 0600 and returns EACCES afterward. Please preserve owner-executable permissions and add an execution-based round-trip regression. No source branch changes were made.

liuchao-001 (changes requested):
I'm withdrawing my approval of `8dd19787` over yuezengwu's executable-bit finding. I missed it; the finding is correct.

At this head:
- `packSkillDirectory` records `0o755` for owner-executable files (`entryMode`, `skill-archive.ts:103-105`).
- `extractSkillArchive` then ignores `header.mode`: it creates every file with `0o600` and `chmod`s it to `0o600` again (`skill-archive.ts:360-361`).

As a result, every `scripts/*.sh` or bundled binary in a synced or pulled Skill loses its execute bit. That is a common Skill layout, so this is a real functional break and not an edge case.

I agree with the suggested fix: carry the owner execute bit through, using `0o700` when `header.mode & 0o100` is set and `0o600` otherwise, and add a round-trip test that actually executes the file. Please treat this review as not approving this head; I'll re-review the fix commit.

bestony:
Fixed at exact head `d13a33e1` (`d13a33e1`). You are both right, and this one traces to my side: the lane's brief did ask for it ("files 0o600, preserve an exec bit as 0o700"), the lane's "modes intact" test pinned `0o600` for everything, and my verification accepted that test instead of executing anything.

**Fix.** `extractSkillArchive` carries only the owner execute bit: `(header.mode & 0o100) ? 0o700 : 0o600`, both when creating the file and in the final `chmod`. Group/other bits are never restored and setuid/setgid/sticky stay rejected, so the private-file policy holds.

**Verified by executing the file, against both heads:**

| | `8dd19787` | `d13a33e1` |
| --- | --- | --- |
| `scripts/run.sh` (0755) before pack | runs | runs |
| after pack → extract | **`EACCES`**, mode 600 | runs, mode **700** |
| a `0777` source script | 600 | **700** (no group/other) |
| a plain `0644` file | 600 | 600 |

I also checked the side effect I was worried about: a Skill with an executable script, synced through `SkillSyncManager` — the materialized script runs, and a **second sync with an unchanged manifest** downloads nothing, reports no local edit and creates no quarantine directory (so normalizing the mode does not make the content digest drift).

The pinned-`0o600` test is replaced by an execution-based round trip (`execFileSync`, skipped on win32) plus the mode matrix above and a sync-level runnable-script test.

`pnpm check` 0 · `pnpm typecheck` 0 · client tests 130 files / 1803 tests (`PORT=18081`) · agent-runtime coverage gate still 100% · CLI 45 files / 526 tests · `pnpm build` 0.

The server half of the same problem (ZIP import assigning every member `0644`) is being fixed on #657, along with a symlink check the ZIP central directory makes possible.

Requesting a fresh exact-head review from you both.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-client-cli-3`._

liuchao-001 (approved):
Approving at exact head `d13a33e1`. This replaces my changes-requested review of `8dd19787`.

The only change since `8dd19787` is `d13a33e1`. `extractFileMode` sets `0o700` when the header has the owner-execute bit (`0o100`) and `0o600` otherwise. It is applied both when the file is created and in the final `chmod`, so umask cannot interfere. Group and other bits, plus setuid/setgid/sticky, are never restored. The file mode is not part of the content digest (`hashSkillDirectory` hashes only paths and bytes), so this can't cause a false local-edit quarantine.

I ran my own execution probe on this head with a throwaway vitest file (not committed): pack a Skill whose `scripts/run.sh` is `0755`, then extract it. The result is `run.sh=700`, `SKILL.md=600`, and the script executes and prints `ran`. `skill-archive.test.ts` and `skill-sync.test.ts` also pass (32 tests with the probe).

CI was still running when I reviewed this head; `ownership-gate` is the only failing check and is waiting on yuezengwu.

yuezengwu (approved):
Revalidated d13a33e1: all four reported regressions are resolved, including executable archive roundtrip (0700; direct execution succeeds). The affected 159 client tests and 14 CLI tests, scoped build/typecheck and repository checks pass. Current technical CI is green. No remaining blocking finding in the reviewed scope.

yuezengwu (commented):
After #655 was squash-merged, this PR was retargeted to main at e328c238. I independently confirmed the current d13a33e1 head cannot merge cleanly: add/add conflicts in docs/design/agent-skills.md, packages/shared/src/__tests__/skill.test.ts and packages/shared/src/skill-manifest.ts. Please update this branch against main, preserve the contract fixes that landed there, resolve the conflicts and run current-head CI. The earlier approval validates the previous reviewed head/base combination; it does not establish a clean, tested integration with the new main. I will re-review the resulting head.

bestony:
@yuezengwu done at exact head `65dbe1e3` — a merge of `main` and nothing else; no `packages/client` or `apps/cli` source changed.

`65dbe1e3` merges `origin/main` (`e328c238`, the squash of #655). The three add/add conflicts you listed were the only ones, all in contract-owned files this PR never owned, and each was resolved by taking **main's** version — so the contract fixes that landed there are preserved exactly:
- `git diff origin/main -- packages/shared docs/design` is **empty** on this head (byte-identical to main);
- `git diff --name-only origin/main...HEAD` lists only `packages/client/` and `apps/cli/`;
- no conflict markers; a dry-run merge of this head into `main` reports 0 conflicts, and GitHub should now show it mergeable.

Integration with the new `main`, re-run by me on `65dbe1e3`: `pnpm check` 0 · `pnpm typecheck` 0 · client tests 130 files / 1803 tests (`PORT=18081` on my host) · agent-runtime coverage gate still 100% · CLI 526 tests · `pnpm build` 0. Current-head CI is running on this push.

Requesting a fresh exact-head review from you and @liuchao-001.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-client-cli-3`._

yuezengwu (approved):
Re-reviewed exact head 65dbe1e3 against main e328c238. All three previously reported merge conflicts are resolved: shared contract files and docs/design match main exactly, client/CLI sources are unchanged by the merge, and an independent dry-run merge produces the current head tree.

Fresh validation after rebuilding shared/client/CLI: 475 shared tests, 159 affected client tests, 14 CLI tests, build, typecheck, pnpm check and diff checks passed. Replaying the original probes confirms the stalled-body deadline, symlink-root protection, retention of the old copy on failed replacement, and owner-only executable installation still work on this head. Current-head GitHub CI is green. No remaining blocking finding. Full repository suites and live provider acceptance were not run by this review.

PR #657, opened 2026-09-19
feat(server): store, validate and serve Agent Skills over S3-compatible storage

bestony:
> **Stacked on #655** (`feat/skill-contract`). Base is that branch so this diff shows only its own work; it will be retargeted to `main` once #655 merges. Design of record: `docs/design/agent-skills.md` on the base branch. The web lane stacks on this one (its journey e2e needs these routes).

## Summary

Server side of **Agent Skills**: PostgreSQL metadata, S3-compatible object storage behind an interface, upload validation with deterministic re-pack, a `SkillService` with ownership enforcement, and three HTTP surfaces. Storage is optional at deploy time.

- **Schema + migration `0047_chemical_hawkeye`** — `agent_skills`: one row per (Agent, name), unique on `(agent_id, lower(name))`, no `account_id` (ownership resolves through `agents.created_by_user_id`, as `mcp_server_authorizations` does). DB-level CHECKs for the name (`^[a-z0-9]+(-[a-z0-9]+)*$` + length ≤ 64, matching the contract after #655's review), description length, sha256 shape, archive bytes, file count, revision, object-key length. `CURRENT_MIGRATION_COUNT` 47 → 48.
- **Object store** — `SkillObjectStore` (put/get/head/delete) + `S3SkillObjectStore` on `aws4fetch` (SigV4 over `fetch`, zero transitive deps, path-style or virtual-hosted, per-request timeout, status → typed error mapping) + in-memory fake. Keys are derived server-side, never accepted from a caller: `<prefix>/accounts/<account>/agents/<agent>/skills/<skill>/<sha256>.tar.gz`. Keying by sha makes a replace "write new → update row → best-effort delete old", so a reader never sees a torn object.
- **Upload validation** (`skill-archive*.ts`) — `tar.gz` and `zip` in, canonical `tar.gz` out. Rejects absolute paths, `..`, backslashes, links, devices, setuid/setgid, duplicates; caps entries and **declared** unpacked size before inflating (zip bomb); strips a single shared top-level directory; ignores `__MACOSX`/`.DS_Store`; requires root `SKILL.md`; rejects reserved names. Re-pack is deterministic (sorted, mtime 0, uid/gid 0, normalized modes) — the stored sha256 is the server's.
- **`SkillService`** — account surface guarded by `createdByUserId` + not-deleted (not-found and not-owned indistinguishable); Computer surface requires `agents.computer_id` = the authenticated Computer and lists enabled Skills only; Agent CLI surface takes the agent id from the session proof, never from the request. Object is put before the row is written and removed if the row write fails. Without a store: listing answers `storage: "unavailable"`, bundle reads/writes fail `SKILL_STORAGE_UNAVAILABLE`, enable/remove still work.
- **Routes** — `api/skills.ts` (user auth, CSRF still applies to cookie uploads), `api/computer-skills.ts` (machine token), `api/runtime-skills.ts` (session proof), shared `api/skill-upload.ts` (octet-stream pass-through, auth before the parser, header preconditions before any body byte, byte-counted read, transfer deadline — modelled on `api/runner-workspace.ts`). **Routes are always registered**, so a deployment without storage never produces a failed request.
- **Config** — `OPENTAG_SKILL_STORAGE_*` group, all-or-nothing, off by default; secret never logged. `docker-compose.yml` gains MinIO + a one-shot bucket init; `docs/deploying.md` and its zh-CN mirror document the group.
- New dependencies: `aws4fetch`, `fflate`, `tar-stream` (third-party notices regenerated).

## Checklist

- [x] 1. Dependencies added; lockfile and third-party notices regenerated
- [x] 2. `agent_skills` schema, migration `0047`, hashes, `CURRENT_MIGRATION_COUNT` = 48
- [x] 3. `SkillObjectStore` interface, S3 adapter, in-memory fake
- [x] 4. Upload validation and deterministic re-pack
- [x] 5. `SkillService` with ownership guards and all three surfaces
- [x] 6. Config env group + `.env.example`
- [x] 7. Routes wired in `app.ts` / `index.ts`
- [x] 8. `docker-compose.yml` MinIO + deploy docs (en + zh-CN)
- [x] 9. Unit and integration tests
- [x] 10. All acceptance criteria verified, `git status` clean
- [x] 11. Follow-up: contract gap resolved upstream, envelope workaround removed

## Acceptance criteria — re-run by the orchestrator on `fe6ee0e3`

| # | Criterion | Result |
| --- | --- | --- |
| 1 | `pnpm check` (notices, workspace contract, migration drift, complexity ratchet) | exit 0 |
| 2 | `pnpm typecheck` | exit 0 |
| 3 | `pnpm --filter @opentag/server test` | exit 0 — 148 files / 2264 tests |
| 4 | `pnpm --filter @opentag/server test:integration` (testcontainers PostgreSQL) | exit 0 — 40 files / 528 tests |
| 5 | `CURRENT_MIGRATION_COUNT` = journal length; exactly one new migration | 48 = 48; one `.sql` |
| 6 | `pnpm build` | exit 0 |
| 7 | No new file > 500 lines | largest 454 (`skill-service.ts`) |
| 8 | Diff only in owned paths | holds |
| 9 | Secret key never reaches a log call | 0 hits; adapter test asserts it |
| 10 | `docker compose config` parses; zh-CN deploy doc mirrors the new section | holds; mirror check passes |

Orchestrator's own read of the security-relevant code: every skill lookup is scoped by `agentId` after the agent guard; the runtime routes never read an agent id from params or body.

## Deviations and findings, recorded

- **Migration regenerated mid-flight.** The first `0047` carried the pre-review name CHECK; because it was unpublished it was regenerated rather than followed by an `0048`.
- **Contract gap found here, fixed in #655.** `ErrorCodeSchema` did not admit `SKILL_*` codes, so `ErrorEnvelopeSchema` rejected every Skill failure (and the web app and client, which `safeParse` it, could not have read one). This lane worked around it locally and reported it; #655 `403aff84` fixed the contract; `fe6ee0e3` removed the workaround and added route assertions that a Skill failure parses through the shared schema.
- **Zip symlinks** are not detectable through `fflate`'s public API. Such an entry is read as a regular file whose content is the link text and re-packed as a regular file, so nothing link-like is ever stored or delivered; the tar path rejects links outright.
- Upload source is derived from the transport: bearer → `cli_upload`, cookie → `web_upload`, session proof → `agent_upload`.
- Non-blocking note: a gzip stream that inflates hugely without yielding tar members is bounded in memory but costs CPU proportional to the 16 MiB input cap.

## Provenance

Written by an agent under `dispatch-opencode` run `tlm1q4`, lane `skill-server-2`. Plan and acceptance criteria authored, and the result verified, by the orchestrator (Claude Code) before push. 17 commits for an 11-item checklist.


packages/server/src/services/skills/skill-archive-reader.ts
```diff
@@ -0,0 +1,185 @@
...
+    await streamed.catch(() => undefined);
+    throw mapArchiveReadError(error);
+  }
+}
+
+function readZipEntries(bytes: Uint8Array): RawSkillEntry[] {
+  const state = { seen: new Set<string>(), declaredBytes: 0, count: 0 };
+  let unzipped: Unzipped;
+  try {
+    unzipped = unzipSync(bytes, {
```
liuchao-001, line 303: P1: the bound here uses only `originalSize`. Stored entries are copied at `size`, and central entries can all point at the same local data, so a 1 MiB zip unpacks to 64 MiB with 0 bytes declared (reproduced). Reject stored `size !== originalSize`, sum `info.size` against `bytes.byteLength`, and re-check the real totals after unzip.

packages/server/src/services/skills/skill-service.ts
```diff
@@ -0,0 +1,454 @@
...
+      prefix: this.#keyPrefix,
+      accountId,
+      agentId: existing.agentId,
+      skillId: existing.id,
+      sha256: normalized.sha256,
+    });
+    await this.#putObject(store, objectKey, normalized.archive, normalized.sha256);
+    let row: SkillRow | undefined;
+    try {
+      [row] = await this.#database
```
liuchao-001, line 324: P2: this update doesn't check that the row is still the one we read. A concurrent replace can re-point the row to a key that the other writer then deletes as `existing.objectKey`. Suggest `and(eq(agentSkills.id, existing.id), eq(agentSkills.revision, existing.revision))`, and treat 0 rows as a conflict.

packages/server/src/config.ts
```diff
@@ -369,6 +369,18 @@ const ServerEnvironmentSchema = z
...
+    /*
+     * Optional S3-compatible object storage for Agent Skill bundles. The five material values are
+     * configured together or not at all; without them, Skill listing still works and every bundle
+     * read or write fails with SKILL_STORAGE_UNAVAILABLE. The bucket is expected to stay private.
+     */
+    OPENTAG_SKILL_STORAGE_ENDPOINT: z.string().trim().optional(),
+    OPENTAG_SKILL_STORAGE_REGION: z.string().trim().min(1).optional(),
+    OPENTAG_SKILL_STORAGE_BUCKET: z.string().trim().min(1).optional(),
+    OPENTAG_SKILL_STORAGE_ACCESS_KEY_ID: z.string().min(1).optional(),
+    OPENTAG_SKILL_STORAGE_SECRET_ACCESS_KEY: z.string().min(1).optional(),
```
liuchao-001, line 381: P2: also add this to `collectKnownSecrets` in `cloud-runtime-composition.ts` so startup-error redaction covers it, as it does for every other secret.

liuchao-001 (changes requested):
Reviewed at `fe6ee0e3`. Covered: the service, archive reader/repacker, upload transport, three route surfaces, S3 adapter, config, migration, and wiring. Ownership scoping holds everywhere: every Skill lookup is filtered by `agentId` after the agent guard, runtime routes take the agent id only from the proof, and cookie uploads still pass the origin/double-submit check because auth runs in `onRequest`. Deterministic repack and the tar path look solid. One blocker and two smaller items are below.

**P1: a zip with overlapping stored entries bypasses the unpacked cap and can OOM the server.** The zip bound only sums `info.originalSize` from the central directory. For a **stored** entry (`compression === 0`), fflate copies `size` (compressed) bytes and ignores `originalSize`. Central-directory entries can also all point at the same local file data. Reproduced with fflate on this branch: a 1 MiB zip whose 64 central entries declare `originalSize: 0` gives `declaredUnpacked: 0`, `actualUnpacked: 67108864`. At the real limits (16 MiB input, 1000 entries) that is about 16 GiB allocated inside `unzipSync`, and the checks after it never run. Any account, or an agent through `/runtime/skills`, can trigger it. Suggested fix inside `filter`:
- reject `compression === 0 && size !== originalSize`
- keep a running sum of `info.size` and reject once it exceeds `bytes.byteLength` (overlapping entries). Deflate entries are already capped at `originalSize` by fflate, so the sum check is what closes the overlap for both kinds.
- after unzip, re-check the real total body bytes against `SKILL_UNPACKED_MAX_BYTES` as a backstop.
A fixture for this belongs in `skill-archive.test.ts`.

**P2: the storage secret is missing from startup-error redaction.** `collectKnownSecrets` (`cloud-runtime-composition.ts`) lists every other secret env var, including the master key and the GCP token, but not `OPENTAG_SKILL_STORAGE_SECRET_ACCESS_KEY`. The adapter never logs it, but a startup failure that echoes config or env would print it. It's a one-line addition.

**P2: a replace built on a stale row can leave that row pointing at a deleted object.** `#replaceSkill` updates `where id = existing.id` with no check that the row still matches, then deletes `existing.objectKey`. The race: A and B both read revision 1 (key K0). A writes K1, updates the row, and deletes K0 last. Meanwhile B re-uploads the original content, so its key is K0 again; B's PUT and row update land before A's delete. The row now says K0, K0 is gone, and every bundle read returns `SKILL_NOT_FOUND "missing from storage"` until the next replace. Guard the update with `and(eq(id), eq(revision, existing.revision))`. On 0 rows, clean up the new object and return a conflict. The count-then-insert against `SKILL_MAX_PER_AGENT` is racy in the same way, but that one is harmless.

Non-blocking: `openBundleForComputer` doesn't require `enabled`, so a Computer can fetch a disabled Skill it isn't told about. It's the same owner, so this only matters if "disabled" is meant to be a hard stop.

bestony:
Addressed at exact head `378d2c23`. The P1 was real and I (the orchestrator) missed it — I read that zip path in my own pass and judged the declared-size bound sufficient. It was not.

**P1 — zip stored/overlapping entries bypass the unpacked cap → `4845d0f2`.** Inside fflate's `filter` (before an entry's bytes exist): reject a stored entry whose `size !== originalSize`; keep a running sum of `info.size` and reject once it exceeds the input length (overlapping members cannot all fit in the archive); cap each entry and the running `originalSize`; after `unzipSync`, re-check the real total as a backstop. The tar.gz path got the same "actual, not declared" treatment: the gunzip output is metered and aborted past the payload ceiling plus framing, so a gzip that inflates hugely without yielding members cannot burn unbounded CPU.
Verified independently of the lane's tests, with hand-built zips (one local entry, N central entries pointing at it) run against **both** readers so the probe is known to detect the bug:

| Case | pre-fix `fe6ee0e3` | `378d2c23` |
| --- | --- | --- |
| your attack: 64 stored entries, `originalSize: 0`, one 1 MiB blob | accepted, 64 MiB allocated | rejected (`inconsistent stored sizes`), 0 MiB growth |
| sizes consistent but overlapping (64 × 1 MiB) | accepted, 64 MiB | rejected (`members overlap`), 1 MiB |
| 1000 entries over a 1 MiB blob | **accepted, 1000 MiB allocated** | rejected, 0 MiB |
| one honest stored entry (control) | accepted | accepted |

**P2 — secret redaction → `57541c3b`.** `OPENTAG_SKILL_STORAGE_SECRET_ACCESS_KEY` is in `collectKnownSecrets`; its test is extended.

**P2 — stale-row replace → `f551cda7`, then `c8dc19b3`.** `replace`, `setEnabled` and `remove` are revision-guarded; 0 rows → 409 "changed concurrently", and the new object is discarded only if no current row references it. Reading that fix I found two more holes of the same class and had them closed before pushing:
- the failed-write `catch` deleted the new key unconditionally — on a same-content replace that *is* the live object. It now goes through the same "unreferenced only" check, and "cannot read the row" means "do not delete".
- `deleteReplacedObject` is check-then-act, so your interleaving survives in a narrower form (A's check sees its own key, B `PUT K0`, A `DELETE K0`, B commits K0). No ordering of A's check fixes that, so every writer now confirms its own object after commit — `head`, and re-`put` from the bytes still in memory if it is gone (`skill_object_restored` warning). Tests drive both interleavings deterministically through the fake store.
  Honest residual: if A's delete lands *after* B's post-commit `head`, the row still dangles. Closing that completely means not deleting inline at all (deferred orphan GC), which I would rather do as a follow-up than grow this PR — say if you want it here.

**Non-blocking — `openBundleForComputer` requires `enabled` → `cb375e71`.** Disabled is now a hard stop on the Computer surface (`SKILL_NOT_FOUND`).

**Also in this head — migration renumbered.** `main`'s #654 added `0047_feishu_pending_activation`, colliding with this branch's `0047`. Per the repo rule, this branch took main's `0047` untouched (sql, snapshot, journal, hashes) and regenerated its own as **`0048_known_gertrude_yorkes`**; `CURRENT_MIGRATION_COUNT` 48 → 49. The PR was DIRTY for that reason and is MERGEABLE again.

Re-run by the orchestrator on `378d2c23`: `pnpm check` 0 (migration drift: 49 migrations) · `pnpm typecheck` 0 · `pnpm build` 0 · server unit 150 files / 2362 tests · integration 41 files / 534 tests (testcontainers PostgreSQL). `skill-service.test.ts` is 582 lines — over the 500-line ratchet, which evidently does not apply to tests since `pnpm check` passes; happy to split it.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-server-2`._

packages/server/src/services/skills/skill-service.ts
```diff
@@ -0,0 +1,449 @@
...
+          fileCount: normalized.fileCount,
+          files: normalized.files,
+          filesTruncated: normalized.filesTruncated,
+          revision: 1,
+          createdAt: now,
+          updatedAt: now,
+        })
+        .returning();
+      if (!row) throw new Error("The Skill insert returned no row");
+      await ensureObjectPresent(this.#database, store, objectKey, row.id, normalized, this.#logger);
```
liuchao-001: If this throws after the insert has committed (for example a transient `head` 5xx or timeout), the `catch` below deletes `objectKey` and leaves the new row dangling. A brand-new id can't be touched by another writer, so either drop this call here or move it after the `try/catch`.

liuchao-001 (commented):
Re-reviewed at exact head `378d2c23`, covering the Skill commits since `fe6ee0e3`. The Feishu changes came in with the `main` merge and are out of scope here. I ran `skill-archive`, `skill-service` and `skills-route` locally: 41/41 pass. `CURRENT_MIGRATION_COUNT` is 49, the journal has 49 entries, and main's `0047` is untouched.

**Resolved:**
- **P1 zip.** Stored entries need `size === originalSize`, the sum of `info.size` is capped at the input length, and there is a backstop on the real bytes. Together these close both the stored-entry case and the overlap case. A deflate entry's output is already capped at its declared `originalSize` by fflate (`{ out: new u8(su) }`), so memory is bounded before the backstop runs. The gunzip meter is a good addition.
- **P2 redaction.** Fixed.
- **P2 stale replace.** The revision guard, the "discard only if unreferenced" cleanup, and the post-commit restore are correct for the interleavings I traced. I'm fine with the residual you describe (a delete landing after the other writer's `head`) going to a deferred orphan-GC follow-up.
- **Disabled Skill on the Computer surface.** Fixed.

**One new issue, from the new code. Please fix before approval: the insert path can now leave a committed row pointing at a deleted object.** In `#insertSkill`, `ensureObjectPresent` runs *inside* the same `try` as the insert. If its `head` (or restore `put`) fails after the row has committed, for example on a transient 5xx or timeout, the `catch` deletes `objectKey` unconditionally and returns an error. The failure sequence:
1. The row exists and is listed.
2. Every bundle read returns `SKILL_NOT_FOUND "missing from storage"`.
3. Retrying the same upload returns `SKILL_NAME_CONFLICT`, unless the client knows to send replace.

That is the exact invariant `skill-object-lifecycle.ts` says it protects. Nobody else can reference a brand-new Skill id, so there's nothing for the check to catch here. Two fixes work: drop `ensureObjectPresent` from the insert path, or move it after the `try/catch` so only a failed *insert* triggers cleanup. Please add a test using the fake store's `head` failure.

Non-blocking: the three "changed concurrently" paths reuse `skillNameConflict`, so a client sees `SKILL_NAME_CONFLICT` for a revision race. The messages are right, but if the web lane branches on the code it will show "name already exists" copy. A dedicated code, or a note in the contract, would avoid that. Splitting `skill-service.test.ts` is optional.

bestony:
Fixed at exact head `0f660f03`. You were right, and the bug was mine, not the lane's: my S5b instruction said "confirm the object after a successful row write in both create and replace", and on the insert path that landed inside the insert's `try`.

**`20b4bd07` — the insert path.** The `try/catch` covers only the insert, and `ensureObjectPresent` is gone from it: no other writer can hold a key under a brand-new skill id, so it repaired nothing and only added a failure mode after the row had committed. On the replace path the post-commit check stays, but outside the update's `catch`: a failing `head`/restore surfaces `SKILL_STORAGE_UNAVAILABLE` and triggers no cleanup of the key the row now references.

**`0f660f03` — one more of the same class, from auditing every delete site rather than just this one.** There are four places an object can be deleted: `discardUnreferencedObject` (guarded), `deleteReplacedObject` (guarded, residual agreed for the GC follow-up), `remove()` (after the row delete returned), and the insert `catch` — which was still unconditional while replace's `catch` already went through the guard. An INSERT can throw *after* the database committed it (connection dropped before the result came back); the row exists and the catch would delete its object. It now routes through `discardUnreferencedObject` too: re-reads the row by id, refuses to delete a key the row references, and treats an unreadable row as do-not-delete. Tests: commit-then-throw keeps the object and `openBundle` works; insert-fails-with-no-row still removes it; `head` failing after a replace fails the call but leaves the object.

`skill-service.test.ts` is split by concern over a shared harness; no test file exceeds 500 lines now.

On your non-blocking point — the three "changed concurrently" paths reusing `SKILL_NAME_CONFLICT`: agreed it deserves its own code (`SKILL_REVISION_CONFLICT`, as MCP has), otherwise the web shows "name already exists" copy for a revision race. It needs a contract change, which would void #655's reviews yet again, so I have recorded it as a follow-up to land right after #655 merges (contract + server + web message map), together with the deferred orphan-object GC.

Re-run by the orchestrator on `0f660f03`: `pnpm check` 0 (49 migrations) · `pnpm typecheck` 0 · `pnpm build` 0 · server unit 151 files / 2365 tests · integration 41 files / 534 tests.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-server-2`._

liuchao-001 (approved):
Approving at exact head `0f660f03`.

- **Insert path.** The `try` now covers only the insert. `ensureObjectPresent` is gone from it, which is correct: nobody else can reference a key under a brand-new id. The `catch` goes through `discardUnreferencedObject`, keyed on the new `skillId`. So a unique violation (no row with that id) still deletes the object, while an insert that committed and then failed keeps it. Nice catch on the dropped-connection case.
- **Replace path.** `ensureObjectPresent` now sits outside the update's `catch`, so a failing `head` or restore after commit can't clean up the key the row references.
- The earlier fixes (zip bounds, gunzip meter, secret redaction, revision guards, enabled-only Computer bundles) are unchanged since my last pass.
- Ran `skill-archive`, `skill-service`, `skill-service-concurrency` and `skills-route` locally: 44/44 pass. CI on this head was still running when I approved.

I'm fine with the two follow-ups going out after #655 merges: `SKILL_REVISION_CONFLICT` (contract, server and web message map) and the deferred orphan-object GC that closes the remaining delete-after-`head` window.

packages/server/src/services/skills/skill-archive-reader.ts
```diff
@@ -0,0 +1,240 @@
...
+  for (const body of Object.values(unzipped)) actualBytes += body?.byteLength ?? 0;
+  if (actualBytes > SKILL_UNPACKED_MAX_BYTES) throw skillArchiveTooLarge();
+  const entries: RawSkillEntry[] = [];
+  for (const [rawName, body] of Object.entries(unzipped)) {
+    if (rawName.endsWith("/") || body === undefined) continue;
+    const path = normalizeMemberPath(rawName);
+    if (isIgnoredSkillPath(path)) continue;
+    if (state.seen.has(path)) throw skillArchiveInvalid(`Skill archive member is duplicated: ${path}`);
+    state.seen.add(path);
+    entries.push({ path, body, mode: 0o644 });
```
yuezengwu: [P2] Preserve executable permissions when normalizing Unix ZIP members. Every ZIP entry is assigned 0644 here, so a legitimate Unix ZIP containing `scripts/run.sh` at 0100755 is re-packed with mode 0644; an equivalent tar.gz keeps 0755. I verified both stored tar headers independently. Skills whose instructions directly invoke a bundled executable therefore stop working after ZIP import, even after the separate client extraction issue in #656 is fixed. Read and validate the supported Unix mode metadata, then carry a safe execute bit into the canonical archive; add a ZIP executable round-trip regression.

packages/server/src/services/skills/skill-archive.ts
```diff
@@ -0,0 +1,141 @@
...
+export async function normalizeSkillArchive(
+  bytes: Uint8Array,
+  format: SkillArchiveFormat,
+): Promise<NormalizedSkillArchive> {
+  if (bytes.byteLength === 0 || bytes.byteLength > SKILL_ARCHIVE_MAX_BYTES) throw skillArchiveTooLarge();
+  const entries = stripSingleRoot(await readSkillEntries(bytes, format));
+  if (entries.length === 0) throw skillManifestInvalid("The Skill archive carries no files");
+  const manifest = requireManifest(entries);
+  const { files, filesTruncated } = toFileList(entries);
+  const archive = await packDeterministic(entries);
```
yuezengwu, line 138: [P2] Enforce the archive size limit again after re-packing. The input guard cannot bound the canonical output because ZIP and tar/gzip have different framing/compression overhead. An independently generated STORED ZIP of 16,776,487 bytes is accepted here and returns a 16,779,914-byte archive, exceeding the 16,777,216-byte contract. A real migrated database upload then fails with PostgreSQL 23514 / agent_skills_archive_bytes_bounds after the object PUT, rather than a typed SKILL_ARCHIVE_TOO_LARGE response. Reject an oversized canonical archive before any storage/database write and cover a near-limit input in a regression test.

yuezengwu (changes requested):
Reviewed exact head `0f660f0339387d200c85d571d2858c0c7c6e0d24`. Requesting changes for two additional archive-normalization failures reproduced independently below; previous review points and agreed lifecycle follow-ups are not repeated.

Validation: 89 affected unit tests (7 files), 7 real PostgreSQL integration tests (2 files), shared/server builds, scoped typechecks, `pnpm check` (49 migrations), and `git diff --check` passed. ZIP/tar probes and a migrated-database upload independently reproduced the findings. No author branch changes; full repository suites and live S3 acceptance were not run.

The [Enforce Patch Coverage job](https://github.com/first-tree-ai/opentag/actions/runs/35449532097/job/105914164572) actually exits while running server unit tests under coverage (`[tests failed]`, 15885/17973 covered server lines); it does not reach the changed-line threshold check. Its log omits individual failed-test output and the run has no artifacts, so I cannot identify that test from this run. Please restore current-head CI and preserve the failure details if it repeats.

bestony:
@yuezengwu all three addressed at exact head `486a3eda` — including the Patch Coverage point, where you were right and I was wrong.

**Patch Coverage.** I had written that job off as the repo-wide floor ratchet without checking. It is not: `coverage.yml` passes on #656 and on unrelated branches, and here the server unit run reports `[tests failed]`. I reproduced the run locally (passes, 166 s vs CI's 398 s) and then timed the skill suite under v8 coverage: `rejects a gzip that inflates past the decompressed-stream ceiling` took **1820 ms** on a fast laptop, so at CI's ~2.4× it crosses vitest's 5 s timeout. That test was my ask in the previous round, specified without thinking about its cost. `486a3eda` makes the ceilings injectable (`limits.maxTarStreamBytes` / `maxUnpackedBytes` / `maxArchiveBytes`, defaulting to the real constants, with one assertion pinning the default) and the test now uses a small ceiling: the slowest skill test under coverage is **93 ms**. If the job is still red on this head I will pull the failing test out of the log rather than guess.

**P2 — ZIP import dropped the execute bit → `ba304048`.** fflate does not expose Unix modes, so there is now a bounded parser for the ZIP central directory (`skill-zip-directory.ts`: EOCD → central headers, every offset/length checked, entry count capped, malformed → `SKILL_ARCHIVE_INVALID`). For Unix-made entries: any execute bit → canonical `0755`, else `0644` — exactly what the tar path normalizes to; DOS-made entries → `0644`. Reading the mode also closes a gap I had recorded as "not detectable, harmless": a ZIP **symlink** member is now rejected, as are special files and setuid/setgid/sticky bits.

**P2 — canonical archive could outgrow the limit → guard in `skill-archive.ts`, test `a7f42f7d`.** After re-packing and before any storage or database write, `archive.byteLength > maxArchiveBytes` → `SKILL_ARCHIVE_TOO_LARGE` (413). The regression uses a stored ZIP just under the input limit with an incompressible payload and asserts a typed 413, zero `put` calls and no row — i.e. no more PostgreSQL 23514 after the PUT, and no orphan object.

**Verified by me with archives built by the real `zip` (Info-ZIP, Unix made-by) and `tar` tools, not the lane's fixtures — against both heads:**

| | `0f660f03` | `486a3eda` |
| --- | --- | --- |
| `scripts/run.sh` at `0755` in a Unix zip → canonical tar mode | `0644` | **`0755`** |
| the same tree as `.zip` and as `.tar.gz` → canonical sha256 | **differ** | **identical** |
| zip with a symlink to `/etc/passwd` (`zip -y`) | **accepted** | rejected — `may not contain links` |
| zip with a setuid script | **accepted** | rejected — setuid/setgid |

Re-run on `486a3eda`: `pnpm check` 0 (49 migrations) · `pnpm typecheck` 0 · `pnpm build` 0 · server unit 151 files / 2371 tests · integration 41 files / 534 tests · skill suite under coverage green. The client half of the execute-bit problem is fixed on #656 (`d13a33e1`).

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-server-2`._

packages/server/src/services/skills/skill-archive-reader.ts
```diff
@@ -0,0 +1,312 @@
...
+ * The canonical mode for a zip member: `0755` when a Unix-made regular file carries any execute bit,
+ * `0644` otherwise — the same normalization the tar path applies. A Unix symlink, a special file, or
+ * any setuid/setgid/sticky bit is rejected; a DOS/Windows entry has no mode to lose, so it is `0644`.
+ */
+function canonicalZipMode(rawName: string, directory: Map<string, ZipDirectoryEntry>): number {
+  const info = directory.get(rawName);
+  if (!info?.madeByUnix || info.unixMode === 0) return 0o644;
+  const type = info.unixMode & S_IFMT;
+  if (type === S_IFLNK) throw skillArchiveInvalid("Skill archive may not contain links");
+  if (type !== S_IFREG && type !== S_IFDIR) {
```
yuezengwu: [P2] Accept ordinary Unix ZIP entries whose file-type bits are absent. Python zipfile.ZipFile.writestr produces create_system=3 with external_attr >> 16 equal to 0600 for a normal SKILL.md on Unix. Here that yields type=0 and rejects the entire supported ZIP as containing special files. I reproduced this at 486a3eda with a two-file archive generated by the unmodified Python standard library; normalization returns SKILL_ARCHIVE_INVALID. Treat absent type bits as an ordinary member (while retaining checks for explicit symlinks/special types and unsafe permissions), and add a permissions-only ZIP regression.

yuezengwu (changes requested):
Re-reviewed exact head 486a3eda. Both prior archive findings are independently resolved: ZIP executables retain their execute bit, and oversized canonical output receives a typed 413 before any object/database write. The 95 affected unit tests, 7 PostgreSQL integration tests, scoped build/typecheck and repository checks pass; current technical CI is green. One new ZIP compatibility regression is reproduced below. Existing review points and agreed follow-ups are not repeated.

bestony:
@yuezengwu fixed at exact head `4fa3e16d`. You're right — Python's `zipfile.writestr` writes `create_system=3` with `external_attr >> 16` equal to the permission bits alone, so treating every non-regular type as a special file rejected a legitimate archive.

`canonicalZipMode` now treats an **absent file-type as an ordinary member**; explicit symlinks, special types, and setuid/setgid/sticky are still rejected. Regressions added: a permissions-only Unix zip (`0600` / `0700`) normalizes to `0644` / `0755`, and a FIFO member is still rejected with `SKILL_ARCHIVE_INVALID`.

Verified against the real repro: a two-file zip built with the unmodified Python standard library normalizes cleanly (SKILL.md → 0644).

Ran: skill-archive 23/23 · server unit 151 files / 2373 tests · `pnpm typecheck` 0 · `pnpm check` 0 · server unit coverage suite green (no `[tests failed]`).

_Pushed by the grok agent (pi) on behalf of @bestony; not the opencode lane._

bestony:
@yuezengwu the file-type-bits regression is fixed at the current head `4fa3e16d` — and you are right that it came from my brief ("any other non-regular, non-directory type → reject"), which is wrong for permission-only modes; my own check had used Info-ZIP archives, which always set the type bits.

A note on provenance, so the history is not confusing: `4fa3e16d` was pushed by @bestony directly from another machine while the dispatched lane was working on the same finding. The lane's equivalent fix was verified locally but **not** pushed — its push was rejected as non-fast-forward and I do not force-push — so the PR head is the owner's commit. The lane is merging it locally; nothing further will be pushed unless a later round needs it.

**`4fa3e16d` verified by me, across ZIP generators rather than one tool:**

| Archive | Expected | Result |
| --- | --- | --- |
| Python `zipfile.writestr`, unmodified stdlib (your case: mode `0600`, type bits 0) | accept | accepted |
| permission-only `0755` member (no type bits) | accept → canonical `0755` | accepted, `0755` |
| `S_IFREG`-typed members | accept | accepted, exec bit kept |
| DOS-made zip (no Unix mode) | accept → `0644` | accepted, `0644` |
| permission-only + setuid (`04755`, no type bits) | reject | rejected — setuid/setgid |
| `S_IFLNK` member | reject | rejected — links |
| `S_IFIFO` / `S_IFCHR` members | reject | rejected — special files |
| Info-ZIP archives from the previous round (exec bit, zip≡tar.gz sha256, `zip -y` symlink, setuid) | unchanged | unchanged |

Full acceptance on `4fa3e16d`, run in a clean worktree: `pnpm check` 0 · `pnpm typecheck` 0 · `pnpm build` 0 · server unit 151 files / 2373 tests · integration 41 files / 534 tests.

Requesting a fresh exact-head review of `4fa3e16d`.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-server-2`._

yuezengwu (approved):
Revalidated 4fa3e16d independently. The original unmodified Python writestr ZIP now imports, permissions-only executables retain 0755, and explicit links/special types and unsafe permission bits remain rejected. I also replayed the earlier executable roundtrip and real-size overflow cases: execution succeeds after extraction, and overflow receives typed 413 with zero storage/database writes. All three reported archive findings are resolved. Validation: 97 affected unit tests, 7 PostgreSQL integration tests, scoped build/typecheck and repository checks pass. No live S3 acceptance or full repository suite is claimed; previously agreed follow-ups remain.

bestony:
New exact head `6bb92b7e` — **no new server behaviour; this head exists because #655 was squash-merged.**

#655 landed in `main` as `e328c238` and its branch was deleted, so GitHub retargeted this PR to `main`, where it showed CONFLICTING and the approvals were cleared. The conflicts were only in contract-owned files (`docs/design/agent-skills.md`, `packages/shared/src/__tests__/skill.test.ts`, `packages/shared/src/skill-manifest.ts`): this branch still carried an older copy of the contract than the one that was merged.

What this head adds on top of the approved `4fa3e16d`:
1. `6bb92b7e` — merge of `origin/main`, resolving every conflict by taking **main's** version. Verified: `git diff origin/main -- packages/shared docs/design` is **empty** (the contract here is byte-identical to main), and `git diff --name-only origin/main...HEAD` lists only this PR's own paths.
2. `d63eccd2` (with `b7f9a706`) — the dispatched lane's own version of the file-type-bits fix, which had been verified but deliberately **not** pushed once @bestony's equivalent `4fa3e16d` became the head. Merged locally: one table-driven `classifyZipMode` with the same behaviour as `4fa3e16d`'s `type !== 0` condition, plus the tests from both sides (+65/−7 in `skill-archive-reader.ts` and `skill-archive.test.ts`). It rides along now only because a push was needed anyway.

Nothing else changed in `packages/server`. Migrations are unchanged (49; this PR's is `0048_known_gertrude_yorkes`).

Re-run by me on `6bb92b7e`: `pnpm check` 0 (migration drift: 49) · `pnpm typecheck` 0 · `pnpm build` 0 · server unit 151 files / 2375 tests · integration 41 files / 534 tests · shared 475 tests · the ZIP generator matrix from my previous comment (Python permission-only, typed, DOS, setuid-without-type, symlink, FIFO, char device) still all as expected. A dry-run merge into `main` reports 0 conflicts.

Sorry for the extra exact-head round — requesting a fresh review from you both.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-server-2`._

yuezengwu (approved):
Re-reviewed exact head 6bb92b7e against main e328c238. The merge is clean and the shared contract matches main exactly. I independently checked the archive-mode refactor and re-ran the original regressions: ordinary Python ZIPs import, safe executable modes survive ZIP/tar normalization and client extraction, unsafe modes are rejected, and canonical archive overflow returns 413 before object or database writes. All previously reported blocking findings are resolved.

Validation on this head: 475 shared tests, 99 affected server unit tests, 7 PostgreSQL integration tests, build, typecheck, pnpm check and diff checks passed. Current-head GitHub CI is now green. Full repository tests and live S3 acceptance were not run by this review.

liuchao-001 (approved):
Approving at exact head `6bb92b7e`. Reviewed the zip-mode work since my `0f660f03` approval: `ba304048`, `a7f42f7d`, `486a3eda`, `b7f9a706`, `4fa3e16d`, plus the #655 merge.

**Checked and good:**
- **`skill-zip-directory.ts`.** The EOCD scan, the field offsets (made-by at 4, name/extra/comment lengths at 28/30/32, external attributes at 38) and the bounds checks all read correctly against the spec, and every arm is a typed `SKILL_ARCHIVE_INVALID`. A ZIP64 sentinel count is rejected, and a sentinel directory offset fails the out-of-bounds check.
- **Mode classification.** Absent type bits → regular file, which is yuezengwu's Python case; explicit symlink and special types rejected; `0o7000` rejected; any execute bit → `0755`. I verified with a real `zipfile.ZipFile.writestr` archive (mode `0600`, no type bits): it normalizes, and a Unix symlink member is rejected with "may not contain links".
- **Canonical-size guard.** After the re-pack, before any store or row write, so the typed 413 replaces the PostgreSQL CHECK failure. Correct placement.
- **Injectable limits.** `resolveSkillReadLimits` validates each value, production leaves them unset, and the default is pinned by an assertion.
- Skill suites at this head: 60/60 pass locally (`skill-archive`, `skill-service`, `skill-service-concurrency`, `skills-route`, `s3-skill-object-store`).

**One non-blocking finding, for a follow-up rather than this PR.** The directory map is keyed by a name the parser always decodes as UTF-8, while fflate decodes a name as latin-1 unless the UTF-8 flag (bit 11) is set (`zh`: `strFromU8(..., !(b2(d, b + 8) & 2048))`). For a member with a non-ASCII name and no UTF-8 flag the two keys differ, `directory.get(rawName)` misses, and `canonicalZipMode` falls back to `0644` — skipping the link, special-file and setuid checks for that member.

Reproduced on this head with a hand-built stored zip (one latin-1 byte in the name, bit 11 clear, `external_attr = 0o120777 << 16`):
```
fflate keys: ["SKILL.md","liénk.txt"]
directory:   [... {"name":"li�nk.txt","madeByUnix":true,"unixMode":41471}]
normalize:   ok, files: [SKILL.md, liénk.txt]   // the symlink member is accepted
```
The stored result is a regular file containing the link text, which is the behaviour this PR already documented as harmless, so it is not a blocker — but the guard silently does not run, and the execute bit is lost for the same names. Worth decoding the name the same way fflate does, or treating a lookup miss as strict rather than as `0644`. Happy for it to ride with the `SKILL_REVISION_CONFLICT` and orphan-GC follow-ups.

PR #658, opened 2026-09-19
feat(web): replace the mock Skills page with real Agent Skills management

bestony:
> **Stacked on #657** (`feat/skill-server`), which is stacked on #655 (`feat/skill-contract`). Base is `feat/skill-server` so this diff shows only the web work; it will be retargeted as the stack merges. It sits on the server lane because the journey e2e exercises the real Skills route.

## Summary

Replaces the mock Skills page with a real one for a single Agent: list, upload (`.zip` / `.skill` / `.tar.gz`), replace-on-conflict after confirmation, enable/disable, delete, download.

- **`apps/web/src/api.ts`** — `agentSkills`, `agentSkill`, `uploadAgentSkill`, `updateAgentSkill`, `removeAgentSkill`, `agentSkillBundleUrl` (built from the typed path helper). The upload is an octet-stream `fetch` with the sha256 / format / replace headers **and the CSRF headers** — it is a cookie-authenticated mutation; `api.csrf.test.ts` covers it.
- **`features/skills/`** — `skills-queries.ts` (live list + mutations with one shared invalidator), `skills-page-model.ts` (pure: format detection, size pre-check, `crypto.subtle` sha256, exhaustive `SKILL_ERROR_CODES` → message map), `skills-page.tsx` + `skill-row.tsx` + `skill-dialogs.tsx`. The page takes only `agentId` and never reads the router.
- **Storage-unavailable is rendered from the list response** (`storage: "unavailable"`): a calm notice, Upload disabled, downloads hidden — no probe request, so a deployment without object storage produces zero failed requests (the e2e server is exactly that).
- A `SKILL_NAME_CONFLICT` opens the replace confirmation with the archive already hashed, then re-submits with `replace: true`.
- Copy in `messages/skills/{en,zh}.json` (identical key sets); mock `skillPreviews`, the old `features/skills-page.tsx` and its mock-only test are removed. The `skills` navigation flag stays off by default.

## Checklist

- [x] 1. Skill API calls with CSRF on the upload, covered by the CSRF test
- [x] 2. `features/skills/` queries, page model, page; route re-pointed; mock removed
- [x] 3. Messages in `en` and `zh`
- [x] 4. Unit/component tests and e2e updates
- [x] 5. All acceptance criteria verified, `git status` clean
- [x] Round 1: merged `feat/skill-server`; **full journey** green

## Acceptance criteria — re-run by the orchestrator on `801f1426`

| Criterion | Result |
| --- | --- |
| `pnpm check` | exit 0 |
| `pnpm typecheck` | exit 0 |
| `pnpm --filter @opentag/web test` | exit 0 — 96 files / 1116 tests |
| `pnpm test:e2e` (build + **full journey**) | exit 0 — 25 passed |
| `pnpm test:e2e:smoke` | 13 passed (round 0) |
| No `skillPreviews`; old page gone; en/zh key parity; no file > 500 lines; no new dependency | holds |
| Diff vs `feat/skill-server` only under `apps/web/` and `e2e/`; nothing under `src/paraglide/` | holds |

## Two things a reviewer should know

1. **The brief's e2e criterion was wrong, and it hid a real dependency.** It named `test:e2e:smoke`, which only matches `*smoke.spec.ts` and never runs the `app.spec.ts` change. The orchestrator ran the full journey on the first version of this branch: `app.spec.ts:412` failed, because it asserts the storage-unavailable notice and that branch's server had no Skills route. Hence this PR stacks on #657 rather than sitting beside it.
2. **An unrelated, pre-existing journey failure is fixed here as its own commit (`801f1426`).** The contact-sheet test expected the heading "Integrations" on `/agents/:id/integrations`; that page's only title is "GitHub repositories" (`integrations_agent_github_title`) — there is no "Integrations" heading on `main`. It had gone unnoticed because CI runs Browser Smoke, not the journey, and on this branch it was first masked by the Skills failure ahead of it in the serial suite. The page is untouched; only the assertion changed. Happy to split it out if you prefer.

Cross-lane check done by the orchestrator: upload and PATCH responses are `SkillDetailSchema` on both sides, and with #655 `403aff84` the shared error envelope admits `SKILL_*` codes, so the conflict → replace flow works against the real server rather than only against mocks.

## Provenance

Written by an agent under `dispatch-opencode` run `tlm1q4`, lane `skill-web-4`. Plan and acceptance criteria authored, and the result verified, by the orchestrator (Claude Code) before push. 6 own commits for a 5-item checklist.


bestony:
New head `70eb73c0` — a merge only; no `apps/web` or `e2e` source changed in this round.

The base `feat/skill-server` moved to `378d2c23` (it now contains `main` `22a85448` — #654, which also edited `apps/web/src/api.ts` and `api.csrf.test.ts`, both touched here — plus the server review fixes and the migration renumbered to `0048`). Git merged it without a textual conflict, so the merge was proven semantically instead: `api.csrf.test.ts` asserts that the set of `browserApi` methods equals its `INVOCATIONS` keys; after the merge it holds both #654's two new Feishu methods and this PR's six Skill methods, and the suite passes. `query/keys.ts` likewise holds both sides.

Re-run by the orchestrator on `70eb73c0`: `pnpm check` 0 · `pnpm typecheck` 0 · web tests 98 files / 1134 tests · **full journey 25/25, twice in a row**.

One thing recorded rather than papered over: the lane's first post-merge journey run had `app.spec.ts:381` ("mobile global return…", a focus-restoration assertion) fail once, which skipped the four serial tests after it; its next two runs and both of mine were clean. That test's code path (`features/shell/shell-transition.ts`, `features/agents/agents-page.tsx`) is untouched by this PR and by the merge, and it passed in every earlier journey run on this branch — so it reads as a pre-existing timing flake (the Playwright config retries once on CI), not a regression. Flagging it in case it is already known.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-web-4`._

apps/web/src/features/skills/skills-page.tsx
```diff
@@ -0,0 +1,180 @@
...
+    } finally {
+      setUploadingName(undefined);
+    }
+  };
+
+  const confirmReplace = async () => {
+    if (!pendingReplace) return;
+    setActionError(undefined);
+    try {
+      await uploadArchive(pendingReplace, true);
```
yuezengwu, line 135: [P1] Keep the pending replacement bound to its original Agent. This state survives agentId changes, but uploadArchive now uses the mutation for the current Agent. I reproduced through the actual App/router: visit B Skills, navigate to A, upload and get a name conflict, press browser Back to B, then click the still-open Replace Skill confirmation. Captured API calls are A/replace=false followed by B/replace=true. If B has a same-name Skill, this overwrites B with the archive selected for A. Clear/remount Agent-scoped transient state and bind/guard pending and in-flight uploads by their originating Agent, including delayed hash/request completions; cover the history-navigation case.

apps/web/src/features/skills/skill-row.tsx
```diff
@@ -0,0 +1,84 @@
...
+            aria-label={m.skills_toggle_label({ name: skill.name })}
+            checked={skill.enabled}
+            disabled={update.isPending}
+            onCheckedChange={(next) => void toggle(next)}
+            transitioning={update.isPending}
+          />
+          {storageAvailable ? (
+            <a
+              className={buttonClassName({ size: "compact", variant: "secondary" })}
+              download={skill.name}
```
yuezengwu: [P2] Give downloaded bundles the canonical .tar.gz extension. skill.name is extensionless, so this link suggests a filename such as demo; the server sends application/octet-stream without Content-Disposition to override it. A local Chromium probe confirms suggestedFilename=demo, and checkSkillArchiveFile({name:"demo", ...}) returns unsupported_format. Consequently a user cannot directly re-upload or transfer a bundle downloaded from this very page. Suggest `${skill.name}.tar.gz` and verify that the downloaded filename is accepted as tar.gz.

yuezengwu (changes requested):
Reviewed exact head `70eb73c0a16a06bb812593b1b5177b54d04227ab`. Requesting changes for two independently reproduced Web issues below: a pending replacement can target a different Agent after browser Back, and downloaded bundles have a filename this page refuses to upload. No existing upstream findings are repeated.

38 existing web tests, shared/web builds, scoped typechecks, `pnpm check`, and `git diff --check` passed. Two additional regression expectations fail on this head using the actual App/router with mocked API calls; a separate local Chromium download probe confirms the extensionless filename. No real Agent data or author source branch was modified. Full E2E and live storage acceptance were not run.

Current CI is also not green: [Enforce Patch Coverage](https://github.com/first-tree-ai/opentag/actions/runs/35449218425/job/105913340716) fails during server tests under coverage, before the changed-line threshold check. The log does not expose the individual failing test.

bestony:
@yuezengwu thank you — both findings are right, and the P1 cuts against the one property this feature exists for (an Agent's Skills are its own). A fix round is starting now; details below the merge note.

**New head `dd3332de` is a merge only** (no `apps/web`/`e2e` source changed): the base `feat/skill-server` moved to `486a3eda`. That merge is also what should clear the red **Enforce Patch Coverage** here — you were right that it fails in server tests under coverage rather than on the threshold. I traced it on #657: `skill-archive.test.ts > rejects a gzip that inflates past the decompressed-stream ceiling` took 1820 ms under v8 coverage locally and times out at CI speed; it now runs in 93 ms. This branch simply carried that server test.

Re-run by me on `dd3332de`: `pnpm check` 0 · `pnpm typecheck` 0 · web tests 98 files / 1134 tests · `pnpm build` 0 · **Browser Smoke 13/13** (the e2e project CI runs).

**Journey e2e — reported as it is, not as I would like it.** It is not stable on my verification host today: across three runs on this head I got 24/25 (`navigation-polish.spec.ts:43`, dock menu not hidden), 25/25, and 19/25 (`app.spec.ts:329`, settings display name). To find out whether that is this PR, I ran the journey three times on `feat/skill-contract` (`0ab6b1c6`), whose `apps/web` and `e2e` are byte-identical to `main`: **3 of 3 failed, each time at `app.spec.ts:329`.** So the instability predates and is independent of this PR; the Skills journey test (`app.spec.ts:412`) passed in every run that reached it. CI does not run the journey project, which is presumably how this has gone unnoticed. I would not treat my earlier "25/25" on this PR as stronger evidence than it is.

**Your two findings — fix in progress on this branch:**
- **P1, pending replace survives an Agent change.** Agreed, and the captured calls (`A/replace=false` → `B/replace=true`) are exactly the cross-Agent write the design forbids. Fix: key the page's transient state to the Agent (remount on `agentId` change), and bind every pending/in-flight upload — including the delayed hash and the request completion — to the Agent it originated from, dropping it if the Agent has changed. Regression through the real App/router with history navigation, as you did it.
- **P2, extensionless download.** Agreed. The link gets `download="<name>.tar.gz"`, with a test that the downloaded filename passes `checkSkillArchiveFile` as `tar.gz`. I will also raise on #657 that the bundle routes should send `Content-Disposition`, so the filename is right for non-browser clients too.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-web-4`._

bestony:
@yuezengwu both fixed at exact head `417c4b0b`.

**P1 — pending replace could be sent to another Agent → `f3983ea2`.** Two layers, because either alone leaves a gap:
1. *State cannot be carried over.* `SkillsPage` now renders `<SkillsPageBody key={agentId} …/>`, so an Agent change remounts the body and drops every dialog, pending archive and error.
2. *Work is bound to where it started.* `PendingReplace` carries its `agentId`; the upload/update/remove mutations take the Agent id explicitly instead of closing over the page's current prop (rows use `skill.agentId`); `confirmReplace` refuses to run when `pendingReplace.agentId !== agentId` or the body is no longer alive; and the async steps (the `sha256Hex` hash, the upload promise) re-check that guard before any `setState` or follow-up request — so a slow hash or a late conflict response cannot open a dialog on another Agent's page.

Four regressions in `skills-page-agent-switch.test.tsx`, driven through the real router with history navigation as you did it: your exact Back scenario (no Replace dialog open on B, **no request ever sent to B**), a hash that resolves after the Agent changed, a conflict that arrives after the Agent changed, and a delete dialog open across an Agent change.

I checked that those tests actually catch the bug rather than trusting them: in a scratch worktree I put the **pre-fix** `skills-page.tsx` / `skills-queries.ts` / `skill-row.tsx` (`dd3332de`) under the new test file — **4 of 4 fail** (dialog still present; `uploadAgentSkill` called once where zero were expected). With this head's sources, 4 of 4 pass.

**P2 — extensionless download → `417c4b0b`.** The link is `download="<name>.tar.gz"`, with a test that this exact filename is accepted by `checkSkillArchiveFile` as `tar.gz`. The server-side half (`Content-Disposition` on the bundle routes, so non-browser clients get the right name too) is recorded as a follow-up for #657's branch.

Re-run by me on `417c4b0b`: `pnpm check` 0 · `pnpm typecheck` 0 · web tests 99 files / 1139 tests · `pnpm build` 0 · Browser Smoke 13/13 · diff vs `feat/skill-server` still only `apps/web/` and `e2e/`. (The journey project remains unstable on my host independent of this PR — see my previous comment; CI does not run it.)

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-web-4`._

yuezengwu (approved):
Revalidated 417c4b0b: both reported Web regressions are resolved. The original actual App/router Back-navigation probe no longer submits a replacement to another Agent; the download filename is accepted by the upload check and independently confirmed as demo.tar.gz in Chromium. The 43 affected existing Web tests plus 2 independent regressions, scoped build/typecheck and repository checks pass. Current technical CI is now green. Approval applies to this exact head; no live storage or full journey acceptance is claimed.

bestony:
New exact head `eab76fa6` — **a merge of `main` only; no `apps/web` or `e2e` source changed.**

#655, #657 and #656 are all squash-merged (`main` = `c8b7f5a6`) and `feat/skill-server` was deleted, so GitHub retargeted this PR to `main`, where it showed CONFLICTING and the approval was cleared. The five add/add conflicts were all **outside** this PR's paths (`docs/design/agent-skills.md`, two `packages/server` files, two `packages/shared` files): this branch still carried older copies of server and contract files it never owned.

`eab76fa6` merges `origin/main` and takes **main's** version of every one of them. Verified:
- `git diff origin/main -- . ':!apps/web' ':!e2e'` is **empty** — everything outside this PR's paths is byte-identical to main;
- `git diff --name-only origin/main...HEAD` lists only `apps/web/` and `e2e/` — the PR diff is now just the web work;
- no conflict markers; a dry-run merge into `main` reports 0 conflicts.

Integration with the new `main`, re-run by me on `eab76fa6`: `pnpm check` 0 · `pnpm typecheck` 0 · web tests 99 files / 1139 tests · `pnpm build` 0 · Browser Smoke 13/13. The lane's full journey run on this head was 25/25 (with the standing caveat from my earlier comment that the journey project is not stable on my host independent of this PR; CI runs Browser Smoke).

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-web-4`._

yuezengwu (approved):
Re-reviewed exact head eab76fa6 against main c8b7f5a6. The retarget/merge conflicts are resolved: Web and E2E sources are unchanged, every file outside those paths matches main exactly, and an independent dry-run merge produces the current head tree.

Fresh validation after rebuilding shared/Web: 45 affected Web tests (including the original two independent regression probes), 475 shared tests, build, typecheck, pnpm check and diff checks passed. Browser Back no longer allows a pending replacement to write to another Agent, and the downloaded archive filename remains accepted by the upload picker. Current-head GitHub CI is green; no remaining blocking finding. Full repository suites, fresh browser E2E and live storage acceptance were not run by this review.

yuezengwu (changes requested):
Reviewed exact head `eab76fa6a3a9b93ae7b18c2805b4c2ebc8203523` against `main@c8b7f5a685c4a231d5464ef86ddefccc5373529d`. I am requesting changes because fresh production-bundle browser validation found two blocking state-lifecycle regressions that the green CI does not cover.

1. **Unknown storage is treated as available, and an initial list failure is rendered as an empty list.**

   In `apps/web/src/features/skills/skills-page.tsx`, `skills.data?.storage !== "unavailable"` evaluates to `true` before the query has returned successful data and after an initial error. In Chromium, Upload is enabled during the initial loading state. When the initial list GET returns 503, the page simultaneously shows the request error, an enabled Upload button, and `No Skills yet`, even though neither storage availability nor an empty list was established.

   Please fail closed until a successful response explicitly reports available storage, and render the empty state only after a successful empty-list response. Add coverage for both initial pending and initial-error states.

2. **A confirmed write is discarded from the UI when the immediately following list refetch fails.**

   In `apps/web/src/features/skills/skills-queries.ts`, POST and PATCH return authoritative `SkillDetail` values and DELETE returns 204, but each `onSuccess` only invalidates the list query. If that refetch returns 503, React Query retains the pre-mutation cache. Reproduced outcomes:

   - POST 201: the new Skill is missing and the old empty state remains.
   - PATCH 200 with `enabled: false`: the toggle remains checked.
   - DELETE 204: the deleted row remains visible.

   Please reconcile each Server-confirmed result into the relevant cache before or alongside invalidation, while still surfacing the refetch error. Add regression coverage for successful POST/PATCH/DELETE followed by a failed list refetch.

These failures were reproduced independently in two production Web bundle + real Chromium harnesses with deterministic same-origin API responses. The passing controls include explicit `storage: unavailable`, all-successful mutation/refetch flows, Agent A pending upload followed by navigation to Agent B and a late 409, and the `.tar.gz` download-to-upload path.

Fresh validation otherwise passed: `git diff --check`, `pnpm check`, build (6/6), typecheck (8/8), focused Web tests (5 files / 43 tests), full Web tests (99 files / 1139 tests), and all current required GitHub checks. No live object-storage or persistent Server acceptance is claimed.

bestony:
@yuezengwu thank you — both findings are right, and neither was reachable from the green CI. Fixed at exact head `f3e60200`.

**1. Unknown storage treated as available; initial failure rendered as an empty list → `7a5bc481`.** The cause was exactly the line you named: `skills.data?.storage !== "unavailable"` is `true` whenever there is no data. Storage is now three states — `unknown | available | unavailable` — derived only from successful data (`storageStateForList` in `skills-page-model.ts`):
- Upload and every storage-dependent row control are enabled only for an explicit `available`;
- the "storage unavailable" notice shows only for an explicit `unavailable` — a failed first load no longer presents a calm deployment notice for what is really a request error;
- "No Skills yet" renders only when a successful response carried an empty list. With no data: loading text while pending, and after an initial error the error banner alone;
- a *background* refetch failure after a successful load keeps the last known list and storage state and adds the error banner, so a list the user was looking at is not blanked.

**2. A confirmed write discarded when the following refetch fails → `654bf5d6`, `f3e60200`.** Each `onSuccess` now reconciles the Server-confirmed result into the cache *before* invalidating (`skills-cache.ts`, pure and unit-tested): upload upserts by `id` in the list endpoint's name order (so a replace updates in place, never duplicates), update replaces by `id`, remove drops by `id`; upload/update also set the detail cache and remove retires it. Three deliberate details:
- the cache key is always the Agent the write targeted, never the page's current Agent — the binding from the previous round is preserved;
- list items are built field-by-field as `Skill`, so `files`/`filesTruncated` from the `SkillDetail` cannot leak into the list cache;
- if the list was never loaded, no list is fabricated (that would assert a storage state the page never read) — the write only invalidates.

The invalidation still runs, and its failure still surfaces as the load-error banner; it does not turn a successful mutation into a failed one.

**Coverage, and proof it bites.** New tests: initial pending; initial 503; success-empty; successful load then failed refetch; and POST 201 / PATCH `enabled:false` / DELETE 204 each followed by a 503 list refetch — asserting the row is present / the toggle is unchecked / the row is gone, the load error is shown, and no mutation error is shown — plus cache-level tests for ordering, replace-in-place, no fabricated list and no detail fields. I put the **pre-fix** `skills-page.tsx` and `skills-queries.ts` (`eab76fa6`) under the new test files: **9 of 24 fail** (all five page-state tests and the four reconciliation tests). With this head's sources, 24 of 24 pass.

Re-run by me on `f3e60200`: `pnpm check` 0 · `pnpm typecheck` 0 · web tests 100 files / 1160 tests · `pnpm build` 0 · Browser Smoke 13/13 · `git diff --name-only origin/main...HEAD` still lists only `apps/web/` and `e2e/`. I did not re-run your production-bundle Chromium harness, and no live object-storage acceptance is claimed.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-web-4`._

yuezengwu (approved):
Re-reviewed exact head `f3e60200cb0f315c63cf65593843f394f06fe8d5` against `main@c8b7f5a685c4a231d5464ef86ddefccc5373529d`. Both findings from review 5259463320 are resolved in the inspected and tested scope: unknown storage stays disabled without claiming an empty list, and confirmed POST/PATCH/DELETE results remain visible after a failed list refetch.

Current-head validation passed 64 affected repository tests plus the two original independent navigation/download probes, Web build, typecheck, `pnpm check` and `git diff --check`. The initial pending/error and successful-write/failed-refetch page regressions all pass. Technical CI has now completed successfully; the remaining ownership status explicitly awaits this approval. No fresh production-bundle Chromium or live object-storage acceptance is claimed.

liuchao-001 (approved):
Reviewed exact head `f3e60200` against `main@1a2f3dd2` (merge base `c8b7f5a6`). Approving: I found nothing blocking, and the two state-lifecycle fixes hold up under reading and under an independent mutation test.

**What I verified myself, not from the description**

- **The regression tests actually bite.** In my own worktree I put the pre-fix `skills-page.tsx` and `skills-queries.ts` (`eab76fa6`) under the new test files: **9 of 24 fail** — all five page-state tests and the four reconciliation tests. Restored to this head, 24 of 24 pass. That matches the author's claim and is the reason I trust the coverage rather than the prose.
- **Suites on my build of this head:** focused Skills + CSRF + mock-pages 6 files / 64 tests pass; full `@opentag/web` 100 files / 1160 tests pass (a first full run of mine overlapped the mutation experiment above, so I discarded it and re-ran clean). All required GitHub checks are green on this head.
- **Diff shape:** `git diff --name-only origin/main...HEAD` is 21 files, all under `apps/web/` and `e2e/`. No `skillPreviews` or `features/skills-page` references survive; `messages/skills/{en,zh}.json` are 34 keys each with identical key sets and no untranslated zh strings.

**On the two findings from review 5259463320**

1. *Fail-closed storage.* `storageStateForList` derives three states from successful data alone, so `unknown` is now structurally distinct from `available`; Upload and the download link key off an explicit `available`, the "not configured" notice off an explicit `unavailable`, and the empty state off `hasData` rather than `skills.length === 0`. The pending / initial-error / success-empty split is exactly right: a failed first load shows the error alone, and a *background* failure after a good load keeps the last list rather than blanking it.
2. *Write survives a failed refetch.* Reconcile-before-invalidate is keyed by the Agent the write targeted, which preserves the previous round's binding. Two details I checked rather than assumed: `listItemOf` builds the list entry field-by-field, so a future `SkillDetail` field is a typecheck error here instead of a leak into the list cache; and `upsertSkill` filtering by `id` cannot duplicate a row on replace, because the Server's `#replaceSkill` updates the existing row rather than inserting a new one. `invalidateQueries` does not reject on a failed refetch, so the confirmed mutation still resolves as success while the refetch failure surfaces as the load banner — which is what the tests assert.

I also looked for the adjacent races: the `alive` ref is re-armed on mount so a StrictMode remount is safe; `invalidateQueries` cancels an in-flight poll refetch by default, so a stale 30s poll cannot land on top of the reconciled cache.

**Non-blocking, take or leave**

- `useAgentSkill` and the detail-cache writes (`setDetail` / `removeDetail`) have no reader today — there is no Skill detail view yet. Harmless, but it is dead weight until one lands; worth a line in the follow-up rather than a change here.
- The `api.ts` comment says the bundle download carries "the Server's `Content-Disposition`". `sendSkillBundle` sets `cache-control`, `content-type`, `content-length` and the sha256 header — no `Content-Disposition`. Same-origin, the `download` attribute is what names the file, so the behaviour is right and the comment is not; the server-side half is already recorded as a follow-up.
- While `sha256Hex` is hashing, `upload.isPending` is still false, so the Upload button stays enabled and a second pick can start a concurrent upload. Small, and the Server would reject the loser, but a busy flag spanning the hash would close it.

Not claimed by this review: live object storage, a full journey run, or a production-bundle browser harness.

PR #659, opened 2026-09-19
feat(cloud): connect Context Tree and enable Session collaboration

yuezengwu:
## Summary

Cloud Agents can connect an existing authorized Context Tree and use `opentag session create/send/list` from a live Cloud execution. Each Session retains its own workspace, Pi history, checkout and unpublished drafts; new Turns use current Agent configuration and fresh execution credentials. Automatic repository creation remains deferred.

- Reuse GitHub authorization and verification, Session models, the Sandbox allocator, and the existing durable-work store. IM and Session collaboration share one Runner FIFO, journal, worker, and save/reclaim boundary.
- Tie Session CLI authority to actual execution and the exact current allocation. Require durable custody before execution and an exact committed terminal result before acknowledgement.
- Preserve custody through cancellation, report-only reconnects, database failures, and retention of active work. Never replay terminal work after allocation loss; atomically replace only superseded never-started custody on the same allocation.
- Scope busy tracking to the existing allocation identity and serialize idle reclaim against dispatch with the existing Sandbox row lock. Preserve current work when retiring old registrations.
- Separate dispatch correlation from logical message identity, share one execution deadline across startup and worker, and observe failed receipts immediately without changing the caller's error result.
- Preserve the newly merged Agent Skills initialization when resolving the Server composition conflict. Fix the composition test to use its existing isolated temporary-workspace helper on macOS.

No new database migration, shared Agent home, message broker, checkpoint history, or cross-Server owner routing. Resources remain 1 vCPU / 1 GiB, with one active Sandbox per Instance.

## Validation

Final head: `a3b9a95b2690c33535dbb47de4fd9c00f0948fa2`.

- Frozen install, `pnpm check`, build, typecheck, and full tests passed on the unchanged clean source.
- Agent Runtime: **456 tests**, **100%** statements/branches/functions/lines.
- Real PostgreSQL integration: **42 files / 547 tests**, including dispatch/reclaim races and custody replacement.
- New fault regressions demonstrated failure before repair and success afterward for throwing socket sends and receipt timeout during activity handoff. The composition test also passed with `TMPDIR` unset, reproducing the original failing environment.
- Exact-head technical GitHub CI passed: Checks, Node 22/24/26 compatibility, Docker, Runner Toolchain, browser/container smoke and Quality Scoreboard. Patch coverage is **2695/2809 = 95.94%**. The ownership/review gate still requires human review.

Fresh native Cloud Run/GCS acceptance on this exact clean Runner image passed **23 checks**: installed Pi and Session CLI collaboration, idempotent child creation, a live child across parent stop, reuse of the same physical Instance without changing its configuration, isolated workspace archives, no platform credential files in archives, and cold recovery of Pi history with current Agent configuration. Each Instance used 1 vCPU / 1 GiB; all temporary Instances, GCS objects/generations and local fixtures were removed. Image digest: `sha256:8a40a1c52578fb72758d887bca3b0105744e6d0a390aaf3459cc2294bfae070a`.

This uses a disposable local Server/PostgreSQL, a controlled SSE model, real installed Pi/CLI, and synthetic IM ingress. The child continuation uses the real Session CLI; its controlled model response spans the parent stop. Two earlier fixture failures are retained: an expected reuse candidate had already expired under the normal idle policy, and an invalid IM delivery to an internal Session was correctly denied. Neither required a product policy or source change. Real GitHub App publication and Slack/Feishu provider delivery remain separate acceptance targets. No staging application configuration was deployed.

## UI validation

This revision changes Cloud error guidance to point to GitHub Integrations and the Agent repository binding, with English/Chinese component tests. Local guidance is preserved. No layout, CSS, or new reusable component is introduced. The prior desktop/mobile fixture checks (320/390/768/1440px, keyboard and alert/status semantics) belong to `37f5a0a1`; no new screenshots are claimed for this revision.

## Deployment and retained boundaries

Deploy Server and Runner as a matched release, Server first. A pre-E8 Server rejects the E8 Runner's auth capability field. For rollback, stop admission, drain/settle/save active E8 environments under a compatible Server, verify those Instances removed, then switch Server/Runner versions. Saved workspace recovery does not reconstruct unacknowledged execution state.

Terminal acknowledgement evidence remains bounded by the existing age/count retention policy. If a terminal record is pruned before its acknowledgement reaches the Runner, the Server will not acknowledge unverifiable work; that journal can block draining until its allocation is removed. Active custody is protected from age pruning. No unbounded terminal-history or additional acknowledgement mechanism is introduced.

An interrupted first Tree clone may require explicit repair; user work is never automatically reset. Local behavior and the database contract remain compatible.

## Checklist

- [x] All seven repository-required commands pass on the current head.


liuchao-001 (changes requested):
Reviewed at head `37f5a0a1`. I read the server collaboration owner, the Runner turns/journal/protocol, the Context Tree path and the credential authority. Credentials and the Context Tree look solid. Internal Sessions are denied on Local, on un-negotiated Runners, after settlement and on stale allocations. No parent IM credential reaches a child. Repo selection is limited to the Agent's own `context_tree` binding, and tokens never land in argv, URLs or `.git/config`. The journal's fsync/rename ordering and capability negotiation in both directions also look right.

The blockers share one theme. Several paths leave a Runner Session-journal entry that the Server will never ack or clear. `drainForRelease` then never reaches zero, and every later `workspace:seal` or idle reclaim of that Sandbox fails.

### Blocking

**1. A re-announced `started` Turn on a stopped Session gets a conflicting terminal state, so its real settle is never acked.**
`cloud-session-collaboration-owner.ts:861-872`
- The `sessionEndedAt / agentStatus !== "active" / suspendedAt` branch writes `failed/session_closed` and rejects, whatever the phase.
- For `phase: "started"`, the Runner ignores the rejection (`cloud-turns.ts:451-456`) and the Turn keeps running.
- When it settles, `settledOutcome()` yields `succeeded` or `failed/turn_*`. That conflicts with `failed/session_closed`, so `#terminalizeDurable` returns `undefined` and no ack goes out (`:977-983`).
- The `reported` entry then replays forever and seal fails.
- Trigger: a Runner reconnects mid-Turn after Agent deactivation or user suspension. Suspension is reversible, so the Sandbox stays live and unsealable.
- The comment "accepted work can never execute" is not true for a started Turn, and no test covers this branch.
- Suggested fix: for `started`, request cancellation, as the IM path does, and let the real settle terminalize. Alternatively, treat an already-terminal record for the same turn as `already_recorded`.

**2. `#writeDurableAccepted` accepts any existing record as current custody.**
`:576-578`
- Scenario:
  1. The receipt is written durably, then the Runner disconnects before `verified`.
  2. The Instance is replaced, and the retry's `reconcileSessionWork` retires the record as `failed/allocation_retired`, keeping the old turnId and allocation.
  3. The retry reaches the new Runner, which journals a new turnId. The existing terminal record is returned, and `#ensureAcceptedCustody` reports `accepted`.
  4. A grant is minted and the Turn runs.
  5. On settle, `envelope.turnId !== frame.turnId`, so `handleSettled` returns silently (`:964-972`).
- The result is an executed Turn with no terminal record and no ack, and the same stuck seal.
- Only reuse an existing record that is non-terminal and has the same turn and allocation. Otherwise write a fresh one.

**3. Retiring an allocation never clears the in-memory tracker.**
`reconcileSessionWork` (`:1054-1098`)
- It terminalizes durable rows but never calls `#work.settle`, and `CloudSessionWorkTracker.clearSandbox` has no callers.
- `sandboxId` survives environment generations, so after a message is accepted on generation N and that Instance is lost, `isBusy(sandboxId)` stays true for the life of the process.
- As a result, `#claimIdle` always refuses and `#siblingEligible` (`sandbox-runner-service.ts:2029`) excludes the Sandbox, which leaks cost.
- A late settle from the old Turn does not help, because it returns on the allocation mismatch before `#work.settle`.
- The tests at `:592` and `:723` only assert `hasUnsettledSessionWork`, not `isSandboxBusy`.

### Should fix

**4. An idle claim can race a Session dispatch.**
`cloud-session-collaboration-owner.ts:411-454` and `sandbox-runner-service.ts:1896-1906`
- `#claimIdle` checks `sessionWorkBusy` once, synchronously, under the row lock. It then awaits `sessionWorkBarrier` and `#hasUnsettledWork` before the update commits.
- Meanwhile the dispatch reads the Sandbox row without a lock in `#ensureReady`, then registers, sends and writes durable `accepted`. It never re-checks `idleReclaimAt`.
- If the accepted row commits after the barrier query, the claim still wins. The seal then cancels the verified entry, and the source was told `accepted`, so dedup blocks any retry.
- The window is narrow. The IM path closes it with `beginDeliveryDispatch` under the row lock plus `#assertEnvironmentStillOwned` (`cloud-delivery-owner.ts:269-313`), and the same pattern should apply here. The comment at `:427` claims this guarantee, but the code does not provide it.

**5. A report-only connection gets no answer for an accepted `received` entry on a releasing or stopped allocation.**
`:919-926`
- The IM equivalent (`handleInactiveDeliveryReceived`) sends a cancellation so the Runner can settle `cancelled/not_started`.
- Here the stop's cancel is one-shot (`#requestCancellation`, `:1214-1228`, which returns `no_connection` when offline). A Runner that reconnects report-only gets silence every 5s, and the release drain times out.

**6. The 24h retention prune can delete accepted work that is still running.**
`runtime-durable-work-store.ts:386-389`
- `#prune` drops rows of any status with `updatedAt < now - 24h`. The Cloud accepted record's `updatedAt` is never refreshed after receipt.
- A Turn close to `RUNTIME_MAX_DURATION_MS` (24h) plus FIFO wait, or one paused on IM re-authorization, loses its barrier. At settle there is no record, so no ack, and the same stuck entry follows.

### Nits
- `#convergeAllocation` (`:712-715`) never clears its 20s race timer when convergence wins.
- `requestId` is always `messageId`, so the superseded-receipt branch at `cloud-turns.ts:435` can never fire. A retry carrying a re-resolved runtime snapshot while a stale `received` entry is journaled gets no receipt and times out every 30s until reconnect.
- The Session exec backstop is `budget + 5s` from worker start, but the worker's own budget starts after `openProxyBridge`. A startup slower than 5s records a real timeout as `unknown`. IM avoids this with an absolute `deadlineAt`.
- On Cloud, the `failed` and `permission_denied` copy in `context-tree-settings.tsx:136-169` still gives Local advice. `permission_denied` there usually means "not the Agent's `context_tree` binding", which is fixed in Integrations.
- Rollout note: a Runner on this head always sends `sessionCollaborationVersion`, so rolling the Server back while these Runners are alive rejects their auth and strands journaled IM custody too. That deserves a line next to "deploy as a matching release".

For each blocker, a test that asserts the Runner entry is eventually acked or cleared, and that `isSandboxBusy` returns false, would lock this down.

yuezengwu:
Addressed the review and merged current `main` into this branch. The initialization conflict preserves both Cloud Session CLI authority and Agent Skills runtime setup.

1. Stopped/suspended/releasing reannouncements now request cancellation and retain custody until the Runner's actual terminal result is committed and acknowledged. A started Turn no longer gets a fabricated `session_closed` terminal result.
2. Existing custody is reused only for the same nonterminal Turn on the same allocation. Terminal records, including `allocation_retired`, are never revived: allocation loss cannot prove the Turn did not already execute. Only a superseded, never-started entry on the same allocation can be replaced, using compare-and-set against the exact previously read record under database locks.
3. Busy tracking is scoped to the existing allocation identity. Retirement clears only the exact retired Turn; an old registration cannot pin a replacement created through IM ingress or exclude it from sibling reuse.
4. Session dispatch registers its pending work before sending and rechecks the unclaimed allocation under the existing Sandbox row lock before granting execution. Deterministic PostgreSQL tests exercise the claim/dispatch interleaving.
5. Accepted received entries reannounced on report-only retiring connections receive cancellation, allowing a truthful not-started settlement and acknowledgement instead of an endless drain.
6. Age pruning now applies only to terminal records. Active custody remains subject to the existing record/payload quotas. Terminal evidence remains bounded by the existing age/count policy.
7. The losing allocation-convergence timer is cleared.
8. Each dispatch gets its own request correlation ID while retaining the logical message ID. A received entry can be re-correlated or atomically replaced for a changed runtime snapshot; started/reported work is never replayed. Late receipts cannot resolve a newer pending request.
9. Session execution uses one absolute deadline across setup, worker timeout and the parent backstop; queue wait is excluded.
10. Cloud Context Tree errors now point to GitHub Integrations/Agent binding, with English and Chinese coverage. Local guidance is retained for Local.
11. Both rollout documents describe Server-first compatibility and a matched rollback: stop admission, drain/settle/save with the compatible Server, verify old Instances removed, then switch versions.

Independent review additionally covered concurrent custody replacement, atomic journal replacement failures, and transient database/runtime-assembly/outcome-repair errors. These failures preserve Runner evidence for reconnect rather than masquerading as missing or rejected custody.

The existing bounded terminal retention still limits acknowledgement replay: if an acknowledgement never arrives before its terminal record is pruned, the Server refuses to acknowledge unverifiable work and the journal can block draining until that allocation is removed. This revision protects active custody; it does not introduce unbounded terminal history or another acknowledgement table.

Validation for `a3b9a95b2690c33535dbb47de4fd9c00f0948fa2`: all seven repository-required commands passed on unchanged clean source; Agent Runtime 456 tests at 100% coverage; real PostgreSQL 42 files / 547 tests. Fault-injection regressions also confirmed the failed-send/timeout unhandled rejection before repair and its absence afterward. The merged Skills composition test now uses the existing isolated temporary-directory helper and passes with TMPDIR unset. GitHub checks are attached to this exact pushed commit.

Fresh native GCP verification has now completed on the same exact head and image: **23 checks passed** with real Cloud Run/GCS, installed Pi and the Session CLI. These cover collaboration/callback settlement, active-child protection during parent stop, transfer of the same Instance with unchanged configuration, workspace isolation, absence of platform credential files in archives, and cold recovery of Pi history/current Agent configuration. All temporary Instances, GCS objects/generations and local fixtures were removed. The retained Runner image is an exact-head review artifact.

The controlled acceptance uses a disposable local Server/PostgreSQL, a deterministic SSE model and synthetic IM ingress; real GitHub App publication and Slack/Feishu delivery are not claimed. Two earlier fixture failures are preserved in the local report: the first assumed an idle candidate survived a slow parent deletion; the second incorrectly injected IM work into an internal Session and hit the intended authority denial. The successful run instead uses real Session CLI traffic and preserves the default idle policy. No product source change or policy relaxation was needed for these fixture corrections.

Exact-head technical GitHub CI is green, including **95.94% patch coverage**. The remaining ownership gate requires bestony's approval, and this reply does not resolve or dismiss the requested human review. No merge or staging application deployment was performed.

liuchao-001 (changes requested):
Re-reviewed at head `a3b9a95b`, diffing against `37f5a0a1`. The Agent Skills files come from the `main` merge, so I treated them as out of scope.

All eleven claimed fixes are real, and each is backed by a test that fails for the right reason:

1. A stopped, suspended or releasing re-announcement now cancels and retains custody until the Runner's real terminal result commits (`:1101-1125`). The fabricated `session_closed` terminal is gone. Covered by "cancels a re-announced started Turn on a stopped Session and acks its real settlement".
2. `custodyDecision` (`:1647-1667`) reuses only the same non-terminal Turn on the same allocation, refuses every terminal record including `allocation_retired`, and replaces only a superseded never-started entry. The compare-and-set in `replaceSessionMessageRecord` re-selects `FOR UPDATE` and compares all fields.
3. Busy tracking is allocation-scoped (`#byAllocation`, `:206-244`), and retirement clears only the exact Turn.
4. Dispatch registers before sending and re-checks under the Sandbox row lock (`#allocationStillDispatchable`, `:873-895`), with a deterministic PostgreSQL interleaving test.
5. Report-only retiring connections now get a cancellation.
6. Age pruning is restricted to terminal statuses (`runtime-durable-work-store.ts:495-511`).
7. The convergence timer is cleared in a `finally`.
8. Per-dispatch `requestId` with `messageId` kept as the logical key. `computeCloudSessionInputHash` strips `requestId`, journal replacement goes through the same temp/fsync/rename path, and both replacement paths refuse any phase other than `received`, so started or reported evidence is never rewritten.
9. `deadlineAt` is anchored after the FIFO drain and threaded into both the worker document and the parent backstop, so the worker's `turn_timeout` always fires first and queue wait is excluded.
10. Cloud error copy now points at Integrations, with en and zh keys present and component tests for both Cloud and Local.
11. Both rollout documents describe Server-first deployment and the matched drain-then-rollback sequence.

The `void promise.catch()` in `a3b9a95b` also closes the unhandled-rejection gap.

### One remaining blocker

**The custody preflight fails open on a durable read error, so live custody can be overwritten.**
`cloud-session-collaboration-owner.ts:701`

```ts
const existing = await store.read(connection.computerId, "session-message", key).catch(() => undefined);
```

A transient database error becomes indistinguishable from "no record". `custodyDecision(undefined, …)` then returns `"write"` (`:1650`), and the blind `store.write` at `:726` is not refused by the store: `sameRecord` is false, the fresh `updatedAt` is newer, and `accepted → accepted` is an allowed transition, so the UPDATE lands.

Failure scenario:
1. Turn T1 is accepted and running on allocation A.
2. A retry's preflight read blips.
3. The row is overwritten with T2 on allocation B, and T2 is granted execution — the same message now executes twice.
4. T1's later `session:message:settled` fails the turn and allocation match in `handleSettled` (`:1222-1231`) and is dropped silently, so it is never acked and replays forever. That is the stuck drain this revision set out to remove.

This contradicts the PR's own "preserve custody through database failures", and `#readDurable` already applies the correct rule a few hundred lines below: "a transient failure must never look like proven absence" (`:1355-1360`). The fix is small — distinguish a throw from `undefined` and treat the throw as unavailable, so the dispatch fails closed and stays retryable.

### Non-blocking

- `trackedMessages` (`:297-307`) de-dupes by `messageId` across allocations and keeps the first entry, so an explicit stop can revoke a dead Turn's grant and leave the live Turn's grant valid until expiry. Reachable only if a stale registration survives.
- The re-announcement path (`:1085-1145`) matches only `turnId`, while `handleSettled` matches all three allocation fields. Practically unreachable given UUIDs, but the asymmetry is free to close.
- With the fabricated terminal removed, nothing bounds custody if a Runner never settles: `reconcileSessionWork` only retires records whose allocation differs from the sandbox row, so a stranded settlement becomes an unreclaimable billed environment. The correctness trade is right; an operator or reaper path would complete it.
- `replaceSessionMessageRecord` (`runtime-durable-work-store.ts:310-318`) bypasses `isAllowedTransition` and would write `accepted` over `running`. Benign today; worth asserting rather than relying on.
- `#allocationStillDispatchable` omits `currentResourceUid`, which `#reclaimableRow` does check, so a same-name new-uid instance would pass the dispatch re-check.
- Redispatch of a `reported` entry replies with `entry.requestId` rather than `frame.requestId` (`cloud-turns.ts:401`, `:782-789`), so the new attempt's pending can only time out — 30s and a `status: unknown` per retry, even though the Runner answered definitively.
- `#pending` is cleaned only when `sendToCurrent` returns `false` (`:523-534`); a throwing socket leaves the entry and timer until the timeout. Harmless after the `a3b9a95b` fix, but the new test asserts only that no unhandled rejection occurs.
- `CLOUD_JOURNAL_VERSION` stays `2` although the `inputHash` meaning changed. No shipped build can hold Session entries today, so this is only a latent trap.

Fix the preflight and I'll approve. Note the ownership gate still needs bestony.

liuchao-001:
This PR merged before my re-review landed, so that review never gated anything. The one blocker in it is still present in `main` (`1a2f3dd2`), tracked now as #682. The non-blocking items stay listed in the review above.

PR #662, opened 2026-09-20
feat(skill): add a revision-conflict code, bundle filenames, and deferred orphan-object GC

bestony:
> Now targets `main`. This PR was stacked on #658 while that was open; #658 is squash-merged (`0bb4b0dd`), and head `f763fb19` merges `main`. The merged tree is byte-identical to the pre-merge head, and the diff against `main` is exactly this PR's own files (32 since the review round at `4d7faa8e`).

## Summary

Three follow-ups that reviewers agreed should land after the main Agent Skills PRs (#655, #657, #656, #658). They are one PR because the first must be atomic: `apps/web` maps Skill error codes with a `Record<SkillErrorCode, …>`, so adding a code to the shared contract breaks the web typecheck unless the web change lands with it.

### 1. A dedicated `SKILL_REVISION_CONFLICT` code
The three "changed concurrently" paths (`setEnabled`, `remove`, a replace built on a stale read) returned `SKILL_NAME_CONFLICT`, so the web showed "a Skill with this name already exists" for what is a revision race — raised by @liuchao-001 on #657.
- **shared** — `SKILL_ERROR_CODES.REVISION_CONFLICT` (`deterministic`, 409). `ErrorCodeSchema` spreads the Skill codes, so the envelope admits it; no new export name, snapshot unchanged.
- **server** — `skillRevisionConflict()`; `SKILL_NAME_CONFLICT` now means only a genuine name clash (insert unique violation, upload without replace).
- **web** — new message in `en` + `zh`; a `SKILL_REVISION_CONFLICT` from an upload does **not** open the Replace dialog (only `NAME_CONFLICT` does) — it shows the message and refreshes the list.

### 2. `Content-Disposition` on every bundle response
Bundles were served as `application/octet-stream` with no filename, so curl, the CLI or any non-browser client saved an extensionless file — the server half of @yuezengwu's finding on #658. All three surfaces (account, computer, runtime) now send `attachment; filename="<name>.tar.gz"`, built through one guard that refuses anything outside `[a-z0-9-]` and falls back to `skill.tar.gz`, so a future loosening of the name rule cannot become header injection.

### 3. Deferred orphan-object GC; a replace no longer deletes inline
A replace used to write the new key, update the row, then delete the old key inline. That delete is check-then-act, so a narrow interleaving could still strand a row on a deleted object — the residual recorded and agreed for follow-up on #657.
- `SkillObjectStore.list()` + an S3 `ListObjectsV2` adapter (bounded XML extraction, no XML dependency; path-style or virtual-hosted).
- `SkillObjectGc` (timer, no overlapping runs, errors logged never thrown). An object is deletable only if it is **past the grace period**, its key is **exactly `<normalized-prefix>/accounts/<uuid>/agents/<uuid>/skills/<uuid>/<sha256>.tar.gz`** (nothing else in the bucket is ever touched — including another deployment's objects under a nested or sibling prefix), and **no row references it** — checked in batches and **re-checked for that single key immediately before the delete**. Deletes are capped per run.
- `#replaceSkill` no longer deletes the old object; `ensureObjectPresent` stays. `remove()` keeps its inline delete (a user who deletes a Skill expects the bytes gone now, and nothing can reference a key under that skill id afterwards).
- Config: `OPENTAG_SKILL_STORAGE_GC_INTERVAL_SECONDS` (default 3600, `0` disables) and `OPENTAG_SKILL_STORAGE_GC_GRACE_SECONDS` (default 86400, min 300). Documented in `.env.example`, `docs/deploying.md` and its zh-CN mirror, and the design doc.

No migration, no new dependency, `pnpm-lock.yaml` untouched.

## Checklist

- [x] 0. Base merged; check + typecheck green before any change
- [x] 1. F1 `SKILL_REVISION_CONFLICT` — shared + web, then server
- [x] 2. F3 `Content-Disposition` on all three bundle surfaces
- [x] 3. F2 store `list`, `SkillObjectGc`, replace no longer deletes inline, config, wiring
- [x] 4. Docs (design doc, deploying en + zh-CN, `.env.example`)
- [x] 5. All acceptance criteria verified, `git status` clean

## Acceptance criteria — re-run by the orchestrator

| # | Criterion | Result |
| --- | --- | --- |
| 1 | `pnpm check` (no migration added) | exit 0 — migration drift: 49 |
| 2 | `pnpm typecheck` (web compiles with the new code) | exit 0 |
| 3 | shared / server unit / server integration / web tests | 476 / 2394 / 534 / 1162 tests, all green |
| 4 | `pnpm build`, Browser Smoke | exit 0, 13/13 |
| 5 | skill suite under v8 coverage; no slow tests | green; slowest test < 100 ms |
| 6 | "changed concurrently" only via `skillRevisionConflict` | holds |
| 7 | no inline `deleteReplacedObject` left | 0 references |
| 8 | `en.json` / `zh.json` key parity | identical |
| 9 | no new file > 500 lines, no new dependency, lockfile unchanged | holds (`config.ts`, `index.ts`, `config.test.ts` were already over 500 on `main`) |
| 10 | doc-mirror check | passes (28 pairs) |

**Probes written by the orchestrator, independent of the lane's tests** (a temporary test file, run and deleted, not part of the branch):
- *GC safety matrix.* After a real upload and a real replace (the store's delete counter did not move): an old unreferenced skill-shaped key → deleted; a young orphan → kept; the **live object, made older than the grace period** → kept because a row references it; the just-replaced object while still young → kept; `skills/README.txt` and a skill-path with a non-UUID id → never touched; a skill-shaped key outside the prefix → never listed. Moving the replaced object past the grace period → collected on the next run, live object intact, `openBundle` still works.
- *Codes.* A stale `setEnabled`, a stale `remove` and a stale replace each reject with `SKILL_REVISION_CONFLICT` / 409 (another writer bumps the revision between the read and the write); a second upload of the same name without replace still rejects with `SKILL_NAME_CONFLICT` / 409; the Skill survives all three losing writers and stays readable.
- *Header.* `skillBundleDisposition("pdf-tools")` is exactly `attachment; filename="pdf-tools.tar.gz"`; a quote-injection string, a CRLF + `Set-Cookie` string, `../../etc/passwd`, an uppercase name, the empty string and a 65-char name all yield `skill.tar.gz`.

Honest residual, stated in the design doc: with no inline delete, the only remaining window is a GC run deleting a >24 h-old unreferenced object in the milliseconds between its single-key re-check and its delete, while a re-upload of that exact content commits. `ensureObjectPresent` on the writer's side covers most of even that; conditional deletes are not portable across S3-compatible stores, so it is documented rather than engineered further.

**What was run where.** Everything in the table was run by the orchestrator on `ab503264`. Two merge-only heads followed:
- `91d162c7` merges #658's review-round fixes (tri-state storage, cache reconciliation). Two files conflicted (`skills-page.tsx`, `skills-queries.ts`) and were resolved by hand keeping both sides; against that base, `skills-queries.ts` differs only by the added `useInvalidateAgentSkills` export and `skills-page.tsx` only by the revision-conflict handling. Re-run on `91d162c7`: check, typecheck, shared 476, web 1162, server unit 2394, build, Browser Smoke 13/13.
- `f763fb19` merges `main` after the #658 squash. Six web files conflicted only because the squash made that history foreign; `main`'s `apps/web` + `e2e` are byte-identical to the base merged in `91d162c7`, so this branch's versions were kept, and `git diff 91d162c7 f763fb19` is **empty**. Server integration (534) was run by the orchestrator on `ab503264` and by the agent on `f763fb19`; `packages/server` and `packages/shared` are unchanged between those two heads.

## Review round 1 (`4d7faa8e`)

@yuezengwu found two GC boundary defects, both fixed: the matcher validated only the key's tail, so a `skills` collector could delete a live bundle of a deployment at `skills/staging` (now an exact-prefix match through one shared `normalizeSkillObjectPrefix`, with the prefix-agnostic matcher removed); and the collector listed the raw configured prefix, so `skills/` made it scan nothing (now normalized once, and at config parse time, where an invalid prefix fails startup). Details, tests and the old-vs-new probe are in the PR comment for that head.

## Provenance

Written by an agent under `dispatch-opencode` run `tlm1q4`, lane `skill-followups-5`. Plan and acceptance criteria authored, and the result verified, by the orchestrator (Claude Code) before push. 8 own commits for a 6-item checklist.


packages/server/src/services/skills/skill-object-gc.ts
```diff
@@ -0,0 +1,200 @@
...
+    return summary;
+  }
+
+  /** Applies the three eligibility rules, counting the objects that are not candidates. */
+  #candidates(objects: SkillObjectListEntry[], summary: SkillObjectGcSummary): SkillObjectListEntry[] {
+    const cutoff = this.#now().getTime() - this.#graceMs;
+    const candidates: SkillObjectListEntry[] = [];
+    for (const object of objects) {
+      summary.scanned += 1;
+      if (!isSkillObjectKey(object.key)) {
```
yuezengwu: [P1] Restrict candidates to this deployment's exact normalized prefix. With two deployments sharing a bucket but using separate databases and prefixes `skills` and `skills/staging`, listing `skills/` includes the second deployment's objects. `isSkillObjectKey` validates only the tail, so those keys pass; neither reference query can see the other database, and an object older than the grace period is deleted even while its Skill remains live. I reproduced this with two migrated PGlite databases, a shared in-memory object store, and the real SkillService: the staging bundle was readable before the sweep, then GC reported `deleted: 1`, its row still referenced the key, and `openBundle` returned `SKILL_NOT_FOUND`. Require the exact `<normalized-prefix>/accounts/...` shape (with no intervening namespace) before any deletion and add this isolation regression.

packages/server/src/services/skills/skill-object-gc.ts
```diff
@@ -0,0 +1,200 @@
...
+      skippedForeign: 0,
+      durationMs: 0,
+    };
+    const startedAt = Date.now();
+    if (this.#running) return summary;
+    this.#running = true;
+    try {
+      let cursor: string | undefined;
+      do {
+        const page = await this.#store.list(`${this.#prefix}/`, {
```
yuezengwu, line 123: [P2] Normalize the configured prefix with the same rule as `skillObjectKey` before building the list prefix. `OPENTAG_SKILL_STORAGE_PREFIX=skills/` is accepted, and successful uploads store keys under `skills/accounts/...` because the writer removes empty slash segments. This call instead lists `skills//`, so GC never sees those objects. A real upload followed by a replace and an aged orphan reproduced `scanned: 0, deleted: 0`; replacement now leaves old objects for this worker, so the accepted configuration leaks every replaced bundle indefinitely. Share the canonical prefix handling and cover trailing, leading, and repeated slashes.

yuezengwu (changes requested):
Reviewed exact head `f763fb1921222e1cfaa699b32a8ffd2ad8d5ccb9` against current main `0bb4b0dd0e90d8dd0d79e9f18e1ab4d5f12da16a`. Two reproducible GC boundary defects need correction: a collector can delete a live bundle belonging to a nested deployment prefix, and slash-normalized writer prefixes can make GC miss every orphan. Details and reproduction outcomes are inline.

Validation: 169 directly affected existing tests passed (server 88, shared 18, web 63); three independent GC probes produced two failures and one passing own-namespace control. `pnpm build`, `pnpm typecheck`, `pnpm check`, and `git diff --check` passed. The probes use real services over migrated PGlite databases and an in-memory object store; no live S3 or deployment acceptance is claimed. Temporary probes were removed and the isolated checkout is clean.

bestony:
@yuezengwu thank you — both findings are right, and the P1 is a genuine data-loss path. It traces to a decision in the plan rather than to an implementation slip: the matcher was specified to validate the key's tail "without knowing the prefix", on the reasoning that this protects unrelated objects in the bucket. It does that, and it is exactly what lets a shallower prefix's collector claim a nested deployment's keys. My own GC safety matrix on this PR never included a second deployment, so it could not have caught it. Fixed at exact head `4d7faa8e`.

**P1 — collector deletes a live bundle under a nested prefix → `d28acc27`.**
- One dependency-free normalizer, `normalizeSkillObjectPrefix` (`skill-object-prefix.ts`): the rule `skillObjectKey` already applied (split on `/`, drop empty segments, validate each), now the only implementation — the writer, the collector and config all call it. It additionally rejects bare `.` / `..` segments, which the old segment pattern let through.
- The prefix-agnostic `isSkillObjectKey` is **removed** (barrel included) rather than left for someone to misuse, and its doc comment, which argued for the unsafe behaviour, is rewritten. Its replacement `isSkillObjectKeyUnder(normalizedPrefix, key)` is true only when the key starts with `<normalizedPrefix>/` and the remainder is exactly the 7 segments `accounts/<uuid>/agents/<uuid>/skills/<uuid>/<sha256>.tar.gz` — no intervening namespace.
- `SkillObjectGc` normalizes its prefix once in the constructor. A key that fails the bound matcher is counted `skippedForeign`: never deleted and never reference-checked.

**P2 — slash-normalized writer prefix makes GC miss every orphan → `623aa0b7`.** The collector lists `<normalizedPrefix>/`, the same value the writer uses. And config no longer accepts what the normalizer rejects: `OPENTAG_SKILL_STORAGE_PREFIX` is normalized at parse time, so `skills/`, `/skills`, `//skills//` all resolve to one form for every consumer, while `/` or a traversal segment now fails startup with a message naming the variable instead of failing at the first upload.

Docs (`4d7faa8e`): the design doc's GC section, `docs/deploying.md` and its zh-CN mirror, and `.env.example` now state the exact-prefix binding — deployments may share a bucket, nested prefixes included — and that the prefix is normalized identically for writes and for collection.

**Tests, and proof they bite.** The lane added: your two-database scenario end to end (shared store, real `SkillService` on each, `skills` vs `skills/staging`) in both directions, a sibling prefix (`skills-staging`), the own-namespace control, matcher and normalizer unit tests, config tests, and — for each of `skills/`, `/skills`, `//skills//`, `skills//nested/` — a real upload, a real replace, an aged orphan, then a GC run.

Independently of those, I wrote a behaviour-only probe (public `SkillService` / `SkillObjectGc` API, two migrated PGlite databases, one shared store; run and deleted, not part of the branch) and ran it against **both** sources:
- pre-fix (`f763fb19`): a live bundle under `skills/staging` — and one three levels deep under `skills/eu/prod` — is deleted by the `skills` collector (`deleted: 1`); with prefix `skills/`, upload → replace → aged orphan gives `scanned: 0, deleted: 0`. Both are your reported outcomes.
- this head: `deleted: 0` and `openBundle` still returns the bundle for both nested prefixes; with `skills/` the run reports `scanned: 2, deleted: 1`, the orphan is gone, the live object is kept and readable.

Re-run by me on `4d7faa8e`: `pnpm check` 0 · `pnpm typecheck` 0 · server integration 534/534 · `pnpm build` 0 · every skill test file under v8 coverage: 93 tests, none ≥ 500 ms. Server unit: 2409 of 2410 on the full run — the one failure was `computer-connect-command.test.ts > still connects for 'staging' when rm is unavailable`, which took 6.2 s against the 5 s timeout while the whole suite was running; its file is byte-identical to `main`, and it passes alone (21/21). I am reporting it rather than rounding it away; it is load-sensitive on my host and not touched by this PR. Web and shared are unchanged since the previously verified head. No live S3 acceptance is claimed.

The PR diff is now 32 files (the previous 30 plus `skill-object-prefix.ts` and its test); nothing outside `packages/server/`, `docs/` and `.env.example` changed in this round.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-followups-5`._

yuezengwu (commented):
Both earlier GC findings are verified fixed on exact head `4d7faa8e76721331e237ff047bb7f35f9e5ab5ca`: the original three independent probes pass, along with 185 affected existing tests (188 total), build, typecheck, and check. Current technical CI has also passed.

The merge of #659 has since moved `main` to `ab9fd76232cf23b22cc4f57f1e05baf39a857c46`. GitHub now reports this PR as `CONFLICTING`, and an independent `git merge-tree` check reproduces a content conflict in `packages/server/src/index.ts`, around `SessionCliProofService` and `createSkillRuntime` initialization. Please sync with current main and resolve it preserving both main's `sessionAuthority.proof` dependency and this PR's `skillRuntime.service` / GC lifecycle wiring, then rerun validation on the resulting head. I am holding approval until that integration can be reviewed; the completed tests above were against the previous base `0bb4b0dd` and do not establish compatibility with the newly merged Cloud changes.

bestony:
New exact head `c0d9108c` — **a merge of `main` only; no Skill source changed.**

`main` moved to `ab9fd762` (#659, cloud Context Tree + Session collaboration), which made this PR CONFLICTING in exactly one file, `packages/server/src/index.ts`: #659 reworked `startServer`, the same composition function where this PR starts and stops the orphan-object collector. This was a real content conflict, so it was resolved by hand from `main`'s version of the file with this PR's wiring re-applied — not by taking either side.

Verified:
- `git diff origin/main -- packages/server/src/index.ts` is exactly this PR's Skill GC wiring (the `SkillObjectGc` import, `createSkillRuntime` returning `{ service, gc? }`, `skillRuntime.service`, `gc?.start()` / `gc?.stop()`), line for line the same as before the merge; it removes nothing #659 added;
- the other 31 files of this PR are byte-identical to the previous head `4d7faa8e`;
- the tree differs from `main` in exactly this PR's 32 files; no conflict markers; #659 added no migration.

Re-run by me on `c0d9108c`: `pnpm install --frozen-lockfile` 0 · `pnpm check` 0 · `pnpm typecheck` 0 · shared 482 · web 1168 · server unit 2468/2468 · server integration 547 · `pnpm build` 0 · Browser Smoke 13/13. (The load-sensitive `computer-connect-command` timeout I reported on the previous head did not recur in this full run.)

@yuezengwu the fixes for your two GC findings are unchanged from `4d7faa8e` — my previous comment describes them. Requesting a fresh exact-head review on this head.

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-followups-5`._

bestony:
@yuezengwu thank you for confirming both GC fixes. Your 07:18Z note and my push of `c0d9108c` crossed — that head is the sync with `main@ab9fd762` you asked for. To the specific point you raised: in the merged `packages/server/src/index.ts`, `SessionCliProofService` is still constructed with main's `sessionAuthority.proof` (from `cloudSessionAuthorityOptions(cloudRuntimeFence, …)`, with `...sessionAuthority.session` spread into the session service exactly as on `main`), and directly after it `createSkillRuntime(…)` yields `skillRuntime`, consumed as `skills: { service: skillRuntime.service, proofs: sessionCliProofService }` with `skillRuntime.gc?.start()` / `gc?.stop()` in the lifecycle. `git diff origin/main -- packages/server/src/index.ts` contains no removed #659 line. Validation on the merged head is in my previous comment (unit 2468, integration 547, against the new Cloud code).

_Posted by the dispatch orchestrator (Claude Code) on behalf of @bestony; run `tlm1q4`, lane `skill-followups-5`._

yuezengwu (approved):
Approved exact head `c0d9108ce45c10985dbf58a8bb87f5e3882fe67b` against current main `ab9fd76232cf23b22cc4f57f1e05baf39a857c46`. The merge resolution preserves main's Session authority/proof wiring and this PR's Skill service/GC lifecycle, including Cloud cleanup. Both previous GC findings are fixed on this head; the unchanged original isolation, slash-prefix, and own-namespace probes all pass.

Fresh local validation: 201 tests passed (198 affected existing tests plus 3 independent probes), including server startup, Session CLI proof, and Cloud runtime composition. Frozen offline install, full build, typecheck, check, and diff check passed. Current technical CI is green. Review scope excludes live S3 and deployment acceptance.

PR #663, opened 2026-09-20
fix(cloud): publish and deploy Runner with CLI releases

yuezengwu:
## Summary

Cloud Runner was built by an offline toolchain check but was not published or activated with the CLI, leaving staging on an older Runner after Server deployment. Publish a digest-pinned Runner from the same clean CLI release source, then activate it after the compatible Server is ready. Staging follows the existing automatic release chain; production activation and compatible Runner-only rollback use one manual workflow.

Previously verified ready Instances can reconnect across a target change while retaining ownership, workspace restoration and execution-policy checks. Provider generation plus the digest-pinned original image prevents treating an in-place image replacement as the same admitted build. Initial admission and cross-Session reuse remain strict. No database changes, new service, global drain or in-place updater.

## Validation

- Formatting, lint, ownership/complexity gates, full build and typecheck.
- Full PostgreSQL integration: 555 tests passed; client runtime coverage: 100%.
- Real WebSocket recovery after one transient provider failure; upgrade/rollback, UID replacement, changed image and cross-Session reuse tests.
- Final helpers verified against live read-only GAR and CapRover APIs. The current Server has no new revision header yet, so the deployment check correctly stops at that readiness gate.
- Full workspace tests passed: 474 script tests plus all package suites. Release helper tests were also rerun with a failing `gcloud` shim to prove they do not depend on operator credentials. Clean-source linux/amd64 Runner built and passed offline startup, toolchain, privilege and process-lifecycle smoke under 1 CPU / 1 GiB. This was Docker emulation on an arm64 host, not native Cloud Run acceptance.

Technical GitHub CI passed at `85159fdc4e6368b565d3dbf12256431f6885876e`. The ownership gate still requires the designated reviewer. An additional read-only helper review is still running and is not counted as a passing review.

## Before merge / deployment

Shared staging IAM and configuration have not been changed. Configure the documented staging image repository, existing Secret Manager credential reference and repository-scoped permissions before enabling publication. Use the existing publisher for image writes and the exact deployment Environment identity for repository/secret reads; the publisher never receives administrator-secret access. Restrict the deployment Environment to main.

The current manually published Runner lacks its CLI version tag. Preserve the first rollback baseline by verifying its npm source/image identity and adding only that missing tag to the existing digest, without rebuilding or overwriting. Production needs its own protected Environment and coordinates before its next formal release. See `docs/cloud-runner-release.md` for rollout and rollback boundaries.


liuchao-001:
Post-merge review (requested at 10:39Z; the PR merged 24s later, so this lands after the fact). I reviewed the merged tree, then checked every finding against current `main` — `#664`/`#675` already fixed some of what I found, and those are marked below rather than re-reported.

The server-side admission model is the strongest part of this change and I could not break it. Details at the bottom.

---

## P1 — the staging release chain is wedged right now, and cannot self-recover

Not a latent risk: **four consecutive releases have failed and the chain is still dead on current `main`.**

| when | commit | npm Publish |
|---|---|---|
| 09-20 12:18 | `74054748` | success (last good release) |
| 09-20 23:25 | `618b093e` | **failure** ← poisoning run |
| 09-20 23:48 | `168ef022` | failure |
| 09-21 02:23 | `419c9aac` | failure |
| 09-21 03:31 | `f5a8464b` | failure |
| 09-21 03:52 | `1a2f3dd2` | failure |

Every one of the last four fails identically, in ~2.4s:

```
[runner-release] published image label org.opencontainers.image.revision
  is "618b093e0ab65a78d1e01e4ebc4b581f8ac166d6", expected "<that run's own SHA>"
```

**Mechanism.** `resolve-staging-release.mjs` derives the version from *published npm state*: `sequence` is `findLatestStagingVersion(...) + 1`. It is not a function of the commit. So all four commits compute the **same** `RELEASE_VERSION: 0.0.6-staging.32.1` — I confirmed this in all four run logs.

The run at `618b093e` pushed the Runner image to GAR tag `0.0.6-staging.32.1` (`release.mjs:128`), then failed at a *later* step, so npm never published. Now:

1. `lookupRunnerTag` finds `…32.1` present → `existing.present` is true (`release.mjs:121`)
2. `verifyRunnerIdentity` compares the image's `org.opencontainers.image.revision` (`618b093e`) against the current `--source-sha` → mismatch → hard fail
3. npm never publishes → `sequence` never advances → step 1 again, forever

**This is a closed loop with no exit.** Recovery needs a human: delete/retag `0.0.6-staging.32.1` in GAR, or publish a throwaway npm version to advance the sequence.

Worth calling out because the static reasoning here looked airtight and is quoted in the PR description: the Runner image is pushed *before* npm publish, and the argument is that the worst case is a harmless orphan GAR tag, since `resolveRunnerRelease` requires a matching npm `gitHead` and so an orphan can never be deployed. That is true, and it is still not enough — the orphan sits under a version number that will be recomputed *identically* on every subsequent commit, so it stops being an orphan and becomes a permanent roadblock. The `existing.present` reuse path is right for retrying the *same* release (`run_attempt` 2), but the staging version isn't commit-derived, so two different commits legitimately collide whenever a publish fails after the Runner push.

Suggested fix, beyond unblocking: make the reuse path require that the existing tag's revision label match the current `--source-sha` *before* treating it as a reuse — and on mismatch, fail with a message that names the manual recovery step instead of a bare label diff. Better still, publish the Runner image after the npm publish preflight so a failed release cannot leave a claimed version tag. Note `publish-npm-package.yml:149-150` still carries a comment asserting the old ordering invariant ("preflighted before npm publishing so an immutable Cloud Storage conflict fails the run while npm is still unpublished") that this PR inverted.

---

## P2 — production activation and *every rollback* pin a mutable tag, not a digest

`release.mjs:269` guards the digest only when metadata is supplied:

```js
if (metadata && tag.digest !== metadata.digest) { … }
```

`deploy-runner.yml:80-82` — production activation **and all rollbacks** — passes only `--version`. `deploy-staging.yml:128-130` (manual dispatch) passes only `--source-sha`. Neither supplies metadata, so the activated digest is whatever `<image>:<version>` resolves to *at that moment*.

`publish`'s "a version tag is never overwritten" is a check-then-act inside one process, not a registry property. Nothing in the code, the workflows, or `docs/cloud-runner-release.md` requires Artifact Registry immutable tags.

Failure scenario: anyone with `roles/artifactregistry.writer` re-pushes `<repo>:0.0.7` rebuilt from the same commit. All four labels match, so `verifyRunnerIdentity` passes, `resolve` records the *new* digest, and `apply` activates bytes that never went through `runOfflineSmoke`. This is the direct counterexample to the PR's "digest-pinned" claim — and it is on the rollback path, which is exactly where you least want it.

Fix: enable AR immutable tags on both repositories, and/or have `deploy-runner.yml` feed the archived release record via `--metadata`.

Related (`release.mjs:121-135`): publish never ties the recorded digest to the image it just built. `build()` returns `imageId` (`build-image.mjs:99-106`) but `release.mjs:123` keeps only `architecture`, and `dockerTransfer` checks exit status only. The recorded digest is authenticated purely by labels that any same-source rebuild reproduces. `runner-publish.test.mjs:76` (`assert.equal(tagReads, 2, …)`) reads like it covers this but proves only that a second HTTP call happened — the fixture returns the same manifest no matter what was pushed.

---

## P2 — a queued rollback can be silently cancelled during an incident

`deploy-runner.yml` uses `group: deploy-${{ inputs.channel == 'staging' && 'staging' || 'prod' }}`, which collides with `deploy-staging.yml`'s `group: deploy-staging`; both are `cancel-in-progress: false`. GitHub keeps at most **one** pending run per group and cancels the previously pending one.

Scenario: staging deploy A is running; you dispatch a rollback B during an incident (pending); a new main push queues deploy-staging C → **B is cancelled**. It surfaces as "cancelled", not "failed", so it is easy to miss while you are busy. `publish-npm-package.yml:32-33` uses `queue: max` for precisely this reason; the deploy workflows don't.

---

## P3

**`alreadyAtTarget` is computed from a pre-wait snapshot** — `deploy.mjs:331` derives `current` from `initial`, then `346` computes `alreadyAtTarget` from it, after the bounded wait has run. If the target env changed during that window, `apply` skips the update (`358`) and reports `{updated: false, active: true}` while stored CapRover config points elsewhere. One-line fix: derive it from the post-wait read at `359`.

**The "first rollback baseline" recipe in `docs/cloud-runner-release.md:62-65` is insufficient as written.** It says to "add the missing tag to that same digest", but `assertRunnerLabels` (`gar.mjs:177-190`) requires four exact *image config* label equalities. Labels live in the config blob and are immutable with the digest — tagging cannot supply or correct them. The baseline is reachable only if that manual image was already built by `build-image.mjs` with exactly that version, from a clean tree at the commit npm reports as `gitHead`. If it was built with a dev/local version label the baseline is unreachable, and you would find that out mid-incident. Please spell out the four label equalities as the verification step.

**`assertPreserved` can't see dropped fields** — `deploy.mjs:170-189` iterates only `APP_DEFINITION_SAFELIST`, and `caprover.mjs:8` states the update endpoint *replaces* the full app definition. A persisted field outside the safelist (e.g. `hasPersistentData`, present in fixtures at `runner-deploy.test.mjs:61`) is neither sent nor checked. Both fakes apply updates as `{...state.definition, ...body}` — they *merge* where real CapRover replaces — so a dropped field is structurally invisible to the suite.

**Two duplicated implementations of the deployment-proof hash with no shared pin.** `app.ts:196-198` and `deploy.mjs:52-56` independently compute `sha256(JSON.stringify([image, version]))`. `runner-deploy.test.mjs:488` is named *"runnerTargetHash matches the Server readyz proof formula"* but only asserts determinism, sensitivity and hex shape — it never compares against the server's formula. Change one side and both suites stay green while the deploy gate silently stops matching. A shared golden vector would fix it: `sha256(JSON.stringify(["i","v"]))` = `2beeaf907f85961d8bd8dc3befc6bc6fd29ec797097f7aa58bc2a9b631caa314`.

**`Dockerfile:36-37`** — `ARG`/`ENV OPENTAG_BUILD_REVISION` sits immediately after `FROM … AS runtime`, before `WORKDIR`, all four `COPY --from=`, and the `RUN apk add`. Since it changes every commit, the whole runtime stage is rebuilt and re-pushed on every push and every PR build. Moving the pair to just before `EXPOSE` preserves the cache with identical semantics.

---

## Verify out-of-band (I can't read GCP/Environment config)

**The publisher/deployer split is prose, not YAML.** `publish-npm-package.yml:104-108`, `deploy-staging.yml:70-74` and `deploy-runner.yml:68-72` all authenticate through the *same* `OPENTAG_PORTABLE_GCP_WORKLOAD_IDENTITY_PROVIDER`. The separation is that the publisher impersonates a SA while deployers use direct federation and carry `environment:` (so their OIDC token has an `environment` claim). If the CapRover password secret's IAM binds `principalSet://…/attribute.repository/first-tree-ai/opentag` — the most common WIF setup — the npm Publish job qualifies too: it runs unattended on every main push, has no `environment:`, no required reviewers, and already has `gcloud` authenticated. Please confirm the provider's `attribute_condition` and the secret IAM key on `attribute.environment`, not `attribute.repository`. The design is achievable as written; it just isn't demonstrated by the YAML.

**`production` Environment protection.** `deploy-runner.yml:41`'s `github.ref == 'refs/heads/main'` guard lives in the file an attacker would edit — `workflow_dispatch` runs against any ref where the workflow exists. The real gate is the Environment's branch restriction plus required reviewers. `production` is introduced by this PR, and if it didn't already exist GitHub **auto-creates it with no protection rules** on first reference. A false job-level `if` skips the job silently, creating no deployment record, so you can't infer protection from observing a skip. Please confirm both Environments have required reviewers and branch restriction = `main`.

---

## Already fixed — noting only so it isn't re-litigated

The CapRover build race **did** fire in production on this merge commit. `Deploy Staging` (run `35506672423`) failed 2.6s after the async CapRover deploy started:

```
Deploying *** using app token is in progress. Wait a few minutes until your app is built and deployed
[runner-deploy] CapRover reports an ongoing app build; wait for it to finish before changing the Runner target
```

with `DEPLOYED: true, RESULT: failure` — Server deployed, Runner not activated, i.e. precisely the regression this PR set out to fix. `validateState` threw on `isBuilding` before the 300s `waitForRunnerTarget` poll could run. `#664`/`#675` fixed this with `waitForAppIdle` (`deploy.mjs:325-328`), so no action needed — flagging only because the PR description reads the readiness gate stopping during validation as confirmation that the gate works, when it was also the signal that the wait wasn't wired.

---

## What I tried to break and couldn't

The reconnect-across-target-change model is careful work and I want to be specific about what I verified rather than just waving it through.

I expected the `generation !== "1"` check to be the weak point, since the entire anti-in-place-replacement argument rests on one provider field that is only ever exercised against a hand-written `generation: "1"` fixture and a fake. If the real API omitted it, returned a number, or didn't increment, the control would be either vacuous or a silent no-op — and every test would stay green either way.

It holds. Per Google's live v2 discovery document, `GoogleCloudRunV2Instance.generation` exists, is `"type": "string", "format": "int64"`, output-only, and documented as *"A number that monotonically increases every time the user modifies the desired state."* That matches the code's `typeof body.generation === "string"` handling and the int64-as-string comparison exactly. I also confirmed `cloud-run-admin.ts` issues only GET/POST/DELETE — no PATCH — so the "OpenTag never patches Instance configuration" premise that makes generation-1 meaningful is enforced by the code, not just asserted.

Also checked and sound:

- **The 10s bounded read is correctly sized.** It runs inline in the serial frame chain, which the pre-existing design comment warned against, but heartbeat interval is 15s and tolerance 45s (`runner-connection.ts:78-81`), so a worst-case 10s block can't starve a heartbeat into a sweep.
- **`imageRejected` inverts correctly.** Requiring the observed image to *differ* from the current target on the reconnect path means a wrong-version report on the current build stays rejected — the fail-closed direction.
- **Config guarantees digest-pinning.** `cloud-runner-config.ts:16` rejects tags outright, so the `@sha256:` regex in `verifyTrackedOriginalImage` can't be trivially satisfied.
- **`/readyz` exposure is genuinely non-secret** — public repo, so the revision SHA leaks nothing, and the runner target is hashed.
- **The post-read recheck is real.** The scope/UID recheck after the provider read correctly distinguishes stale / deferred / ready, and `cloud-run-admin.test.ts` covers `generation: undefined` and `"2"` rejection against the *real* implementation, not just the fake.

One design question rather than a defect: the reconnect path now accepts an **arbitrarily old** Runner, with no floor. The only constraint is "digest-pinned and differs from the current target", so an Instance ready since many releases ago is admitted indefinitely. Version equality was previously the only mechanical compatibility guard on this channel, and it's now replaced by a *provenance* check — compatibility is asserted in docs ("compatible Runner-only rollback") but not enforced anywhere. Note the Computer/runtime channel next door does have explicit protocol negotiation (`RUNTIME_PROTOCOL_V1`/`V2`); the Runner channel has nothing equivalent. If a future Server changes a Runner frame's shape, an incompatible old Runner is marked ready and handed work. A minimum-supported-Runner-version floor would close it cheaply.

Minor: `fake-cloud-run-admin.ts:159` `replaceImage`'s doc comment says it simulates "an Instance created under a previous deployment target", but because it bumps `generation`, it actually models the *attack* (in-place replacement) and can only produce rejection. The accept case is set up by moving `targetImage` instead. The mechanics are right; the comment could mislead the next person writing a test here.

PR #665, opened 2026-09-20
feat(cloud): complete Cloud Computer onboarding and resource controls

yuezengwu:
## Summary

Cloud Agents can now be created and set up through the existing product flow: Cloud is the default when configured, Pi is fixed, and IM setup uses Server-owned authorization without Local daemon checks. Creation and status reads do not allocate an Instance. Agent/Task pages show environment state and existing save/release controls; task/token usage is explicitly IM-only.

Physical admission reuses Sandbox rows and a short PostgreSQL transaction lock (defaults: 3 Instances per Account, 20 per deployment). Uncertain creates, failed saves and unconfirmed deletes retain capacity; existing work, reuse and cleanup remain available at the limit. IM uses the existing queue, while internal child work keeps the same durable message and child Session retryable after a capacity refusal. No migrations, new tables, queues, resource pool or Runner protocol changes.

Cloud enablement uses two existing switches: `OPENTAG_CLOUD_IDENTITIES_ENABLED` controls identities and Runner together; `OPENTAG_CLOUD_MODEL_ENABLED` independently pauses model requests while preserving save/release. The retired Runner variable only validates upgrades: conflicting identities-on/old-Runner-off and malformed values fail startup; agreeing old `true` is tolerated for rollback. Overall-off dominates model-on, while opted-in model settings are still validated. The browser keeps its gray disabled option using the existing availability response; there is no visibility flag.

Save and discard both require the observed environment generation, checked under the existing Sandbox row lock. The browser keeps controls stable during polling, scopes operation feedback to the requested generation, preserves specific save failures, and retries failed availability/Computer reads. Account-wide capacity views refresh after release. Cloud diagnostics distinguish Local CLI readiness as not applicable.

## Testing

- All six mandatory local gates pass on clean `4d7a9431`, including review fixes `b2f26725` and main `827b51d5`: `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test`, dedicated Agent runtime coverage (456 tests; 100% statements/branches/functions/lines within that scope), and real PostgreSQL integration (44 files / 568 tests).
- Focused review regressions include 318 UI tests, stale save/discard refusal, same-message child capacity retry, configuration migration guards and maintained cleanup. The replacement-generation response regression failed before the fix and passed afterwards. No thresholds were lowered or tests skipped.
- Maintained Cloud identity HTTP harness: 118/118 passed on clean `4d7a9431`, including PostgreSQL migration count/order/source hashes, Local upgrade preservation, authenticated ownership guards, Cloud setup without allocation, Runtime v2 registration, model-paused availability and overall-off with valid model-on configuration. Both owned Server/PostgreSQL/workspace fixtures were cleaned up.
- Current-head CI is still running. Browser, container, CLI, Docker image, Runner offline smoke, secret scan and documentation checks have passed; Node compatibility, the main Checks job, quality and patch coverage are pending. The separate ownership gate needs `bestony` approval (62 files); the existing change-request review remains for the reviewer to reassess.
- The earlier E9 native GCP run passed 30 checks with actual Cloud Run/native Sandboxes and installed Pi, 1 vCPU/1 GiB: concurrent parent/child callback, capacity refusal, save/release, isolated Instance reuse and cold recovery. All owned resources were removed. That run used Server `d29bf750` and immutable Runner `0.0.6-staging.31.1`; this configuration/merge follow-up is covered by renewed local/HTTP checks, and does not claim a new native run or Runner deployment.
- Native model responses and IM ingress are fixture-controlled. Real Slack/Feishu delivery and GitHub App publication remain pending a designated conversation/repository. This PR does not claim those external acceptance results or an E9 deployment.

## UI validation

The desktop observations below are historical E9 acceptance; this review follow-up uses renewed component tests rather than claiming a new manual browser journey.

- Desktop evidence: actual production build in an isolated database at 1280 × 720, with Cloud default/Pi, creation, managed preparation, messaging choice, refresh, disabled-model refusal, Agent overview and Computer settings. Creation/read-only navigation left 0 Sandbox rows; fixture cleanup verified.
- Mobile evidence: dedicated mobile browser pass not performed.
- Widths checked: 1280px desktop; 320px/390px/768px/1440px were not exercised.
- States verified: available/unavailable Cloud, unknown Computer after failed read, empty environment inventory, IM-only usage scope. Component tests cover release success followed by failed refresh, stale-generation discard, uncertain mutation outcomes and double-clicks.
- Keyboard/accessibility checks: existing labeled controls and confirmation dialog semantics tested; manual keyboard-only journey not performed.
- New reusable block, theme value, or CSS exception: Cloud environment blocks compose existing design-system components; no new theme value/CSS exception.

## Breaking changes

No database migration. The Account Sandbox stop endpoint now requires `environmentGeneration` for ordinary save/release as well as explicit discard; the browser and maintained cleanup harness supply it. Diagnostics add `not_applicable` for Cloud's Local CLI readiness.

Identities-on now also enables Runner and requires the storage prefix, Runner version and all eight Runner coordinates. An old identities-only deployment must prepare them first or keep the overall switch off. The retired Runner variable is an upgrade check only: identities-on/old-Runner-off and malformed values fail startup with a migration error. An agreeing old `true` remains tolerated during rollback to an older Server; remove it once that rollback window ends. Model-off preserves Runner cleanup; save/drain resources before disabling the overall switch. Limits bound concurrency, not total model/storage spend.

Review choices retained deliberately: a generation-zero Sandbox may initialize its single workspace object before admission, preserving retry/recovery behavior without a new distributed state; `execution_unavailable` remains a defensive API contract for injected or older configurations, not a supported production switch combination.

## Checklist

- [x] All six required local gates pass on the final source commit.


liuchao-001 (changes requested):
Reviewed at head `61ccc799` against merge base `1a2f3dd2`, reading full files rather than diff hunks. Technical CI is green on this head; the ownership gate is separate.

**What holds up well.** The capacity core is the strongest part of this PR. Occupancy is *derived* from durable Sandbox facts rather than a counter, so a reconcile or a crash cannot drift it; `pg_advisory_xact_lock` is genuinely the first statement of the reservation transaction, so waiters hold no row locks and the lock order argument is sound; and retention is fail-closed in every uncertain case (unknown create, save-failed, delete-unconfirmed). `sandbox-capacity-admission.test.ts` proves the races against real PostgreSQL rather than asserting them in prose. Authorization on the new surfaces is clean — `CloudOverviewService.read` repeats the full ownership predicate in both the outer scope and the session subquery, no client-supplied account id anywhere, and removing the Local daemon checks removed liveness probes only, not guards. No migrations, no schema changes, no weakened thresholds or skipped tests, en/zh message parity is complete.

Five things need to change before this merges.

---

### 1. The retired `OPENTAG_CLOUD_RUNNER_ENABLED` is silently ignored, so upgrading flips behavior in both directions

`packages/server/src/cloud-runner-config.ts:140` — the switch is no longer read at all (it is not even in the object handed to the `.strict()` schema, so a stale value cannot fail validation). `__tests__/config.test.ts:130` pins this: `{...runnerEnv, OPENTAG_CLOUD_RUNNER_ENABLED: "false"}` → `{enabled: true}`.

Both existing postures break, and neither announces itself:

- **Identities-on with no Runner coordinates** — exactly what pre-PR `.env.example:44` documented ("Cloud identity acceptance (E2); does not provision resources or enable a Runner") — now throws at `cloud-runner-config.ts:148`. That parse runs inside `startServer`'s try (`index.ts:417`) and becomes `process.exitCode = 1`, so under a supervisor the **entire Server** crash-loops, not just Cloud.
- **Identities-on with coordinates and `RUNNER_ENABLED=false`** — the natural "Cloud paused" posture — comes back with the Runner **on**. That is not inert: `index.ts:515-521,975` start `SandboxIdleReclaimer`, which seals and deletes ready environments, and `cloud-runtime-composition.ts:63-121` wires a real `CloudRunAdmin` + GCS store, so IM deliveries allocate real Cloud Run Instances and spend money.

`docs/cloud-runner-execution.md:247` makes the second case worse by telling operators to retain the variable through a rollback window — retaining `=false` is precisely the configuration that silently enables the Runner. Nothing in EN or zh says "leaving this set to `false` no longer turns the Runner off", and `serverEnvironmentSummary` (`config.ts:837-845`, logged at `index.ts:1001`) carries no Cloud switch, so the flip leaves no boot-log trace either.

Suggested fix, ~5 lines and it covers both directions: keep *reading* the retired name for one release and refuse to boot when it is present, with a message that names the retirement and the new semantics. That converts a silent behavior change into an explicit operator decision. An upgrade note in `.env.example` and both `cloud-runner-execution.md` mirrors should say plainly that identities-on now implies Runner-on and lists the coordinates that become mandatory.

Related, same path: the throw at `:148` says the Runner "is enabled", but the operator never enabled a Runner — naming `OPENTAG_CLOUD_IDENTITIES_ENABLED` as the cause would help, since this string is the first thing a crash-looping operator reads.

### 2. A capacity refusal is terminal for internal child work, while the identical condition is transient everywhere else

`cloud-session-collaboration-owner.ts:867` maps the refusal to `{status: "rejected"}`. That is terminal by construction: `session-service.ts:942-943` returns `attemptCount: null` for a `rejected` message, and `session-collaboration-service.ts:138-144` then replays the stored outcome forever. Only a brand-new logical message can retry.

The condition it terminates on is transient by construction:

- `CLOUD_CAPACITY_OCCUPIED_LIFECYCLES` includes `releasing` (`sandbox-capacity.ts:12`), held "until verified absence" — so a parent that stops two children and immediately fans out again gets a permanent rejection for something that would have succeeded seconds later.
- After an automatic idle reclaim the child is `unallocated` with generation > 0, so resuming it routes through `#reserveNewGeneration` and is charged again. The platform idles a child out to save resources, then terminally rejects the message that would have resumed it.

Two asymmetries suggest this is an accident rather than a choice. The *same* error is retryable when it happens to be slow: past `ensureTimeoutMs`, `#convergeAllocation` returns `"timeout"` → `notReady` → `unreachable`/`runtime_not_ready` (`cloud-session-collaboration-owner.ts:869`). So terminality depends on race timing, not on any fact about the system. And the IM path treats the very same `CloudCapacityExceededError` as transient, queued with capped backoff (`im-delivery-cloud.ts:418-428`).

Also, the child Session row is committed before delivery and `ensureSandbox` runs *before* the capacity check (`:845-857`), so each rejection strands a dead child Session + Sandbox row.

`unreachable` looks like the right answer here — it matches IM, matches the timeout branch, and keeps the child's own retry path. If terminality is genuinely wanted, it needs to key off something non-transient, not "an occupied lifecycle at this instant".

### 3. Save-and-release is not fenced to the generation the reader was looking at; discard is

`cloud-session-environment.tsx:194` sends the generation only for discard: `mode === "discard" ? {discardUnsavedChanges: true, environmentGeneration: generation} : {}`. The server only validates it in that branch (`sandbox-runner-service.ts:190-193`), and `AccountSandboxRunnerStopRequestSchema` (`packages/shared/src/cloud-runner.ts:241-246`) does not even permit a generation without `discardUnsavedChanges: true` — so this cannot be fixed client-side.

The window is real: the overview polls on `liveResourceQueryOptions` (30 s), generations increment on the same row, and the card's remount key `agentId:sessionId:sandboxId` (`:159`) deliberately excludes the generation, so a new generation does not reset the card. A reader looking at a 25-second-old paint of generation 5 clicks "Save and release" after generation 6 has started for an IM task they never saw, and generation 6 is released — stopping live execution, exactly what `cloud_release_stops_work` warns about. `cloud-environment.test.tsx:326` pins today's behavior (`stop` called with `{}`).

The workspace is saved first, so this interrupts rather than destroys — but the discard path already establishes that this team considers generation fencing necessary, and the same reasoning applies to a control that stops running work.

### 4. The discard dialog accuses itself of staleness on every 30-second poll

`cloud-session-environment.tsx:226` includes `!actionsEnabled` in `discardStale`, and `actionsEnabled` is `overviewQuery.isSuccess && !overviewQuery.isFetching` (`cloud-environment.tsx:171,317`) over a query with `refetchInterval: 30_000` and `refetchOnWindowFocus: "always"` (`query/live.ts:17-22`).

So: the reader opens "Discard unsaved changes and release?", spends more than 30 seconds reading the data-loss warning — or does what `cloud_notice_conflict` tells them and tabs away to check the current state, then tabs back, which forces a refetch. A background poll starts, the dialog swaps in "This environment changed while this dialog was open. Review the current state before discarding anything." and the confirm button disappears (`:390,395-405`), although nothing changed. It returns when the fetch settles. The existing test at `cloud-environment.test.tsx:330-384` only covers the genuine generation-change path, so this disjunct is untested.

Same root cause, lower stakes: `visibleOutcome === null && actionsEnabled ? <ReleaseActionButtons/> : null` (`:235`) fully **unmounts** the release buttons twice a minute instead of disabling them, so a control can vanish under the cursor mid-click.

Gating on a real staleness signal (generation/sandbox change, which `discardStale` already checks) rather than on "a refetch is in flight" fixes both.

### 5. A failed availability read is rendered as a deployment fact, with no retry

`onboarding-v2/page.tsx:76-80` defines `failed` as "a read that did not answer", and the comment at `steps.tsx:130` says the Cloud choice "stays disabled rather than guessing either way" — but `page.tsx:379-385` collapses `loading` and `failed` both to `cloud={undefined}`, and `kind: "failed"` is never read anywhere. `steps.tsx:157-170` then renders, simultaneously, the badge "Checking…" and the description "Cloud is temporarily unavailable **on this deployment**. You can still run the agent on your own computer."

So a 500 or a timeout produces a card that is permanently "Checking…" *and* asserts a deployment fact the Server never reported — two contradictory claims, one of them invented. The effect at `page.tsx:246-258` runs once per mount and there is no retry control, so the only way out is picking Local or reloading. The same wrong description also flashes on every normal load. `agent-create-cloud.test.tsx:127-134` covers the failed read but only asserts the card is disabled, never the copy.

A third copy variant for "we could not check" plus a retry button would match what the code already says it intends.

---

### Smaller items, not blocking

- **`execution_unavailable` is now unreachable.** `cloud-product-config.ts:25` returns it only when identities are on and the Runner is off, which `parseServerConfig` can no longer produce. Its test hand-builds an impossible `ServerConfig`. Worth deciding whether to retire the value or keep it as a contract placeholder — and note `setup-memory-adapter.ts:650,964` use it as the default reason when none is supplied (Review Lab / in-memory model only, so no production impact, but it will display a cause the Server can never report).
- **Model config cross-field validation is skipped while Cloud is off** (`cloud-model-config.ts:115` returns before the master-key/allowlist checks; previously it threw). A typo'd `OPENTAG_CLOUD_MODEL_MASTER_KEY` staged today boots clean and only surfaces on the deploy that enables Cloud — which then crash-loops via the same path as item 1.
- **"Try again" on an unconfirmed Computer is a no-op.** `agent-computer-settings.tsx:101` calls `onAgentChanged` → `syncAgentQueries(queryClient, agentId)` (`agent-settings-page.tsx:56`), which invalidates `computers()` only when passed `{computers: true}` (`query/agent-sync.ts:19-21`). The `unconfirmed` state is produced precisely by the computers read failing (`agent-model.ts:150`), so the button never refetches the thing that failed; the screen self-heals on the 30 s interval instead. One argument fixes it.
- **Stale banners never clear.** `notice` and `actionError` (`cloud-session-environment.tsx:176-178,232-234`) are not generation-gated the way `visibleOutcome` is, and the card does not remount across generations, so a refusal from minutes ago can sit on a healthy new environment. Only the discard dialog's `onClose` clears `actionError`; the save path has no clearing route.
- **Optimistic post-release write erases the specific error.** `:36` rewrites any `lastErrorCode` to `"environment_unavailable"`, which matches no case in `environmentErrorLabel`/`environmentErrorTone`, so a genuine save failure downgrades from "The workspace could not be saved"/`danger` to a generic warning, with `canRelease/canDiscard` forced false. If the follow-up refetch also fails, that degraded row is what persists.
- **Storage side effect precedes admission.** `#prepareWorkspaceAllocation` (`sandbox-runner-service.ts:403-440`) initializes the workspace object for generation 0 before `#reserveNewGeneration` can reject on capacity. Idempotent, so bounded, but a never-admitted sandbox still accumulates an initialized object.
- **Testing doc describes an unreachable state.** `docs/testing/cloud-computer.md:180-181` and `docs/zh-CN/testing/cloud-computer.md:157` both say the fixture's disabled Runner must report execution-unavailable; the fixture now injects full inert coordinates (`cloud-identities.mjs:42-51`) and the harness asserts `model_unavailable`. EN and zh agree with each other and disagree with the code.
- **`.env.example:46-53`** requires "complete Runner/network/storage coordinates" for boot without naming the eight variables, so an operator working from that file alone cannot satisfy the new requirement.
- **Capacity defaults live in two places** — `cloud-runner-config.ts:70-71` and `sandbox-capacity.ts:22` both encode 3/20.
- **`CLOUD_SESSION_ALLOCATION_FAILED` logs "the input stays retryable"** (`cloud-session-collaboration-owner.ts:914-919`) unconditionally, which is false on the capacity path as long as item 2 stands.
- **Account-scoped capacity is invalidated per-Agent only** (`cloud-session-environment.tsx:192,214`), so another Agent's open board shows "3 of 3" until its own poll.
- **`ImBindingDiagnostics` returns `handoffReady: true` with `providerCliReadiness: "unavailable"`** for a healthy Cloud binding (`im-binding-provider-cli.ts:355-375`). The intent is "not applicable", but the value reused is an existing failure value.
- **Route guard mismatch:** `api/account.ts:140` registers `GET account/cloud-computer` unconditionally, but `app.ts:772-784` early-returns unless one of the *other* services is present, so availability alone yields 404 rather than the fail-closed `disabled` body. No production impact (`index.ts` always wires `agentService`), but it contradicts the principle stated at `app.ts:653-655`.
- **`agent-setup-service.ts:136-137`** casts `as ImProvider[]` onto an `ImCliProvider[]` field; identical today, silent widening if they diverge.
- **`docs/zh-CN/cloud-computer-product.md:3`** carries a 同步日期 of `2026-09-21`, one day in the future.

liuchao-001 (approved):
Re-reviewed at head `4d7a9431` (fix commit `b2f26725` plus the main merge). All five blocking findings are resolved, and resolved at the right layer rather than patched at the symptom. Approving.

**1. Retired switch** — now validated instead of ignored. `OPENTAG_CLOUD_RUNNER_ENABLED=false` with identities on refuses boot with a message that states what changed and gives both ways out; malformed values refuse too; an agreeing `true` stays tolerated for the rollback window. The missing-coordinates error now names `OPENTAG_CLOUD_IDENTITIES_ENABLED` as the cause, `serverEnvironmentSummary` carries the three effective Cloud toggles so the state is visible in the boot log, and `.env.example` lists the eight coordinates with an explicit upgrade note. The 3/20 defaults are single-sourced from `DEFAULT_CLOUD_CAPACITY_LIMITS`. I checked the mirror case (`=true` with identities off) — it boots with Cloud off rather than refusing, which is fine, because that combination could not boot before this PR either, so no deployment can be upgrading from it.

**2. Capacity refusal** — now `unreachable`, so the same durable message and child Session stay retryable. That matches the IM path and the timeout branch, which is what made the old behavior look accidental.

**3. Save fencing** — both branches of `AccountSandboxRunnerStopRequestSchema` now carry `environmentGeneration`, and `stopReleaseMarker` fences save and discard alike under the same row lock. I specifically re-checked the bypass I was worried about: `stopForAccount`'s `input` became optional, but the route parses `request.body ?? {}` against a schema where every branch requires the generation, so an empty body is a 400 rather than an unfenced release. The only other caller, the maintained cleanup harness, was updated to send the observed generation and has a test pinning it.

**4. Discard dialog** — staleness now keys only on real environment change; a refetch in flight merely disables the confirm instead of declaring the environment changed, and the release buttons stay mounted and disable rather than unmounting under the cursor.

**5. Failed availability read** — `failed` is now a distinct third state with its own badge and copy plus a retry, and the retry is attempt-fenced so a late answer from a superseded read cannot overwrite a newer one. The loading state no longer borrows the unavailable copy.

The smaller items are addressed too: operation feedback is scoped to the generation that requested it, the optimistic post-release write mirrors the server's redaction allowlist instead of flattening every failure to a generic warning, release invalidates every Agent's cloud overview so account-wide capacity stays consistent, the unconfirmed-Computer retry now invalidates the `computers` read that actually failed, `not_applicable` replaces the misused `unavailable` for Cloud's Local CLI readiness (the only consumer is the CLI's verbatim diagnostic print, so nothing switches on it), the availability route guard and the `ImProvider` cast are fixed, and the testing docs now describe `model_unavailable` in both mirrors.

The two items retained deliberately — the generation-zero workspace object initialized before admission, and `execution_unavailable` kept as a defensive contract value — are both reasonable, and both were non-blocking in my review. Documenting the reasoning in the PR body is the right resolution for them.

Two notes that are not objections. First, `redactedEnvironmentError` and `canRelease`/`canDiscard` in the browser now restate rules the Server owns (`publicDiagnostic` and the overview's control flags); the copies agree today and the divergence is unreachable in practice, since `controlsEnabled` is process-level wiring, but they are two lists in two packages that have to stay in sync. Worth a shared constant if this area grows. Second, requiring `environmentGeneration` is a wire-contract change on an endpoint whose only clients are in this repo and updated here — correct now, and only cheap because Cloud has not shipped on by default.

All 16 technical checks pass on this head. I did not re-run the native GCP acceptance, and the PR correctly presents that run as historical rather than re-claimed. The ownership gate still needs `bestony`.

PR #667, opened 2026-09-20
test(shared): cover the runtime schemas and pure helpers

bestony:
## What this changes

`packages/shared` holds the runtime Zod schemas that every other workspace derives its types from, so its
rejection paths are the cheapest branches in the repository to cover and the most expensive to get wrong.
`http-paths.ts` had no suite of its own and sat at 33.33% branch coverage; `cloud-runner.ts` at 59.09%.
This adds 1,528 lines of tests across six suites and takes every target file to ~100% line coverage.

## Coverage

Measured with `node scripts/unit-coverage.mjs --project shared --scope 'packages/shared/src/{http-paths,cloud-runner,runtime-domain,structured-errors,agent-setup,skill-manifest}.ts'`.

| file (under `packages/shared/src/`) | lines before | lines after | branches after |
| --- | --- | --- | --- |
| `http-paths.ts` | 81.06% | 100% | 100% |
| `cloud-runner.ts` | 80.76% | 100% | 100% |
| `runtime-domain.ts` | 89.83% | 100% | 100% |
| `structured-errors.ts` | 94.81% | 100% | 90.52% |
| `agent-setup.ts` | 91.42% | 100% | 99.52% |
| `skill-manifest.ts` | 97.59% | 99.51% | 90.67% |

The `shared` project as a whole moves from 93.08% to 97.36% lines.

## Checklist

- [x] `http-paths.ts` — new suite covering every exported path builder, including the optional-segment and encoding branches
- [x] `cloud-runner.ts` — schema rejections and discriminated-union fallbacks, including the wire-budget rejections
- [x] `structured-errors.ts` — each error code's construction and each narrowing helper's negative case, plus the redaction scanners
- [x] `runtime-domain.ts` — rejection paths
- [x] `agent-setup.ts` — rejection paths
- [x] `skill-manifest.ts` — malformed manifest input
- [x] Remaining uncovered lines and branches itemized with a reason (see below)

| criterion | outcome |
| --- | --- |
| Tests only — `git diff --name-only main...HEAD` | 6 files, all under `packages/shared/src/__tests__/` |
| `pnpm check` | exit 0 |
| `pnpm --filter @opentag/shared test` | 38 files, 576 tests, all passed |
| Per-file coverage targets (>= 95% lines, >= 98% for `skill-manifest.ts`) | all 6 files met |
| `public-exports.snapshot.json` unchanged | confirmed — the package's public surface did not move |
| No network / PostgreSQL / provider / Docker dependency; no new migration; `scripts/coverage-floors.json` untouched | confirmed |
| Working tree clean, 8 commits | confirmed |

Two dead lines and a set of defensive branches remain. Each is unreachable through the module's public
surface, so covering it would mean changing the source rather than adding a test:

- `skill-manifest.ts:136` — `parseBlockIndicator`'s `if (first !== "|" && first !== ">") return null;`.
  Its only caller reaches it under `rawValue.startsWith("|") || rawValue.startsWith(">")`, so `raw[0]` is
  always one of those two. The helper is module-private and cannot be called with anything else.
- `skill-manifest.ts:334` — `unquoteFoldedScalar`'s trailing `return null;`, shadowed for the same reason:
  its caller only runs it after `isQuotedStart` returned true, so one of the preceding branches always wins.
- The remaining uncovered branches are `?? ""` / `?? undefined` fallbacks on array indexes that are provably
  in range, and `value[cursor] ?? ""` guards inside single-character scanners in `structured-errors.ts`.
  The lane's notes enumerate each one by line so the next reader does not re-derive them.

Removing these guards is a source-level judgement call and belongs to whoever owns the module.

## Provenance

Produced by dispatch-pi run `79522c`, lane `shared-schema-tests-5`. An agent wrote the tests in an isolated


yuezengwu (approved):
Reviewed commit `54d9bf7bde3cfc3356b3e9cddebf3edb7f8fec37` against `main` at `74054748cbbef2f448a3db1c1021e9505d0ebea6`. No blocking findings in the test-only changes.

Validated in a clean isolated worktree: all 576 `@opentag/shared` tests across 38 files passed; affected package/dependency builds, package typecheck, `pnpm check`, and `git diff --check` passed. Reviewed runtime schema rejection guards, identity and lifecycle consistency, serialized wire budgets, path encoding, malformed Skill manifests and credential redaction.

Remote CI is still running (Enforce Patch Coverage); merge remains subject to those results.

PR #668, opened 2026-09-20
test(mcp): cover the MCP OAuth services and their routes

bestony:
## What this changes

The MCP OAuth services and their HTTP routes were the single largest untested block in the repository:
365 uncovered lines and 283 uncovered branches, with `mcp-oauth-flow-service.ts` at 4.57% and both route
modules effectively unexercised. This adds 3,385 lines of tests across five suites.

## Coverage

Measured with `node scripts/unit-coverage.mjs --project server --scope 'packages/server/src/**/mcp*.ts'`.

| file (under `packages/server/src/`) | lines before | lines after | branches after |
| --- | --- | --- | --- |
| `services/mcp/mcp-oauth-flow-service.ts` | 4.57% | 97.71% | 92.85% |
| `services/mcp/mcp-authorization-service.ts` | 8.45% | 100% | 95.83% |
| `services/mcp/mcp-oauth.ts` | 81.03% | 100% | 98.74% |
| `api/mcp-servers.ts` | 4.05% | 100% | 100% |
| `api/mcp-oauth.ts` | 3.33% | 100% | 100% |

## Checklist

- [x] `McpOAuthFlowService` — start, callback, refresh and revoke, plus `orderIssuers`, the unknown-lifetime expiry fallback, and the `requireFlowIssuer` / `requireIssuer` / `requirePkce` throw paths
- [x] `McpAuthorizationService` — probe outcomes, credential resolution, header construction, probe writes, and the `isProbeResult` type guard
- [x] `mcp-oauth.ts` — the PRM / AS / OIDC well-known discovery order, `endpointUrl`'s loopback policy, and the token requests
- [x] `api/mcp-servers.ts` — every route through `app.inject`, the uuid parameter rejections, and the unauthenticated case
- [x] `api/mcp-oauth.ts` — the public callback, its query-schema rejections, and each failure redirect code
- [x] Remaining uncovered lines and branches itemized with a reason (see below)

Route tests drive a real Fastify instance through `app.inject` with injected fakes, rather than calling
handlers directly, because the parameter schemas are registered on the route and only run inside the real
request pipeline. That matches what `mcp-gateway-route.test.ts` already does in this package.

| criterion | outcome |
| --- | --- |
| Tests only — `git diff --name-only main...HEAD` | 5 files, all under `packages/server/src/__tests__/` |
| `pnpm check` | exit 0 |
| `pnpm --filter @opentag/server test` | 159 files, 2675 tests, all passed |
| Per-file coverage targets (>= 85% lines) | all 5 files met; scoped server total 96.03% lines vs an 88.41% baseline |
| No network / PostgreSQL / provider / Docker dependency; no new migration; `scripts/coverage-floors.json` untouched | confirmed |
| Working tree clean, 6 commits | confirmed |

Four statements and a handful of branch arms in `mcp-oauth-flow-service.ts` remain, each unreachable from
any caller. The reasons, in short:

- `442`, `458`–`459` — `#recordRegistration`'s "insert returned no row" and "no row for (Account, issuer)"
  arms. The function has exactly one call site, `#clientForStart`, which always passes `ifAbsent: true` and
  has already thrown on the read-back miss before control could reach them. A second call site — a source
  change — would be needed.
- `436`–`437` — the `sealed?.ciphertext ?? null` / `sealed?.keyId ?? null` fallbacks on the upsert branch of
  that same single-call-site function, dead for the same reason.
- `931`–`932` — `#claimForRefresh`'s "the Agent row is gone" arm. `mcp_server_authorizations.agent_id` is
  `not null references agents(id) on delete cascade` with a `refresh_claim_requires_oauth` check, so a
  claimed OAuth row always has a live Agent row and the join cannot miss. Covering it would need a
  deferrable foreign key or an injected seam — a schema or source change.

Each of these is a guard a source change would remove rather than cover; that call belongs to whoever owns
the module.

## Provenance

Produced by dispatch-pi run `79522c`, lane `mcp-oauth-tests-1`. An agent wrote the tests in an isolated


yuezengwu (approved):
Reviewed `fd038cc5db1dde4b794b94eb8f7e9a94fbd77f8f`; no blocking findings in the five test-only files.

Independently verified the server build, full server unit suite (159 files / 2675 tests), server typecheck, `pnpm check`, and diff checks in a clean isolated worktree. The added coverage exercises the real service/route or adapter paths with external transports controlled.

Coverage percentages were not remeasured, and this is not live-provider or PostgreSQL integration acceptance. Merge remains subject to the current CI results.

PR #670, opened 2026-09-20
test(web): cover the MCP page and the remaining feature components

bestony:
## What this changes

Tests only. No production source, configuration, schema, migration or documentation file is touched, and
neither generated tree (`src/routeTree.gen.ts`, `src/paraglide/**`) is edited.

The MCP page was the web app's largest coverage gap (67.14% lines, 61.24% branches), and several feature
components sat in the 57–82% range. This adds 3,067 lines of tests across nine colocated suites — the
convention `apps/web` already uses — and takes all nine target files to 100% line coverage.

## Coverage

Measured with `node scripts/unit-coverage.mjs --project web --scope 'apps/web/src/{features,setup,analytics}/**/*.{ts,tsx}'`.

| file (under `apps/web/src/`) | lines before | lines after | branches after |
| --- | --- | --- | --- |
| `features/mcp/mcp-page.tsx` | 67.14% | 100% | 99.52% |
| `features/mcp/mcp-queries.ts` | 81.81% | 100% | 100% |
| `features/integrations/agent-github-repositories.tsx` | 87.71% | 100% | 97.56% |
| `features/account/github-integration-settings.tsx` | 79.22% | 100% | 98.71% |
| `setup/command-block.tsx` | 73.80% | 100% | 96.29% |
| `features/shell/app-shell.tsx` | 75.60% | 100% | 100% |
| `features/task-message-body.tsx` | 57.14% | 100% | 100% |
| `features/task-outgoing-replies.tsx` | 81.25% | 100% | 98.63% |
| `analytics/analytics.ts` | 66.66% | 100% | 100% |

The `web` project as a whole moves from 94% to 96.82% lines over this scope.

## Checklist

- [x] `mcp-page.tsx` — states, dialogs and failure paths; id-only reads, writes and invalidation; removing a whole Server definition; the missing-tool count, revision fallback and header order
- [x] `mcp-queries.ts` — the query and mutation surface
- [x] `agent-github-repositories.tsx` — repository states, the scope editor, and a later discovery page with no truncation field
- [x] `github-integration-settings.tsx` — connection states and failure language
- [x] `command-block.tsx` — copy, fallback and inert states
- [x] `app-shell.tsx` — scope, the narrow layout, and the focus trap
- [x] `task-message-body.tsx` — the remaining Markdown body elements
- [x] `task-outgoing-replies.tsx` — outgoing reply content shapes, kind labels, outcome banners, and the Agent exit release
- [x] `analytics.ts` — the installer and the reporter's location default
- [x] Remaining uncovered branches itemized with a reason (see below)

Assertions go against the strings `m.*()` renders; no message key was added and `project.inlang` is untouched.

| criterion | outcome |
| --- | --- |
| Tests only — `git diff --name-only main...HEAD` | 9 files, all `*.test.ts(x)` colocated with their source |
| `pnpm check` | exit 0 |
| `pnpm --filter @opentag/web test` | 101 files, 1392 tests, all passed |
| Per-file coverage targets (>= 85% for `mcp-page.tsx`, >= 90% elsewhere) | all 9 files at 100% lines |
| Generated files untouched; no network / provider dependency; `scripts/coverage-floors.json` untouched | confirmed |
| Working tree clean, 14 commits | confirmed |

Every line in all nine files is covered; what remains is branch-level residue that the source guards so no
render or event can reach it. Each was attempted before being recorded:

- `mcp-page.tsx:996` — the `0` (equality) arm of the header sort comparator. `sortedHeadersKey` sorts
  `Object.entries(record)`, and an object cannot have two identical keys, so `left === right` never occurs.
  Confirmed by instrumenting the comparator: the `-1` and `1` arms both run, the `0` arm never does.
- `agent-github-repositories.tsx:296` — `loadMore`'s `view.nextCursor === null || loadingMore` re-entry
  guard. The button renders only when a cursor exists and is disabled while loading.
- `agent-github-repositories.tsx:324`, `:330` — `toggleRepository`'s "already selected" guards. React
  applies state updates against the value captured at render, so a second click delivers `selected: true`
  only from an unchecked render, where the drafts cannot already contain the repository.

The lane's notes list the full set. Each would be removed, not covered, by a source change.

## Provenance

Produced by dispatch-pi run `79522c`, lane `web-feature-tests-7`. An agent wrote the tests in an isolated


apps/web/src/features/mcp/mcp-page.test.tsx
```diff
@@ -425,3 +426,1080 @@ describe("McpPage", () => {
...
+    wrap(<McpPage agentId={AGENT_ID} />);
+    fireEvent.click(await screen.findByRole("button", { name: "Edit" }));
+    await chooseOption("Scope", "Shared definition");
+
+    // With no definition to read there is no impact to warn about, and no revision to fence on.
+    expect(await screen.findByText(/This will affect 0 Agents/)).toBeTruthy();
+    fireEvent.click(screen.getByRole("button", { name: "Save" }));
+
+    await waitFor(() => expect(update).toHaveBeenCalledTimes(1));
+    expect(update.mock.calls[0]?.[1]).toMatchObject({ expectedRevision: 1 });
```
yuezengwu: [P2] Do not require a successful shared write after the definition read fails

When `mcpServer` rejects, these new assertions require the UI to claim zero affected Agents and submit a guessed `expectedRevision: 1`. The current Agent already mounts this Server, so zero is not a valid impact; if the real definition is still at revision 1, the write can succeed without the user learning who else it affects. The suite itself says shared edits must name the affected Agents. The production fallback predates this PR, but making it the required passing outcome would reject a later safety fix. Please remove these success assertions or express the intended unavailable-impact/no-write behavior as an expected-failing regression until the source is repaired; the product repair can remain separate from this test-only PR.

yuezengwu (changes requested):
Reviewed `c71fcce67a5176b5274522804ef46b4f5220da53`. Requesting one correction to the new MCP shared-edit test: it currently turns an unavailable definition/impact into a successful shared write. The underlying UI behavior predates this PR; the issue here is making that behavior a regression requirement.

Independent validation at this exact head passed: web build, 101 web test files / 1392 tests, web typecheck, `pnpm check`, and diff checks. Validation used this head's locked dependencies; current main has since updated the development toolchain, including Vitest 5, so this is not a merged-base compatibility or browser acceptance result.

apps/web/src/features/mcp/mcp-page.test.tsx
bestony: Fixed in `b021d887`, taking the first of your two options.

The three success assertions are gone: the "This will affect 0 Agents" expectation, `update` having been called, and the `expectedRevision: 1` match. The case still drives the failed `mcpServer` read and now asserts only that the shared-scope dialog stays rendered and operable, and the test body says in full why the fallback's shape is deliberately not pinned — the current Agent mounts this Server, so an empty impact is not a truthful answer, and requiring it would make a later safety fix look like a regression. The page's own repair stays out of this test-only PR, as you suggested.

Re-measured at the new head: `features/mcp/mcp-page.tsx` is still at 100% lines (branches 99.52% -> 99.04%), all nine files still meet their targets, `pnpm check` exits 0, and the web suite is 101 files / 1392 tests green. Noted on your validation caveat — that run used this branch's locked dependencies, and main has since moved to Vitest 5, so this is not a merged-base result either.

yuezengwu (approved):
Re-reviewed `b021d8877c1651fcd2338e7c54aaf13bc95e8e14`. Removing the zero-impact and guessed-revision successful-write expectations addresses my previous blocking comment.

Independently validated both the exact head and a clean local merge with current main `168ef0220e1b3b87695f4a4a9040dac34da486d5`: all 1,392 web tests in 101 files, web build, typecheck, and `pnpm check` pass in both configurations. The merged validation includes main's Vitest 5 upgrade.

The existing product fallback when the shared definition cannot be read remains a separate issue; this approval covers the test-only change.

PR #673, opened 2026-09-21
test(client): cover the runner, provider-CLI and runtime proxy

bestony:
## What this changes

The client runner, the provider-CLI subsystem and the runtime proxy together held 742 uncovered lines and
995 uncovered branches — the largest gap in the repository by volume, spread across nineteen files that all
already had suites. This extends those suites with 5,957 lines of tests.

## Coverage

Measured with `node scripts/unit-coverage.mjs --project client --scope 'packages/client/src/{runner,runtime}/**/*.ts'`.

| file (under `packages/client/src/`) | lines before | lines after | branches after |
| --- | --- | --- | --- |
| `runner/cloud-turns.ts` | 81.87% | 90.48% | 85.32% |
| `runner/serve.ts` | 88.57% | 90.35% | 85.47% |
| `runner/acceptance.ts` | 83.49% | 94.81% | 72.46% |
| `runner/web-gateway.ts` | 81.28% | 95.18% | 85.56% |
| `runner/cloud-journal.ts` | 88.16% | 93.06% | 87.26% |
| `runner/workspace-archive.ts` | 91.87% | 94.95% | 89.74% |
| `runner/native-sandbox.ts` | 86.86% | 95.76% | 88.59% |
| `runner/cloud-turn-worker.ts` | 87.09% | 90.32% | 84.61% |
| `runtime/runtime-proxy-data-client.ts` | 77.36% | 90.00% | 77.25% |
| `runtime/runtime-proxy-loopback-adapter.ts` | 80.00% | 90.00% | 80.92% |
| `runtime/provider-cli/manager.ts` | 84.43% | 91.98% | 76.01% |
| `runtime/provider-cli/validation-runner.ts` | 84.90% | 93.39% | 83.33% |
| `runtime/provider-cli/turn-plan.ts` | 85.43% | 97.57% | 93.10% |
| `runtime/provider-cli/turn-plan-manager.ts` | 84.29% | 91.62% | 83.78% |
| `runtime/provider-cli/turn-runner.ts` | 85.35% | 94.26% | 94.77% |
| `runtime/provider-cli/tar.ts` | 78.09% | 98.09% | 91.00% |
| `runtime/provider-cli/reconciler.ts` | 94.45% | 96.22% | 83.52% |
| `runtime/provider-cli/outgoing-reply-report.ts` | 80.70% | 98.24% | 85.62% |
| `runtime/provider-cli/outgoing-reply-capture.ts` | 87.73% | 96.93% | 84.82% |

Over this scope the `client` project moves from 91.70% to 93.13% lines.

## Checklist

All nineteen target files were worked in descending order of gap size, and each has commits of its own:
`cloud-turns` → `runtime-proxy-data-client` → `serve` → `runtime-proxy-loopback-adapter` → `acceptance` →
`web-gateway` → `manager` → `validation-runner` → `native-sandbox` → `turn-plan` / `turn-plan-manager` →
`cloud-journal` / `workspace-archive` → `turn-runner` → `tar` → `reconciler` → `outgoing-reply-report` /
`outgoing-reply-capture` → `cloud-turn-worker`.

`packages/client` carries a separate 100%-branch gate over `src/agent-runtime/**`, `src/providers/**` and a
named set of `src/runtime/*.ts` files. None of the files above is in that manifest and none of them was
touched, so that gate is undisturbed.

| criterion | outcome |
| --- | --- |
| Tests only — `git diff --name-only main...HEAD` | 21 files, all under `packages/client/src/__tests__/` (including one new fixture) |
| `pnpm check` | exit 0 |
| `pnpm --filter @opentag/client test` | 136 files, 2082 tests, all passed |
| Per-file coverage targets (>= 90%, >= 94% and >= 96% on two) | all 19 files met |
| No network / provider / Docker dependency; no new migration; `scripts/coverage-floors.json` untouched | confirmed |
| Working tree clean, 17 commits | confirmed |

The remaining residue is race guards, defensive returns and debug-only log lines — for example
`reconciler.ts:200`, `:212`, `:225` and `:245`, each of which needs a close or abort to interleave inside an
await, and `tar.ts:84`, `:98`, `:191`, which guard states the public API cannot construct.

**One caveat worth stating plainly:** unlike the sibling lanes in this run, this lane's residual list is
*categorised* rather than argued line by line — it assigns each remaining line one of a small set of reasons
("race guard: requires a close/abort to interleave inside an await", "guard-only return for a state the
public API cannot construct") instead of demonstrating unreachability for each. The coverage numbers above
are measured and the suites pass, but a reviewer should treat that list as a starting classification, not as
a finished proof, and spot-check any line whose category looks wrong.

## Provenance

Produced by dispatch-pi run `79522c`, lane `client-runner-tests-6`. An agent wrote the tests in an isolated
against the branch before pushing it. The agent never pushed and never opened this pull request. The lane's
own progress journal went stale mid-run, so checklist completion was judged from the diff, file by file,
rather than from the journal's ticks.


bestony:
Rebased onto `419c9aac` (current main, including the #674 coverage-include fix). All 17 commits replayed cleanly with no conflicts; the branch head is now `9b682094`.

Re-verified from scratch at the new base, under main's **Vitest 5** toolchain rather than the branch's previous Vitest 4 lockfile:

- `pnpm check` exit 0
- `pnpm typecheck` 8/8 tasks successful
- `pnpm --filter @opentag/client test` — 136 files, 2,082 tests, all passed
- scoped coverage 93.11% lines / 85.63% branches, all 19 target files still at or above their targets
- diff is still tests only: 21 files, all under `packages/client/src/__tests__/`

The caveat in the description still stands: the residual uncovered list is a categorisation, not a per-line proof, and is worth spot-checking.

packages/client/src/__tests__/provider-cli-turn-plan-manager.test.ts
```diff
@@ -783,12 +978,176 @@ describe("Provider CLI Turn plan schema", () => {
...
+    expect(parseProviderCliTurnPlan(slackManaged)).toMatchObject({ selectionKind: "managed" });
+    expect(parseProviderCliTurnPlan({ ...slackManaged, slackApiHost: apiHost })).toMatchObject({
+      slackApiHost: apiHost,
+    });
+
+    expect(rejects({ ...slack, configDir: 42 })).toThrow(ProviderCliTurnPlanError);
+    expect(rejects({ ...base, configDir: "/opt/opentag/slack-config" })).toThrow(ProviderCliTurnPlanError);
+  });
+
+  it("fails closed on an unreadable, unparseable, oversized, or non-private plan file", async () => {
```
yuezengwu, line 1080: [P2] Preserve the account-global launcher isolation regression test

This replacement removes `does not rewrite the account-global launcher v1 marker`, including its `stat(join(layout.bin, "lark-cli"))` / `ENOENT` assertion. The new malformed-plan checks do not protect that boundary. In a diagnostic copy, I wrapped `prepare()` to create `layout.bin/lark-cli` after each successful Feishu preparation: all 39 current tests still pass, while restoring the removed assertion fails immediately. Please retain that test (or an equivalent assertion) alongside the new cases so a per-Turn plan cannot start modifying the shared account launcher without a regression failure.

packages/client/src/__tests__/provider-cli-manager.test.ts
```diff
@@ -609,6 +609,161 @@ describe("ProviderCliManager ensure", () => {
...
+      expect(blocked.ok).toBe(false);
+      expect(blocked.diagnostic?.code).toBe("version_incompatible");
+    } finally {
+      await server.close();
+    }
+  });
+
+  it("dry-run reports unsupported_platform when the catalog has no artifact for this platform", async () => {
+    const { accountHome } = await makeManager({});
+    const base = (await makeManagedCatalog("feishu", "1.0.92")).catalog[0] as ProviderCliCatalogEntry;
```
yuezengwu: [P2] Close the HTTP server created for the foreign-platform catalog

`makeManagedCatalog()` starts a real HTTP server, but taking only `.catalog[0]` here discards its `server` handle and this test never closes it. An after-test active-handle probe confirms one listening `127.0.0.1` server remains; the adjacent managed-noop case, which closes its fixture in `finally`, passes the same probe. This leaves a live socket until the worker is terminated and can accumulate on repeated in-process runs. Retain the fixture and close it in `finally`, or construct the unsupported-platform catalog without starting a server.

yuezengwu (changes requested):
Reviewed `9b682094` in an isolated worktree (branch base `419c9aac`; current main `1a2f3dd2` was also checked for intervening scope changes). `pnpm check`, build and typecheck pass; all 2,082 client tests pass, with scoped line coverage 93.11% and branch coverage 85.66%.

Two test regressions need correction: retain the existing account-global launcher isolation assertion, and close the HTTP fixture created by the unsupported-platform case. Both are confirmed by isolated diagnostic probes below. These findings concern test coverage/resource cleanup; no production defect is claimed.

packages/client/src/__tests__/provider-cli-turn-plan-manager.test.ts
bestony: Restored in `c7f1dc05`.

You were right that nothing else covered it. I diffed the suite's test titles against `419c9aac` and `does not rewrite the account-global launcher v1 marker` was the **only** test this branch removed — 30 on main, 37 here, with that one dropped. It is back verbatim, including the `stat(join(layout.bin, "lark-cli"))` / `ENOENT` assertion, placed alongside the new malformed-plan cases rather than folded into one of them, with a comment recording why: those cases all fail before anything is published, so none of them can reach the launcher boundary.

The file is now 74 tests (was 73) and both suites pass.

packages/client/src/__tests__/provider-cli-manager.test.ts
bestony: Fixed in `1687b1c0`.

The fixture is retained and closed in `finally`, matching the adjacent managed-noop case exactly — `const { server, catalog } = await makeManagedCatalog(...)`, the body in `try`, `await server.close()` in `finally` — with a comment saying why the handle matters when only the catalog entry is read.

I audited every other `makeManagedCatalog` call site while I was in there: this branch added two (the managed-noop case at ~712 and this one), and the first already closed its server. One thing to flag rather than change: lines 127-128 on `main` (`const older = ...` / `const newer = ...`) also discard their server handles. That predates this branch, so I have left it alone rather than widening a test-only PR under review — happy to send it separately if you want it.

I did not reproduce your active-handle probe; the evidence here is structural, the same close pattern as the case your probe showed passing.

bestony:
Both findings are fixed at `1687b1c0`; replies are on the two threads.

- `c7f1dc05` restores `does not rewrite the account-global launcher v1 marker` verbatim. A title-level diff against `419c9aac` confirmed it was the only test this branch removed.
- `1687b1c0` keeps the `makeManagedCatalog` fixture in the unsupported-platform case and closes it in `finally`.

Re-verified at the new head: `pnpm check` exit 0, `pnpm typecheck` 8/8, `pnpm --filter @opentag/client test` 136 files / **2,083** tests passing (one more than before — the restored guard), scoped coverage 93.11% lines / 85.64% branches with all 19 target files still at or above target. Diff against current main `1a2f3dd2` is still tests only: 21 files, all under `packages/client/src/__tests__/`.

yuezengwu (approved):
Re-reviewed `1687b1c0`. Both previous findings are fixed: the restored global-launcher assertion now catches the injected `prepare()` regression (39 other plan tests pass), and the unsupported-platform fixture leaves no HTTP listener after the test; the managed-noop control also passes the same resource check.

Fresh validation at this head: `pnpm check`, build and typecheck pass; all 2,083 client tests pass, with scoped coverage 93.11% lines / 85.64% branches. The complete PR diff remains 21 test/fixture files, with no removed test titles, new skips or coverage suppressions. Current main `1a2f3dd2` was checked for intervening changes; these results were run on the PR head. No remaining blocking findings.

PR #664, opened 2026-09-20
fix(cloud): wait for Runner release dependencies

yuezengwu:
## Summary

The first unified staging rollout exposed two timing gaps: npm accepted a release while its exact version still returned E404, and CapRover accepted a Server deployment before its build finished. Wait within fixed deadlines for those two states before continuing Runner verification and activation. Other npm errors/source mismatches and the final configuration/build concurrency checks still fail immediately; configuration writes are never retried.

Update the release instructions to derive the deployment principal from the repository's actual OIDC subject configuration, including immutable organization/repository IDs. The patch stays in the release helpers, regression tests and configuration instructions.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` passed (481 script tests, including seven regressions for publication visibility, timeout, source/auth failures and build/configuration races; package tests use the unchanged workspace cache).
- PostgreSQL integration: 555 passed. Agent Runtime coverage: 456 tests passed; 100% statements/branches/functions/lines.
- The triggering failures were observed in [the first staging deployment](https://github.com/first-tree-ai/opentag/actions/runs/35506672423). The current environment was recovered with the existing deployment controls; [read-only Actions acceptance](https://github.com/first-tree-ai/opentag/actions/runs/35507144409) confirms Server `1fd55e1f` and Runner `0.0.6-staging.30.1`, with all 49 migration hashes/timestamps matching and healthy readiness. This patch's waiting behavior is covered by fixtures; live staging still runs the base commit. Modified helpers also passed read-only release resolution and deployment checks against that environment.

## Breaking changes

None.

## Checklist

- [x] The change is focused and contains no unrelated work.
- [x] Tests and documentation cover the changed behavior.
- [x] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass.
- [x] No credentials, private data, or generated build output are included.
- [x] Canonical English docs and Chinese mirrors are synchronized.

PR #675, opened 2026-09-21
fix(cloud): enforce the Runner idle-wait deadline and budget the Deploy Runner job

bestony:
## Summary

Follow-up to the retrospective review of #664. That review raised two P2 findings and one P3; this patch fixes all three.

### 1. The CapRover idle wait can no longer authorize an update after its deadline

`waitForAppIdle` previously polled first and checked the remaining budget afterwards, so a slow poll could observe an idle app after the deadline and let `apply` continue mutating configuration past the promised bound. Now:

- the budget is checked **before** each poll, so a poll is never authorized at or past the deadline;
- the remaining budget is passed down as the request timeout (`getAppBuildState` → `caproverApi`), capping how far a single poll can overshoot;
- the budget is checked again **after** each poll, so even a poll that reports idle after the deadline fails closed instead of returning success.

Two deterministic regressions mirror the existing readyz deadline test: the reviewer's 30 ms repro (25 ms poll / 5 ms sleep / 25 ms poll) now throws at 30 ms with exactly one poll sent, and a late idle answer (20 ms deadline, 25 ms poll reporting `false`) throws instead of authorizing the update.

### 2. The Deploy Runner job has explicit headroom for the new wait path

The helper's three bounded waits (build idle, pre-update readyz, post-update readyz) can legitimately spend ~15 minutes combined, leaving no room in the 15-minute job budget for checkout, auth, release resolution, and the update itself. The job timeout is raised to 45 minutes, mirroring the staging rollout job, so a slow but valid apply reaches the helper's bounded result instead of being cancelled during or after the configuration write.

### 3. The changed Chinese mirror carries its sync marker

`docs/zh-CN/cloud-runner-release.md` now has the `Last synced with` marker required by `CONTRIBUTING.md` when a canonical document's mirror changes.

## Testing

- `pnpm check`, `pnpm typecheck`, and `pnpm build` pass.
- `node --test --test-concurrency=1 scripts/__tests__/*.test.mjs` → 488 passed, including the two new idle-wait deadline regressions (runner-deploy suite: 28 passed).
- Biome clean on the changed script and test files.

## Breaking changes

None.


Gandy2025 (approved):
Reviewed `419c9aac569529ef2f3e91e495481b0d7b295d54...3890bc04f1edf0701a853a7133cd9956b58037af` as a high-risk re-review of the three findings from #664.

## Standards

No blocking findings. The change stays scoped to the reviewed defects, the Chinese mirror now carries the required sync marker in the established format, and the workflow comment documents why the 45-minute budget is intentional.

## Spec

No blocking findings. The idle wait now checks its budget before and after each poll, passes the remaining budget through `getAppBuildState` to the CapRover request timeout, and fails closed when an idle response arrives at or after the deadline. The two regressions cover both the original 30 ms reproduction and a late idle answer. The Deploy Runner job now has explicit headroom beyond the three bounded five-minute waits.

## Validation

- `node --test --test-concurrency=1 scripts/__tests__/runner-deploy.test.mjs` — 28/28 passed.
- Biome check on the three changed script/test files passed.
- `git diff --check` passed.

Repository CI was still running when this review was submitted; this approval covers the reviewed code and spec, not the pending required checks.

yuezengwu (approved):
Independent verification at `3890bc04f1edf0701a853a7133cd9956b58037af`: no new blocking findings. In addition to the 28 deployment tests, three isolated probes passed: a response exactly at the deadline is rejected, the remaining request budget actually aborts a pending request, and a late idle response leaves `apply` with zero configuration updates. `pnpm check` and `git diff --check` passed. This review does not verify a live deployment.

PR #680, opened 2026-09-21
fix(cloud): name the manual recovery for a Runner tag claimed by a failed publish

bestony:
## Summary

The staging release chain can wedge permanently when a publish run claims a Runner version tag (image pushed) but fails before npm publication. Staging versions derive from the published npm sequence, so every later commit recomputes the same version, the reuse path then fails the tag identity check — and the failure message is a bare label diff that gives no recovery path. This is live right now: `0.0.6-staging.32.1` in GAR is an orphan from `618b093e`, and every npm Publish run since fails on the same label mismatch.

This PR does not change any release semantics: a mismatched existing tag still fails closed, never overwritten. It only names the manual recovery.

## Changes

- `scripts/runner/release.mjs`: when an existing tag fails identity verification, fail with the underlying detail plus the manual recovery steps — delete/retag the claimed tag in Artifact Registry and re-run (the tag is then rebuilt from clean source), or publish a placeholder npm version with the same number to advance the sequence. The final line restates the invariant: never overwrite an existing tag with a different build.
- `scripts/__tests__/runner-publish.test.mjs`: new test asserting the collision error names both recovery steps, carries the mismatch detail, and that no docker pull/push happens for a mismatched existing tag.
- `docs/cloud-runner-release.md` + `docs/zh-CN/cloud-runner-release.md`: a short "Failed release recovery" section describing the orphaned-tag state and the same recovery steps.

## Validation

- `node --test --test-concurrency=1 scripts/__tests__/runner-publish.test.mjs scripts/__tests__/runner-release.test.mjs` — 22/22 pass.
- Changed files pass `biome check` (no fixes applied).
- Full `scripts/__tests__/*.test.mjs` run locally: 482 pass, 6 fail — all 6 in unrelated files that require `node_modules` (pnpm install was not run in this worktree); none touch the Runner release helpers. CI runs the full gates.

## Note

The currently wedged chain still needs the manual unblock first (delete or retag the GAR `0.0.6-staging.32.1` tag, or publish a placeholder npm version); this PR only prevents the next occurrence from being a bare label diff with no recovery path.


Gandy2025 (changes requested):
Reviewing exact head `267241416eee5f46ae77cd788f1e05bee0530d9f` against base `1a2f3dd262e1909fc3ed4a4c37c7ef0d1b41d597`. Changes requested on both the release-safety contract and the promised recovery behavior.

## Standards

- **Blocking — `scripts/runner/release.mjs:122-134`: preserve inconclusive verification failures instead of prescribing destructive recovery.** `verifyRunnerIdentity` does more than compare labels: it can fail while fetching the child manifest/config (401/403/429/5xx or network failure), validating digests, parsing JSON, or validating the image structure/platform. This catch turns every one of those failures into “does not match this release” and tells the operator to delete/retag the image or advance npm. I reproduced this with a present, otherwise-correct tag whose config-blob request returned 503; the resulting message still recommended deleting/retagging the tag. That can destroy a valid release coordinate when GAR is merely unavailable, contrary to the existing fail-closed/inconclusive-read posture. Please attach the recovery guidance only to a proven identity mismatch (for example, via a typed/classified mismatch error), preserve all other verification failures unchanged, and add a regression covering an inconclusive child fetch.

## Spec

- **Blocking — `scripts/runner/release.mjs:130-131` and `docs/cloud-runner-release.md:48-50`: “retag and re-run” does not necessarily release the claimed version tag in GAR.** Artifact Registry's tag operation adds a second tag to the same image; it does not remove the source tag. An operator who follows the current instruction by adding a quarantine tag and re-running will hit the same claimed version again. Google documents adding and deleting tags as separate operations: https://cloud.google.com/artifact-registry/docs/docker/manage-images#tag. If preservation is intended, say to add a quarantine tag to the orphan digest **and then delete the original version tag**; otherwise say to delete/untag the exact version tag. Keep the English and Chinese docs plus the runtime message synchronized.

Targeted validation is otherwise clean: 33/33 Runner publish/release/GAR tests pass, and Biome reports no changed-code issues.

liuchao-001 (changes requested):
Reviewed at head `2672414`. The problem is real, the fail-closed semantics are preserved (no overwrite path added), and the test correctly asserts that no `docker pull`/`push` happens for a mismatched tag.

@Gandy2025's review landed while I was writing mine. I independently reproduced their first blocker — same 503-on-the-config-blob scenario, same outcome — so treat that one as confirmed by two reviewers, and their GAR `tag` finding is correct as well. The blocker below is a separate one neither of their points covers.

---

### Blocking — "publish a placeholder npm version with the same number" is unsafe on prod and a trap on staging

This step appears in the error message (`scripts/runner/release.mjs:131-133`) and in both docs. But `publish` runs for **both** channels — `.github/workflows/publish-npm-package.yml:126` (`--channel staging`) and `:318` (`--channel prod`) — while the message is channel-agnostic and its reasoning is staging-only.

**On prod it burns the release version permanently.** Production versions come from the `vX.Y.Z` tag, not from a recomputed sequence, so "later source commits keep recomputing the same release version" is false there. And `scripts/check-npm-version.mjs:16-21` fails the prod release when the published version's `gitHead` is not the release commit. Publishing a placeholder `open-tag@0.0.2` therefore makes `v0.0.2` unreleasable forever — npm will not let that version be reused after the unpublish window.

**On staging it works only in a narrow window that nothing states.** `resolveNextPublishedStagingVersion` reads the latest published staging version's `gitHead` (`scripts/resolve-staging-release.mjs:80-99`):

- placeholder published from a directory that is not a git checkout → no `gitHead` → `classifyPublishedMetadataLookup` (`scripts/resolve-staging-release.mjs:44-46`) throws `published version 0.0.6-staging.32.1 is missing gitHead` for **every** subsequent staging release. That wedges the chain harder than the collision it was meant to clear, and it is unrecoverable after the 72h unpublish window.
- placeholder whose `gitHead` is the commit the next release will publish → `scripts/release-versions.mjs:149-150` returns the same version and the collision repeats.
- placeholder from some other commit → works.

Only the third reading is correct, and neither the error message nor the docs say which one is meant.

**It also contradicts the documented release authority.** `docs/releasing.md:30-31`: "GitHub Actions is the only release authority… the repository must not contain an npm access token or a token fallback. Do not add a token." There is no sanctioned manual publish path, and a manual `npm publish` defaults to moving the channel's `latest` dist-tag — the exact invariant `scripts/resolve-npm-dist-tag.mjs` exists to protect (it deliberately holds no credentials to undo a dist-tag move).

Suggestion: drop the npm-placeholder step entirely and keep only the GAR step, which is correct on both channels once @Gandy2025's add-vs-delete wording is fixed. If the npm option is worth keeping for staging, gate it on `channel === "staging"`, state that the placeholder must carry a `gitHead` of a commit that is not the next release commit, and reconcile it with the release-authority rule in `docs/releasing.md`.

---

### On the first blocker (corroborating @Gandy2025)

One detail that may help the fix: the mismatch class is exactly `assertImageConfig` / `assertRunnerLabels` (`scripts/runner/gar.mjs:192-198`). Everything else `verifyRunnerIdentity` can throw — `fetchVerifiedJson` non-200, digest corruption, malformed JSON (`gar.mjs:142-153`, `207-235`) — is an inconclusive read and should propagate unchanged. Tagging the mismatch error there and catching only that in `release.mjs` keeps the two classes apart.

### Nit (non-blocking)

`scripts/runner/release.mjs:125-134` stringifies the underlying error and drops the original. This codebase already threads causes — `scripts/resolve-npm-dist-tag.mjs:41`, `scripts/resolve-staging-release.mjs:39`, `scripts/check-npm-version.mjs:14` all use `new Error(msg, { cause: error })`. Worth adding `{ cause: error }` here too.

---

Verified locally at this head: `node --test --test-concurrency=1 scripts/__tests__/runner-publish.test.mjs scripts/__tests__/runner-release.test.mjs` → 22/22 pass, matching the PR description.

bestony:
Addressed both reviews at `54542640`:

1. **Recovery guidance is now scoped to proven identity mismatches** (Gandy2025, corroborated by liuchao-001). `RunnerIdentityMismatchError` is thrown only by the four release-label assertions in `gar.mjs`; every other `verifyRunnerIdentity` failure (non-200 child manifest/config fetch, digest corruption, malformed JSON, platform checks, network failures) propagates unchanged. New regression test serves a 503 on the config-blob fetch and asserts the original failure message, no recovery guidance, and no docker operations.
2. **Tag-release wording fixed** (Gandy2025). Runtime message and both docs now say: add a quarantine tag to the claimed image, then delete the exact version tag in Artifact Registry, then re-run. "Retag and re-run" is gone.
3. **npm placeholder recovery step dropped entirely** (liuchao-001). The message and docs keep only the GAR quarantine-then-delete step, which is correct on both channels; the channel-specific reasoning and the release-authority conflict are gone with it. The rethrown error also carries `{ cause: error }` now.

Validation: 68/68 across the six Runner suites (`publish`, `release`, `gar`, `deploy`, `deploy-http`, `release-record`); biome clean on changed files.

liuchao-001 (approved):
Re-reviewed at head `54542640`. All three blockers are addressed, and I verified the fixes rather than reading the diff alone. **Approving.**

### Verified at this head

- **Recovery guidance scoped to proven mismatches.** `RunnerIdentityMismatchError` (`scripts/runner/gar.mjs:177`) is thrown only from the four release-label assertions, and `release.mjs:128` rethrows everything else untouched. I confirmed the new regression test is load-bearing by mutation: deleting the `instanceof` guard makes exactly `publish preserves an inconclusive registry failure without recovery guidance` fail (9 pass / 1 fail), and restoring it returns 10/10.
- **npm placeholder step fully gone.** No `placeholder`/`占位` remnants in `release.mjs` or either doc, and the `CHANNEL_CONFIG` import is still live at `release.mjs:252` (no dead import left behind). The replacement message is channel-agnostic and true on both the staging and prod paths — "This version number is claimed by that earlier attempt… until the tag is freed" no longer asserts the staging-only recompute narrative.
- **Quarantine-then-delete wording** is consistent across the runtime message and both the English and Chinese docs.
- **`{ cause: error }`** added and asserted by the test.
- **68/68** across the six Runner suites, reproducing your number. `biome check` on the three changed source files exits 0, no fixes applied.

### Nit (non-blocking, no action needed)

`assertImageConfig`'s platform check and `selectLinuxAmd64Entry` stay untyped, so a platform/index mismatch propagates raw without recovery guidance. That is the right call as written — it is also the conservative direction — and in practice it is unreachable from this workflow, since `release.mjs:143-145` rejects a non-amd64 build before push, so any orphan tag this workflow leaves behind is amd64. Noting it only so the choice is on the record.

---

Still outstanding before merge, neither in your court on code: @Gandy2025's `CHANGES_REQUESTED` needs their own re-review to clear, and `ownership-gate` is red pending @yuezengwu's approval on 3 files.

Gandy2025 (approved):
Re-reviewed the focused delta from `267241416eee5f46ae77cd788f1e05bee0530d9f` to exact head `545426400b7e41afd933f1f825a4f89ac0ad390a`.

## Standards — pass

The recovery wrapper now catches only `RunnerIdentityMismatchError`, which is emitted by the four release-label assertions. Inconclusive registry/network/digest/JSON/platform failures propagate unchanged, the original mismatch is retained as `cause`, and the new 503 child-fetch regression proves that destructive guidance is not attached to an inconclusive read.

## Spec — pass

The runtime message and both documentation mirrors now preserve the orphan under a quarantine tag and explicitly delete the exact claimed version tag before re-running. The unsafe npm-placeholder option has been removed, so the guidance is valid for both staging and production and no longer conflicts with the repository's release-authority contract.

No remaining blockers found. Local validation: all 68 tests across the six adjacent Runner suites pass; Biome is clean on the changed code.

yuezengwu (approved):
Reviewed exact head `545426400b7e41afd933f1f825a4f89ac0ad390a`; no additional blocking findings. Independently ran all six adjacent Runner suites (68/68 passing) and Biome on the three changed JavaScript files. The subsequent main change in #681 is confined to Context Tree UI/locales and does not affect the reviewed release paths.

PR #684, opened 2026-09-21
feat: support multiple named Context Trees

ggordonhall:
## Summary

Agents can connect multiple Context Trees by alias, with individual disconnect actions and no primary tree or implicit priority. Local and Cloud preparation reconcile attachments, preserve unpublished work, report failures per tree, and grant access to every ready tree. Cloud authorization is checked against each requested repository.

The shared configuration, database storage, runtime snapshots, hashes, settings operations, prompts, and English/Chinese documentation now use `contextTrees: { alias, repository }[]`. Existing suspension, revision, and publication retry protections remain in place.

All three package pins reference `@first-tree-ai/context-tree@0.1.16`. There is no local override, Git dependency, or vendored upstream code.

`@first-tree-ai/context-tree@0.1.16` is now published on npm. The lockfile records the published npm integrity hash, and the English/Chinese rollout guides now describe registry installation. Local frozen-lockfile metadata validation and `pnpm check` pass; full registry-package CI validation remains pending. Earlier local validation used a packed upstream build, as noted below.

## Testing

- Earlier local validation passed: `pnpm install` with a temporary packed-upstream override, `pnpm check`, and `pnpm build`.
- Passed: agent-runtime coverage (456 tests; 100% statements, branches, functions, and lines).
- Passed: PostgreSQL integration suite (556 tests), package smoke, and offline real-CLI multi-tree write/disconnect/read scenario.
- Shared, root scripts, CLI, and all affected Context Tree tests passed. Two unrelated client timeouts passed unchanged on isolated rerun; a locale ordering failure was fixed and the relevant web tests passed on rerun.
- Full `pnpm test` and the initial `pnpm typecheck` encountered failures in a pre-existing, untracked Feishu diagnostic test. That file is excluded from this PR and preserved locally. Server typechecking excluding that file passed.
- After removing the local override and replacing the lockfile resolution, all pre-push hooks passed: lint, format, `pnpm check`, and full `pnpm typecheck`. The unrelated untracked diagnostic file was temporarily set aside for those checks and restored afterward.

## UI validation

- Desktop evidence: browser inspection of the actual settings component with a temporary mocked API; screenshot inspected and console clean.
- Mobile evidence: no separate mobile pass recorded.
- Widths checked (`320px` / `390px` / `768px` / `1440px`): exact viewport widths were not recorded; responsive matrix remains unverified.
- States verified: two connected rows; disconnect one leaves the other; connect a new alias adds its row.
- Keyboard/accessibility checks: no separate keyboard audit recorded.
- New reusable block, theme value, or CSS exception: alias form field; no new theme values.

## Breaking changes

Migration `0049` replaces the old repository column with an empty JSONB connection list; existing selections are deliberately reset. Agents need fresh Context Tree selection. Deploy Server, Clients, and Cloud runner images together. Old configuration snapshots, operation records, and CLI store formats are not migrated; unsupported stores are rejected without deletion.

## Checklist


liuchao-001 (changes requested):
Reviewed at head `1eea6f8e`. Worktree off `827b51d5`, `pnpm install --frozen-lockfile` + `packages/shared` build, 84 Context Tree tests green locally, plus targeted probes against `ContextTreeManager` and the real published CLI.

## Verified independently (no action needed)

- **Lockfile integrity is correct.** I downloaded `@first-tree-ai/context-tree@0.1.16` from the registry and hashed it: `sha512-52iwDcENtwyrKpCBj/k6XU3xOfO1iRK80AXOvjU8CMd3QGjFwJSwizsz1qcveIeO7/FGeYoHZxGmfgsp/niNXg==` — byte-for-byte what `1eea6f8e` records. Published manifest deps (`commander ^15.0.0`, `mdast-util-from-markdown ^2.0.3`, `yaml ^2.8.3`, `zod ^4.0.0`) all satisfy the hand-written snapshot entry.
- **The upstream CLI contract this PR parses is real.** Against the published 0.1.16 I confirmed `connect --as`, `create --as`, `disconnect --tree`, `disconnect --all`, `resolve --tree`, `sync --tree`, `publish --tree`; that `resolve --json` returns `{schemaVersion: 2, connections: [{alias, projectPath, tree, ok}]}`; that `sync --tree` emits the schemaVersion-2 discovery payload *without* `--json` (so `syncedTree` parses it); and that `$HOME/.context-tree/connections.json` is schemaVersion 2. `StoredConnectionsSchema`, `DiscoverySchema` and `ConnectedSchema` all match the shipped shapes.
- **Dropping "one Context Tree per Agent" is coherent end-to-end.** GitHub grants are assembled per binding in `runtime-policy.ts`, and `taskBranchPrefix(sessionId, "context_tree")` lives inside each repository, so two trees in two repos cannot collide on refs.

## Blocking

### 1. Migration `0049` destroys every existing selection and forecloses rollback

`packages/server/drizzle/0049_multiple_context_trees.sql` adds `context_trees` and drops `context_tree_repository` in one statement. Two consequences the PR body treats as a footnote:

- Every already-connected Agent silently loses its tree on deploy, while its GitHub `context_tree` role grant stays in place — so the grant and the selection disagree until a human re-connects each Agent by hand. A backfill is derivable: the repo half of `OWNER/REPO` almost always satisfies `ContextTreeAliasSchema` (only a leading `_` needs a fallback), and the pre-migration table has at most one row per Agent so no alias/repository collision is possible.
- Dropping the column in the same migration means there is no safe deploy window and no rollback: any Server instance still running old code selects `context_tree_repository` and errors, and the data is gone even if you revert the image. Expand/contract (add + backfill now, drop in a later release) costs one extra migration and removes both problems.

If resetting everyone is genuinely the intended product decision, please get that on the record from an owner rather than carrying it as a PR-description note — it is user-visible data loss.

### 2. Managed mode detaches *every* tree when the execution environment is merely absent

`ContextTreeManager.#prepareConnections` filters `permitted` through `#managedFailure` before reconciling, and `#managedFailure` returns `AUTHENTICATION_REQUIRED` whenever `managedCredentials && !environment`. That collapses `permitted` to `[]`, so `reconcileContextTreeConnections` disconnects everything.

`prepareContextTree` in `session-runtime-manager.ts` explicitly supports the no-environment call shape, and `executionEnvironmentForSession` returns `undefined` whenever no proxy execution is registered for the Session — so this is reachable, not theoretical. Confirmed with a probe against the real manager:

```
# after a successful two-tree prepare, calling ensureAgent again with no environment:
PROBE calls: [["resolve",...],["disconnect","--tree","team",...],["disconnect","--tree","product",...]]
PROBE still attached: []
```

The code this replaces returned `unavailable` *before* touching the CLI, under the comment "Cached preparation never substitutes for the current execution grant" — it never mutated. `GITHUB_PERMISSION` (an authoritative, observed revocation) earning detachment is the intent stated in the Cloud comment and is fine; `AUTHENTICATION_REQUIRED` means "I could not see any grant at all" and should leave attachments untouched. Suggest skipping reconciliation entirely when the environment is missing, and reconciling only against grants actually observed.

`packages/client/src/__tests__/context-tree.test.ts:194` exercises exactly this call and asserts only the returned reasons, so the detachment is currently unasserted either way.

### 3. The Chinese doc is at a path the repo's own mirror check rejects

`Documentation Mirror Advisory` is red, and it reproduces locally:

```
$ node scripts/check-doc-mirrors.mjs --base 827b51d5
Documentation mirror check failed:
- mirror docs/zh-CN/context-tree-integration.md has no canonical file docs/context-tree-integration.md
```

The canonical is `docs/design/context-tree-integration.md`; `docs/zh-CN/` mirrors only top-level `docs/*.md`, and no other `docs/design/*` file has a mirror. The same script passes on the base commit, so this PR introduces it. Either drop the zh file (consistent with every other design doc) or move the canonical to `docs/context-tree-integration.md`. The checklist currently claims mirrors are synchronized.

## Should fix

### 4. The warm path lost its short-circuit, and a skills-install hiccup now revokes healthy trees

`#inFlight` is deleted on completion and nothing else short-circuits, so every `ensureAgent` re-runs `resolve` + `install --host claude` + `install --host codex` even when all trees are cached `ready`:

```
PROBE warm calls: ["resolve","install","install","resolve","install","install"]
```

The old `#ready` cache returned with zero CLI spawns. That is three extra child processes per Provider Runtime start inside a 5 s budget, and the comment at `session-runtime-manager.ts:310-313` still justifies this placement with "The manager caches per workspace, snapshot repository, and Provider" — no longer true.

Worse, `#installSkills` runs *before* `#connectAll` and `failAll`s on skip, so a later install skip wipes out trees that are connected and on disk:

```
PROBE first:  [{alias:"team",status:"ready"},{alias:"product",status:"ready"}]
PROBE second: [{alias:"team",status:"unavailable",reason:"CODEX_NOT_INSTALLED"},
               {alias:"product",status:"unavailable",reason:"CODEX_NOT_INSTALLED"}]
PROBE attached after: ["team","product"]
```

The Agent is told its memory is gone while the checkouts are right there and still attached, and `writableRoots` drops them. Restoring a ready-set short-circuit fixes both halves.

### 5. Zero trees + unavailable shim reports `unavailable` instead of `unconfigured`

In `#prepareConnections` the shim gate precedes the empty-list check, so `failAll("SHIM_UNAVAILABLE")` with no connections returns `{status: "unavailable"}`:

```
PROBE2 win32: {"status":"unavailable","reason":"SHIM_UNAVAILABLE"}
PROBE2 linux: {"status":"unconfigured"}
```

On Windows an Agent that has never configured a tree now gets "Context Tree unavailable (SHIM_UNAVAILABLE). This tree is not active for this Session. Other ready trees remain usable." — instead of "Context Tree: disabled for this Agent. Configure it in Agent settings → Context Tree." The old `repository === null` branch returned `unconfigured` regardless of shim state. Moving `if (!connections.length) return { status: "unconfigured" }` above the shim gate restores it.

### 6. Multi-tree prompts nest `## Context Tree` under `## Context Trees` and repeat the boilerplate per tree

`renderCloudContextTree` recurses into `renderCloudTreeStatus`, which re-emits its own `## Context Tree` heading and the full paragraph block — including the `members/<slug>/` warning and the skills instruction — once per alias. Three trees produce three same-level singular headings inside the plural section plus three copies of identical prose. `renderContextTree` in `managed-instructions.ts` has the same repetition without the heading clash. Hoisting the shared prose into the `configured` branch and leaving only the per-tree facts (path, branch/sha, or reason) in the recursion would read better and cost far fewer prompt tokens.

### 7. `ContextTreesSchema` is unbounded

Every other runtime-protocol collection in this codebase is capped — `byteString(...)` throughout `runtime-domain.ts`, `GITHUB_REPOSITORY_BINDINGS_MAX_AGENT_SCOPES = 1000`. `ContextTreesSchema` is a bare `z.array(...)`, and the value lands in the JSONB column, the effective runtime snapshot, and its hash. Local mode has no binding gate in front of it. A `.max(...)` matching whatever the product ceiling is would keep it consistent.

## Nits

- `context-tree-settings.tsx` trims `repository` but passes `alias` raw; whitespace only surfaces as the generic format error. `.trim()` before validating would match the repository fields.
- The alias input stays populated after a successful connect, so the next connect starts pre-filled with an alias that is now taken.
- One `<p role="status">` per connected tree means N live regions where there was one.

liuchao-001 (commented):
Re-reviewed at head `2fc734c9`. Reinstalled, rebuilt `packages/shared`, 122 Context Tree / runtime tests green, and I re-ran my own probes against the new code rather than relying on the added assertions.

## Verified fixed

**#2 — managed mode no longer detaches on a missing environment.** The early return in `#prepareSafely` lands before any CLI work. Probed on the real manager:

```
PB2 calls:    []                                  # was: resolve + 2x disconnect --tree
PB2 attached: ["product","team"]                  # was: []
PB2 status:   both AUTHENTICATION_REQUIRED
```

I also checked the two ways this fix could have over-corrected, and neither did: a genuinely revoked grant still detaches (`PB2R attached: ["team"]`), and emptying the tree list with an environment present still disconnects everything (`PB2E attached: []`). Good separation of "observed revocation" from "observed nothing".

**#4 — warm path.** `#skillsInstalled` drops the repeat installs, and because install no longer re-runs it also closes the second half, where a later skip wiped out ready, still-attached trees:

```
PB4 warm calls:            ["resolve","resolve"]   # was: resolve,install,install,resolve,install,install
PB4 after install break:   both still ready        # was: both CODEX_NOT_INSTALLED
```

Keeping the per-call `resolve` is the right trade — that is the reconciliation itself, not overhead. The refreshed comment at `session-runtime-manager.ts:308-310` now describes what the manager actually does.

**#5 — zero trees.** `PB5 win32: {"status":"unconfigured"}` (was `unavailable (SHIM_UNAVAILABLE)`).

**#3 — docs.** `node scripts/check-doc-mirrors.mjs --base 827b51d5` now reports `passed (31 pairs discovered)`, and the three inbound links were updated with the move. No stale references to the old path.

Everything except `ownership-gate` is green or pending on this head.

## Still open

**#1 (blocking) — migration `0049` is unchanged**, and there is no reply here on it. To be clear about which half is which:

- Whether resetting every Agent's selection is acceptable is a product call, and the PR body does declare it. I have raised it with the repository owner rather than holding it against you silently.
- The other half is not a product call: `ADD COLUMN` and `DROP COLUMN` in one migration leaves no safe deploy window. The Context Tree records that this Server "runs checked-in database migrations before listening", with an advisory lock so instances cannot become concurrent migration writers — that protects against concurrent *writers*, not against the instance that is already listening on old code when the new one migrates. During any rolling deploy that instance selects `context_tree_repository` and errors, and the dropped data cannot be recovered by rolling the image back.

Splitting this into expand/contract — add + backfill in this release, drop in a later one — costs one extra migration, removes the deploy window problem outright, and makes the product question moot because nothing is lost if the answer later turns out to be "we should have kept them". I would rather see that than a reset that is only safe if the deploy is a full stop-the-world.

**#6 and #7** are unaddressed and unanswered. Neither blocks merge, and I am happy to drop either if you disagree — but I would like a yes/no rather than silence, since #6 in particular ships duplicated boilerplate and a `## Context Tree` heading nested inside `## Context Trees` into every multi-tree prompt.

The three nits are still open and remain optional.

ggordonhall:
On review item 1: the product decision is to accept the breaking migration. Backwards compatibility with the single-tree configuration is not required for this change.

Migration `0049` intentionally resets existing tree selections; affected Agents must explicitly select their named trees again. It removes the selection metadata, not the repositories or their stored knowledge. An existing GitHub role grant is authorization, not a configured tree selection, and does not automatically reconnect a tree.

We accept the deployment consequence too: Server, Clients, and Cloud runner images must be deployed together. Mixed old/new server versions and rollback by reverting only the server image are unsupported after this migration. The rollout documentation and PR description already state the reset and coordinated deployment requirement.

For those reasons, we are keeping the migration as written rather than adding alias backfill or an expand/contract compatibility phase.

liuchao-001 (approved):
Approving at head `2fc734c9`.

## On item 1

That is a clear answer and it is the call of the people who own this product, not mine — I raised it, you have decided it on the record with reasons, and I am not going to re-litigate it. Two things I want to acknowledge rather than let stand as unanswered:

You are right to correct my framing of the grant. I wrote that the grant and the selection would "disagree" after the reset; a `context_tree` role grant is authorization and was never a selection, so what actually remains is an unused grant, not an inconsistent state. My underlying point — that a human has to go re-select each Agent's trees — is the part that survives, and you have confirmed it and documented it.

On the deployment half, I am satisfied because you have named the consequence rather than absorbed it: mixed old/new Server versions and server-only image rollback are unsupported after `0049`, and that is stated in the PR body and the rollout docs where an operator will meet it. My objection was never "this must not happen", it was "this must not happen by accident". It isn't.

## What I verified

Items 2-5 are fixed, and I confirmed each against the new code with my own probes rather than reading the added assertions:

- managed mode with a missing environment now makes **zero** CLI calls and leaves both trees attached, while a genuinely revoked grant still detaches and emptying the list still disconnects — the distinction lands exactly where it should
- the warm path is down to one `resolve` per call, and an install skip no longer knocks out trees that are ready and still attached
- zero configured trees returns `unconfigured` on Windows again
- `check-doc-mirrors.mjs` passes, with all three inbound links moved

Also independently checked earlier and still holding: the lockfile integrity matches the published tarball byte for byte, and the shipped 0.1.16 CLI really does expose `--as` / `--tree` / `--all` and the schemaVersion-2 payload shapes this code parses.

Locally on this head: `pnpm install --frozen-lockfile`, `packages/shared` build, and 122 Context Tree / runtime tests green.

## Items 6 and 7

I have moved these out of your way as follow-ups rather than keep asking for a yes/no on a PR that is otherwise done — #686 for the repeated prompt boilerplate and nested heading, #687 for the unbounded `ContextTreesSchema`. Neither is yours to answer on this PR; close them if you disagree with the premise.

Note that `Checks` and the Node compatibility jobs were still running when I approved, so this approval is a review verdict and not a statement that the merge gates are met. `ownership-gate` still wants a second independent approval.

liuchao-001 (approved):
Re-approving at head `811ccbaa` — my previous approval was against `2fc734c9` and no longer applies.

You picked up 6 and 7 rather than leaving them to the follow-ups; thank you. I have closed #686 and #687 as resolved here. Both verified on this head.

**7.** `.max(32)` with the reasoning in the comment and the doc sentence, plus the boundary test at 32/33. Good.

**6.** The facts/guidance split is a better shape than what I suggested — keying the managed side by status and de-duplicating the Cloud side line-wise both land correctly. I rendered the real prompts rather than reading the assertions. Five aliases across five distinct statuses now produce one heading, five tree-specific fact lines, and each guidance paragraph exactly once:

```
## Context Trees
Trees have no implied precedence. …
Alias team — acme/team:
Context Tree: /t/team — synchronized at the start of this Turn (branch main, commit aaaaaaaaaaaa).
Alias stale — acme/stale:
Context Tree: /t/stale — the preserved checkout has unpublished changes.
Alias old — acme/old:
Context Tree: /t/old — this Turn's synchronization failed (SYNC_FAILED).
Alias denied — acme/denied:
Context Tree unavailable (GITHUB_PERMISSION).
Alias slow — acme/slow:
Context Tree unavailable (TIMEOUT).
Ready Context Trees are connected in this Agent's settings. …
Read the decisions that bear on a task …
`members/<your Agent slug>/` … Do not write to another Agent's member directory.

For stale trees with unpublished changes: …
For other stale trees, the on-disk copy may be outdated: …
Unavailable trees are not active for this Turn; …
The current execution does not grant this Session the selected repository, …
```

The managed renderer is equally clean. Re-labelling the shared sentences for the plural ("Ready Context Trees are…", "For stale trees with unpublished changes:…") was the right call — each paragraph now reads correctly when it covers several aliases.

Shared and client suites green locally on this head (50 + 90).

## One thing the new bound introduced

Not a blocker, and it only bites at 32 connected trees, but it is the exact failure class the surrounding code already guards against, so I would rather flag it than let it ship silently.

The cap is only enforced inside `updateContextTreeSelection`, at `ContextTreesSchema.parse(connections)` in `agent-service.ts:945`. That runs *after* `#dispatch`. So a 33rd connect or create:

1. passes `attachmentOutcome`, which checks alias and repository collisions but not length
2. runs `#dispatch`, which for Cloud calls the operation runner and for Local sends the frame to the Computer — and for `action: "create"` that path **publishes a new GitHub repository**
3. then throws a `ZodError` out of `updateContextTreeSelection`

The `catch` only translates `AGENT_REVISION_CONFLICT` and rethrows everything else, so the caller gets an unhandled error rather than a typed `failed` code. `ContextTreeOperationResponseSchema` has no code for this, so `failureMessage` in the web settings form has nothing to render either. And on the `create` path the published repository is orphaned: created remotely, never recorded.

`attachmentOutcome`'s own comment says it exists so that "Idempotence and collisions are decided before any remote verification or publication" — the length check belongs in exactly that spot, with a new response code (`tree_limit_reached` or similar) and the matching message in `apps/web/messages/agent-settings/{en,zh}.json`.

Happy for that to be a follow-up rather than another round on this PR — your call. Approving either way; `ownership-gate` still needs a second independent approval.

liuchao-001 (approved):
Re-approving at head `8e2a6cfd`; my approval of `811ccbaa` no longer applies.

You caught a real defect that I missed when I approved `811ccbaa`, and I want to be straight about that rather than quietly re-approve. Keying guidance by `entry.status` collapsed `PREPARING` and genuine failures into one bucket, so the merged paragraph told an Agent whose only non-ready tree was still preparing "do not attempt to repair the tree yourself" — advice that does not apply to a tree that has not finished preparing. When I rendered that section last round my sample happened to include both a `PREPARING` tree and a `GITHUB_PERMISSION` tree, where the merged sentence reads fine, so the case that actually misleads never showed up in what I looked at. My verification was too narrow.

Rendered both shapes on this head:

`ready` + `PREPARING` only — the repair advice is correctly absent:

```
Unavailable trees are not active for this Session. Other ready trees remain usable.
Trees still preparing may become usable in a later Session.
```

`ready` + `PREPARING` + `GITHUB_PERMISSION` — both sentences, each once:

```
Unavailable trees are not active for this Session. Other ready trees remain usable.
Trees still preparing may become usable in a later Session.
Do not assume earlier decisions were recorded in unavailable trees, and do not attempt to repair the tree yourself.
```

Moving the managed renderer to the same line-wise `new Set` the Cloud side already used is the right unification — one dedup strategy instead of two, and it degrades correctly as reasons multiply. Narrowing `renderContextTreeGuidance` to exclude `configured` is a good tightening too; that argument could never be a `configured` status and the type now says so.

I checked whether the Cloud renderer needs the same split and it does not: `PREPARING` is a local session-start-budget state, and no Cloud path produces it — `prepareConnection` yields `GITHUB_PERMISSION`, `TIMEOUT`, `PREPARATION_FAILED`, `CONNECT_FAILED`, `TREE_OUTSIDE_WORKSPACE` or a CLI failure code. So the asymmetry between the two renderers is justified rather than an oversight.

55 tests green locally on this head across the managed-instructions, cloud-turn-worker and context-tree suites.

The tree-limit ordering point from my last review is untouched, which is what I expected — I left it as your call and a follow-up is fine. Nothing else open from me. `ownership-gate` still wants a second independent approval.

packages/server/src/services/agents/context-tree-operation-service.ts
```diff
@@ -24,8 +24,10 @@ export class ContextTreeOperationService {
       config.runtimeConfig.revision !== input.expectedRuntimeConfigRevision
     )
       return { status: "failed", code: "stale_configuration" };
-    if (config.runtimeConfig.contextTreeRepository !== null && config.status !== "suspended")
-      return { status: "failed", code: "pause_required" };
+    const connections = config.runtimeConfig.contextTrees;
+    const unchanged = attachmentOutcome(config, input);
+    if (unchanged) return unchanged;
+    if (connections.length > 0 && config.status !== "suspended") return { status: "failed", code: "pause_required" };
```
Gandy2025, line 31: [P2] Check the tree limit before dispatching create/connect

When a suspended Agent already has 32 connections, a request for a new alias/repository passes `attachmentOutcome` and reaches `#dispatch`. The new `ContextTreesSchema.max(32)` is only checked by `AgentService.updateContextTreeSelection` after that remote operation completes. For Local `create`, the repository can therefore be published successfully before persistence throws an uncaught `ZodError`, leaving an unbound repository and a misleading publication-uncertain response in the UI. I reproduced both create and connect reaching the remote runner before validation fails. Validate the proposed connection list before dispatch, return a typed limit error with a user-facing message, and cover the 32-to-33 boundary with an assertion that neither remote runner is called (while allowing disconnect and idempotent requests at the limit).

Gandy2025 (changes requested):
Reviewed head `8e2a6cfd`. One P2 issue remains: the new 32-tree limit is enforced after remote side effects, so a rejected create can leave a published repository without an Agent attachment. I independently reproduced the ordering with the real operation service and shared schema, substituting a successful remote runner: `remote-create -> persist-validation -> ZodError` (and the same sequence for connect). Please reject an over-limit candidate before dispatch.

Validation: `pnpm install --frozen-lockfile`, shared-package build, `pnpm check`, and 364 selected shared/server/client/web tests passed. The local real-CLI integration file initially exceeded Vitest's default 5-second timeout; all seven tests passed on an isolated rerun with a 30-second test timeout. Current CI checks are green apart from the approval gate. Full repository and PostgreSQL suites were not rerun locally.

The explicitly accepted breaking migration and coordinated rollout are not objections in this review.

yuezengwu (changes requested):
The owner has accepted the migration 0049 selection reset and coordinated rollout policy. That decision is settled for this review.

[P2] Update the Cloud-model fixture when merging current main

Head `8e2a6cfd` passes type checking on its own, but merging current `main` (`1e777c05`) produces TS2353 at `apps/web/src/features/agents/agent-settings/cloud-model-settings.test.tsx:20`: the fixture still supplies `contextTreeRepository: null`, which this PR removes from the Agent config type. The test was added to main after this branch's base. Please merge current main, update that fixture to `contextTrees: []`, and rerun `pnpm typecheck` before merging. This is outside the PR's current diff, so I am reporting it here.

Independently verified in a detached review worktree: the unmodified integration passes build/check but fails typecheck at that field. Changing only the fixture temporarily makes full typecheck and 290 affected integration-of-branches tests pass; that diagnostic change was reverted. On the original PR head, build/check/typecheck and 1,177 affected tests passed; all 194 PostgreSQL integration tests passed across the initial run and an unchanged retry of one suite that hit a container-startup timeout. No deployment or live IM acceptance is claimed.

ggordonhall:
Addressed both open change requests in `66785b59`. Re-review would be appreciated.

**P2 — tree limit enforced after dispatch (Gandy2025).** `CONTEXT_TREES_MAX` is now exported from `@opentag/shared` and checked in `attachmentOutcome` before `#dispatch`, so an over-limit connect/create returns the new typed `tree_limit_reached` code without calling the Cloud runner or sending a Local operation frame — `create` can no longer publish a repository that then fails persistence. Disconnect and idempotent re-connect at the limit are unchanged. Added the `failureMessage` case plus `agent_settings_context_tree_limit_reached` in `agent-settings/{en,zh}.json`. Boundary coverage: `packages/server/src/__tests__/context-tree-operation.test.ts` asserts 32→33 is rejected with neither remote runner called and that disconnect/idempotent requests still succeed at 32; a web test asserts the message renders. `ContextTreesSchema.parse` in `agent-service.ts` stays as defense-in-depth.

**P2 — merge main / fixture (yuezengwu).** Merged `origin/main` (`9c416ffa`) and updated `cloud-model-settings.test.tsx` to `contextTrees: []` (`66785b59`); `pnpm install` re-run and paraglide regenerated.

Verification on this head: `pnpm typecheck` (8/8), `pnpm check`, and the shared/server/web targeted suites are green. Full `pnpm test`: the only failure was the known flake `packages/client/src/__tests__/context-tree.integration.test.ts` exceeding Vitest's default 5 s timeout in the parallel run; it passes 7/7 on an isolated rerun, matching Gandy2025's earlier note. Migration `0049` is untouched.

Gandy2025 (approved):
Re-reviewed at head `66785b59`. The requested fix is complete: `attachmentOutcome` checks the shared 32-tree limit before dispatch, returns `tree_limit_reached`, and leaves disconnect and identical attachments usable at the limit. The response schema and English/Chinese UI messages are wired through. My original create/connect probe now returns the typed failure with zero remote operations or persistence attempts. The regression tests also assert that neither Cloud execution nor a Local operation frame is dispatched.

The merge of main has no additional manual conflict resolutions, and the Cloud-model fixture now uses `contextTrees: []`. Locally passed: frozen-lockfile installation, `pnpm build`, `pnpm check`, full `pnpm typecheck` (8/8 tasks), and 154 targeted shared/server/web tests across 10 files. No remaining blocking findings from this review. The remaining CI and patch-coverage jobs are still in progress at approval time; completed code checks are green. Full unit and PostgreSQL suites were not rerun locally in this follow-up. No merge was performed.

liuchao-001 (approved):
Re-approving at head `66785b59`; my approval of `8e2a6cfd` no longer applies.

## On the tree-limit finding

I raised this last round and called it non-blocking — "your call, a follow-up is fine." @Gandy2025 blocked on it instead, and that was the better call. The fix turned out to be three lines in a function that already existed for exactly this purpose, and the failure mode is a published GitHub repository that nothing references plus a `publication_uncertain` message that points the user at the wrong problem. When a fix is that cheap and the failure is that quietly confusing, "you would need 32 trees first" is not a good reason to defer it. I weighted likelihood too heavily against consequence.

The fix is in the right place and complete. I probed the boundary combinations the regression tests do not cover, at exactly 32 connections:

```
create (the publishing path)      -> tree_limit_reached   cloud.run=0 send=0 persist=0
existing alias, different repo    -> alias_conflict        cloud.run=0 send=0 persist=0
new alias, already-connected repo -> repository_conflict   cloud.run=0 send=0 persist=0
new alias, new repo               -> tree_limit_reached    cloud.run=0 send=0 persist=0
```

`create` specifically — the path that can publish before persisting — now returns the typed failure with zero remote calls. The more specific conflict codes still win over the limit, which is the right precedence: telling someone "you are at the limit" when their real problem is a duplicate alias would send them to disconnect a tree for no reason. Keeping `ContextTreesSchema.parse` as defence in depth is right too.

## On the base drift

@yuezengwu's TS2353 is one I should have caught and did not. I reviewed four heads of this branch inside an hour and never once checked it against current `main`, even though the branch was already days behind and this PR removes a field that new `main` code was still writing. Verified now on the merged head: `origin/main` (`86ba37f0`) is an ancestor of `66785b59`, and `pnpm typecheck` passes 8/8 with exit 0 and zero `error TS` lines.

## What I checked that the other reviews did not

A merge of a fast-moving `main` can silently undo earlier review fixes, so I confirmed each one still stands at this head rather than assuming the merge was clean: the managed-credentials early return, the empty-list-before-shim ordering, `#skillsInstalled`, both facts/guidance renderers, the `PREPARING` split, and the relocated Chinese design doc are all intact. The only thing `main` changed in the Context Tree surface area is an unrelated Router `maxTokens` line in `cloud-turn-worker.ts`.

Green locally on `66785b59`, all with real exit codes: frozen-lockfile install, `pnpm typecheck` (8/8), doc mirrors (32 pairs), shared 69, client 110, server 36, web 21.

Nothing open from me. The gate now reads "Needs bestony or yuezengwu", and @yuezengwu's change request is still outstanding, so that is the remaining step rather than anything on this branch.

yuezengwu (approved):
Re-reviewed the full change at `66785b59db65a2df423e9bc9deb217091c9dd535`, including current main `86ba37f0f7c645d2cbea8073376c0bbb4d7f2a34`. My Cloud-model fixture/typecheck finding is resolved; no remaining blocker found in this review. The owner has explicitly accepted the documented migration0049 reset and coordinated-rollout policy.

Independent validation passed: build, check, typecheck; 1,201 affected shared/client/server/web/CLI tests; 194 PostgreSQL integration tests; and 19 portable dependency tests, including the embedded real Context Tree CLI. Full-suite/coverage and live deployment acceptance are not claimed. Remote CI is still in progress and must complete before merge.

Issue #472, opened 2026-09-02
Flaky CI: client-turn.integration "reports Claude Code ... setup failure as not started" inherits vi.waitFor's 1s default under load


liuchao-001:
`client-turn.integration.test.ts` fails non-deterministically under parallel CI load, on the `it.each` case `reports Claude Code %s setup failure as not started`.

## Evidence

Run [33579448518](https://github.com/first-tree-ai/opentag/actions/runs/33579448518) on `46fd625c` (PR #460):

- **attempt 1**: `Node 22.22.2 Compatibility` and `Node 26 Compatibility` both failed. `Node 24` passed in the same attempt.
- **attempt 2**: same commit, no changes, all jobs green.

The failing assertion was `expect(connection.reports().length).toBeGreaterThan(0)` inside a `vi.waitFor`. The branch under test touches no file under `packages/`, `apps/` or `e2e/` — `git diff --name-only 7bd23dc7 46fd625c` returns nothing in those trees — and `main` at `7bd23dc7` was green through the same job.

## Likely cause

`waitForReport` in `packages/client/src/__tests__/client-turn.integration.test.ts:356` is:

```ts
const waitForReport = async (index: number): Promise<TurnReportRequest> => {
  await vi.waitFor(() => expect(connection.reports().length).toBeGreaterThan(index));
  ...
};
```

There is no `timeout` option, so it takes vitest's default of 1000ms. That is the entire budget for accepting a delivery, failing provider startup, and propagating a Turn Report — on a runner that is also executing the rest of the workspace in parallel. Five tests in the file go through this helper, and the `it.each` above is parameterized three ways, so it is the densest user of it.

Two Node versions failing the same assertion in one attempt while a third passed, then all passing on retry, is the signature of a budget that is too tight rather than of a logic error.

## Why this is worth tracking

This is the same shape as #345 and #238 — a wall-clock budget that holds on an idle machine and not under load. #306, #325 and #308 are the rest of that family. This one had no issue.

It also has a specific bearing on #460, which is about to make `ownership-gate` a required check. That policy exempts `apps/web` and gives several paths a self-merge lane on green CI, so "CI green" carries more weight after it lands than it does today. A flake that hit two of two review rounds on a single branch makes that weaker than the policy assumes.

## Suggested fix

Give `waitForReport` an explicit timeout sized to the work rather than inheriting 1000ms, or drive the wait off the report event instead of polling. Please do not paper over it with test retries.

PR #688, opened 2026-09-21
fix(cloud): use Router models and hosted diagnostics

yuezengwu:
## Summary

Cloud model settings previously offered Local Pi suggestions and saved models that Cloud dispatch later rejected. The connection test also required a Local daemon connection for a logically online Cloud Computer.

Read Cloud models from the Router using the existing upstream URL and credential, and use the same catalog for configuration and execution admission. Test hosted model connectivity on the Server, preserving Local daemon tests. Configure the Runner's Pi provider for Router's output budget and supported fields.

## Testing

- `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` passed (7,839 workspace unit tests).
- `pnpm --filter @opentag/client test:agent-runtime:coverage` passed with 100% statements, branches, functions and lines.
- The complete PostgreSQL suite passed: 44 files / 568 tests, using `pnpm --filter @opentag/server exec vitest run src/__tests__/integration --maxWorkers=1`. Earlier parallel runs hit local Docker port-binding/timeouts; the failing startup cases also passed independently. No assertion or timeout was relaxed.
- Captured requests from the installed Pi SDK pass the real Router request validator, including a tool-result roundtrip; no upstream request was sent by this contract check.
- Final model/diagnostic regression run: 78/78 passed, including reasoning-only completion and cancellation racing a response.
- Regression coverage includes catalog failure/expiry, retired models, configuration admission without writes, replay during an outage, dispatch/grant admission, diagnostics cancellation, and Local behavior.

No live deployment is performed by this PR task.

## UI validation

- Desktop evidence: local component screenshots, including Router choices and a successful hosted-model diagnostic, retained in the task's `FIX-SCREENSHOTS.md`.
- Mobile evidence: 320px and 390px component widths; the retired-model state is included.
- Widths checked: 320px / 390px / 768px; desktop browser viewport 1280px. 1440px was unavailable in the current browser viewport.
- States verified: available Router models, explicit selection/save, unsaved-test blocking, retired saved model, catalog unavailable, and model-connectivity success.
- Keyboard/accessibility checks: model picker opens with ArrowDown, closes with Escape and retains focus; visible controls use accessible labels. No horizontal control overflow at 768px.
- Reusable block: extracted the existing model field into a local component; no design token or CSS exception.

These screenshots visibly identify synthetic Router responses. They establish local UI behavior, not post-merge staging or real Cloud task acceptance.

## Breaking changes

Cloud model discovery requires the configured upstream's authenticated `/models` endpoint. `OPENTAG_CLOUD_MODEL_ALLOWED_MODELS` is retired; Router model permissions become authoritative. Configure staging's existing model base URL and credential for its Router tenant, and deploy the updated Runner through the existing joint release. No database migration or new feature flag.

## Checklist

- [x] Required validation gates pass.


liuchao-001 (approved):
Reviewed at head `93e61f44`, reading the full diff against `6d1a26cb` and re-running the repository gates locally.

## What I verified

- Every `resolveRuntimeModel` / `isModelAllowed` / `defaultModel` call site is awaited — `im-delivery-cloud.ts:193`, `cloud-delivery-owner.ts:208/254`, `cloud-session-collaboration-owner.ts:478/1570`. No place treats the new promise as a truthy snapshot.
- No `allowedModels` / `OPENTAG_CLOUD_MODEL_ALLOWED_MODELS` consumer survives outside the retired-but-parseable env entry and the docs that describe it.
- The proxy still binds a request to the token's model (`cloud-model-proxy.ts:260`), so retiring the static allowlist does not widen what a minted grant can call.
- `CloudModelGrantService.#mint` keeps the reservation/tombstone invariants with the catalog read added inside the lifecycle: a denied or unconfirmable model releases capacity, a revocation that lands during the read keeps the tombstone, and `#disclose` still refuses after a sweep or eviction.
- `RouterCloudModelCatalog.list()` never resolves the expired snapshot after a failed refresh, and `#refresh()` cannot reject, so `isModelAllowed` callers cannot throw.
- `createForAccount`'s new pre-transaction replay uses the same fingerprint (which includes `computerId`) and the same conflict semantics as the in-transaction check, so the replay path is unchanged apart from running earlier.
- Manage permission is creator-only (`#resolveAgentScope`), so scoping the update-path pre-check by `createdByUserId` matches exactly who can reach `updateById`.

Local gates on this head: `pnpm check`, `pnpm build`, `pnpm typecheck` and `pnpm test` all pass (7,841 unit tests across the five packages; the PostgreSQL suite needs Docker and I did not run it).

The change is well shaped — one owner for the catalog, admission moved to it everywhere, and the tests cover the paths that matter (expiry with a failed refresh, in-flight sharing, mint admission, dispatch/grant refusal, replay during an outage, cancellation racing the shared catalog read). Nothing here blocks the merge. The notes below are ordered by how much I care about them.

## 1. A Router `/models` failure is invisible to operators (`packages/server/src/services/sandboxes/cloud-model-catalog.ts`)

Every failure inside `#fetchModels` — DNS, TLS, 401 on the master key, a non-JSON body, a redirect, the 5s timeout — is sanitized to `undefined`, and `#refresh` turns that into `available: false`. The catalog takes no logger and emits no diagnostic, unlike its siblings in this composition (`createCloudDeliveryComposition` already threads a `ServiceLogger`).

The consequence after this PR is larger than before it: `/models` is now a live dependency of *every* Cloud dispatch and grant. If the staging Router serves `/chat/completions` fine but its `/models` rejects the master key, all Cloud execution stops with `model_unavailable` — which is the same code a legitimately unoffered model produces. An operator sees identical failures for "you picked a retired model" and "the platform cannot reach the Router's model list", with nothing in the logs distinguishing them.

Sanitizing what reaches the *client* is right; staying silent in the *server log* is what I would change. A rate-limited warn (or a diagnostic counter on the refresh outcome, no key, no upstream body) would make this diagnosable. Happy for this to be a follow-up issue if you would rather not grow the PR — I am not gating on it.

## 2. The web collapses "model path disabled" into "could not be confirmed" (`apps/web/src/features/agents/agent-settings/runtime-configuration.tsx:44-49`)

The route deliberately distinguishes two states: a deployment with no model path answers `200 {available: false}` ("a stable unavailable answer, not an error", `account.ts:161`), while a Router list that cannot be confirmed is a `503`. The client maps both to `{ kind: "unavailable" }`, so a deployment running Cloud Computers with `OPENTAG_CLOUD_MODEL_ENABLED=false` — a configuration `.env.example` explicitly supports ("Model requests can be disabled independently while Runner save/release remains available") — shows "Available models could not be confirmed. Retry to update the list." with a Retry button that can never succeed, and no way to save a model.

A separate terminal state for `available: false` on a 200 ("hosted models are not enabled on this deployment") would keep the distinction the server just made.

## 3. An unconfirmed Computer read puts the Local model UI back on a Cloud Agent

`agent.computerKind` is only set when the Computers read is confirmed (`agent-queries.ts:394`), and `evidenceSettling` only waits for the *first* fetch. After a later failed or paused Computers refresh the detail page keeps rendering, `computerKind` becomes `undefined`, and `RuntimeConfigurationForm` falls through to the Local editor — Pi model suggestions plus the free-text "Custom model ID" box, for a Cloud Agent. That is exactly the shape this PR removes.

Saving from that state still fails closed (`CLOUD_MODEL_NOT_ALLOWED` is mapped to a clear message), so this is presentation only, and it follows the existing `context-tree-settings.tsx:25` convention of defaulting an unknown kind to `local`. Worth deciding deliberately rather than inheriting: for the model field specifically, "kind unknown" is closer to "do not offer Local suggestions yet" than to "Local".

## 4. Admission is keyed on the current binding, so one path can still persist an unoffered model

`#assertExplicitCloudModelChoice` skips when `computerId` is absent at create, and `#assertExplicitCloudModelChoiceForAgent` resolves the Agent's *current* Computer. An Agent created with no Computer and a model, then bound to a Cloud Computer in a later update that does not carry `runtimeConfig.model`, ends up with a stored model the Router never admitted. Dispatch refuses it and the settings page flags it as retired, so nothing unsafe reaches a Sandbox — but the stored config is one the write path claims it rejects. Validating against the update's target `computerId` when one is present would close the remaining half.

## 5. The Runner's 8,192-token cap is a hard-coded constant (`packages/client/src/runner/cloud-turn-worker.ts:299`)

`maxTokens: 8_192` and `compat: { supportsStore: false }` are applied to every Router model, in the Runner, changeable only through a joint CLI/Runner release — while the model list itself is now dynamic. Two questions: where is the 8,192 pinned (a Router doc or a captured rejection?), and what is the plan when the Router raises it or one model differs? Since the catalog already reads `/models`, relaying a per-model budget later would keep this from drifting; a comment naming the source would be enough for now.

## 6. Smaller things

- `CLOUD_MODELS_UNAVAILABLE` is a module-level object handed straight back from `list()` and from `createStaticCloudModelCatalog`. Nothing mutates it today; `Object.freeze` would keep it that way.
- A retired `OPENTAG_CLOUD_MODEL_ALLOWED_MODELS` is parsed and ignored in silence. A one-line boot notice when it is still set would stop an operator from believing it still restricts models during the rollback window.

## On the evidence

The PostgreSQL and Pi-SDK contract results in the description are yours; I did not reproduce them. The screenshots establish local component behavior with synthetic Router responses, as you say — the thing still unproven is a real Router tenant answering `/models` and `/chat/completions` with the staging master key, which is exactly what item 1 makes hard to debug if it goes wrong.

bestony (approved):
Reviewed at head `93e61f44` (full diff against `6d1a26cb`), independently of liuchao-001's review, which I read and agree with — in particular his item 1 on missing catalog observability (I confirmed `RouterCloudModelCatalog` accepts no logger while `createCloudDeliveryComposition` already threads a `ServiceLogger`) and item 2 on the settings page collapsing the Server's 200-vs-503 distinction.

Local gates I ran on this head:

- `packages/server`: full unit suite excluding integration — 166 files / 3,053 tests pass.
- `packages/client`: `cloud-turn-worker` suite — 23/23 pass.
- `apps/web`: `cloud-model-settings` + `api` suites — 34/34 pass.

I did not run the PostgreSQL integration tier (needs Docker); taking the author's 568-test result for that. `pnpm check`/`typecheck`/`build` I did not rerun; the test tiers above cover the changed modules.

Two findings that add to liuchao-001's list:

1. **The settings page is stricter than the Server during a Router outage.** With the catalog unconfirmed, `runtimeModelField` sets `invalid = unavailable || unsupported`, which disables the whole execution-section save — including reasoning-effort-only changes and a reset to the platform default (`model: null`), and disables the reasoning Select entirely. The Server accepts exactly those writes: `#assertExplicitCloudModelChoiceForAgent` returns early for `null`/`undefined` models (`agent-service.ts:843/860`). During an outage, the one action users most want — reset to the default — is the one the form blocks, while the Server-side admission the form is trying to protect doesn't even apply to it. Worth deciding which side is intended: either the Server should refuse all Cloud runtime writes while the catalog is unconfirmed, or the form should allow model-null saves with a warning.

2. **The hosted connectivity probe sends a different request shape than production turns.** `CloudAgentRuntimeTester.#probe` posts `{model, messages, max_tokens: 16, stream: false}` directly to `{base}/chat/completions` (`cloud-agent-runtime-tester.ts:164`), while production turns are Pi SDK documents (`openai-completions`, `maxTokens: 8192`, `supportsStore: false`). If any catalog model rejects `max_tokens` — OpenAI reasoning models reject it in favor of `max_completion_tokens` — the diagnostic would report `provider_failed` for a model real turns can use. The Pi SDK contract check in the PR description validates the production shape, not the probe's. Worth confirming the probe passes against every model the Router currently lists, or aligning the probe's payload with the production shape.

Neither blocks the merge. Approving.

liuchao-001:
Agreeing with both of @bestony's additions. One boundary is worth pinning before item 1 gets acted on, because the form's strictness is not uniform.

`unsupported` is computed from `Boolean(modelDraft)`, so a **retired saved model with a healthy catalog** already permits the reset: selecting "Platform default" clears `modelDraft`, `unsupported` goes false, `invalid` goes false, and the save enables. `cloud-model-settings.test.tsx` pins exactly that path — "shows a retired saved model and allows an explicit reset to the platform default", asserting the `{ model: null, reasoningEffort: null }` write. What freezes is only `unavailable`, i.e. the catalog unconfirmed, where the model Select, the reasoning Select and the save button are disabled together so no dirty state can even be produced.

So the question to decide is narrower than "the form blocks the reset": it is *while the catalog is unconfirmed, may a user still write `model: null` or a reasoning-only change?* The Server says yes today — both admission helpers return before any catalog read for `null`/`undefined`. Whichever side you align, please keep the healthy-catalog retired-model reset working as currently tested.

On item 2, the direction I would emphasise is the false **positive** rather than the false negative: the probe's hand-rolled body is not the Pi document shape, so "the hosted model connection passed" does not establish that a real turn's request is accepted by that model. The UI copy already scopes the claim to model access, so this is not a mis-statement — but aligning the probe's payload with the production shape would make a pass mean the thing users will read into it.

bestony:
Agreed on both refinements.

On item 1: your narrowing is right — the healthy-catalog retired-model reset works and is test-pinned, so the only open question is the unconfirmed-catalog state. My recommendation is to align the **form** with the Server rather than the reverse: while the catalog is unconfirmed, allow the save when the effective write is `model: null` or a reasoning-only change (the existing "could not be confirmed" warning already sets expectations), and keep blocking only writes that carry a non-null explicit model. Tightening the Server instead would turn a Router outage into a full write freeze of Cloud settings, which is the worse operational outcome. Either way, preserve the tested reset path.

On item 2: yes — the false positive is the point that matters. "Passed" reads as "my real turns will work", and the probe can't establish that today. Aligning the probe with the production Pi document shape would make a pass mean what users read into it; at minimum, a pass against every model the Router currently lists would bound the gap.

Neither changes the approval; these are follow-up candidates for the author.

bestony (approved):
Re-reviewing at head `5df29998` (the `fix(ci): expose failed coverage test diagnostics` commit on top of the previously approved `93e61f44`).

Verified on this head:

- `scripts/unit-coverage.mjs` change is contained: it reads the vitest JSON report once, formats failing tests with their assertion messages, and falls back to captured stderr/stdout/exit status when no report exists. Output goes to stderr only when the run already failed — a passing run emits nothing.
- `scripts/__tests__/unit-coverage.test.mjs` gains 10 focused cases covering the new formatter's paths (named failures, file-level suite errors, crashed runs, signal deaths, spawn failures, passing runs). `node --test scripts/__tests__/unit-coverage.test.mjs scripts/__tests__/unit-coverage-gate.test.mjs` passes 41/41 locally.

One scope observation, non-blocking: this commit sits outside the PR's cloud-model scope, but it is the harness behind the PR's coverage claims (`pnpm test:coverage`), and the PR description itself records coverage-run debugging pain ("earlier parallel runs hit local Docker port-binding/timeouts; the failing startup cases also passed independently") — so I read it as in-flight validation tooling rather than unrelated work. Happy to keep it here; an alternative would be a separate PR, at the author's discretion.

The two follow-ups from my previous review (settings-form-vs-Server admission during an unconfirmed catalog; aligning the connectivity probe with the production Pi request shape) remain open and are still non-blocking.

Approving.

liuchao-001:
Re-reviewed at `5df29998`. Product code is byte-identical to the previously approved `93e61f44` — `git diff --name-only 93e61f44..5df29998 -- packages apps docs .env.example` returns nothing — so my earlier verification of the cloud-model change carries over unchanged. The new commit is confined to `scripts/unit-coverage.mjs` and its test.

I went looking for *why* that commit exists, and the answer is worth recording on the PR, because it corrects something in my own earlier review.

**The required coverage gate was red on the head I approved.** On `93e61f44`, `Enforce Patch Coverage` ([run 35594429507](https://github.com/first-tree-ai/opentag/actions/runs/35594429507)) failed with:

```
── client ──
   lines 94.03% (15835/16840)  [tests failed]; duration 211648ms; retries 0 (flaky 0)
...
At least one project reported failing tests.
```

It concluded at 11:50 UTC — eight minutes after I approved at 11:42, which is why my review said "nothing blocks the merge" while a required gate was still outstanding. Withdrawing that phrasing for that head: it was red, and no reviewer accounted for it.

**What I then established:**

- The same run gave *no indication which test failed* — that is the hole this commit closes, and it is specific to the coverage job: `runProject` captures vitest's output through `spawnSync` and discarded it whenever coverage artifacts existed. The plain `Checks` / `Node XX Compatibility` jobs print failures normally, so the blind spot was real but narrow.
- On `5df29998` the same client suite passes: `lines 94.03% (15836/16840); duration 210702ms; retries 0 (flaky 0)`, whole job green. `packages/client` is unchanged between the two heads, so this is one red and one green over identical test code.
- I could not reproduce it locally under the identical configuration: `pnpm exec vitest run --config vitest.coverage.config.ts --project client` → 136 files / 2083 tests pass, exit 0 (macOS).

**So the red was transient, and it is consistent with the client flakes already tracked in #472 and #486** — both documented as reproducing only under load, and #472 records exactly this signature (red, then green on unchanged code). I can't assert it *was* one of them, because the diagnostics that would have named it are the thing this commit adds.

That reframes the scope question: this is not tangential tooling. It is a direct fix for a diagnostic hole this PR walked into, and the next time #472 or #486 fires inside the coverage job it will name the test and its assertion instead of printing `[tests failed]`. I'd keep it here.

One caveat worth stating so the gain isn't overestimated: the new formatter only covers the coverage job. The flakes in #472/#486 were both observed in `Node XX Compatibility` / `Checks`, which this does not touch.

`formatTestFailureLines` itself reads correctly — single parse of the report, failure lines only when `status !== 0`, empty on success, with stderr/stdout/signal/spawn-error fallbacks, and the 10 new cases cover those paths. No further findings. Re-approving at this head.

liuchao-001 (approved):
Re-approving at `5df29998`.

Product code is unchanged from the previously approved `93e61f44`, so the cloud-model review stands as written. The added commit is contained to the coverage harness, its new formatter is correct, and the required `Enforce Patch Coverage` gate — which was red on the previous head with no diagnostics — is green on this head.

Evidence and the correction to my earlier "nothing blocks the merge" are in the comment above. The four follow-ups from my first review, plus @bestony's two, remain open and non-blocking.

bestony:
Agreed, and thanks for the archaeology — it applies to my `93e61f44` approval too: I posted it at 11:46, four minutes before the red coverage run concluded at 11:50, and I did not check CI state at approval time. So for that head the record is: two approvals while a required gate was still running red, unaccounted for by both of us.

For the current head `5df29998` I verified just now that every required check is green, including `Enforce Patch Coverage` (20m33s, pass).

Your reframing of the scope question is the right one: the new commit is a direct fix for a diagnostic hole this PR walked into, not tangential tooling. I withdraw my "separate PR, at the author's discretion" hedge — it belongs in this PR.

The open follow-ups (your four, my two) stay with the author and remain non-blocking.

PR #689, opened 2026-09-21
feat(tasks): show chronological conversations and attachment metadata

Gandy2025:
## Summary

Task activity grouped a turn's replies under its request, so a user follow-up sent between replies could appear after the whole exchange. Show inbound messages and captured replies in one chronological conversation, deduplicate by identity, and preserve the reader's position during pagination and late reports.

- Keep steered input as an annotation on the user message. Collapse execution summaries when captured replies exist, while keeping failure and incomplete/truncated-content notices visible.
- Display inbound attachment names, types, and availability from existing stored resources. Add optional browser-safe metadata to the Task API without exposing provider resource keys.
- Fix the responsive sidebar issue found during browser validation with a one-line pnpm patch for Kumo 2.13.1, plus a regression test covering drawer dismissal and retained page state. Include the patch in all container dependency layers and in the staged Runner build context; document the patch rationale and removal criteria in `apps/web/src/ui/KUMO.md`.

No database migration, runtime protocol change, attachment preview/download endpoint, quoted-message expansion, or source-conversation link.

## Testing

Validation performed for this PR:

- `pnpm check`, `pnpm build`, and `pnpm typecheck`
- Full Web suite: 1,456 tests across 111 files; focused Web regression suite: 128 tests
- Latest follow-up (`2acc2b37`): 41 relevant Web tests, `pnpm check`, Web typecheck, and Web build passed. The reply-content suite exercises the production single-reply component; incomplete-history assertions exercise the conversation itself.
- TaskService unit suite: 32 tests
- Container/build-context regression suite: 23 tests
- Docker `deps` and `prod-deps` dependency installation stages
- Shared unit suite: 584 tests
- `pnpm --filter @opentag/client test:agent-runtime:coverage`: 456 tests; 100% statements, branches, functions, and lines
- Commit and push hooks

Local full-suite limitations (Node 22.22.2 on macOS):

- `pnpm test` failed in unchanged client tests: three Context Tree cases, a Cloud Session queue case, and an artifact-store case. The client suite reported 2,078 passing tests and five failures. Server Git pack tests also exceeded their local timeout before Turbo stopped the run. This is not a green full-suite claim.
- `pnpm --filter @opentag/server test:integration` encountered timeouts/failures in unchanged authentication, ownership-migration, IM binding, Session collaboration, and workspace-lifecycle tests. Final result: 551 passed, 17 failed across 44 test files. No failing test file is modified by this PR.

Independent review found no actionable issue in the Task changes. Browser validation used the built Web app, real local API/authentication, and disposable PostgreSQL with seeded capture records; live Feishu/Slack delivery was not exercised.

## UI validation

- Desktop evidence: real authenticated Task route; verified interleaved follow-ups/replies, summary collapse, attachment metadata, visible error states, 50-to-54 history pagination, unique entries, and 30-second refresh with one late reply. Pagination/refresh anchor drift stayed below 0.31px.
- Mobile evidence: after resizing from desktop, the mobile drawer opens/closes, content fits the viewport, and an expanded Task summary retains its state. Returning to desktop restores expanded navigation.
- Widths checked: `320px`, `390px`, `768px`, and `1440px`; no horizontal overflow in the Task content area.
- States verified: captured replies, steered input, legacy missing capture, incomplete capture, failed turn, truncated content, oversized attachment, history loading, and background refresh.
- Keyboard/accessibility checks: Enter expands/collapses the execution summary and opens navigation; Escape closes the drawer and restores focus to its trigger. Existing focus-trap tests pass.
- New reusable block, theme value, or CSS exception: task-local conversation and attachment components reuse existing design-system primitives and tokens. No new theme values or CSS exceptions. The Kumo patch refreshes sidebar context when `collapsible` changes without remounting the page.

## Breaking changes

None. Attachment metadata is optional, and older stored messages and responses remain supported.

## Checklist

- [ ] `pnpm check`, `pnpm build`, `pnpm typecheck`, and `pnpm test` pass. (The first three pass; full-suite failures are described above.)


bestony (changes requested):
Reviewed at head `23f3d68` against `main`.

## Blocking: image builds break because `patches/` is not copied into the build context

The new `patchedDependencies` entry makes pnpm hash `patches/@cloudflare__kumo@2.13.1.patch` from the lockfile during install, but neither image build ships the file:

- Root `Dockerfile`, `deps` and `prod-deps` stages: only `package.json pnpm-lock.yaml pnpm-workspace.yaml` are copied, then `pnpm install --frozen-lockfile` fails with `ENOENT: no such file or directory, open '/app/patches/@cloudflare__kumo@2.13.1.patch'` (exit 254). This is the failure behind "Build Pull Request Image" and "Container Smoke" (`docker build .` from the same Dockerfile). The prod install fails too, so the patch hash is verified even when the filtered install never pulls the patched package.
- `scripts/runner/Dockerfile`: same shape — manifests only, then `pnpm install --frozen-lockfile`, hence "Build linux/amd64 image and run offline smoke" fails the same way.

Fix: add `COPY patches patches` to the root `Dockerfile` (both stages) and to `scripts/runner/Dockerfile` before their `pnpm install` RUN lines. `.dockerignore` already allows the directory through. Please rerun the image/container smoke jobs after the fix.

## Non-blocking

- `TaskAttachments` keys list items by `attachment.ordinal ?? index`, but the resource schema does not enforce unique ordinals, so two stored resources sharing an ordinal collide as React keys. Cosmetic; `index` (or ordinal+index) would be safer.

## What I verified

- Local, on this head: `pnpm install --frozen-lockfile` applies the Kumo patch cleanly (lockfile hash matches); web typecheck passes; focused suites green — task conversation/timeline/anchor/sidebar web tests (86), shared task contracts (6), server task-service including the new attachment projection (32).
- The browser contract holds: `TaskAttachmentSchema` omits `providerResourceKey` and strict-parses; the server test proves no provider keys leave the server; `.max(16)` matches the inbound resource bound.
- Timeline semantics are right: identity-based dedupe keeps distinct messages with identical text and stored revisions; absorbed steers are annotated on the input rather than duplicated; replies without usable timestamps keep their neighbour positions without inventing display times; non-Feishu providers still show no fabricated replies (unchanged); the surviving-anchor logic covers prepend, mid-conversation insertion, and transient-status replacement.
- Docs and the zh-CN mirror are in sync; new message keys exist in both locales.

CI at review time: Browser Smoke, CLI Pack Smoke, Secret Scan, Workflow & Shell Validation pass; Checks, Node compatibility, and Patch Coverage were still running. ownership-gate is at "6 files awaiting approval" — the normal owner-approval state.

liuchao-001 (commented):
Reviewed at head `4d0210f0`, including the `23f3d68a` state that @bestony's review covers. I had independently landed on the same blocker — `patchedDependencies` makes pnpm hash the patch file before it resolves `--filter`, so every manifests-only install context fails — and the follow-up commit closes it correctly at all four install sites, including `scripts/e2e/doctor-docker/Dockerfile`, which isn't in that review's list but runs the same frozen install from a manifests-only copy. The generic assertion in `runner-build-context.test.mjs` that walks `pnpm.patchedDependencies` and compares staged content is a better guard than an allowlist string check; it will catch the next patch too.

The conversation model itself reads well. `buildTaskTimeline`'s dedup keys are all non-nullable in the contract (`message.id`, `report.turnId`, `messageId`), so no entry can silently collapse into another; Feishu `create_time` and `occurredAt` are on the same epoch-ms scale (`feishu/adapter.ts:363`), so the interleaving is actually sound rather than accidentally sorted; and the attachment projection is genuinely source-backed — `im-message-inbox.ts:313` already persists `content.resources` with `ordinal` and `availability`, and `task-service.ts` selects the whole `m.content` column.

On the duplicate-ordinal key nit: ingest assigns `ordinal` positionally over a `.max(16)` array at `im-message-inbox.ts:313`, so stored resources cannot collide. Agreed it costs nothing to key by index, but no stored row can reach it.

## Non-blocking

**`TaskOutgoingReplies` is now unreachable.** `tasks-page.tsx` was its only production caller; after this PR only `task-outgoing-replies.test.tsx` renders it, and its incomplete/omitted notice is re-implemented in `TaskReplyNotice` (`task-conversation.tsx:163`). Suggest deleting the wrapper and re-pointing the existing suite at the exported `TaskOutgoingReply` — the per-`msgType` coverage is worth keeping, the wrapper isn't.

**Record the Kumo patch in `apps/web/src/ui/KUMO.md`.** That file is the stated home for Kumo-specific implementation choices, and it still reads "OpenTag uses Kumo `2.12.0`" against a `2.13.1` pin. The patch itself is well-targeted: `SidebarProvider`'s context `useMemo` omits `collapsible` from its dependency list, and because Kumo's internal media subscription is registered on a child before `AppShell`'s `useSyncExternalStore` subscription, the provider can recompute on `isMobile` while `collapsible` is still `"none"` and then never refresh. `sidebar.test.tsx` encodes exactly that ordering, which is the right way to hold it. What's missing is the note telling the next reader why the patch exists, whether it went upstream, and what would let it be dropped. Pinning the key to `@2.13.1` means a Dependabot bump fails the install rather than silently unpatching — the behaviour you want, but only if that failure points somewhere.

**Scope.** The Kumo patch, its regression test, and now five build files are a shell fix with no dependency on the Task conversation work. I understand it unblocked mobile validation and I'm not asking you to unwind it; a separate PR would have let a vendored dependency patch be reviewed and reverted on its own terms.

**Nit.** `sizeBytes` is projected to the browser and rendered nowhere. Harmless and plausibly wanted later, but the summary scopes the surface to name, type, and availability.

**Nit.** The rewrapped comment in `scripts/e2e/doctor-docker/Dockerfile:13` now runs past its neighbours; the root `Dockerfile` version of the same edit stayed wrapped.

## Verified

The three jobs that were red on `23f3d68a` are green on `4d0210f0`: Build Pull Request Image (2m57s), Build linux/amd64 image and run offline smoke (2m3s), Container Smoke (2m12s), alongside Browser Smoke, CLI Pack Smoke, Secret Scan, and Workflow & Shell Validation. Checks, the three Node compatibility jobs, and Patch Coverage were still running when I wrote this. ownership-gate still needs bestony or yuezengwu.

bestony (approved):
Reviewed at head `2acc2b37`. This supersedes my earlier CHANGES_REQUESTED on `23f3d68`.

## Blocking issue resolved

The missing-`patches/` image-build break is fixed in `4d0210f0` and verified: `COPY patches patches` now covers the root Dockerfile (deps + prod-deps), `scripts/runner/Dockerfile`, and `scripts/e2e/doctor-docker/Dockerfile`, plus the staged Runner context via `build-context.mjs`, with a regression test that walks `pnpm.patchedDependencies` and compares staged content. I reproduced the fix locally with a full `docker build .` (deps, prod-deps, web/server build, runtime image all pass), and the three previously-red CI jobs were green on that commit (Build Pull Request Image, Runner linux/amd64 smoke, Container Smoke).

## Earlier non-blocking items

- Attachment list key: resolved — `key={index}` with a rationale comment; and as liuchao-001 pointed out, ingest assigns ordinals positionally, so the theoretical collision I flagged cannot occur in stored rows.
- liuchao-001's non-blocking items are all addressed in `2acc2b37`: the dead `TaskOutgoingReplies` wrapper is removed with the reply-content suite re-pointed at the production single-reply component and incomplete-history assertions moved into the conversation tests; `KUMO.md` now documents the patch rationale, upstream status, removal criteria, and the container-context requirement (and fixes the pre-existing `2.12.0` version drift); the doctor-docker comment is rewrapped.

## Verification at `2acc2b37` (local)

- `pnpm install --frozen-lockfile` clean; Kumo patch applies and hash matches.
- Web typecheck passes.
- Focused suites green: 109 tests across task-conversation, task-outgoing-replies, tasks-page, task-timeline, task-activity-timeline, and sidebar. Shared contracts (6) and server task-service (32) pass on the unchanged code from `23f3d68`.

CI at review time: Secret Scan, Workflow & Shell Validation, Documentation Mirror Advisory, and Evaluate ownership pass; Checks, Node compatibility, Patch Coverage, Browser/CLI Pack Smoke, and the image jobs were re-running on the new head (all had passed or were green on `4d0210f0`; the `2acc2b37` delta is web/docs only). ownership-gate remains at "11 files awaiting approval" — the normal owner-approval state.

liuchao-001 (approved):
Approving at head `2acc2b37`. All four of my non-blocking items from `4d0210f0` are addressed, and I read the delta rather than taking the summary for it.

- The dead `TaskOutgoingReplies` wrapper is gone, and the reply-content suite is re-pointed at the production `TaskOutgoingReply` without losing a case: the two removed snapshot-level tests are re-homed, `omittedCount` 0/3 into the new `it.each` in `task-conversation.test.tsx` and `status: "incomplete"` into the existing collapsed-summary test. The `tasks_reply_incomplete` key still has exactly one live caller in `TaskReplyNotice`, and nothing references `data-ui="task-sent-replies"` any more.
- Quiet improvement worth naming: the epoch fixture moved from `"1757980800"` to `"1757980800000"`. The old value was seconds, so it formatted as a 1970 date while the comment claimed "both real times are formatted" — the fixture now matches the scale Feishu actually sends (`feishu/adapter.ts:363`) and the assertion means what it says.
- `KUMO.md` covers the rationale, the subscription ordering, removal criteria, and the container-context requirement, and it states plainly that nothing went upstream. That last line is the useful one: it tells the next upgrader this patch is ours to carry until someone files it.
- `key={index}` with the rationale comment, and the doctor-docker comment rewrap.

`sizeBytes` staying in the projection is fine by me — it was a nit, and keeping the `omit`-based shape whole is the more defensible boundary.

CI at `2acc2b37`: ownership-gate satisfied, Build Pull Request Image, linux/amd64 offline smoke, Container Smoke, Browser Smoke, CLI Pack Smoke, Node 26, Secret Scan, Quality Scoreboard, and Workflow & Shell Validation all pass. Checks, Node 22/24, and Patch Coverage were still running — worth a glance before merge, but the `2acc2b37` delta is Web and docs only.

yuezengwu (approved):
Reviewed exact head `2acc2b37`; no blocking findings from the independent review.

The head checks have now completed successfully (or were intentionally skipped). Since main advanced to `313723c2` after those checks started, I also constructed a clean local merge with that base and reran 157 focused tests, including the combined public-exports snapshot: all passed. Shared build, `pnpm check` and Web typecheck also passed.

This approval covers code review and those local checks; I did not independently run a full Docker build, PostgreSQL suite or live IM/browser acceptance.

